PR #4378, opened 2018-11-27
Feature ret scalars

heckad:
Add new option to method all in the Query object. They responsible for return list of scalars inside of results objects when `len(self._entities) == 1 and isinstance(self._entities[0], _ColumnEntity)`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea
 *.pyc
 *.pyo
 *.egg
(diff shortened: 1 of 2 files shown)


zzzeek:
hi there -

For this PR, it needs:

* all whitespace changes outside of the feature itself need to be reverted. 

* revert the changes to .gitignore

* unit tests are required, please see https://www.sqlalchemy.org/develop.html#pullrequests

* I disagree this should be a flag on all().   There is a method named `.all_scalar()` that has been proposed.

* it would very helpful to create a feature issue as well (see guidelines)

thanks!

Issue #4377, opened 2018-11-27
PostgreSQL domain with array reflected to sqltypes.NULLTYPE


jsynowiec:
I have two pgsql domains: `isoweekday` that is based on INT and `isoweekdays` that is an array of the former. Both domains add some checks like no duplicate values, no empty list, only values from <1, 7> range.

```sql
CREATE DOMAIN isoweekday AS INT
CHECK(...);

CREATE DOMAIN isoweekdays AS isoweekday[]
CHECK(...);
```

I've created custom types, like so:

```py
class ISOWeekDayDomain(types.UserDefinedType, types.Integer):
    def get_col_spec(self, **kwargs):
        return 'isoweekday'


class ISOWeekDaysDomain(types.UserDefinedType, ARRAY):
    def __init__(self, **kwargs):
        if 'item_type' not in kwargs:
            kwargs['item_type'] = ISOWeekDayDomain

        super(ISOWeekDaysDomain, self).__init__(**kwargs)

    def get_col_spec(self, **kwargs):
        return 'isoweekdays'
```

and then used them to define a column:

```py
Column(
    ISOWeekDaysDomain, nullable=False, server_default=text('ARRAY[1,2,3,4,5,6,7]')
)
```

Creating migrations using Alembic went well, and it created the following statement:

```py
    op.add_column(
        u'config',
        sa.Column(
            'dow',
            ISOWeekDaysDomain(),
            server_default=sa.text(u'ARRAY[1, 2, 3, 4, 5, 6, 7]'),
            nullable=False,
        ),
    )
```

After applying migrations on the database, I'm getting the following warning:
```
sqlalchemy/dialects/postgresql/base.py:2683: SAWarning: Did not recognize type 'isoweekday[]' of column 'dow'
```

Running alembic yields a new database migration that alters the previously added column with `sqltypes.NULLTYPE`.

```py
op.alter_column(u'tent_crawling_config', 'dow',
           existing_type=sa.NullType(),
           nullable=False,
           existing_server_default=sa.text(u'ARRAY[1, 2, 3, 4, 5, 6, 7]'))
```

The `domains` dictionary passed to `_get_column_info` contains both created domains and during execution it enters the [dialects/postgresql/base.py#L2677](https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/dialects/postgresql/base.py#L2677) block.

What am I missing? How to register my custom types that map domains so the reflection API recognises them?

[...]

PR #4380, opened 2018-11-27
fix/arrays in psql domains

jsynowiec:
This adds handling of PostgreSQL domains with array data types to the db reflection (`_get_column_info`).

Tests against PostgreSQL database are passing (at least of the test_reflection unit).
```
53 passed, 4 skipped in 3.18 seconds
```
There are some failing tests in the other units but those are also failing on the `master` branch for me.

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -2600,6 +2600,12 @@ class PGDialect(default.DefaultDialect):
 
     def _get_column_info(self, name, format_type, default,
                          notnull, domains, enums, schema, comment):
+        def _handle_array_type(attype):
+            return (
+                attype.replace('[]', ''), # strip '[]' from integer[], etc.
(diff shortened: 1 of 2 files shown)


zzzeek:
OK this is a bug but I'm having trouble connecting this to your description in #4377 where you are working with named domains. In this test, the domain name "arraydomain" is still lost, how were you planning to handle the Alembic comparison part of this?

jsynowiec:
You're right. I got carried away and missed that I've also did what you've suggested in your response to the #4377. I've added my user defined types to `ischema_names` at some point and somehow didn't connect those two things (it was late).

What do you want to do about the other, seems unrelated now thing about reflecting a domain with array data type?

zzzeek:
the thing you hvae here is a great PR so we can just put it through and merge it.  I like the changes to be linked to an issue but we'll just amend #4377 to express how that went.   also the "ischema_names" thing has never been documented because I've never made it "official", but it's been there for many years so maybe that's the API for now.

PR #4402, opened 2018-12-10
Fix the "greatest" example.

konstantint:
The current example code does not pass `**kw` down to the `compiler.process` calls, thus the example does not work when invoked with, `literal_binds=True`.

Besides, the calls to `process` each argument twice are wasteful, and reusing the built-in `case` expression instead of hard-coding the SQL statements is slightly nicer overall, isn't it?

### DIFF
--- a/lib/sqlalchemy/ext/compiler.py
+++ b/lib/sqlalchemy/ext/compiler.py
@@ -85,8 +85,8 @@ method which can be used for compilation of embedded attributes::
     @compiles(InsertFromSelect)
     def visit_insert_from_select(element, compiler, **kw):
         return "INSERT INTO %s (%s)" % (
-            compiler.process(element.table, asfrom=True),
-            compiler.process(element.select)
+            compiler.process(element.table, asfrom=True, **kw),
+            compiler.process(element.select, **kw)
         )
 
     insert = InsertFromSelect(t1, select([t1]).where(t1.c.x>5))
@@ -119,10 +119,11 @@ below where we generate a CHECK constraint that embeds a SQL expression::
 
     @compiles(MyConstraint)
     def compile_my_constraint(constraint, ddlcompiler, **kw):
+        kw['literal_binds'] = True
         return "CONSTRAINT %s CHECK (%s)" % (
             constraint.name,
             ddlcompiler.sql_compiler.process(
-                constraint.expression, literal_binds=True)
+                constraint.expression, **kw)
         )
 
 Above, we add an additional flag to the process step as called by
@@ -265,13 +266,13 @@ A synopsis is as follows:
 
       @compiles(coalesce)
       def compile(element, compiler, **kw):
-          return "coalesce(%s)" % compiler.process(element.clauses)
+          return "coalesce(%s)" % compiler.process(element.clauses, **kw)
 
       @compiles(coalesce, 'oracle')
       def compile(element, compiler, **kw):
           if len(element.clauses) > 2:
               raise TypeError("coalesce only supports two arguments on Oracle")
-          return "nvl(%s)" % compiler.process(element.clauses)
+          return "nvl(%s)" % compiler.process(element.clauses, **kw)
 
 * :class:`~sqlalchemy.schema.DDLElement` - The root of all DDL expressions,
   like CREATE TABLE, ALTER TABLE, etc. Compilation of ``DDLElement``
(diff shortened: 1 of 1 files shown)


zzzeek:
There's an intent to how it draws out the raw "CASE" string, which is as an example is intended to illustrate that plain strings can be used in these kinds of places.

However, the case() looks nicer and there are other examples of using raw strings, I think it will be fine.

However also, if we are fixing `**kw`, there's five more to fix:
```

classic@photon2 ext]$ grep -nHe 'process(' compiler.py

# add **kw
compiler.py:88:            compiler.process(element.table, asfrom=True),

# add **kw
compiler.py:89:            compiler.process(element.select)

# this one needs to send literal_binds into **kw in case it's already there
compiler.py:124:            ddlcompiler.sql_compiler.process(

# add **kw
compiler.py:268:          return "coalesce(%s)" % compiler.process(element.clauses)

# add **kw
compiler.py:274:          return "nvl(%s)" % compiler.process(element.clauses)

# these are fixed with your existing patch
compiler.py:357:            compiler.process(arg1),
compiler.py:358:            compiler.process(arg2),
compiler.py:359:            compiler.process(arg1),
compiler.py:360:            compiler.process(arg2),

```

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get this PR into gerrit so we can run tests and reviews and stuff

sqla-tester:
Change has been squashed to Gerrit review: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1038

zzzeek (changes requested):
please add **kw for the following sections, thanks!

```
classic@photon2 ext]$ grep -nHe 'process(' compiler.py

# add **kw
compiler.py:88:            compiler.process(element.table, asfrom=True),

# add **kw
compiler.py:89:            compiler.process(element.select)

# this one needs to send literal_binds into **kw in case it's already there
compiler.py:124:            ddlcompiler.sql_compiler.process(

# add **kw
compiler.py:268:          return "coalesce(%s)" % compiler.process(element.clauses)

# add **kw
compiler.py:274:          return "nvl(%s)" % compiler.process(element.clauses)

# these are fixed with your existing patch
compiler.py:357:            compiler.process(arg1),
compiler.py:358:            compiler.process(arg2),
compiler.py:359:            compiler.process(arg1),
compiler.py:360:            compiler.process(arg2),
```

konstantint:
Added extra changes.

I followed your suggestion of passing `ddlcompiler` kwargs to `sql_compiler`'s `process` call blindly on line 124 - I'm not familiar with the details myself to judge whether this is actually consistent.

zzzeek:
> Added extra changes.
> 
> I followed your suggestion of passing `ddlcompiler` kwargs to `sql_compiler`'s `process` call blindly on line 124 - I'm not familiar with the details myself to judge whether this is actually consistent.

it should be fine as usually build-in DDL compiler methods don't have any **kw anyway, e.g. the kw here would usually be blank, we just have **kw to future proof it.   Definitely the various kwargs that are meaningful to SQLCompiler vs. hypothetical kwargs that would be meaningful directly inside of DDLCompiler (I don't think the latter exists right now) should be kept free of duplication so that either side can safely consume the same set of kwargs.

lib/sqlalchemy/ext/compiler.py
```diff
@@ -85,8 +85,8 @@ def __init__(self, table, select):
     @compiles(InsertFromSelect)
     def visit_insert_from_select(element, compiler, **kw):
         return "INSERT INTO %s (%s)" % (
-            compiler.process(element.table, asfrom=True),
-            compiler.process(element.select)
+            compiler.process(element.table, asfrom=True, **kw),
```
zzzeek: there's a thing that can happen sometimes with this which is that the kw would already have "asfrom" inside of it, and this would raise an error.  however insertfromselect in this example would not be embedded inside a FROM clause itself so we can keep this as is.  we would not want asfrom=True for the element.select visit either.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 24ee93f63e21fccae6cbc1cc1c154dd56f1e1963 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 24ee93f63e21fccae6cbc1cc1c154dd56f1e1963 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1038

zzzeek (approved):

sqla-tester:

Code-Review+2

sqla-tester:

Workflow+1

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1038 has been **merged**. Congratulations! :)

PR #4488, opened 2019-02-08
Add type stubs link to the docs

ilevkivskyi:
### Description

This is a follow-up on https://github.com/sqlalchemy/sqlalchemy/issues/4315

This PR adds a link to type stubs for SQLAlchemy to the docs. The package also includes a [mypy](https://github.com/python/mypy) plugin for more precise type checking. The project is still an early alpha but may be already interesting for some people.

### Checklist

This pull request is:

### DIFF
--- a/doc/build/core/type_basics.rst
+++ b/doc/build/core/type_basics.rst
@@ -26,6 +26,9 @@ Functions that accept a type (such as :func:`~sqlalchemy.schema.Column`) will
 typically accept a type class or instance; ``Integer`` is equivalent
 to ``Integer()`` with no construction arguments in this case.
 
+If you are using PEP 484 static type checkers for Python, you can use
+`type stubs for SQLAlchemy (alpha) <https://github.com/dropbox/sqlalchemy-stubs>`_.
+
 .. _types_generic:
 
 Generic Types


zzzeek:
wow that is pretty amazing.    I took a deep dive into mypy when I reworked my [publishthing](https://github.com/sqlalchemyorg/publishthing) tool, and I used explicit type annotations and tried to get as far as possible with mypy strict.     Clearly this is the future of Python and I've started the cognitive work on what Python3-only SQLAlchemy will look like....I feel like skipping the py2 compatible annotations and going right for the full syntax.   It's going to be an immense job because even on a small tool like publishthing, it was a lot of work.

but yes I can see this plugin is specific to the SQLAlchemy typing system, seems like it should be more prominent but let's start with it there for now.  thanks!

zzzeek:
I'm going to try making it a sidebar, maybe that will look more spiffy

zzzeek:
is this plugin ORM-specific ?   maybe it should be mentioned where ORM models are defined.

zzzeek:
it looks like it's specific to Declarative.  going to add the note over there for now.

zzzeek:
b51f045ee555ecf5e9e3ee7383d00d5d52ca3f39 thanks!

ilevkivskyi:
> is this plugin ORM-specific ? maybe it should be mentioned where ORM models are defined.

It currently only supports precise types for ORM models, but there are plans to add better support for `Table(...)` definitions, for example currently `users.c.name` is just `Column[Any]`, we are planning to  infer `Column[str]` in future versions.

I was not sure where to add the link, thanks for finding a better place in the docs!

PR #4514, opened 2019-02-23
Add Pipenv* and .idea to .gitignore

villebro:
I started working on the `text().expressions()` feature and noticed that `.gitignore` doesn't include `.idea`(PyCharm) or `Pipfile*` (Pipenv). I assume these are fairly commonly used tools, and should be ignored by default.


zzzeek:
thanks!

PR #4520, opened 2019-03-04
Add missing attribute in hybrid.py docs

daaain:
### Description

Just adding a missing attribute in the docs which made me scratch my head for a few minutes!

I might have misunderstood something, but using this:

```python
@Charge.overrides.expression
    def user_id(cls):
```

Gives me this error:

```sh
    @Charge.overrides.expression
E   AttributeError: type object 'Charge' has no attribute 'overrides'
```

But if I add the property it's fine:

```python
@Charge.user_id.overrides.expression
    def user_id(cls):
```

Just for the sake of completeness (in case it matters), I also have a getter in the class before that:

```python
@Charge.user_id.getter
    def user_id(self):
```

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/ext/hybrid.py
+++ b/lib/sqlalchemy/ext/hybrid.py
@@ -509,7 +509,7 @@ reference the instrumented attribute back to the hybrid object::
 
         last_name = Column(String)
 
-        @FirstNameOnly.overrides.expression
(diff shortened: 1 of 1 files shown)


zzzeek:
ah yes.  thanks!

PR #4568, opened 2019-03-25
CheckIn and checkouts not displaying on set echo_pool=True. Look…

jaskirat1208:
As mentioned in SQLAlchemy docs, 
'sqlalchemy.pool' - controls connection pool logging. set to logging.INFO or lower to log connection pool checkouts/checkins.
However, the checkouts and checkins to pool were not shown when I set echoflag to True. As in the source code, changed logging.debug to logging.info so that it could match what documentation suggested.

### DIFF
--- a/lib/sqlalchemy/pool/base.py
+++ b/lib/sqlalchemy/pool/base.py
@@ -495,7 +495,7 @@ class _ConnectionRecord(object):
         )
         _refs.add(rec)
         if echo:
-            pool.logger.debug(
+            pool.logger.info(
                 "Connection %r checked out from pool", dbapi_connection
             )
         return fairy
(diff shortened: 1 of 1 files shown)


zzzeek:
Hi there -

Please create issues for these things so we can discuss them *before* writing pull requests.   As stated in the guidelines above, we do not accept "one line" pull requests without tests and without a corresponding issue describing what the bug or feature is.

In this case, the bug here is the documentation is wrong, not the code.   The "INFO" level logs pool invalidations and similar connection issues" "DEBUG" logs all checkins and checkouts.  In addition, the create_engine.echo_pool flag accepts the string "debug" as does Pool.echo, in the same manner as the engine.echo flag.   These are all documentation issues.    Please re-submit as an issue so we can track this.  If you prefer to implement a documentaion PR, files to edit include doc/build/core/engines.rst, lib/sqlalchemy/engine/__init__.py, lib/sqlalchemy/pool/base.py.  Thanks!

PR #4570, opened 2019-03-25
Add case insensitivity feature to GenericFunction.

### LINKED ISSUE #4569: GenericFunction register is case-sensitive
The `GenericFunction` instances are registered in `sqlalchemy.sql.functions._register`, which is a `collections.defaultdict(dict)`. The `sqlalchemy.sql.functions.func()` function uses the `get()` method to retrieved registered functions, which means that the search is case-sensitive.

This case-sensitivity can have some drawbacks, e.g.:

```python
from sqlalchemy import func
from sqlalchemy.sql.functions import GenericFunction
from sqlalchemy.types import Text, Integer

class CaseSensitiveFunc(GenericFunction):
    type = Text

class CaseInsensitiveFunc(GenericFunction):
    type = Integer

print(func.CaseSensitiveFunc().type, func.CaseInsensitiveFunc().type)  # Out: (Text(), Integer())
print(func.CaSeSeNsItIveFuNc().type, func.CaSeInSeNsItIveFuNc().type)  # Out: (NullType(), NullType())
```
In this example, the `GenericFunction(s)` are not retrieved by the `sqlalchemy.sql.functions.func()` function when their letter case is not exactly the same as in their definitions. So in this case, they are just sent to the DBMS as they are and the DBMS silently interpret them without considering the letter case, which can lead to some weird bugs (see here for example: https://github.com/geoalchemy/geoalchemy2/issues/215).

Some simple changes in `sqlalchemy.sql.functions` could add the possibility to register a `GenericFunction` without being case-sensitive. In this case we would have the following:

```python
from sqlalchemy import func
from sqlalchemy.sql.functions import GenericFunction
from sqlalchemy.types import Text, Integer

class CaseSensitiveFunc(GenericFunction):
    type = Text
    case_sensitive = True

class CaseInsensitiveFunc(GenericFunction):
    type = Integer
    case_sensitive = False

print(func.CaseSensitiveFunc().type, func.CaseInsensitiveFunc().type)  # Out: (Text(), Integer())
print(func.CaSeSeNsItIveFuNc().type, func.CaSeInSeNsItIveFuNc().type)  # Out: (NullType(), Integer())
```

Here the case-insensitive function is retrieved by `sqlalchemy.sql.functions.func()`.

adrien-berchet:
### Description
Fixes: #4569

Following this discussion: https://github.com/geoalchemy/geoalchemy2/pull/216
The goal of the PR is to allow the possibility to register case-insensitive `GenericFunction(s)`.
By default, the case-sensitivity is kept in order to keep the current behavior.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/sql/functions.py
+++ b/lib/sqlalchemy/sql/functions.py
@@ -8,6 +8,9 @@
 """SQL function API, factories, and built-in functions.
 
 """
+from functools import partial
+import warnings
+
 from . import annotation
 from . import operators
 from . import schema
@@ -33,15 +36,17 @@ from .selectable import Alias
 from .selectable import FromClause
 from .selectable import Select
 from .visitors import VisitableType
+from .. import exc as sa_exc
 from .. import util
 
 
+_RegistryList = partial(util.defaultdict, list)
+
 _registry = util.defaultdict(dict)
-_case_insensitive_functions = util.defaultdict(dict)
+_case_sensitive_reg = util.defaultdict(_RegistryList)
 
 
-def register_function(identifier, fn, package="_default",
-                      case_sensitive=True):
+def register_function(identifier, fn, package="_default"):
     """Associate a callable with a particular func. name.
 
     This is normally called by _GenericMeta, but is also
@@ -51,11 +56,37 @@ def register_function(identifier, fn, package="_default",
 
     """
     reg = _registry[package]
-    reg[identifier] = fn
-    if not case_sensitive:
-        _case_insensitive_functions[package][identifier.lower()] = identifier
-    elif identifier.lower() in _case_insensitive_functions[package]:
-        del _case_insensitive_functions[package][identifier.lower()]
+    raw_identifier = identifier
+    identifier = identifier.lower()
+
+    # Check if a function with the same lowercase identifier is registered.
+    if identifier in _case_sensitive_reg[package]:
+        if (
+            raw_identifier not in _case_sensitive_reg[package][identifier]
+        ):
+            warnings.warn(
+                "GenericFunction(s) '{}' are already registered with "
+                "different letter cases and might interact with {}.".format(
+                    _case_sensitive_reg[package][identifier],
+                    raw_identifier),
+                sa_exc.SADeprecationWarning)
+        else:
+                warnings.warn(
+                    "The GenericFunction '{}' is already registered and "
+                    "is going to be overriden.".format(identifier))
+
+        # If a function with the same lowercase identifier is registered, then
+        # these 2 functions are considered as case-sensitive.
+        if len(_case_sensitive_reg[package][identifier]) == 1:
+            old_fn = reg[identifier]
+            del reg[identifier]
+            reg[_case_sensitive_reg[package][identifier][0]] = old_fn
+        reg[raw_identifier] = fn
+    else:
+        reg[identifier] = fn
+
+    if raw_identifier not in _case_sensitive_reg[package][identifier]:
+        _case_sensitive_reg[package][identifier].append(raw_identifier)
 
 
 class FunctionElement(Executable, ColumnElement, FromClause):
@@ -446,8 +477,12 @@ class _FunctionGenerator(object):
             package = None
 
         if package is not None:
-            func = _registry[package].get(fname) or _registry[package].get(
-                _case_insensitive_functions[package].get(fname.lower()))
+            if (
+                len(_case_sensitive_reg[package].get(fname.lower(), [])) > 1
+            ):
+                func = _registry[package].get(fname)
+            else:
+                func = _registry[package].get(fname.lower())
             if func is not None:
                 return func(*c, **o)
 
@@ -577,13 +612,11 @@ class _GenericMeta(VisitableType):
         if annotation.Annotated not in cls.__mro__:
             cls.name = name = clsdict.get("name", clsname)
             cls.identifier = identifier = clsdict.get("identifier", name)
-            cls.case_sensitive = case_sensitive = clsdict.get(
-                "case_sensitive", True)
             package = clsdict.pop("package", "_default")
             # legacy
             if "__return_type__" in clsdict:
                 cls.type = clsdict["__return_type__"]
-            register_function(identifier, cls, package, case_sensitive)
+            register_function(identifier, cls, package)
         super(_GenericMeta, cls).__init__(clsname, bases, clsdict)
 
 
--- a/lib/sqlalchemy/sql/functions.py
+++ b/lib/sqlalchemy/sql/functions.py
@@ -60,7 +60,7 @@ def register_function(identifier, fn, package="_default"):
     identifier = identifier.lower()
 
     # Check if a function with the same lowercase identifier is registered.
-    if identifier in _registry[package]:
+    if identifier in _case_sensitive_reg[package]:
         if (
             raw_identifier not in _case_sensitive_reg[package][identifier]
         ):
--- a/lib/sqlalchemy/sql/functions.py
+++ b/lib/sqlalchemy/sql/functions.py
@@ -60,7 +60,22 @@ def register_function(identifier, fn, package="_default"):
     identifier = identifier.lower()
 
     # Check if a function with the same lowercase identifier is registered.
-    if case_sensitive_reg[identifier]:
+    if identifier in reg:
+        if raw_identifier in case_sensitive_reg[identifier]:
+            warnings.warn(
+                "The GenericFunction '{}' is already registered and "
+                "is going to be overriden.".format(identifier),
+                sa_exc.SAWarning)
+            reg[identifier] = fn
+        else:
+            # If a function with the same lowercase identifier is registered,
+            # then these 2 functions are considered as case-sensitive.
+            # Note: This case should raise an error in a later release.
+            reg.pop(identifier)
+
+    # Check if a function with different letter case identifier is registered.
+    elif identifier in case_sensitive_reg:
+        # Note: This case will be removed in a later release.
         if (
             raw_identifier not in case_sensitive_reg[identifier]
         ):
@@ -69,11 +84,6 @@ def register_function(identifier, fn, package="_default"):
                 "different letter cases and might interact with {}.".format(
                     list(case_sensitive_reg[identifier].keys()),
                     raw_identifier))
-            # If a function with the same lowercase identifier is registered,
-            # then these 2 functions are considered as case-sensitive.
-            if len(case_sensitive_reg[identifier]) == 1:
-                reg.pop(identifier, None)
-                reg.update(case_sensitive_reg[identifier])
 
         else:
             warnings.warn(
@@ -81,13 +91,12 @@ def register_function(identifier, fn, package="_default"):
                 "is going to be overriden.".format(identifier),
                 sa_exc.SAWarning)
 
-        reg[raw_identifier] = fn
+    # Register by default
     else:
         reg[identifier] = fn
 
-    # Add the raw_identifier to the case-sensitive registry
-    if raw_identifier not in case_sensitive_reg[identifier]:
-        case_sensitive_reg[identifier][raw_identifier] = fn
+    # Always register in case-sensitive registry
+    case_sensitive_reg[identifier][raw_identifier] = fn
 
 
 class FunctionElement(Executable, ColumnElement, FromClause):
@@ -480,12 +489,9 @@ class _FunctionGenerator(object):
         if package is not None:
             reg = _registry[package]
             case_sensitive_reg = _case_sensitive_registry[package]
-            if (
-                len(case_sensitive_reg.get(fname.lower(), [])) > 1
-            ):
-                func = reg.get(fname)
-            else:
-                func = reg.get(fname.lower())
+            func = reg.get(fname.lower())
+            if func is None and fname.lower() in case_sensitive_reg:
+                func = case_sensitive_reg[fname.lower()].get(fname)
             if func is not None:
                 return func(*c, **o)
 
--- a/lib/sqlalchemy/sql/functions.py
+++ b/lib/sqlalchemy/sql/functions.py
@@ -71,9 +71,9 @@ def register_function(identifier, fn, package="_default"):
                     raw_identifier),
                 sa_exc.SADeprecationWarning)
         else:
-                warnings.warn(
-                    "The GenericFunction '{}' is already registered and "
-                    "is going to be overriden.".format(identifier))
+            warnings.warn(
+                "The GenericFunction '{}' is already registered and "
+                "is going to be overriden.".format(identifier))
 
         # If a function with the same lowercase identifier is registered, then
         # these 2 functions are considered as case-sensitive.
--- a/lib/sqlalchemy/sql/functions.py
+++ b/lib/sqlalchemy/sql/functions.py
@@ -8,7 +8,6 @@
 """SQL function API, factories, and built-in functions.
 
 """
-from functools import partial
 import warnings
 
 from . import annotation
@@ -40,10 +39,10 @@ from .. import exc as sa_exc
 from .. import util
 
 
-_RegistryList = partial(util.defaultdict, list)
-
 _registry = util.defaultdict(dict)
-_case_sensitive_reg = util.defaultdict(_RegistryList)
+_case_sensitive_registry = util.defaultdict(
+    lambda: util.defaultdict(dict)
+)
 
 
 def register_function(identifier, fn, package="_default"):
@@ -56,25 +55,26 @@ def register_function(identifier, fn, package="_default"):
 
     """
     reg = _registry[package]
+    case_sensitive_reg = _case_sensitive_registry[package]
     raw_identifier = identifier
     identifier = identifier.lower()
 
     # Check if a function with the same lowercase identifier is registered.
-    if identifier in _case_sensitive_reg[package]:
+    if case_sensitive_reg[identifier]:
         if (
-            raw_identifier not in _case_sensitive_reg[package][identifier]
+            raw_identifier not in case_sensitive_reg[identifier]
         ):
             util.warn_deprecated(
-                "GenericFunction(s) '{}' are already registered with "
+                "GenericFunction(s) {} are already registered with "
                 "different letter cases and might interact with {}.".format(
-                    _case_sensitive_reg[package][identifier],
+                    list(case_sensitive_reg[identifier].keys()),
                     raw_identifier))
             # If a function with the same lowercase identifier is registered,
             # then these 2 functions are considered as case-sensitive.
-            if len(_case_sensitive_reg[package][identifier]) == 1:
-                old_fn = reg[identifier]
-                del reg[identifier]
-                reg[_case_sensitive_reg[package][identifier][0]] = old_fn
+            if len(case_sensitive_reg[identifier]) == 1:
+                reg.pop(identifier, None)
+                reg.update(case_sensitive_reg[identifier])
+
         else:
             warnings.warn(
                 "The GenericFunction '{}' is already registered and "
@@ -86,8 +86,8 @@ def register_function(identifier, fn, package="_default"):
         reg[identifier] = fn
 
     # Add the raw_identifier to the case-sensitive registry
-    if raw_identifier not in _case_sensitive_reg[package][identifier]:
-        _case_sensitive_reg[package][identifier].append(raw_identifier)
+    if raw_identifier not in case_sensitive_reg[identifier]:
+        case_sensitive_reg[identifier][raw_identifier] = fn
 
 
 class FunctionElement(Executable, ColumnElement, FromClause):
@@ -478,12 +478,14 @@ class _FunctionGenerator(object):
             package = None
 
         if package is not None:
+            reg = _registry[package]
+            case_sensitive_reg = _case_sensitive_registry[package]
             if (
-                len(_case_sensitive_reg[package].get(fname.lower(), [])) > 1
+                len(case_sensitive_reg.get(fname.lower(), [])) > 1
             ):
-                func = _registry[package].get(fname)
+                func = reg.get(fname)
             else:
-                func = _registry[package].get(fname.lower())
+                func = reg.get(fname.lower())
             if func is not None:
                 return func(*c, **o)
 
--- a/lib/sqlalchemy/sql/functions.py
+++ b/lib/sqlalchemy/sql/functions.py
@@ -93,7 +93,7 @@ def register_function(identifier, fn, package="_default"):
                 "different letter cases and might interact with '{}'. "
                 "GenericFunction objects will be fully case-insensitive in a "
                 "future release.".format(
-                    list(case_sensitive_reg[identifier].keys()),
+                    sorted(case_sensitive_reg[identifier].keys()),
                     raw_identifier))
 
         else:
--- a/lib/sqlalchemy/sql/functions.py
+++ b/lib/sqlalchemy/sql/functions.py
@@ -60,31 +60,32 @@ def register_function(identifier, fn, package="_default"):
     identifier = identifier.lower()
 
     # Check if a function with the same lowercase identifier is registered.
-    if identifier in _case_sensitive_reg[package]:
+    if identifier in _registry[package]:
         if (
             raw_identifier not in _case_sensitive_reg[package][identifier]
         ):
-            warnings.warn(
+            util.warn_deprecated(
                 "GenericFunction(s) '{}' are already registered with "
                 "different letter cases and might interact with {}.".format(
                     _case_sensitive_reg[package][identifier],
-                    raw_identifier),
-                sa_exc.SADeprecationWarning)
+                    raw_identifier))
+            # If a function with the same lowercase identifier is registered,
+            # then these 2 functions are considered as case-sensitive.
+            if len(_case_sensitive_reg[package][identifier]) == 1:
+                old_fn = reg[identifier]
+                del reg[identifier]
+                reg[_case_sensitive_reg[package][identifier][0]] = old_fn
         else:
             warnings.warn(
                 "The GenericFunction '{}' is already registered and "
-                "is going to be overriden.".format(identifier))
+                "is going to be overriden.".format(identifier),
+                sa_exc.SAWarning)
 
-        # If a function with the same lowercase identifier is registered, then
-        # these 2 functions are considered as case-sensitive.
-        if len(_case_sensitive_reg[package][identifier]) == 1:
-            old_fn = reg[identifier]
-            del reg[identifier]
-            reg[_case_sensitive_reg[package][identifier][0]] = old_fn
         reg[raw_identifier] = fn
     else:
         reg[identifier] = fn
 
+    # Add the raw_identifier to the case-sensitive registry
     if raw_identifier not in _case_sensitive_reg[package][identifier]:
         _case_sensitive_reg[package][identifier].append(raw_identifier)
 
--- a/lib/sqlalchemy/sql/functions.py
+++ b/lib/sqlalchemy/sql/functions.py
@@ -8,8 +8,6 @@
 """SQL function API, factories, and built-in functions.
 
 """
-import warnings
-
 from . import annotation
 from . import operators
 from . import schema
@@ -35,7 +33,6 @@ from .selectable import Alias
 from .selectable import FromClause
 from .selectable import Select
 from .visitors import VisitableType
-from .. import exc as sa_exc
 from .. import util
 
 
@@ -43,6 +40,10 @@ _registry = util.defaultdict(dict)
 _case_sensitive_registry = util.defaultdict(
     lambda: util.defaultdict(dict)
 )
+_CASE_SENSITIVE = util.symbol(
+    name="case_sensitive_function",
+    doc="Symbol to mark the functions that are switched into case-sensitive "
+    "mode.")
 
 
 def register_function(identifier, fn, package="_default"):
@@ -60,18 +61,26 @@ def register_function(identifier, fn, package="_default"):
     identifier = identifier.lower()
 
     # Check if a function with the same lowercase identifier is registered.
-    if identifier in reg:
+    if identifier in reg and reg[identifier] is not _CASE_SENSITIVE:
         if raw_identifier in case_sensitive_reg[identifier]:
-            warnings.warn(
+            util.warn(
                 "The GenericFunction '{}' is already registered and "
-                "is going to be overriden.".format(identifier),
-                sa_exc.SAWarning)
+                "is going to be overriden.".format(identifier))
             reg[identifier] = fn
         else:
             # If a function with the same lowercase identifier is registered,
             # then these 2 functions are considered as case-sensitive.
             # Note: This case should raise an error in a later release.
-            reg.pop(identifier)
+            util.warn_deprecated(
+                "GenericFunction '{}' is already registered with "
+                "different letter case, so the previously registered function "
+                "'{}' is switched into case-sensitive mode. "
+                "GenericFunction objects will be fully case-insensitive in a "
+                "future release.".format(
+                    raw_identifier,
+                    list(case_sensitive_reg[identifier].keys())[0],
+                ))
+            reg[identifier] = _CASE_SENSITIVE
 
     # Check if a function with different letter case identifier is registered.
     elif identifier in case_sensitive_reg:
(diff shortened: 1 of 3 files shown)


zzzeek:
looks very nice, let's run on CI

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 0a8b385270d4009996c56cecb5a956bfe2f7357e of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 0a8b385270d4009996c56cecb5a956bfe2f7357e: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1178

adrien-berchet:
Tests were failing because they are run in parallel in the CI. I forced the `sqlalchemy.sql.functions._registry` to be reset at each method teardown in order to ensure that this registry is in the proper state for each test.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 9c84dc82b8937dcbd9219ea6601820cf5402be5d of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 9c84dc82b8937dcbd9219ea6601820cf5402be5d added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1178

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 60f2da20902258573c5b50bd0ecefa256c0b24c8 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 60f2da20902258573c5b50bd0ecefa256c0b24c8 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1178

zzzeek:
I wrote a tool to do the imports called [zimports](https://github.com/sqlalchemyorg/zimports), all you do is install it, cd into the sqlalchemy checkout and type "zimports".  or give it the filenames you care about.  I dont think anyone uses it but it works pretty great over here.

adrien-berchet:
Nice tool, thx :)

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision f64e7afcc392ff72a5a8be79bc84dce100a38bee of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset f64e7afcc392ff72a5a8be79bc84dce100a38bee added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1178

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 537fa192fa77f979fe2502bacf8809b20c29b9c7 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 537fa192fa77f979fe2502bacf8809b20c29b9c7 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1178

zzzeek:
keep in mind im just waiting for CI to be done i haven't looked at this yet

adrien-berchet:
While I was testing this branch with my other works, I realized that the `sqlalchemy.sql.functions.GenericFunction` class was registered in `sqlalchemy.sql.functions._register` (I realized this because now the `geoalchemy2` package raises a warning at import time since it also defines a class `GenericFunction` which inherits from `sqlalchemy.sql.functions.GenericFunction`). This behavior seems like an unintended result to me since calling `func.GenericFunction()` has no sens (I tried to remove it and all the tests are still passing). What do you think about it? Should we remove it from `sqlalchemy.sql.functions._register`?

adrien-berchet:
> keep in mind im just waiting for CI to be done i haven't looked at this yet

Yup. I just found a small bug while I was testing it :)

zzzeek:
> While I was testing this branch with my other works, I realized that the `sqlalchemy.sql.functions.GenericFunction` class was registered in `sqlalchemy.sql.functions._register` (I realized this because now the `geoalchemy2` package raises a warning at import time since it also defines a class `GenericFunction` which inherits from `sqlalchemy.sql.functions.GenericFunction`). This behavior seems like an unintended result to me since calling `func.GenericFunction()` has no sens (I tried to remove it and all the tests are still passing). What do you think about it? Should we remove it from `sqlalchemy.sql.functions._register`?

I don't see the word "_register" anywhere in the sqlalchemy.sql package at all so I'm not sure what we are referring towards.

Are you referring to `_registry` ?  That's how the `GenericFunction` classes we make are located as part of `func.XYZ()`.   That is the intended result.  It's a result of the metaclass that is part of GenericFunction's base.     Are you saying, there's a `func.GenericFunction` being generated?   OK, the metaclass should probably try to be smart about that to skip the base class itself from being in the registry.

adrien-berchet:
> Are you referring to `_registry` ?

Sorry for the typo, I was indeed referring to `_registry`.

> Are you saying, there's a func.GenericFunction being generated?

Yes

> OK, the metaclass should probably try to be smart about that to skip the base class itself from being in the registry.

I also think so.
Should I try to include a fix in this PR? Though I think this is not directly related to this PR (it just pointed it out) and I am not used to these metaclass things...

zzzeek:
don't include it if it's not breaking things right now, no.  we can addrss that separately and thanks for the help!

zzzeek:
by "breaking things" i mean blocking what we are trying to do here.

adrien-berchet:
Ok
As far as I can see, it does not break anything. The only issue I noticed is the warning raised when importing `geoalchemy2`. But it is no big deal.

sqla-tester:

is it ready that I should find time to look at this?

adrien-berchet:
Yes, this time it looks good to me, at least functionally.

sqla-tester (commented):

(9 comments)

OK I have a bunch of thoughts, and I'm not 100% sure I'm getting this right.   The main thing I want to try to figure out, if possible is, no "len()" checks, they are less peformant than bools and also logic that is comparing "len() == 1" in order to determine a future mutation is going to be subject to race conditions.   So I've proposed changing the datastructure, but I haven't spent time testing it or anything so it could be off.

lib/sqlalchemy/sql/functions.py
```diff
@@ -33,10 +36,14 @@
 from .selectable import FromClause
 from .selectable import Select
 from .visitors import VisitableType
+from .. import exc as sa_exc
 from .. import util
 
 
+_RegistryList = partial(util.defaultdict, list)
```
sqla-tester: **mike bayer** (zzzeek) wrote:

I would want this to be a set() or a dict(), so that we don't need to test for containment before calling append() as we do with a list.  I was going to suggest a set() but as I look at the registry logic below I think it might be simpler to make it a dict that actually stores the function objects as well, to ease the transition into case sensitive mode.

lib/sqlalchemy/sql/functions.py
```diff
@@ -33,10 +36,14 @@
...
+
 _registry = util.defaultdict(dict)
+_case_sensitive_reg = util.defaultdict(_RegistryList)
```
sqla-tester: **mike bayer** (zzzeek) wrote:

Let's call this "_case_sensitive_registry" for some naming reasons below.

Also with nested defaultdicts I just name them all at once for clarity:

   _case_sensitive_registry = util.defaultdict(
        lambda: util.defaultdict(dict)
    )

it looks like functools.partial is a C implementation now but the Python version illustrates there is a lot of overhead to calling it, especially for what it is.

lib/sqlalchemy/sql/functions.py
```diff
@@ -49,7 +56,38 @@ def register_function(identifier, fn, package="_default"):
 
     """
     reg = _registry[package]
-    reg[identifier] = fn
+    raw_identifier = identifier
+    identifier = identifier.lower()
+
+    # Check if a function with the same lowercase identifier is registered.
+    if identifier in _case_sensitive_reg[package]:
```
sqla-tester: **mike bayer** (zzzeek) wrote:

so all the _case_sensitive_reg[package] lookups lets do them just once above, see where it says "reg = _registry[package]", add another one, "case_sensitive_reg = _case_sensitive_registry[package]" (this is why I went with "registry" and not "reg" above).   that should reduce overhead / verbosity in the rest of this.

lib/sqlalchemy/sql/functions.py
```diff
@@ -49,7 +56,38 @@ def register_function(identifier, fn, package="_default"):
...
+            raw_identifier not in _case_sensitive_reg[package][identifier]
+        ):
+            util.warn_deprecated(
+                "GenericFunction(s) '{}' are already registered with "
+                "different letter cases and might interact with {}.".format(
+                    _case_sensitive_reg[package][identifier],
+                    raw_identifier))
+            # If a function with the same lowercase identifier is registered,
+            # then these 2 functions are considered as case-sensitive.
+            if len(_case_sensitive_reg[package][identifier]) == 1:
```
sqla-tester: **mike bayer** (zzzeek) wrote:

the len() here makes me uncomfortable for two reasons.  One is, it's much better if we can do boolean checks, like, "if collection:", rather than "if len(collection) == 1".   The other is, race conditions / thread safety - this code is not threadsafe because the length can grow while we are still checking it for length one, or two threads can try to call "del" at the same time.

So when I originally suggested a set(), I see that doesn't work because this is considering the first element in the list to be the official "this was the case insensitive version".   I think there's less chance of race complexity if we just mirror all the case-sensitive names/ functions here, so that the transfer to "_registry" with case sensitive can be done idempotently.

if we store all the fns in both registries, then this logic should be able to be:

   case_sensitive_reg = _case_sensitive_registry[package]

   # ...

   if case_sensitive_reg[identifier]:
        # remove case insensitive name
        old_fn = reg.pop(identifier, None)

        # populate with all case sensitive names/ fns
        reg.update(case_sensitive_reg[identifier])

do I have that right?   I've revised my thinking here several times so I could be in a fog by now.

lib/sqlalchemy/sql/functions.py
```diff
@@ -49,7 +56,38 @@ def register_function(identifier, fn, package="_default"):
...
+            warnings.warn(
+                "The GenericFunction '{}' is already registered and "
+                "is going to be overriden.".format(identifier),
+                sa_exc.SAWarning)
+
+        reg[raw_identifier] = fn
+    else:
+        reg[identifier] = fn
+
+    # Add the raw_identifier to the case-sensitive registry
```
sqla-tester: **mike bayer** (zzzeek) wrote:

so if this is a dict, it becomes:

    _case_sensitive_reg[[package][identifier][raw_identifier] = fn

lib/sqlalchemy/sql/functions.py
```diff
@@ -440,7 +478,12 @@ def __call__(self, *c, **kwargs):
             package = None
 
         if package is not None:
-            func = _registry[package].get(fname)
+            if (
+                len(_case_sensitive_reg[package].get(fname.lower(), [])) > 1
```
sqla-tester: **mike bayer** (zzzeek) wrote:

does this have to be len > 1 ?  this is a more critical spot.

can we simply look in _registry[package] for the exact name match first, and then if not found fall back to lower()?  because the majority of use cases will be all-lowercase use and there should be no second check:

  reg = _registry[package]
  func = reg.get(fname, None)
  if func is None:
      func = reg.get(fname.lower())
  if func is not None:
       # ...

let me know if that works

test/sql/test_deprecations.py
```diff
@@ -17,19 +23,31 @@
...
 from sqlalchemy.testing import eq_
 from sqlalchemy.testing import fixtures
 from sqlalchemy.testing import mock
+from sqlalchemy.testing.assertions import expect_warnings
 
 
 class DeprecationWarningsTest(fixtures.TestBase):
     __backend__ = True
 
+    def setup_method(self):
```
sqla-tester: **mike bayer** (zzzeek) wrote:

hmmm is setup_method / teardown_method a pytest thing?  I think the test suite uses setup() / teardown() across the board so we should keep consistent.

test/sql/test_deprecations.py
```diff
@@ -135,6 +153,105 @@ def test_table_useexisting(self):
...
+
+        class MYFUNC(GenericFunction):
+            type = DateTime
+
+        assert isinstance(func.MYFUNC().type, DateTime)
+        assert isinstance(func.MyFunc().type, DateTime)
+        assert isinstance(func.mYfUnC().type, DateTime)
+        assert isinstance(func.myfunc().type, DateTime)
+
+        assert "myfunc" in reg['_default']
```
sqla-tester: **mike bayer** (zzzeek) wrote:

nit: for all the "in" / "not in" we have little functions in_(), not_in_() which can be imported from "sqlalchemy.testing".   We should probably have isinstance / not_isinstance too.    There's a not-well-attended to convention that we don't want to use "assert " in test suites even though yes it is all over the place.

test/sql/test_deprecations.py
```diff
@@ -135,6 +153,105 @@ def test_table_useexisting(self):
...
+        with pytest.raises(AssertionError):
+            assert isinstance(func.mYfUnC().type, Integer)
+        with pytest.raises(AssertionError):
+            assert isinstance(func.myfunc().type, Integer)
+
+        assert "myfunc" not in reg['_default']
+        assert "MYFUNC" in reg['_default']
+        assert "MyFunc" in reg['_default']
+        assert "myfunc" in cs_reg['_default']
+        assert cs_reg['_default']['myfunc'] == ['MYFUNC', 'MyFunc']
```
sqla-tester: **mike bayer** (zzzeek) wrote:

for this one we use sqalchemy.testing.eq_(), that one is pretty consistent.

zzzeek:
also it looks like I need to improve code formatting in my gerrit <-> github integration thing, my apologies for the code not being formatted correctly in some of these

lib/sqlalchemy/sql/functions.py
adrien-berchet: I didn't want to duplicate the functions so I prefered to just store the names mapping. But you are right, it will be easier to make the future transition. I will update that to use a `dict()`.

lib/sqlalchemy/sql/functions.py
adrien-berchet: Ok for the name.

I was not aware of this overhead, thanks for the details.

lib/sqlalchemy/sql/functions.py
adrien-berchet: You are right, it will be far better.

lib/sqlalchemy/sql/functions.py
adrien-berchet: I also don't really like this `len()` but I did not figure out how I could achieve the same result with another way, except by writing a specific class to handle the function registry or by using a third registry with booleans to tag the case-sensitve and case-incensitive identifiers. Because here, the `len(...) == 1` means 'One (and only one) function with the given `identifier` was already registered and now we want to register a second one with a different letter case, so we must switch to case-sensitive mode'. But after that this second function have been registered, all the next functions with the same `identifier` must be considered as case-sensitive. If we just look if `identifier` is in `case_sensitive_reg`, it does not mean there is only one function registered, so we miss something. For example, if we use your snippet to register `MyFunc`, then `myfunc`, then `MYFUNC`, then the last one will pop the second one, which is not the intended result.

Sorry if I am not clear to describe this.

lib/sqlalchemy/sql/functions.py
adrien-berchet: Ok

test/sql/test_deprecations.py
adrien-berchet: Yes, this is a `pytest` feature to execute the setup and teardown methods for each test method, not only at the beginning and the end of the test class. I needed that because the content of the registry at the beginning of a given test (e.g. `test_deprecated::test_case_sensitive`) was dependant of the number of threads used for running the tests. This is not because of thread safety, it is just because the previous tests were updating the registry, so the tests were actually mutually dependant.

test/sql/test_deprecations.py
adrien-berchet: Ok, I will update that.

test/sql/test_deprecations.py
adrien-berchet: Ok

lib/sqlalchemy/sql/functions.py
adrien-berchet: Here again, the `len()` thing is just to detect if this identifier is in case-sensitive or case-insensitive mode. If we always get `fname` first, then we break things.

For example: if `myfunc` and `MyFunc` are registered (so the identifier `myfunc` is switched to case-sensitive mode and is associated with the 2 case-sensitive identifiers `myfunc` and `MyFunc`), then `func.MYFUNC` will actually return the function registered with the `myfunc` case-sensitive identifier instead of an unregistered `Function`.

adrien-berchet:
I pushed a new commit that fixes most of your comments. The main pending point is the `len()` thing...
I may have an idea to deal with it:
* when a function is registered with an unknown identifier, it is only registered in `_registry` (not in `_case_sensitive_registry`).
* when a second function must be registered with a known identifier with different letter case, we must be switched to case-sensitive mode. In this case, we completely remove this identifier from `_registry` and add a new entry into `_case_sensitive_registry`.
* in the `func()` method, we can just try to get it from `_registry`, then from `_case_sensitive_registry` if is was not found and then returning a unregistered `function` if is was still not found.

This way we get ride of `len()` calls. What do you think about it? I can try it this week.

adrien-berchet:
After some thoughts I realized it would not be so long to try it so I pushed another commit to test what I described. I just had to change the first point in order to keep the `raw_identifier` of the first registered function.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 655aa8007478162a2c71d089f9a6f6cb7b9562a9 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 655aa8007478162a2c71d089f9a6f6cb7b9562a9 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1178

test/sql/test_deprecations.py
zzzeek: that's what setup() and teardown() do.   setup() and teardown() are per test.  setup_class() and teardown_class() are per test class.  Within py.test this might be part of its "nose" compatibility, SQLAlchemy just dropped nose support in 1.3.    But in any case, any future changes to test suite idioms should be done all at once and I don't want to introduce any pytest-specific idioms (since we did that with Nose, and then suddenly Nose decided to go away, so never again with that).

adrien-berchet: Ok, `setup()` and `teardown()` seems to work but not `tear_down()`. So if I just change `tear_down()` it seems ok.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision ea8a189943ad7bb9c40f7ed9c214b178103fe800 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset ea8a189943ad7bb9c40f7ed9c214b178103fe800 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1178

lib/sqlalchemy/sql/functions.py
```diff
@@ -49,7 +55,48 @@ def register_function(identifier, fn, package="_default"):
 
     """
     reg = _registry[package]
```
sqla-tester: **mike bayer** (zzzeek) wrote:

this looks GREAT, way simpler than the previous version and won't break under concurrency.

lib/sqlalchemy/sql/functions.py
```diff
@@ -49,7 +55,48 @@ def register_function(identifier, fn, package="_default"):
...
-    reg[identifier] = fn
+    case_sensitive_reg = _case_sensitive_registry[package]
+    raw_identifier = identifier
+    identifier = identifier.lower()
+
+    # Check if a function with the same lowercase identifier is registered.
+    if identifier in reg:
```
sqla-tester: **mike bayer** (zzzeek) wrote:

If we can try the "_CASE_SENSITIVE" symbol idea in [2] below, this line would become:

    if identifier in reg and identifier[reg] is not _CASE_SENSITIVE:

lib/sqlalchemy/sql/functions.py
```diff
@@ -49,7 +55,48 @@ def register_function(identifier, fn, package="_default"):
...
+        if raw_identifier in case_sensitive_reg[identifier]:
+            warnings.warn(
```
sqla-tester: **mike bayer** (zzzeek) wrote:

util.warn() I think (just for consistency)

lib/sqlalchemy/sql/functions.py
```diff
@@ -49,7 +55,48 @@ def register_function(identifier, fn, package="_default"):
...
+                "The GenericFunction '{}' is already registered and "
+                "is going to be overriden.".format(identifier),
+                sa_exc.SAWarning)
+            reg[identifier] = fn
+        else:
+            # If a function with the same lowercase identifier is registered,
+            # then these 2 functions are considered as case-sensitive.
+            # Note: This case should raise an error in a later release.
```
sqla-tester: **mike bayer** (zzzeek) wrote:

I think you mean to have a warning here, see note [1] in the test suite where it should be checking for the warning

lib/sqlalchemy/sql/functions.py
```diff
@@ -49,7 +55,48 @@ def register_function(identifier, fn, package="_default"):
...
+            reg.pop(identifier)
```
sqla-tester: **mike bayer** (zzzeek) wrote:

reg.pop(identifier, None) in case two threads are here after both passing the check.

[2] However, I have an idea which is to add a little more information.   We instead say:

   reg[identifier] = _CASE_SENSITIVE

Where _CASE_SENSITIVE is a util.symbol() defined at the module level.   This gives us a clue to make a quicker decision during the lookup and also is less ambiguous.

lib/sqlalchemy/sql/functions.py
```diff
@@ -49,7 +55,48 @@ def register_function(identifier, fn, package="_default"):
...
+
+    # Check if a function with different letter case identifier is registered.
+    elif identifier in case_sensitive_reg:
+        # Note: This case will be removed in a later release.
+        if (
+            raw_identifier not in case_sensitive_reg[identifier]
+        ):
+            util.warn_deprecated(
+                "GenericFunction(s) {} are already registered with "
+                "different letter cases and might interact with {}.".format(
```
sqla-tester: **mike bayer** (zzzeek) wrote:

..."GenericFunction objects will be fully case-insensitive in a future release."

i've been trying to make all the warnings hit home w/ the users :)

lib/sqlalchemy/sql/functions.py
```diff
@@ -49,7 +55,48 @@ def register_function(identifier, fn, package="_default"):
...
+                    list(case_sensitive_reg[identifier].keys()),
+                    raw_identifier))
+
+        else:
+            warnings.warn(
```
sqla-tester: **mike bayer** (zzzeek) wrote:

util.warn()

lib/sqlalchemy/sql/functions.py
```diff
@@ -440,7 +487,11 @@ def __call__(self, *c, **kwargs):
             package = None
 
         if package is not None:
-            func = _registry[package].get(fname)
+            reg = _registry[package]
```
sqla-tester: **mike bayer** (zzzeek) wrote:

So here, this is a *very* common codepath and it occurs such that an extremely common outcome is that the function is not registered, and we will be creating a new Function object.

The next *very* common codepath is that someone is calling upon one of the existing generic functions like count() and there's only one of them registered.

So I want to reduce the lookups here for these codepaths; if we make use of the _CASE_SENSITIVE symbol in [2] above, I think this can become:


    func = _registry[package].get(fname.lower())
    if func is _CASE_SENSITIVE:
        func = _case_sensitive_registry[package]get(fname)

    if func is not None:
       return func(*c, **o)

That way we don't add any new lookups unless they are using the deprecated / unusual case sensitive path.  does that work?

test/sql/test_deprecations.py
```diff
@@ -135,6 +156,104 @@ def test_table_useexisting(self):
...
+        assert isinstance(func.mYfUnC().type, DateTime)
+        assert isinstance(func.myfunc().type, DateTime)
+
+        in_("myfunc", reg)
+        not_in_("MYFUNC", reg)
+        not_in_("MyFunc", reg)
+        in_("myfunc", cs_reg)
+        eq_(set(cs_reg['myfunc'].keys()), set(['MYFUNC']))
+
+        with testing.expect_deprecated():
```
sqla-tester: **mike bayer** (zzzeek) wrote:

this is note [1], this isn't actually asserting that a deprecation is being emitted unless at least the beginning of the deprecation message is put here; there's no warning actually being triggered right now here.

sqla-tester (commented):

(11 comments)

looks GREAT, I have more ideas.

Let me know if you want me to just take over if this is getting tiring but otherwise feel free to keep on it.

test/sql/test_deprecations.py
```diff
@@ -135,6 +156,104 @@ def test_table_useexisting(self):
...
+        assert isinstance(func.RePlAcEaBlE_fUnC().type, Integer)
+        assert isinstance(func.replaceable_func().type, Integer)
+
+        in_("replaceable_func", reg)
+        not_in_("REPLACEABLE_FUNC", reg)
+        not_in_("Replaceable_Func", reg)
+        in_("replaceable_func", cs_reg)
+        eq_(set(cs_reg['replaceable_func'].keys()), set(['REPLACEABLE_FUNC']))
+
+        with testing.expect_deprecated():
```
sqla-tester: **mike bayer** (zzzeek) wrote:

same here as [1]

test/sql/test_functions.py
```diff
@@ -219,6 +230,34 @@ class myfunc(GenericFunction):
...
+
+        class replacable_func(GenericFunction):
+            type = Integer
+            identifier = 'replacable_func'
+
+        assert isinstance(func.Replacable_Func().type, Integer)
+        assert isinstance(func.RePlAcaBlE_fUnC().type, Integer)
+        assert isinstance(func.replacable_func().type, Integer)
+
+        with expect_warnings():
```
sqla-tester: **mike bayer** (zzzeek) wrote:

the expect_warnings() here should have at least part of the text of the warning that's being emitted to match it up.   At the moment the expect_warnings() context manager is otherwise not asserting that a warning was definitely emitted (this is a bug in the test fixture).

adrien-berchet:
> Let me know if you want me to just take over if this is getting tiring but otherwise feel free to keep on it.

It's ok, these talks are very instructive for me so it is much more enjoyable than tiring :-)
But you would probably be much faster than me so if you prefer you can take it over of course.

lib/sqlalchemy/sql/functions.py
adrien-berchet: Indeed, I forgot this one. I update here and the tests.

lib/sqlalchemy/sql/functions.py
adrien-berchet: Ok

lib/sqlalchemy/sql/functions.py
adrien-berchet: I like it!

lib/sqlalchemy/sql/functions.py
adrien-berchet: I like this design, it is much more clear and explicit. I update it.

lib/sqlalchemy/sql/functions.py
adrien-berchet: Ok

lib/sqlalchemy/sql/functions.py
adrien-berchet: Indeed, with the _CASE_SENSITIVE symbol it is much better.

test/sql/test_deprecations.py
adrien-berchet: Ups, I didn't see that. I updated the `expect_*` statements.

test/sql/test_deprecations.py
adrien-berchet: Ok

test/sql/test_deprecations.py
adrien-berchet: Ups, I didn't see that. I updated the `expect_*` statements.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 0156bfd88a27bc5a740407987271126d70b4afaf of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 0156bfd88a27bc5a740407987271126d70b4afaf added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1178

adrien-berchet:
I am not able to reproduce the failures reported by the CI with Python 2.7. Is it possible to get the docker image used to run the tests please? Or do you have any idea of what is causing this?

sqla-tester:

> adrien-berchet@github wrote:
 > 
 > I am not able to reproduce the failures reported by the CI with
 > Python 2.7. Is it possible to get the docker image used to run the
 > tests please? Or do you have any idea of what is causing this?

yeah that is a very strange occurrence, and also the sqlalchemy2 gerrit seems to be running which it's not supposed to here so ill try to resolve all this today.

sqla-tester:

> > adrien-berchet@github wrote:
 > >
 > > I am not able to reproduce the failures reported by the CI with
 > > Python 2.7. Is it possible to get the docker image used to run
 > the
 > > tests please? Or do you have any idea of what is causing this?
 > 
 > yeah that is a very strange occurrence, and also the sqlalchemy2
 > gerrit seems to be running which it's not supposed to here so ill
 > try to resolve all this today.

yeah it's kind of a worst case scenario problem where the faliure only happens with the complete tox run, not individual py.test runs and even not if I enable xdist, im going to have it print out what it's seeing but yeah this would be very tough for people new to the test suite to figure out...

sqla-tester:

OH.   it's that sqlalchemy2 job getting in and spreading mischief.   sqlalchemy2 is taking out that "lib/" folder.   I thought and thought, ok HOW can this break things.  How can it?  it can't break anything, people install a package, lib/ is gone.   Nothing will break, right?    This breaks!   when your CI checks both out in the same directory and the .pycs hang around.

sqla-tester:

recheck

adrien-berchet:
> OH. it's that sqlalchemy2 job getting in and spreading mischief. sqlalchemy2 is taking out that "lib/" folder. I thought and thought, ok HOW can this break things. How can it? it can't break anything, people install a package, lib/ is gone. Nothing will break, right? This breaks! when your CI checks both out in the same directory and the .pycs hang around.

Oh, that's a tricky CI bug...

sqla-tester (commented):
**Adrien Berchet** (adrien-berchet) wrote:

(2 comments)

Some tests are still failing after your last commit. It seems it is due to a wrong order when keys are transformed into list. Maybe we should just sort the case-sensitive keys?

* test/sql/test_deprecations.py (line 288): Remove one empty line

lib/sqlalchemy/sql/functions.py
```diff
@@ -49,7 +56,57 @@ def register_function(identifier, fn, package="_default"):
...
+        # Note: This case will be removed in a later release.
+        if (
+            raw_identifier not in case_sensitive_reg[identifier]
+        ):
+            util.warn_deprecated(
+                "GenericFunction(s) '{}' are already registered with "
+                "different letter cases and might interact with '{}'. "
+                "GenericFunction objects will be fully case-insensitive in a "
+                "future release.".format(
+                    list(case_sensitive_reg[identifier].keys()),
```
sqla-tester: **Adrien Berchet** (adrien-berchet) wrote:

Maybe we should replace `list` by `sorted` so that the key order would be always the same, whatever the Python version is used. The deprecation test would then be easier to write.

adrien-berchet:
I see that you pushed a new commit on gerrit and some tests are still failing. How can I get these changes from gerrit to my branch to propose a fix?

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 37d4f3322b6bace88c99b959cb1916dbbc57610e of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 37d4f3322b6bace88c99b959cb1916dbbc57610e added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1178

sqla-tester:

sorry this has fallen behind.   I will be getting to this soon.

adrien-berchet:
> sorry this has fallen behind. I will be getting to this soon.

No problem, I will see when you have some time to get back to this :)

adrien-berchet:
Oh, I realize that my last commit discarded the change log you introduced in patch set 9, sorry about that...

sqla-tester:

Code-Review+2 Workflow+1

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1178 has been **merged**. Congratulations! :)

sqla-tester:

Code-Review+2 Workflow+1

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1241 has been **merged**. Congratulations! :)

zzzeek:
great job,thanks for the above and beyond effort on this!

adrien-berchet:
Glad this helps :)
Thanks for you feedback, it was very instructive.
See you in the next PR!

PR #4577, opened 2019-03-27
Raise awareness about partial Ranges support

ticosax:
As soon as we specify the bounds,
the data we fetch from the db is not what we expect

postgresql logs
```
statement: BEGIN
statement: INSERT INTO data_table (range) VALUES (daterange('2013-03-23'::date, '2013-03-24'::date, '()'))
statement: COMMIT
statement: SELECT data_table.range
	FROM data_table

```

test result
```
AssertionError: [(DateRange(empty=True),)] != [(DateRange(datetime.date(2013, 3, 23), datetime.date(2013, 3, 24), '()'),)]

```

Note, at this stage I'm not sure it is sqlalchemy or psycopg2 or myself responsibility

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/test/dialect/postgresql/test_types.py
+++ b/test/dialect/postgresql/test_types.py
@@ -2628,7 +2628,7 @@ class _NumRangeTests(object):
         )
 
 
-class _DateRangeTests(object):
+class _DateRangeDefaultBoundsTests(object):
 
     _col_type = DATERANGE
     _col_str = "DATERANGE"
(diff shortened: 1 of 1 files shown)


zzzeek:
im not sure I understand all the context here but it looks like you are experimenting with some psycopg2 specific objects (e.g.  extras.DateRange is from psycopg2, not SQLAlchemy).   SQLAlchemy just cares about passing through the data correctly I'm not sure what we gain by testing psycopg2.

ticosax:
Hi @zzzeek , I had the same suspicion, but I couldn't confirm it.

This line seems suspicious, I will investigate there.
https://github.com/psycopg/psycopg2/blob/master/tests/test_types_extras.py#L1441

I'm Closing for now, thanks for your feedback.

ticosax:
for the record the issue was me (as usual).

If we save `Range(1, 3, '[]')` we get back from psycopg `Range(1, 4 '[)')` which is logically equivalent . But unfortunately, both ranges are not equals to each other in python. This is a bit misleading.

Issue #4663, opened 2019-05-06
The first connection against a postgres engine differs in state from all subsequent connections


MatthewWilkes:
Hi there,

When working on https://github.com/niteoweb/pyramid_deferred_sqla/issues/2 we discovered a potential minor bug in the `PGDialect` setup code. We're using the psycopg2 feature `set_session(readonly=True)` to ensure that certain transactions do not cause database writes. This is being done on the underlying connection objects as the readonly feature is not exposed through sqlalchemy as the isolation level is.

This is working well in the general case, however the very first connection that is set-up has already had statements run against it, specifically `show standard_conforming_strings` in
https://github.com/sqlalchemy/sqlalchemy/blob/fd2ecb5f87c1c0132263b5a35067c4bb76160fb2/lib/sqlalchemy/dialects/postgresql/base.py#L2432

The `DefaultDialect` has a similar pattern for feature detection, specifically `select version()` in https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/engine/default.py#L288-L290 and other commands in that function. These are explicitly rolled back in https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/engine/default.py#L315 but there is no explicit roll back in the PGDialect initialisation.

For users who are trying to set an isolation level, this isn't a problem as psycopg2's `connection.set_isolation_level(...)` is allowed in this state. However, `connection.set_session(...)` is not, raising an error as the transaction is already in progress.

I suggest we add an explicit rollback to the end of the `PGDialect` feature detection to match the one in `DefaultDialect` as it will allow the assumption that newly created connections to a postgres database have not yet had an implicit transaction begun.

I'm very much aware that sqlalchemy is not intended to support direct manipulation of the underlying connection options, however it appears that an attempt was made to keep the set-up code from affecting user transactions in some places and therefore the fact this dialect does affect the first user transaction appears to be an oversight.

I've attached a simple script that demonstrates this, which will require customisation to the connection string depending on your local environment.

[minrep_py.txt](https://github.com/sqlalchemy/sqlalchemy/files/3148676/minrep_py.txt)

MatthewWilkes:
@zzzeek I just spent some time working on a more generic test, before realising you'd already put a better one than the one I wrote in a later comment. Oops. This PR has been updated with your suggested changes.

sqla-tester:
[...]

PR #4664, opened 2019-05-06
Add additional rollback to PGDialect feature detection

MatthewWilkes:
Address problem reported in bug #4663 that causes minor inconsistency in the underlying postgresql transaction state for the first connection
to a new engine.

### Description
As described in #4663, the first transaction created against a postgres database is responsible for doing engine feature detection. All the feature-detection queries done in `BaseDialect` are guarded by an explicit rollback, however on postgresql > 8.2 there is an additional query done in the `PGDialect` which is not so guarded. This adds an additional rollback at the end of that check.

### Checklist

This pull request is:

- [X] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/doc/build/changelog/unreleased_14/4663.rst
+++ b/doc/build/changelog/unreleased_14/4663.rst
@@ -1,0 +1,7 @@
+.. change::
+   :tags: bug, orm, postgresql
+   :tickets: 4663
+
+   Added additional rollback during engine creation for postgresql
+   engines that prevents the first connection starting having an
+   implicit transaction created earlier than usual.
--- a/lib/sqlalchemy/engine/default.py
+++ b/lib/sqlalchemy/engine/default.py
@@ -312,8 +312,6 @@ class DefaultDialect(interfaces.Dialect):
         ):
             self._description_decoder = self.description_encoding = None
 
-        self.do_rollback(connection.connection)
-
     def on_connect(self):
         """return a callable which sets up a newly created DBAPI connection.
 
--- a/lib/sqlalchemy/engine/strategies.py
+++ b/lib/sqlalchemy/engine/strategies.py
@@ -197,6 +197,7 @@ class DefaultEngineStrategy(EngineStrategy):
                 )
                 c._execution_options = util.immutabledict()
                 dialect.initialize(c)
+                dialect.do_rollback(c.connection)
 
             event.listen(pool, "first_connect", first_connect, once=True)
 
--- a/test/dialect/postgresql/test_dialect.py
+++ b/test/dialect/postgresql/test_dialect.py
@@ -500,6 +500,12 @@ class MiscBackendTest(
                 "c %s NOT NULL" % expected,
             )
 
+    def test_initial_transaction_state(self):
+        from psycopg2.extensions import STATUS_IN_TRANSACTION
+        engine = engines.testing_engine()
+        with engine.connect() as conn:
+            assert conn.connection.status != STATUS_IN_TRANSACTION
+
 
 class AutocommitTextTest(test_execute.AutocommitTextTest):
     __only_on__ = "postgresql"
--- a/test/engine/test_execute.py
+++ b/test/engine/test_execute.py
@@ -660,6 +660,15 @@ class ExecuteTest(fixtures.TestBase):
         eq_(conn._execution_options, {"autocommit": True})
         conn.close()
 
+    def test_initialize_rollback(self):
+        """test a rollback happens during first connect"""
+        eng = create_engine(testing.db.url)
+        with patch.object(eng.dialect, "do_rollback") as do_rollback:
+            assert do_rollback.call_count == 0
+            connection = eng.connect()
+            assert do_rollback.call_count == 1
+        connection.close()
+
     @testing.requires.ad_hoc_engines
     def test_dialect_init_uses_options(self):
         eng = create_engine(testing.db.url)


zzzeek:
hey there -

why isn't this rollback placed [here](https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/engine/strategies.py#L199)?  All DBAPIs are required to implicitly start transactions, so I would assume this applies to all of them, PostgreSQL being just the one that complains the most.

zzzeek:
a generic case would be needed in https://github.com/sqlalchemy/sqlalchemy/blob/master/test/engine/test_transaction.py most likely.     Most DBAPIs won't have any reliable way to report on this condition so it might have to use mock.patch.

zzzeek:
OK some of the most relevant tests for this kind of area are in test/engine/test_execute.py, such as https://github.com/sqlalchemy/sqlalchemy/blob/master/test/engine/test_execute.py#L663 .

zzzeek:
OK furher update, we *do* [do the rollback](https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/engine/default.py#L315) it's just PG is doing additional work after it.    So the do_rollback() should be moved still but need to figure out a way to test it.

zzzeek:
it's not even clear what state the Postgresql methods are adding,  except for the "show standard_conforming_strings" call as well as the get_oids() call, so there's no good way to mock out testing for state here.  so the existing psycopg2 test should remain, however, the do_rollback() still needs to be moved out and additionally the following which tests only what it is that we already do, it can't really detect in every case that the DBAPI connection has no transactional state on it:

```
diff --git a/test/engine/test_execute.py b/test/engine/test_execute.py
index e18cdfad4c..f88f08fda6 100644
--- a/test/engine/test_execute.py
+++ b/test/engine/test_execute.py
@@ -672,6 +672,18 @@ class ExecuteTest(fixtures.TestBase):
             eq_(conn._execution_options, {})
             conn.close()
 
+    @testing.requires.ad_hoc_engines
+    def test_dialect_init_rolls_back(self):
+        eng = create_engine(testing.db.url)
+
+        with patch.object(eng.dialect, "do_rollback") as evt:
+            conn = eng.connect()
+            dbapi_conn = conn.connection.connection
+
+        eq_(evt.mock_calls, [mock.call(dbapi_conn)])
+
+        conn.close()
+
     @testing.requires.ad_hoc_engines
     def test_generative_engine_event_dispatch_hasevents(self):
         def l1(*arg, **kw):

```

zzzeek:
correction, the *proposed* psycopg2 test, since postgresql is very picky

MatthewWilkes:
@zzzeek Yeah, as you say, there is an existing rollback but PGDialect does additional work after. I was assuming that it would be the responsibility of the dialect to do a rollback if it caused an additional one to be needed, rather than moving the initial rollback.

The other way of looking at it, which appears to be what you're suggesting, is that we shouldn't have an additional rollback in the `PGDialect` in addition to the one in `DefaultDialect`, we should move that out into one consistent place. That means it's no longer the responsibility of the dialect to ensure that its underlying session machinery is clean once it's done the feature detection.

My instinct was that refactoring the cleanup for all dialects to deal with the fact that one dialect makes the connection dirty again after the cleanup was overly invasive as it changed the responsibility of the dialect.

To answer your first question about implicitly starting transactions, my understanding of this is that it's different to the implicit begin. A sqlalchemy level transaction is still created, the problem is that the feature detection causes the postgresql implicit transaction to be created earlier than expected.

On a side note, this is the first time I've even looked at the sqlalchemy source code, let alone submitted a patch to your project and I'm no expert in database mechanic internals. I'm very happy to clarify the issue I'm seeing and the intention I had when writing code, but equally I don't have strong feelings about how this should be fixed and am happy to take direction. To that end, I'm happy to do the refactoring I mentioned earlier if you think that's best.

zzzeek:
> @zzzeek Yeah, as you say, there is an existing rollback but PGDialect does additional work after. I was assuming that it would be the responsibility of the dialect to do a rollback if it caused an additional one to be needed, rather than moving the initial rollback.

It's assumed that dialect.initialize() does things with the connection, so an important thing like making sure connection is rolled back is best done outside the specific dialects, DefaultDialect was trying to do it but due to subclassing, it should just be moved outside, *or* moved to a different method inside DefaultDialect like "_initiailize()", that does the call to .initialize() and then the rollback, but strategies.py is the only caller in any case (note that I will also have to port this to 2.0 separately as that code has been refactored).

> 
> The other way of looking at it, which appears to be what you're suggesting, is that we shouldn't have an additional rollback in the `PGDialect` in addition to the one in `DefaultDialect`, we should move that out into one consistent place. That means it's no longer the responsibility of the dialect to ensure that its underlying session machinery is clean once it's done the feature detection.

yes we try to keep the dialects as free of additional steps as possible.

> 
> My instinct was that refactoring the cleanup for all dialects to deal with the fact that one dialect makes the connection dirty again after the cleanup was overly invasive as it changed the responsibility of the dialect.

I had that instinct also but seeing that as we are already doing the do_rollback(), it's safe to assume not much really changes.   the Oracle dialect has a similar thing happening also (runs SQL after super.initialize) but I fully expect it to be OK.

> 
> To answer your first question about implicitly starting transactions, my understanding of this is that it's different to the implicit begin. A sqlalchemy level transaction is still created, the problem is that the feature detection causes the postgresql implicit transaction to be created earlier than expected.

this would be dialect.do_rollback() on the DBAPI connection as it is now.

> 
> On a side note, this is the first time I've even looked at the sqlalchemy source code, let alone submitted a patch to your project and I'm no expert in database mechanic internals. I'm very happy to clarify the issue I'm seeing and the intention I had when writing code, but equally I don't have strong feelings about how this should be fixed and am happy to take direction. To that end, I'm happy to do the refactoring I mentioned earlier if you think that's best.

zzzeek:
> On a side note, this is the first time I've even looked at the sqlalchemy source code, let alone submitted a patch to your project and I'm no expert in database mechanic internals. I'm very happy to clarify the issue I'm seeing and the intention I had when writing code, but equally I don't have strong feelings about how this should be fixed and am happy to take direction. To that end, I'm happy to do the refactoring I mentioned earlier if you think that's best.

I would move dialect.do_rollback() to from DefaultDialect.initialize() into strategies.py as in the link above, keep the additional test you've proposed, and also add the test that I have a diff against above.  then we run through CI to make sure it all works.

MatthewWilkes:
Great, shall do. I'm travelling at the moment, so don't expect this today, but I'll get it to you in the next day or so and bump this thread when done.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 23e809ca037721c26b2e0e380bb46535ae4baa54 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 23e809ca037721c26b2e0e380bb46535ae4baa54: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1259

MatthewWilkes:
Urgh, had a line that was 1 character too long. Resubmitted.

zzzeek:
I'm using [black](https://github.com/python/black/) and [zimports](https://github.com/sqlalchemyorg/zimports/) to do code formatting and it is saving me enormous amounts of time, I have now enabled them with [pre-commit](https://github.com/pre-commit/pre-commit) locally so in most cases I don't have flake8 failures anymore, the only case is black won't shorten literal lines and it also has a few scenarios where it doesnt shorten without some manual rework of the code.      I will soon be adding my .pre-commit-config.yaml file and encouraging people to start using it.   it helps a lot

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision e544fe671d443ed06b210ba1cd1d7ba9c5653831 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset e544fe671d443ed06b210ba1cd1d7ba9c5653831 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1259

MatthewWilkes:
Yeah, most of my projects are now using black in a pre-commit hook. I think adding the config is a good idea.

It looks like the PEP8 check has failed again, with an overly long line in `lib/sqlalchemy/sql/selectable.py` but that's not a file I've touched and I don't see a long line in there, so suspect something went wrong there.

zzzeek:
sometimes I push commits directly with long lines lets try to rebase

zzzeek:
yup selectable is failing in master, moment

zzzeek:
i've pushed to the gerrit, which doesn't seem to add any comments here but it's going through

MatthewWilkes:
Great!

sqla-tester:

Code-Review+2 Workflow+1

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1259 has been **merged**. Congratulations! :)

sqla-tester:

Code-Review+2 Workflow+1

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1260 has been **merged**. Congratulations! :)

zzzeek:
thanks for the efforts!

MatthewWilkes:
And thank you for the brilliant software, and for helping me contibute to it.

PR #4679, opened 2019-05-13
Sql

auvipy:


.gitreview
```diff
@@ -1,4 +1,4 @@
 [gerrit]
 host=gerrit.sqlalchemy.org
 project=sqlalchemy/sqlalchemy
-defaultbranch=master
+defaultbranch=rel_2_0
```
auvipy: is it correct?

auvipy (commented):
will fix the merge conflicts

.gitreview
zzzeek: if the PR is against master then that should stay at "master".

zzzeek:
So I need PRs that have just one or two things changed each, against current master.  These PRs look like they are getting all confused in merging across branches or something.   I know it's a little tricky to get git to do the right thing.

auvipy:
Ok, I am going to come with smaller chunks of changes on top of master. Will close this after coming with that. Allow me a lil bit of time.

auvipy:
are you done with those?

zzzeek:
the 2.0 branch is not really ready right now as a lot of what I was going to put in 2.0 I'm actually putting in 1.4.  I'm going to keep getting as much of the infrastructure into 1.4 as I can without major backwards incompatibility.  then i *think* what I will try first is to build out some of sqlalchemy.future in 1.4, that will be the basis for things to go into 2.0.    I'm still not really sure how I want to organize 1.4  / 2.0 branches because I'm concerned about changes being easy to port between the two.

zzzeek:
in 1.4 / master, I do want to remove older things that are deprecated, anything marked deprecated before 1.1 or so we can just remove at this point.   but if you contribute they have to be small PRs so they can be managed more easily.

auvipy:
OK, i will be sending once deprecation once, on 1.4

auvipy:
Many small PR's are coming to address this! against master

PR #4724, opened 2019-06-12
Fixes 4717 - Reflection of indexes in PostgreSQL doesn't include DESC / NULLS ordering

### LINKED ISSUE #4717: Reflection of indexes in PostgreSQL doesn't include DESC / NULLS ordering.
As of sqlalchemy 1.3.4, it looks like the Postgres dialect doesn't reflect the ASC/DESC or NULLS FIRST/LAST state of an index.   E.g.  `Index(col_a, col_b.desc())` is reflected back as `Index(col_a, col_b)`.

After diving into it a little, it looks like the query `PGDialect.get_indexes()` needs to pull the "ix.indoption" column as part of the query.  Per some references such as [this stackoverflow thread](https://stackoverflow.com/questions/18121103/how-to-get-the-index-column-orderasc-desc-nulls-first-from-postgresql),  ix.indoption appears to be a int2vector, where each integer is a mask corresponding to the columns/expressions in the index, where bit 0 indicates "DESC" and bit 1 indicates "NULLS FIRST".  

This seems poorly documented in postgres, and I hate to suggest hardcoding something so obscure; but short of rewriting the entire get_indexes() method to call and parse the output of `pg_get_indexdef()`, there doesn't seem to be much other way to get this information.

(I'm also not entirely clear how to pass this information back via get_indexes() to the reflection code. If this change seems reasonable, and someone can give me a pointer on that part, I'm willing to give a shot at making a patch for this).

eli-collins:
This fixes #4717, adding support to the postgresql dialect to reflect per-column sort order for indexes.

*I may not have made this commit on top of the appropriate branch, please let me know if I should rebase & resubmit*.

### Description

1) `Dialect.get_indexes()` was expanded to support general reflection
of per-column sort order via "column_sorting" keyword (details added to get_indexes docstring).  This keyword signals `Inspector._reflect_indexes()` to adapt the column objects via .desc(), .nullslast(), etc before adding them to the resulting Index.  

This should be generic enough to work for any database.  Additionally, since the dictionary is keyed off the column name, when/if the framework is expanded in the future to support general expressions, this should still hold up: dialects can either just omit expressions from the "column_sorting" dictionary (and handle appending sort modifiers as part of the returned expression), or just use the sql.text() expressions themselves as "column_sorting" key. 

2) PGDialect.get_indexes() now reflects per-column sort order from "pg_index.indoption" in system schema. (This feature wasn't present before pg 8.3).   I added a new unittest for index sort order reflection; and on my limited system, all index reflection tests are passing.

### Checklist
This pull request is:

- [ ] A short code fix
	- Please include: `Fixes: #4717` in the commit message
	- please include tests.

### DIFF
--- a/lib/sqlalchemy/engine/reflection.py
+++ b/lib/sqlalchemy/engine/reflection.py
@@ -469,6 +469,11 @@ class Inspector(object):
         unique
           boolean
 
+        column_sorting
+          optional dict mapping column names to dict of sorting options.
+          each sorting options dict may set any of the following keywords
+          to True: ``asc``, ``desc``, ``nullsfirst``, ``nullslast``.
+
         dialect_options
           dict of dialect-specific index options.  May not be present
           for all dialects.
@@ -869,6 +874,7 @@ class Inspector(object):
         for index_d in indexes:
             name = index_d["name"]
             columns = index_d["column_names"]
+            column_sorting = index_d.get("column_sorting", {})
             unique = index_d["unique"]
             flavor = index_d.get("type", "index")
             dialect_options = index_d.get("dialect_options", {})
@@ -897,8 +903,11 @@ class Inspector(object):
                         "%s key '%s' was not located in "
                         "columns for table '%s'" % (flavor, c, table_name)
                     )
-                else:
-                    idx_cols.append(idx_col)
+                    continue
+                c_sorting = column_sorting.get(c)
+                if c_sorting:
+                    idx_col = self._create_sorted_expr(idx_col, **c_sorting)
+                idx_cols.append(idx_col)
 
             sa_schema.Index(
                 name,
@@ -906,6 +915,19 @@ class Inspector(object):
                 _table=table,
                 **dict(list(dialect_options.items()) + [("unique", unique)])
             )
+
+    @classmethod
+    def _create_sorted_expr(cls, expr, asc=False, desc=False, nullsfirst=False,
+                            nullslast=False):
+        if asc:
+            expr = expr.asc()
+        if desc:
+            expr = expr.desc()
+        if nullsfirst:
+            expr = expr.nullsfirst()
+        if nullslast:
+            expr = expr.nullslast()
+        return expr
 
     def _reflect_unique_constraints(
         self,
--- a/test/dialect/postgresql/test_reflection.py
+++ b/test/dialect/postgresql/test_reflection.py
@@ -907,6 +907,74 @@ class ReflectionTest(fixtures.TestBase):
             ],
         )
 
+    @testing.fails_if("postgresql < 8.3", "index ordering not supported")
+    @testing.provide_metadata
+    def test_index_reflection_with_sorting(self):
+        """reflect indexes with sorting options set"""
+        metadata = self.metadata
+
+        t1 = Table(
+            "party",
+            metadata,
+            Column("id", String(10), nullable=False),
+            Column("name", String(20)),
+            Column("aname", String(20)),
+        )
+        metadata.create_all()
+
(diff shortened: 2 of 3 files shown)


zzzeek:
this should be on master, I have scripts to do the backporting

lib/sqlalchemy/engine/reflection.py
```diff
@@ -469,6 +469,11 @@ def get_indexes(self, table_name, schema=None, **kw):
         unique
           boolean
 
+        column_sorting
+          optional dict mapping column names to dict of sorting options.
+          each sorting options dict may set any of the following keywords
+          to True: ``asc``, ``desc``, ``nullsfirst``, ``nullslast``.
+
```
zzzeek: add a version note, this should be good for 1.3 so let's assume 1.3.5:

`    .. versionadded:: 1.3.5`

test/dialect/postgresql/test_reflection.py
```diff
@@ -907,6 +907,74 @@ def go():
...
+        t1 = Table(
+            "party",
+            metadata,
+            Column("id", String(10), nullable=False),
+            Column("name", String(20)),
+            Column("aname", String(20)),
+        )
+        metadata.create_all()
+
+        # check ASC, DESC options alone
```
zzzeek: I'm sure this is imitating the other tests but i tend to try to use a newer pattern here:

```

with testing.db.connect() as conn:
    conn.execute("create index ...")
    conn.execute("create index ...")
```

test/dialect/postgresql/test_reflection.py
```diff
@@ -907,6 +907,74 @@ def go():
...
+        # check ASC w/ NULLS options
+        testing.db.execute("""
+          create index idx3 on party 
+                (name ASC NULLS FIRST, aname ASC NULLS LAST)  
+        """)
+
+        # reflect data
+        m2 = MetaData(testing.db)
+        t2 = Table("party", m2, autoload=True)
+        assert len(t2.indexes) == 3
```
zzzeek: another thing that is newer and likely not everywhere in this old test suite:

`    eq_(len(t2.indexes), 3)
`

since then, py.test detects and instruments `==` signs, unittest added assertEquals(), bah...we're just trying to stick w/ the one :)

test/dialect/postgresql/test_reflection.py
```diff
@@ -907,6 +907,74 @@ def go():
...
+
+        # Make sure indexes are in the order we expect them in
+        r1, r2, r3 = sorted(t2.indexes, key=lambda idx: idx.name)
+
+        assert r1.name == "idx1"
```
zzzeek: eq_ for these too

test/dialect/postgresql/test_reflection.py
```diff
@@ -907,6 +907,74 @@ def go():
...
+        assert r3.name == "idx3"
+
+        # "ASC NULLS LAST" is implicit default for indexes,
+        # and "NULLS FIRST" is implicit default for "DESC".
+        # (https://www.postgresql.org/docs/11/indexes-ordering.html)
+
+        def compile_exprs(exprs):
+            return list(map(str, exprs))
+
+        assert compile_exprs([
```
zzzeek: eq_ eq_ eq_ :)

also, if you are using [pre-commit](https://pre-commit.com/), this code will automatically format itself when you commit as we pull in the black code formatter.

zzzeek (changes requested):
just cosmetic style stuff, looks great!

eli-collins:
Rebased onto master, and added all those changes!  Wasn't sure how to do that without closing this one and starting a new PR -- #4725

PR #4759, opened 2019-07-10
sql.sqltypes.Enum: More efficient copy for python 2.7

### LINKED ISSUE #4758: Importing declarative schema classes with large enums is slow under python 2.7
Importing ≈10 classes, each with an enum with ≈300 choices takes more than 2 seconds running with python 2.7 on a quad core Intel i5 @ 2.2 GHz.

Note that the underlying performance problem appears to have been fixed as of python 3.6.8.

Reproduction script:

``` python
import enum
import uuid
from datetime import datetime

import babel
from sqlalchemy import Column, Enum, Integer
from sqlalchemy.ext.declarative.api import DeclarativeMeta

class Currency(enum.Enum):
    """Standard unit of money.

    Instances can be referenced via ISO 4217 currency codes, e.g.,
    `Currency.EUR`.
    """

    locals().update(
        {code: object() for code in list(babel.Locale("en", "US").currencies)}
    )

print(len(Currency.__members__))
 # prints 297 with Babel==2.4.0

class CurrencyClassType(DeclarativeMeta):
    def __init__(cls, *args, **kwargs):
        setattr(cls, "__tablename__", uuid.uuid4().hex)
        setattr(cls, "id", Column(Integer, primary_key=True))
        setattr(cls, "currency", Column(Enum(Currency, native_enum=False)))
        super(CurrencyClassType, cls).__init__(*args, **kwargs)

start = datetime.now()
Base = declarative_base()

 # This loop has the same effect as declaring N(=10) classes with a
 # random class name, a random table name, an `id` primary key column,
 # and a `currency` enumerated column, using the `Currency` enumeration
 # above.
for i in xrange(10):
    # Simulate declaring a class, inheriting from `Base`
    CurrencyClassType(uuid.uuid4().hex, (Base, ), {})

print(datetime.now() - start)
 # prints roughly 2.5 seconds
```

HeyBillFinn:
### Description

Before this commit, an enumeration with a few hundred options (e.g.,
representing 3-letter ISO 4217 currency codes) that appeared on ≈10
tables could take a couple of seconds to import and declare the classes.
After this commit, the same declared schema classes can be imported in
less than a second.

This performance issue is not present when running with Python 3.6.8.

Reproduction script, run using `Python 2.7.12`, `Babel==2.4.0`,
`SQLAlchemy==1.3.5` on a quad core Intel i5 @ 2.20GHz running Ubuntu
16.04.01.

``` python
import enum
import uuid
from datetime import datetime

import babel
from sqlalchemy import Column, Enum, Integer
from sqlalchemy.ext.declarative.api import DeclarativeMeta

class Currency(enum.Enum):
    """Standard unit of money.

    Instances can be referenced via ISO 4217 currency codes, e.g.,
    `Currency.EUR`.
    """

    locals().update(
        {code: object() for code in list(babel.Locale("en", "US").currencies)}
    )

print(len(Currency.__members__))
 # prints 297 with Babel==2.4.0

class CurrencyClassType(DeclarativeMeta):
    def __init__(cls, *args, **kwargs):
        setattr(cls, "__tablename__", uuid.uuid4().hex)
        setattr(cls, "id", Column(Integer, primary_key=True))
        setattr(cls, "currency", Column(Enum(Currency, native_enum=False)))
        super(CurrencyClassType, cls).__init__(*args, **kwargs)

start = datetime.now()
Base = declarative_base()

 # This loop has the same effect as declaring N(=10) classes with a
 # random class name, a random table name, an `id` primary key column,
 # and a `currency` enumerated column, using the `Currency` enumeration
 # above.
for i in xrange(10):
    # Simulate declaring a class, inheriting from `Base`
    CurrencyClassType(uuid.uuid4().hex, (Base, ), {})

print(datetime.now() - start)
 # prints roughly 2.5 seconds
 # prints less than 200 milliseconds
```

Fixes #4758.

### Checklist

This pull request is:

- [x] A performance improvement

### DIFF
--- a/lib/sqlalchemy/sql/sqltypes.py
+++ b/lib/sqlalchemy/sql/sqltypes.py
@@ -1426,12 +1426,14 @@ class Enum(Emulated, String, SchemaType):
             self.enum_class = enums[0]
             if self.values_callable:
                 values = self.values_callable(self.enum_class)
+                objects = [
+                    self.enum_class.__members__[k]
+                    for k in self.enum_class.__members__
+                ]
             else:
-                values = list(self.enum_class.__members__)
-            objects = [
-                self.enum_class.__members__[k]
-                for k in self.enum_class.__members__
-            ]
+                enum_members = self.enum_class.__members__.copy()
+                values = enum_members.keys()
+                objects = enum_members.values()
             return values, objects
         else:
             self.enum_class = None


zzzeek:
hi there -

for a performance related thing we'd want to probably have a test in test/aaa_profiling/.    

A tip for timing performance, the improvement here is in Enum(), so use the timeit module and just Enum() directly, there's no need for babel or anything else, e.g.:
```

import timeit
import enum

from sqlalchemy import Enum


class Currency(enum.Enum):
    locals().update(
        {("code%d" % i): object() for i in range(4000)}
    )

def go():
    e = Enum(Currency, native_enum=False)

print(
    timeit.timeit("go()", "from __main__ import go", number=10000)
)
```

I can confirm the "enum" module in Python 2 has a performance problem here.  There's no issue under Python 3.      Is there a reason Python 2 "enum" is so slow and any chance you can propose a fix for them?   that way everyone gains from the performance improvement and I don't have to worry about it being reverted here.

zzzeek:
so here is the test suite:

https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1349

note it does not use the "enum" module.  We will now apply your patch and see if it reduces callcounts.

zzzeek:
so with the .copy(), the patch doubles the function call count from what it usually is:

AssertionError: Adjusted function call count 16050 not within 5.0% of expected 8042, platform 3.7_sqlite_pysqlite_dbapiunicode_cextensions. Rerun with --write-profiles to regenerate this callcount.


AssertionError: Adjusted function call count 20048 not within 5.0% of expected 12040, platform 2.7_sqlite_pysqlite_dbapiunicode_nocextensions. Rerun with --write-profiles to regenerate this callcount.

without the .copy(), the callcounts stay exactly identical, which is strange.   so we need to lose that .copy().    also I need to figure out what we are testing here because both approaches are the identical speed without using the Python 2 "enum" module.

zzzeek:
so running profiles against enum.Enum, which I can't add to the test suite in the final commit because it is a dependency, the callcounts for a 100 element enum blow out for the old approach:

Adjusted function call count 23101 not within 5.0% of expected 1489, platform 2.7_sqlite_pysqlite_dbapiunicode_nocextensions. Rerun with --write-profiles to regenerate this callcount.


looking for some key culprits, there is some kind of `__setitem__` thing going on in enum for unknown reasons:

```


  10605    0.002    0.000    0.002    0.000 /usr/lib64/python2.7/collections.py:90(__iter__)
      103    0.000    0.000    0.023    0.000 /home/classic/.venv2/lib/python2.7/site-packages/enum/__init__.py:366(__members__)
        1    0.000    0.000    0.000    0.000 /home/classic/dev/sqlalchemy/test/../lib/sqlalchemy/sql/elements.py:4340(__new__)
    10300    0.011    0.000    0.011    0.000 /usr/lib64/python2.7/collections.py:71(__setitem__)
  105/104    0.000    0.000    0.000    0.000 /home/classic/.venv2/lib64/python2.7/abc.py:148(__subclasscheck__)
```

those 5-digit counts are not there if we don't use the old way.   so need to see what enum is doing here it looks quite complicated.

zzzeek:
(note this is all work that the "enum" maintainers could be doing)

zzzeek:
seems to be just the __getitem__ that's hitting it.  performance is still fine if we do:

                values = list(self.enum_class.__members__)
                objects = self.enum_class.__members__.values()

if we want to patch here and test against this, we need a mock enum class that illustrates a slow `__getitem__` call.

zzzeek:
The performance problem is because it uses Python 2 OrderedDict which is horrible for performance and even with your patch as is, the class performs vastly slower than Python 3's enum because iterating OrderedDict is necessary and it's extremely slow.

It looks like you *are* the maintainer of the Python 2 enum library, so why not just use a reasonable OrderedDict and fix the issue for everyone?

zzzeek:
Hi why not just do this?  the issue is that ``__members__`` is copying the dictionary every time, performance issue solved very simply like this:
```

diff --git a/lib/sqlalchemy/sql/sqltypes.py b/lib/sqlalchemy/sql/sqltypes.py
index 6a520a2d59..37301123de 100644
--- a/lib/sqlalchemy/sql/sqltypes.py
+++ b/lib/sqlalchemy/sql/sqltypes.py
@@ -1424,13 +1424,14 @@ class Enum(Emulated, String, SchemaType):
 
         if len(enums) == 1 and hasattr(enums[0], "__members__"):
             self.enum_class = enums[0]
+            members = self.enum_class.__members__
             if self.values_callable:
                 values = self.values_callable(self.enum_class)
             else:
-                values = list(self.enum_class.__members__)
+                values = list(members)
             objects = [
-                self.enum_class.__members__[k]
-                for k in self.enum_class.__members__
+                members[k]
+                for k in members
             ]
             return values, objects
         else:

```

much easier for me to commit / test that

HeyBillFinn:
Thanks for the detailed feedback. I tried your patch locally, and it works great. Thanks for finding a cleaner way. Let me know if you want me to do anything to help land the changes.

PR #4764, opened 2019-07-15
Enable use of psycopg2.execute_values()

### LINKED ISSUE #4623: Use psycopg2 execute_values() instead of execute_batch()
Hi,
When working with psycopg2 and setting flag use_batch_mode, a call to psycopg2.extras.execute_batch() is used instead of cursor.executemany(). But there's a faster function - [psycopg2.extras.execute_values()](http://initd.org/psycopg/docs/extras.html#psycopg2.extras.execute_values).
execute_batch works by concatenating single insert statements (using ';') and sending them together to the server.
execute_values works by creating a single insert statement with multiple tuples, which is much faster
(for both the page_size parameter determines the number of rows that will be inserted in each call to the database).
In my case I saw a 40% improvement using execute_values().

ydinari:
### Description

Fixes: #4623
In SQLCompiler.visit_insert(), put logic to generate the VALUES clause into function (generate_values_placeholders_str) to enable overriding just that logic in derived classes.
In PGCompiler_psycopg2 override generate_values_placeholders_str() to enable use of psycopg2.execute_values() which is faster than execute_batch().
Extend flag use_batch_mode to indicate which psycopg2 method to use.
In PGDialect_psycopg2.do_executemany(), call the correct psycopg2 method based on the use_batch_mode flag and compiler variables.
Add tests for psycopg2.execute_values().

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/psycopg2.py
+++ b/lib/sqlalchemy/dialects/postgresql/psycopg2.py
@@ -596,11 +596,13 @@ class PGCompiler_psycopg2(PGCompiler):
     def generate_values_placeholders_str(
             self, crud_params, returning_clause_exists):
 
-        # Currently not using psycopg2.execute_values() when there's a returning clause; need to add support
+        # Currently not using psycopg2.execute_values() when there's a
+        # returning clause; need to add support
         # for receiving multiple return values from insert query
         # Note: self.inline is true iff there are multiple parameters sets to
         # the query
-        if self.inline and not returning_clause_exists and self.dialect.psycopg2_executemany_mode is EXECUTEMANY_VALUES:
+        if self.inline and not returning_clause_exists and \
+                self.dialect.psycopg2_executemany_mode is EXECUTEMANY_VALUES:
             self.execute_values_insert_template = "(" + \
                 ", ".join([c[1] for c in crud_params]) + ")"
             return " VALUES %s"
@@ -693,24 +695,27 @@ class PGDialect_psycopg2(PGDialect):
                     EXECUTEMANY_BATCH.name,
                     EXECUTEMANY_VALUES.name):
                 raise exc.ArgumentError(
-                    "Unsupported value for 'executemany_mode': %s" % executemany_mode)
+                    "Unsupported value for 'executemany_mode': %s" %
+                    executemany_mode)
             self.psycopg2_executemany_mode = util.symbol(executemany_mode)
         else:
             self.psycopg2_executemany_mode = None
 
         if not isinstance(executemany_page_size, int):
-            raise exc.ArgumentError("Wrong type for 'executemany_page_size': %s" % type(
-                executemany_page_size).__name__)
+            raise exc.ArgumentError(
+                "Wrong type for 'executemany_page_size': %s" %
+                type(executemany_page_size).__name__)
 
         if executemany_page_size <= 0:
             raise exc.ArgumentError(
-                "Wrong value for 'executemany_page_size': %s" % executemany_page_size)
+                "Wrong value for 'executemany_page_size': %s" %
+                executemany_page_size)
 
         self.psycopg2_executemany_page_size = executemany_page_size
         self.psycopg2_batch_mode = use_batch_mode
 
-        # use_batch_mode supported for backward compatibility. To avoid having to check two flags,
-        # set executemany_mode flag here and use it only
+        # use_batch_mode supported for backward compatibility. To avoid having
+        # to check two flags, set executemany_mode flag here and use it only
         if not self.psycopg2_executemany_mode:
             self.psycopg2_executemany_mode = \
                 EXECUTEMANY_BATCH if self.psycopg2_batch_mode else \
@@ -875,7 +880,8 @@ class PGDialect_psycopg2(PGDialect):
                 page_size=self.psycopg2_executemany_page_size)
 
         else:  # EXECUTEMANY_VALUES of non-insert query, or EXECUTEMANY_BATCH
-            self._psycopg2_extras().execute_batch(cursor, statement, parameters)
+            self._psycopg2_extras().execute_batch(cursor, statement,
+                                                  parameters)
 
     @util.memoized_instancemethod
     def _hstore_oids(self, conn):
--- a/lib/sqlalchemy/dialects/postgresql/psycopg2.py
+++ b/lib/sqlalchemy/dialects/postgresql/psycopg2.py
@@ -52,6 +52,21 @@ psycopg2-specific keyword arguments which are accepted by
 
     :ref:`psycopg2_unicode`
 
+* ``execution_mode``: This flag determines the psycopg2 methods used for executing queries,
+  enabling faster execution for queries with multiple parameters. It replaces use_batch_mode
+  (which is still supported for backward compatibility).
+  Possible values are: 
+  None - Flag not used, ``use_batch_mode`` determines behavior
+  'single_statement'  - Every parameters set is a single query that is sent to the database.
+                        Uses ``psycopg2.extras.execute`` or ``psycopg2.extras.executemany``.
+                        Equivalent to use_batch_mode=False
+  'statements_batch' - Multiple queries are sent together to the database.
+                       Uses ``psycopg2.extras.execute_batch``.
+                       Equivalent to use_batch_mode=True.
+  'values_batch' - Multiple parameters sets are packed together into a single query that is sent to the database.
+                   This flag currently supports only insert queries, and uses ``psycopg2.extras.execute_values``.
+                   For other queries it behaves the same as 'statements_batch'.
+
 * ``use_batch_mode``: This flag allows using psycopg2 faster methods for
   executing queries with multiple parameters (usually INSERT queries).
   If equals True or 'execute_batch', ``psycopg2.extras.execute_batch`` is used.
@@ -647,6 +662,7 @@ class PGDialect_psycopg2(PGDialect):
         client_encoding=None,
         use_native_hstore=True,
         use_native_uuid=True,
+        execution_mode=None,
         use_batch_mode=False,
         **kwargs
     ):
@@ -657,7 +673,14 @@ class PGDialect_psycopg2(PGDialect):
         self.use_native_uuid = use_native_uuid
         self.supports_unicode_binds = use_native_unicode
         self.client_encoding = client_encoding
+        self.psycopg2_execution_mode = execution_mode
         self.psycopg2_batch_mode = use_batch_mode
+
+        # use_batch_mode supported for backward compatibility. To avoid having to check two flags,
+        # set execution_mode flag here and use it only
+        if not self.psycopg2_execution_mode:
+            self.psycopg2_execution_mode = 'statements_batch' if self.psycopg2_batch_mode else 'single_statement'
+
         if self.dbapi and hasattr(self.dbapi, "__version__"):
             m = re.match(r"(\d+)\.(\d+)(?:\.(\d+))?", self.dbapi.__version__)
             if m:
@@ -682,7 +705,7 @@ class PGDialect_psycopg2(PGDialect):
         self.supports_sane_multi_rowcount = (
             self.psycopg2_version
             >= self.FEATURE_VERSION_MAP["sane_multi_rowcount"]
-            and not self.psycopg2_batch_mode
+            and self.psycopg2_execution_mode == 'single_statement'
         )
 
     @classmethod
@@ -803,18 +826,21 @@ class PGDialect_psycopg2(PGDialect):
             return None
 
     def do_executemany(self, cursor, statement, parameters, context=None):
-        if self.psycopg2_batch_mode == 'execute_values' and context and context.compiled.execute_values_insert_template:
+        if self.psycopg2_execution_mode == 'single_insert':
+            cursor.executemany(statement, parameters)
+        elif self.psycopg2_execution_mode == 'values_batch' and \
+                context and \
+                context.compiled.execute_values_insert_template:
+
             self._psycopg2_extras().execute_values(
                 cursor,
                 statement,
                 parameters,
                 template=context.compiled.execute_values_insert_template,
                 page_size=context.compiled.execute_values_page_size)
-        # Support True for backward compatibility
-        elif self.psycopg2_batch_mode in ('execute_batch', True):
+
+        else:  # 'values_batch' of non-insert query, or 'statements_batch'
             self._psycopg2_extras().execute_batch(cursor, statement, parameters)
-        else:
-            cursor.executemany(statement, parameters)
 
     @util.memoized_instancemethod
     def _hstore_oids(self, conn):
--- a/lib/sqlalchemy/dialects/postgresql/psycopg2.py
+++ b/lib/sqlalchemy/dialects/postgresql/psycopg2.py
@@ -77,7 +77,7 @@ psycopg2-specific keyword arguments which are accepted by
 
     :ref:`psycopg2_batch_mode`
 
-* ``execute_values_page_size``: When ``executemany_mode`` is ``'executemany_values'``, this flag
+* ``executemany_page_size``: When ``executemany_mode`` is ``'executemany_values'``, this flag
   determines the maximum number of rows to insert in a single request to the database.
   If there are more rows than that, they will be sent in several requests (each
   containing a single query). Default value is 10000.
@@ -589,30 +589,17 @@ class PGExecutionContext_psycopg2(PGExecutionContext):
 
 
 class PGCompiler_psycopg2(PGCompiler):
-    def __init__(
-        self, dialect, statement, column_keys=None, inline=False, **kwargs
-    ):
-        self.dialect = dialect
-        self.multiple_rows = inline
-        self.execute_values_insert_template = None
-
-        super(
-            PGCompiler_psycopg2,
-            PGCompiler_psycopg2).__init__(
-            self,
-            dialect,
-            statement,
-            column_keys,
-            inline,
-            **kwargs)
+    execute_values_insert_template = None
 
     # Override SQLCompiler.generate_values_placeholders_str - enable use of
     # psycopg2 execute_values()
     def generate_values_placeholders_str(
             self, crud_params, returning_clause_exists):
+
         # Currently not using psycopg2.execute_values() when there's a returning clause; need to add support
         # for receiving multiple return values from insert query
-        if self.multiple_rows and not returning_clause_exists and self.dialect.psycopg2_executemany_mode is EXECUTEMANY_VALUES:
+        # Note: self.inline is true iff there are multiple parameters sets to the query
+        if self.inline and not returning_clause_exists and self.dialect.psycopg2_executemany_mode is EXECUTEMANY_VALUES:
             self.execute_values_insert_template = "(" + \
                 ", ".join([c[1] for c in crud_params]) + ")"
             return " VALUES %s"
@@ -688,7 +675,7 @@ class PGDialect_psycopg2(PGDialect):
         use_native_hstore=True,
         use_native_uuid=True,
         executemany_mode=None,
-        execute_values_page_size=10000,
+        executemany_page_size=10000,
         use_batch_mode=False,
         **kwargs
     ):
@@ -707,7 +694,7 @@ class PGDialect_psycopg2(PGDialect):
         else:
             self.psycopg2_executemany_mode = None
 
-        self.psycopg2_execute_values_page_size = execute_values_page_size
+        self.psycopg2_executemany_page_size = executemany_page_size
         self.psycopg2_batch_mode = use_batch_mode
 
         # use_batch_mode supported for backward compatibility. To avoid having to check two flags,
@@ -873,7 +860,7 @@ class PGDialect_psycopg2(PGDialect):
                 statement,
                 parameters,
                 template=context.compiled.execute_values_insert_template,
-                page_size=self.psycopg2_execute_values_page_size)
+                page_size=self.psycopg2_executemany_page_size)
 
         else:  # EXECUTEMANY_VALUES of non-insert query, or EXECUTEMANY_BATCH
             self._psycopg2_extras().execute_batch(cursor, statement, parameters)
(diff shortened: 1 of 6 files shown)


zzzeek:
vonage, wow I am a customer in fact.

OK, this looks pretty good, it will need pep8 formatting fixes as there are tests that will fail right now without them.   Do you want me to show you how that can be mostly automated?  if you install [pre-commit](https://pre-commit.com/) it will run code formatting automatically, however there might be some long lines here you'd have to fix manually.

I'll run this through gerrit so you can see.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision b7f6fb5f522db7e03c9cce7bc21222aa0843d166 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change b7f6fb5f522db7e03c9cce7bc21222aa0843d166: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1355

zzzeek:
I'm going to be out of town for a week so thanks very much for your patience.

lib/sqlalchemy/dialects/postgresql/psycopg2.py
```diff
@@ -551,13 +590,55 @@ def _log_notices(self, cursor):
 
 
 class PGCompiler_psycopg2(PGCompiler):
-    pass
+    def __init__(
+        self, dialect, statement, column_keys=None, inline=False, **kwargs
+    ):
+        self.dialect = dialect
+        self.multiple_rows = inline
+        self.execute_values_insert_template = None
```
zzzeek: this can be set as a class-level variable to start wtih, as None, that would allow you to not have to rewrite the whole constructor:

```
class PGCompiler_psycopg2(PGCompiler):
    execute_values_insert_template = None

```

lib/sqlalchemy/dialects/postgresql/psycopg2.py
```diff
@@ -551,13 +590,55 @@ def _log_notices(self, cursor):
...
+                self,
+                crud_params,
+                returning_clause_exists)
 
 
 class PGIdentifierPreparer_psycopg2(PGIdentifierPreparer):
     pass
 
 
+class EnumPsycopg2ExecutionMode(Enum):
```
zzzeek: we support Python 2, can't use Enum.  Please use util.symbol:

```
from .. import util
SINGLE_STATEMENT = util.symbol("single_statement")
```

lib/sqlalchemy/dialects/postgresql/psycopg2.py
```diff
@@ -610,6 +691,8 @@ def __init__(
         client_encoding=None,
         use_native_hstore=True,
         use_native_uuid=True,
+        execution_mode=None,
+        page_size=10000,
```
zzzeek: these names are both too generic if they are at the dialect level.    execution_mode needs to refer to "executemany" and "page_size" needs to refer to "execute_values".

lib/sqlalchemy/dialects/postgresql/psycopg2.py
```diff
@@ -766,11 +861,21 @@ def on_connect(conn):
             return None
 
     def do_executemany(self, cursor, statement, parameters, context=None):
-        if self.psycopg2_batch_mode:
-            extras = self._psycopg2_extras()
-            extras.execute_batch(cursor, statement, parameters)
-        else:
+        if self.psycopg2_execution_mode == EnumPsycopg2ExecutionMode.SINGLE_STATEMENT:
```
zzzeek: use "is" for comparing symbols, not ==

lib/sqlalchemy/dialects/postgresql/psycopg2.py
```diff
@@ -551,13 +590,55 @@ def _log_notices(self, cursor):
...
 
 class PGIdentifierPreparer_psycopg2(PGIdentifierPreparer):
     pass
 
 
+class EnumPsycopg2ExecutionMode(Enum):
+    """
+    Enum type to describe psycopg2 execution mode of multiple parameters queries
+    """
+    SINGLE_STATEMENT = 'single_statement'
```
zzzeek: Can we use a more descriptive set of names here ?    I'm not sure what "single statement" describes without reading the source code.   In all three cases, we are invoking a "statement", and we are calling a method on psycopg2 only once, so from that perspective, they are all "single statement".  Within the internals of psycopg2 executemany(), the default implementation runs the given statement for each parameter set, so that's also not "single".

I think a set of parameters / names that are more descriptive might be:   "executemany_mode = (EXECUTEMANY_DEFAULT|EXECUTEMANY_BATCH|EXECUTEMANY_VALUES)".   that way they are all specific to executemany and they indicate which ones are special and also map directly to the names used by psycopg2; someone using this API will likely be referring to psycopg2s documentation.

zzzeek (commented):
hi !   I see you've been working on this a lot, thanks!  I made a bunch of initial review comments.

lib/sqlalchemy/dialects/postgresql/psycopg2.py
ydinari: OK. You mean I guess all those variables?

lib/sqlalchemy/dialects/postgresql/psycopg2.py
ydinari: OK.
page_size can be used for execute_batch as well. I didn't use it in this pull request to keep the notion that no change should be made when not using the new flags. However, it will improve the performance considerably. So, if we want to add it to the call to execute_batch(), maybe we should give the flag a more generic name (like executemany_page_size) ?

lib/sqlalchemy/dialects/postgresql/psycopg2.py
ydinari: But what about the other two variables? I still need to override the constructor to get those

ydinari: And also, is this class guaranteed to have only one instance? If not it's problematic to use a class variable to hold the template

lib/sqlalchemy/dialects/postgresql/psycopg2.py
ydinari: ok

lib/sqlalchemy/dialects/postgresql/psycopg2.py
ydinari: ok

lib/sqlalchemy/dialects/postgresql/psycopg2.py
ydinari: ok

lib/sqlalchemy/dialects/postgresql/psycopg2.py
zzzeek: So there are the "other two variables", "dialect", and "multiple_rows", for which I don't see the purpose for these assignments.  The PGCompiler_psycopg2 object already has a .dialect instance variable you can access as "self.dialect".   The "inline" keyword argument is also available on the object as just that, "self.inline", there is no need to copy it to an additional instance variable.

For "execute_values_insert_template", it appears you only modify execute_values_insert_template in the generate_values_placeholders_str method.   You can safely default this to None at the class level.  When you later invoke "self.execute_values_insert_template = 'some value'", that sets an *instance level* variable which overrides the class-level default version.    SQLAlchemy uses this pattern all over the place:

```
Python 3.7.3 (default, May 11 2019, 00:38:04) 
[GCC 9.1.1 20190503 (Red Hat 9.1.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> class Foo:
...    some_arg = None
... 
>>> 
>>> f1 = Foo()
>>> f1.some_arg
>>> f2 = Foo()
>>> f2.some_arg = 'a'
>>> f3 = Foo()
>>> f3.some_arg = 'b'
>>> f1.some_arg
>>> f2.some_arg
'a'
>>> f3.some_arg
'b'

```


If I'm misreading or misinterpreting what I'm seeing here, that is very possible.    So let me know if I'm doing that here.

lib/sqlalchemy/dialects/postgresql/psycopg2.py
zzzeek: just that it has "executemany" in the name should be enough.

zzzeek:
send me a ping here when you want me to pull this through CI and review, thanks!

lib/sqlalchemy/dialects/postgresql/psycopg2.py
ydinari: OK. I missed the dialect variable, and didn't know about this behavior of class variables.

ydinari: As for multiple_rows, I wasn't sure that reading self.inline the correct way to get this info. But since you didn't comment about that, I guess it is :)

lib/sqlalchemy/dialects/postgresql/psycopg2.py
ydinari: ok

ydinari:
Hi, check out last commit! thanks

ydinari:
Hi,
Any progress with the review?
Thanks

zzzeek:
OK, i just looked and the changes look very nice, thanks for making them just like I said.   I still have to decide if I like those names, because they are a little cumbersome, but that's on me.   it may be that we remove the "psycopg2_" prefix perhaps.

we need to decide:

- it's easier to introduce this in 1.4 but that would mean it won't be available for a while.   This change is likely small enough to be in 1.3 but needs some verbiage around the switch from use_batch_mode to psycopg2_executemany_mode; this goes in the docs with a "..versionchanged" section and also I like to emit a deprecation warning for the old flag.   I have a loose policy that deprecations aren't introduced mid-minor release but as this is an "experimental" feature, a warning that the flag is "superseded" might not be so bad.

- if we do this in 1.3, in 1.4 we can perhaps default it to the highest performing option.    A new minor release is allowed to have bigger regressions when it first comes out (usually as betas, which nobody tests anyway) but that's a handy way for more dramatic behavioral changes to come out and get tested.

- I'm not sure if I saw test coverage for the ArgumentError part of this.  Also, I wonder if the pattern where they pass a string that gets converted to a symbol could somehow be a sqlalchemy.util function, and I feel like I want to look elsewhere in the codebase for this pattern too (oh, [it is](https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/pool/base.py#L194) , maybe a separate commit can make that improvement separately)

I think if you can confirm you really were hoping to have this in 1.3 that would set me up for the next step.

ydinari:
Hi,
* Yes I really would like it to be in 1.3 so we can start using it soon.
* Let me know what you decide about using the flag by default or not so I'll know what to put in the versionchanged section, and also change the flag's default value if needed.
* I didn't understand what you meant about an sqlalchemy.util function to check for correct argument for the flag. The link shows code doing pretty much the same kind of checks I do, and there's no call to a util function.
* Arguments tests added

zzzeek:
I will try to get this is in 1.3.7.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 715cfbea4d9f8f4e5ef239b0921a8dceb36ccbfb of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 715cfbea4d9f8f4e5ef239b0921a8dceb36ccbfb added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1355

zzzeek:
at this point it's faster if I take it into gerrit fully and add the remaining changelog notes and other doc tweaks etc.

zzzeek:
symbol parsing feature is in https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1405/

ydinari:
Cherry picked the argument parsing commit and used it.
Are you doing the doc changes?
Thanks

zzzeek:
ill base this one in and take over, thanks

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision c3d3a36f7eb66c86d14ed9c1c31b4b48bd204855 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset c3d3a36f7eb66c86d14ed9c1c31b4b48bd204855 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1355

lib/sqlalchemy/dialects/postgresql/psycopg2.py
```diff
@@ -52,11 +52,42 @@
 
     :ref:`psycopg2_unicode`
 
-* ``use_batch_mode``: This flag allows ``psycopg2.extras.execute_batch``
-  for ``cursor.executemany()`` calls performed by the :class:`.Engine`.
-  It is currently experimental but
-  may well become True by default as it is critical for executemany
-  performance.
+.. _psycopg2_executemany_mode:
```
sqla-tester: **mike bayer** (zzzeek) wrote:

I'm reworking the organization here since this is inside of a bulleted list stating all parameters,

lib/sqlalchemy/dialects/postgresql/psycopg2.py
```diff
@@ -551,13 +589,41 @@ def _log_notices(self, cursor):
...
+    # psycopg2 execute_values()
+    def generate_values_placeholders_str(
+            self, crud_params, returning_clause_exists):
+
+        # Currently not using psycopg2.execute_values() when there's a
+        # returning clause; need to add support
+        # for receiving multiple return values from insert query
+        # Note: self.inline is true iff there are multiple parameters sets to
+        # the query
+        if self.inline and not returning_clause_exists and \
```
sqla-tester: **mike bayer** (zzzeek) wrote:

is it guaranteed this will be an executemany()?

lib/sqlalchemy/dialects/postgresql/psycopg2.py
```diff
@@ -551,13 +589,41 @@ def _log_notices(self, cursor):
...
+                self.dialect.psycopg2_executemany_mode is EXECUTEMANY_VALUES:
+            self.execute_values_insert_template = "(" + \
```
sqla-tester: **mike bayer** (zzzeek) wrote:

this is the string we would generate anyway..right?   Why not have the base compiler just store this string in all cases if that's the case, then we don't need this method here.

sqla-tester (commented):

(5 comments)

Just some comments as I go through this.

* lib/sqlalchemy/dialects/postgresql/psycopg2.py (line 193): so the plan is that this gets deprecated since it's being superseded by a new system.

lib/sqlalchemy/dialects/postgresql/psycopg2.py
```diff
@@ -766,11 +867,22 @@ def on_connect(conn):
...
+
+            self._psycopg2_extras().execute_values(
+                cursor,
+                statement,
+                parameters,
+                template=context.compiled.execute_values_insert_template,
+                page_size=self.psycopg2_executemany_page_size)
+
+        else:  # EXECUTEMANY_VALUES of non-insert query, or EXECUTEMANY_BATCH
+            self._psycopg2_extras().execute_batch(cursor, statement,
```
sqla-tester: **mike bayer** (zzzeek) wrote:

why aren't we using pagesize here?

zzzeek:
hi -

to sum up changes I've made:

1. the biggest is that I think for now this is better as a short regular expression operation, which is done right in do_executemany(), rather than changing the base compiler class to farm out this hook.   It's a lot less code and doesn't add a funny looking concept to the base compiler that only matters for one specific DBAPI.  

2. I've changed the names of things, since we're in psycopg2 already we don't need "psycopg2_executemany_mode", it's just "executemany_mode".    The values then for this are None, "batch", and "values".

3. I've consolidated the unit tests into a single fixture, deprecated use_batch_mode

4. page_size applies to both execute_batch and execute_values

5. due to the above, I think we are safer deferring to psycopg2's defaults, even though 100 is quite low.  This is a power user feature so the docs clearly state that you should probably also set executemany_page_size too to a number like 10000 if you are using execute_values.

6. in an upcoming patch Im going to break page size into two variables for execute_batch and execute_values, as we have fallback logic here; I think 10000 is too high for execute_batch so these need to be separate

sqla-tester:

I think I'm happy with this now.  I'll try to merge tomorrow if I have time unless you have suggestions.

lib/sqlalchemy/dialects/postgresql/psycopg2.py
ydinari: Do you mean when self.inline is true? This is what I deduced from the code, I hope that's right

lib/sqlalchemy/dialects/postgresql/psycopg2.py
ydinari: We can, I didn't want to change the behavior for people not using the new flag.
If you want I'll add it

ydinari:
Hi, so where do I see the changes you made?
And do you need anything from me?

zzzeek:
> Hi, so where do I see the changes you made?
> And do you need anything from me?

oh, OK so this PR is mirrored in my gerrit server at https://gerrit.sqlalchemy.org/1413 and one of the main reasons I do that is so that I can also push to it, so it has moved past the PR now.   i just need you to flag anything that isn't going to work for you at this point.

sqla-tester:

Code-Review+2 Workflow+1

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1355 has been **merged**. Congratulations! :)

sqla-tester:

Code-Review+2 Workflow+1

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1413 has been **merged**. Congratulations! :)

ydinari:
Hi,
Great!
When will 1.3.7 be released?
Also, looking at the new implementation of do_executemany(), I see there is no check whether the query contains a RETURNING clause. Is that ok, because currently sqlalchemy doesn't generate insert statements with both multiple parameters and RETURNING?
BTW, what kind of account do you have with Vonage?
Thanks,
Yuval

zzzeek:
it's released.    Right now there aren't any codepaths that will automatically use a "returning" statement with executemany(), and the current behavior is that fetchone() doesn't work for these in any case since no DBAPI supports fetching rows after executemany.    I have a residential account with vonage.

ydinari:
Hi,
There is a problem with the regular expression used in do_executemany - it doesn't work if the query contains a newline character (which it does in our case).
Possible solutions is write it like this:
`_insert_values_match = re.compile(r".* VALUES (\(.+\))", re.DOTALL).match`
Or:
`_insert_values_search = re.compile(r" VALUES (\(.+\))").search`
Do you want me to open a new PR or will you do it?
Thanks

zzzeek:
pretty unfortunate timing!   as always, the test case is what's needed.

how are you creating the insert with the newline ?   it looks like if things like CTEs are there and such which could add newlines.

if you can send a PR that includes tests then I can get that in.

can we think of any scenario where the " VALUES ()" could appear twice in a single INSERT statement ?

zzzeek:
also can you confirm the current behavior is that it's degrading to execute_batch, right?  eg. it isn't failing, just isn't using the optimization.

zzzeek:
also, I will need you to fully sign off and test this *completely* in your environment before I can release again.    I don't usually put features this significant in a point release.  it is very risky.

zzzeek:
I've created new tests and revised the approach to be more like the approach you orgiinally had in https://gerrit.sqlalchemy.org/#/c/1419/ .   Pretty sure this one will work consistently.

ydinari:
Hi,
OK, I'll test it in our environment. On which branch should I be?
As for what you asked:
- In our every sql statement starts with some comment after which there's a newline. But in the general case I think sql statements may include newline anywhere.
- I don't know whether sql syntax allows multiple VALUES keyword in a single statement.
- Yes, the behavior is that execute_batch() is called (but then the executemany_values_page_size flag is ignored).

ydinari:
Hi,
It works in my environment.
There's an issue you decide if important or not: the debug log show the query as it is before the replacement of the VALUES clause, which can be confusing.
Thanks

zzzeek:
> Hi,
> It works in my environment.
> There's an issue you decide if important or not: the debug log show the query as it is before the replacement of the VALUES clause, which can be confusing.
> Thanks

you mean, it shows the regular INSERT query with a single VALUES clause, right ?  I guess earlier when we did this up at the compiler phase, it would show the modified query, but then it wouldn't show the template at all.  psycopg2 is not unique in this particular optimization, they just make it something the end-user has to do explicitly, if you are using for example mysqlclient, they are doing it [here](https://github.com/PyMySQL/mysqlclient-python/blob/master/MySQLdb/cursors.py#L21) and [here](https://github.com/PyMySQL/mysqlclient-python/blob/master/MySQLdb/cursors.py#L229), which IMHO is what psycopg2 should have done for us rather than burdening us with all this explicitness, but I can certainly relate to when a maintainer doesn't want to take on a complex new behavior.

anyway the point is that MySQL is doing this same thing and we can't see that either so we can likely get away with considering this to be a driver detail.

zzzeek:
as far as newlines, yes a SQL statement can have newlines, but the assumption I had was that these were INSERTs created by the compiler, which don't have newlines in them.  However, as my test in the patch demonstrates, adding a scalar SELECT as one of the values will put a newline right in the middle of the VALUES which makes the regexp approach not really work (even mysql's approach wouldn't work with that).

zzzeek:
and yes if you have an event handler adding comments, that too.   Is that how you are adding comments ?    The new patch includes a test for the case where someone's event handler has modified the query beyond where we know how to find the VALUES anymore.  this should not be the norm, but if an event handler really rewrote the whole query, it's going to degrade for now.

zzzeek:
OK I'm going to merge it..take two!

zzzeek:
after a few more tweaks, forgot to remove the regex..etc

ydinari:
In our code there's an `@event.listens_for(Engine, "before_cursor_execute", retval=True)` decorated method the prepend a comment with a newline to the existing statement.
So this will be release 1.3.8?

zzzeek:
yup 1.3.8 is next

ydinari:
Hi,
When will it be released?
Thanks

zzzeek:
was released a few days ago, all good?

PR #4769, opened 2019-07-17
Small text error in json field doc

kataev:
Small fix json field doc


zzzeek:
haha I added extra options and never changed "two"

zzzeek:
thank you!

PR #4792, opened 2019-08-01
type_ck integration

### LINKED ISSUE #4784: Default MetaData kwarg naming_convention can create issues with certain databases
* this is a docs ticket; assign to @jvanasco 
* reference group discussion: https://groups.google.com/forum/#!topic/sqlalchemy/8xOu-Mlcobo

The alembic docs reference this naming convention: https://alembic.sqlalchemy.org/en/latest/naming.html
```
convention = {
  "ix": "ix_%(column_0_label)s",
  "uq": "uq_%(table_name)s_%(column_0_name)s",
  "ck": "ck_%(table_name)s_%(constraint_name)s",
  "fk": "fk_%(table_name)s_%(column_0_name)s_%(referred_table_name)s",
  "pk": "pk_%(table_name)s"
}

metadata = MetaData(naming_convention=convention)
```

a Boolean/Enum column will try to create a constraint on sqlite by default, but `constraint_name` is not defined. this will cause an error in `Engine.create_all(`

Tasks:

[ ] have the docs use custom callables to offer an automatic constraint name.
[ ] long term: descope into another issue a fallback option for constraint name. 

see referenced discussion for details of above.

### LINKED ISSUE #4790: Improve error messages on create_table
There are a handful of messages that could be improved.

1. `sqlalchemy.exc.InvalidRequestError: Naming convention including %(constraint_name)s token requires that constraint is explicitly named.`  This doesn't give much of a context.  

Suggestion for improvement:  Extend message with debug info such as:

    Naming convention including %(constraint_name)s token requires that constraint is explicitly named. Constraint Identifiers: Table- `user`; Column(s)- `x`.

2. `CheckConstraint('z > 10')` raises `IndexError: list index out of range` ; this could be caught with a similar message to the above.

debug info can be generated within `ConventionDict` via:

            _debug_info = ' Constraint Identifiers: '
            try:
                _debug_info += "Table- `%s`;" % self.const.table.name
                _columns = ["`%s`" % i.name for i in self.const.columns]
                _debug_info += " Column(s)- %s." % ", ".join(_columns)
            except:
                pass

jvanasco:
This is the first attempt at #4784 and #4790. I expect requested changes on the error messages at the very least, but wanted to post it for review.

### Description

* type_ck has been added
* `_prefix_dict` was renamed `_base_prefix_dict` along with a note to better describe how prefix overrides work (such as 'ck' -> 'type_ck')
* a small change to `_get_prefixes`: only use the `type_` prefix if no name was presented. this was more backwards compatible after writing some test cases.
* test cases! test_metadata has a lot of rewrites and additions to NamingConventionsTest , starting with `test_schematype_ck_name_boolean` and all the way down to `test_schematype_no_ck_name_boolean_no_name`
* newly added- `ConventionDict.debug_info(`
* handled edgecase & test for a constraint with no columns, but a column_name token. improved error message, also added  "debug_info".
* improved error message for having convention that requires a constraint name, but no constraint name available; also added  "debug_info".

### Checklist
This pull request is:
- [x] A short code fix
        - Fixes: #4784
        - Fixes: #4790

### Of Importance...

I added a docstring to each test describing the use-case.  These were either explicit from the various tickets or inferred by my best guesses.  They need review.

I also added 2 tests for dealing with PEP-435 enums. The sqlalchemy docs note the classname will be used as the constraint_name, however there were no previous tests concerning this.

### DIFF
--- a/doc/build/core/constraints.rst
+++ b/doc/build/core/constraints.rst
@@ -454,11 +454,11 @@ event-based approach is included, and can be configured using the argument
 the :class:`.Index` class or individual :class:`.Constraint` classes as keys,
 and Python string templates as values.   It also accepts a series of
 string-codes as alternative keys, ``"fk"``, ``"pk"``,
-``"ix"``, ``"ck"``, ``"uq"`` for foreign key, primary key, index,
-check, and unique constraint, respectively.  The string templates in this
-dictionary are used whenever a constraint or index is associated with this
-:class:`_schema.MetaData` object that does not have an existing name given (including
-one exception case where an existing name can be further embellished).
+``"ix"``, ``"ck"``, ``"type_ck"``, ``"uq"`` for foreign key, primary key, index,
+check, "type bound" check, and unique constraint, respectively.  The string templates
+in this dictionary are used whenever a constraint or index is associated with this
+:class:`._schema.MetaData` object that does not have an existing name given
+(including one exception case where an existing name can be further embellished).
 
 An example naming convention that suits basic cases is as follows::
 
@@ -466,6 +466,7 @@ An example naming convention that suits basic cases is as follows::
       "ix": 'ix_%(column_0_label)s',
       "uq": "uq_%(table_name)s_%(column_0_name)s",
       "ck": "ck_%(table_name)s_%(constraint_name)s",
+      "type_ck": "ck_%(table_name)s_%(column_0_name)s",
       "fk": "fk_%(table_name)s_%(column_0_name)s_%(referred_table_name)s",
       "pk": "pk_%(table_name)s"
     }
@@ -683,8 +684,20 @@ one column present, the scan does use a deterministic search, however the
 structure of the expression will determine which column is noted as
 "column zero".
 
+Some backends do not natively support core SqlAlchemy datatypes like
+:class:`.Boolean` and  :class:`.Enum`, so SqlAlchemy enforces standard
+behaviors by automatically generating a CHECK constraint.  These "Type Bound"
+column types will prefer the specialized ``type_ck`` naming convention when
+available and needed.  The ``type_ck`` prefix is only used for auto-generation; 
+if a constraint name is explicitly provided, the naming convention with a ``ck``
+prefix will be used.  This topic is covered in the next section.
+
 .. versionadded:: 1.0.0 The :class:`.CheckConstraint` object now supports
    the ``column_0_name`` naming convention token.
+
+.. seealso::
+
+ :ref:`_naming_schematypes`
 
 .. _naming_schematypes:
 
@@ -730,22 +743,35 @@ MySQL.
 
 The CHECK constraint may also make use of the ``column_0_name`` token,
 which works nicely with :class:`.SchemaType` since these constraints have
-only one column::
+only one column.  The ``column_0_name`` token is the recommended token when
+dealing with auto-generated "type bound" constraints if you do not want to
+explicitly name them::
 
     metadata = MetaData(
-        naming_convention={"ck": "ck_%(table_name)s_%(column_0_name)s"}
+        naming_convention={"ck": "ck_%(table_name)s_%(constraint_name)s",
+                           "type_ck": "ck_%(table_name)s_%(column_0_name)s",
+                           }
     )
 
     Table('foo', metadata,
-        Column('flag', Boolean())
+        Column('flag_explicit', Boolean(name='explicit_flag_constraint')),
+        Column('flag_auto', Boolean()),
     )
 
 The above schema will produce::
 
     CREATE TABLE foo (
-        flag BOOL,
-        CONSTRAINT ck_foo_flag CHECK (flag IN (0, 1))
+        flag_explicit BOOL,
+        flag_auto BOOL,
+        CONSTRAINT ck_foo_explicit_flag_constraint CHECK (flag_explicit IN (0, 1)),
+        CONSTRAINT ck_foo_flag_auto CHECK (flag_auto IN (0, 1)),
     )
+
+Note how the type-bound :class:`.Boolean` created a contraint name using  the
+"type_ck" naming convention template when no constraint name was provided, and
+created a constraint with the "ck" naming convention when a name was provided.
+
+.. versionchanged:: 1.3 Introduced the special ``type_ck`` prefix.
 
 .. versionchanged:: 1.0 Constraint naming conventions that don't include
    ``%(constraint_name)s`` again work with :class:`.SchemaType` constraints.
--- a/test/sql/test_metadata.py
+++ b/test/sql/test_metadata.py
@@ -4911,8 +4911,10 @@ class NamingConventionTest(fixtures.TestBase, AssertsCompiledSQL):
 
         assert_raises_message(
             exc.InvalidRequestError,
-            r"Naming convention including %\(constraint_name\)s token "
-            "requires that constraint is explicitly named.",
+            r'Naming convention including "%\(constraint_name\)s" token '
+            r"requires that constraint is explicitly named. "
+            r'Check naming convention "ck_%\(table_name\)s_%\(constraint_name\)s", '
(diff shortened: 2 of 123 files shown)


doc/build/core/constraints.rst
```diff
@@ -697,20 +703,22 @@ normally by using a convention which includes ``%(constraint_name)s``
 and then applying a name to the type::
 
     metadata = MetaData(
-        naming_convention={"ck": "ck_%(table_name)s_%(constraint_name)s"}
+        naming_convention={"type_ck": "type_ck_%(table_name)s_%(constraint_name)s"}
```
zzzeek: I think I might want to keep the rendered example prefix as "ck_", because, when database identifiers get too long, things start getting weird.   at this point we have good truncation logic for these names but especially on a backend like Oracle names are truncated very short.

doc/build/core/constraints.rst
```diff
@@ -725,20 +733,32 @@ which works nicely with :class:`.SchemaType` since these constraints have
 only one column::
 
     metadata = MetaData(
-        naming_convention={"ck": "ck_%(table_name)s_%(column_0_name)s"}
+        naming_convention={"ck": "ck_%(table_name)s_%(constraint_name)s",
```
zzzeek: I'd change the code formatting here, if you want, we can just pull this into gerrit and I can push additional tweaks and whatnot.

test/sql/test_metadata.py
```diff
@@ -4755,16 +4792,52 @@ def test_schematype_ck_name_boolean_not_on_name(self):
...
+        eq_(
+            [c for c in u1.constraints if isinstance(c, CheckConstraint)][
+                0
+            ].name,
+            "_unnamed_",
+        )
+        # but is hit at compile time
+        assert_raises_message(
+            exc.InvalidRequestError,
+            r"Naming convention including \%\(constraint_name\)s token "
```
zzzeek: OK so, more monkey wrenches, when the user gets this error, the thing they likely *want* to do is add a "type_ck" convention to their dictionary, because now they can!   So the message should refer to that.   Since this is completely confusing, you an also add a "code" to the exception to point them to a doc blurb at https://docs.sqlalchemy.org/en/13/errors.html that explains the whole story.     this one is a real quagmire.

test/sql/test_metadata.py
```diff
@@ -4755,16 +4792,52 @@ def test_schematype_ck_name_boolean_not_on_name(self):
...
+            r"requires that constraint is explicitly named."
+            r" Constraint Identifiers: Table- `user`; Column\(s\)- `x`.",
```
zzzeek: also I'd put the "table" / "column" info more in sentence format above , "naming convention "ck_%(foo)s_%(bar)s" failed to apply to CHECK constraint on table X with column Q since the constraint has no name, and the convention uses the "%(constraint_name)s" token.     Consider using the "type_ck" naming convention for type-bound CHECK constraints so that no name is necessary.   more inforamtion at https://sqlalche.me/foobar"

test/sql/test_metadata.py
```diff
@@ -4755,16 +4792,52 @@ def test_schematype_ck_name_boolean_not_on_name(self):
...
+            schema.CreateTable(u1).compile,
+            dialect=default.DefaultDialect(),
         )
 
+    def test_schematype_ck_name_enum(self):
+        """If a CheckConstraint is EXPLICITLY declared for a Enum column, it
+        should use the `ck_` prefix and not the `type_ck`."""
+        m1 = MetaData(
+            naming_convention={"ck": "ck_%(table_name)s_%(constraint_name)s",
```
zzzeek: have you tried out the [pre-commit](https://pre-commit.com/) tool?   Like everything, took me awhile to be open towards it but all of my commits are auto-black formatted now, which is critical because otherwise they get auto-formatted in a later commit and make a mess.

test/sql/test_metadata.py
```diff
@@ -4780,11 +4853,190 @@ def test_schematype_ck_name_enum(self):
...
+                               "type_ck": "type_ck_%(table_name)s_%(column_0_name)s",
+                               }
+        )
+        u1 = Table("user", m1, Column("x", Enum("a", "b")))
+        # constraint is not hit
+        eq_(
+            [c for c in u1.constraints if isinstance(c, CheckConstraint)][
+                0
+            ].name,
+            "_unnamed_",
```
zzzeek: I hate the "_unnamed_" thing, can we maybe just do an all 1.4 commit here that whacks that whole "_defer_none_name" thing (assuming this is possible, I want to have a default "type_ck" convention to help with this) ?   I can work on that since it's my mess to clean up.

zzzeek (commented):
I put some thoughts in but I havent gone through all of it yet.     this is a hard one because the naming thing is so weird.

doc/build/core/constraints.rst
jvanasco: let me handle most of this in github for a bit first. i still have issues with gerrit and I'll lose too much time on getting up-to-speed.

test/sql/test_metadata.py
jvanasco: will do!

test/sql/test_metadata.py
jvanasco: will do!

test/sql/test_metadata.py
jvanasco: i can git it a shot. if not, i'll just run black manually.

test/sql/test_metadata.py
jvanasco: 1. I dislike the unnamed thing too.
2. I'm not sure `_defer_none_name` can be dropped.  if it can, I think it will be a lot of work and break many test cases.  a relevant test-case is the new one I added for your edge-case of columnless constraints -- `[test_schematype_ck_no_columns_on_constraint](https://github.com/sqlalchemy/sqlalchemy/pull/4792/files/6182721f8556a08485f1b089f24cf30f4705d1ed#diff-85c40f646c916a1455f90ca632061d28R5006-R5035)` (you may need to scroll all the way to the bottom; github is highlighting the code but not autoscrolling to the section)

recreated here...

            u1 = Table(
                "user", m1,
                Column("x", Integer),
                CheckConstraint('z > 10')
            )

This will automatically raise the error on `Table` - instead of `.compile` like the other cases - because the constraint name doesn't exist but it also isn't deferred.

doc/build/core/constraints.rst
jvanasco: ok.

test/sql/test_metadata.py
jvanasco: @zzzeek where are the short urls maintained?

zzzeek: you make up a random four letter code, you add it as "code='ABCD'" to the InvalidRequestError() or whatever it is, then you write it in https://github.com/sqlalchemy/sqlalchemy/blob/master/doc/build/errors.rst

jvanasco:
integrated the changes and updated some more error messages.

i didn't run black on a few files (notably tests, but IIRC also schema) because it was clear black had not been run on them before, and the actual changes would have been hidden.

one of the bigger changes in this, is tracking the "active convention", which is really an active naming convention template.  this is required for the improved error messages.

LouisTrezzini:
This looks great! Do you have any plans to have this feature merged in an upcoming release?

zzzeek:
someday... :)   I really have a backlog

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision e407db8c3b59e93d5823f224a05ab0f32a463b5b of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Failed to create a gerrit review, git squash against branch 'master' failed

zzzeek:
hi @jvanasco , do a rebase on this and I can pull it into gerrit.

jvanasco:
Closing this because the merging/etc is a nightmare. I will generate a new PR and pull it into gerrit.

PR #4801, opened 2019-08-09
New cx_Oracle dialect connection option to handle text encoding errors

### LINKED ISSUE #4799: cx_Oracle dialect does not offer options to handle corrupt data
This issue has been described in details at https://github.com/oracle/python-cx_Oracle/issues/343
In summary, there are cases when Oracle contains "corrupt data", meaning data which cannot be encoded with UTF8 properly. This should be rare, but when it happens, then the current behavior of SQLAlchemy+cx_Oracle to error out. See error below.

Code:
```
sql = f"SELECT {_fields} FROM {self.schema}.{self.name} {_where}"
for r in self.edw.engine.execute(sql):
```
Error thrown when hitting corrupted chars:
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xed in position 8: invalid continuation byte
```

It would be preferable if SQLAlchemy would allow different handlings of this string encoding issue, just like it is standard in Python (eg "replace", "xmlcharrefreplace", etc. see https://docs.python.org/3/library/codecs.html#error-handlers).

Luckily cx_Oracle allows for this, see https://cx-oracle.readthedocs.io/en/latest/user_guide/sql_execution.html#querying-corrupt-data

BUT to take advantage of these options, SQLAlchemy's cx_Oracle dialect would need to be enhanced, by allowing a new setting for the selection of the strategy to handle these encoding errors.

I am going to make a PR for it, if that's okey.

How the new functionality would work:
1. There would be a new `create_engine()` parameter in https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/dialects/oracle/cx_oracle.py named `text_encoding_errors`
2. This parameter's default value would be None for backwards compatibility
3. The parameter's value would be supplied to the `encodingErrors` parameter of cx_Oracle's `cursor.var()`, see https://cx-oracle.readthedocs.io/en/latest/user_guide/sql_execution.html#querying-corrupt-data
4. The possible values of this parameter would be of the standard Python text encoding error handlers, see https://docs.python.org/3/library/codecs.html#error-handlers

zoltan-fedor:
Added a new connection option named `text_encoding_errors` to allow the handling of text encoding errors when using the cx_Oracle dialect.

Fixes: #4799 

### Description
There are cases when Oracle contains "corrupt data", meaning data which cannot be encoded with UTF8 properly. This should be rare, but when it happens, then the current behaviour of SQLAlchemy+cx_Oracle to error out.
To allow different handling this error the above connection option has been added

See more detials https://github.com/sqlalchemy/sqlalchemy/issues/4799

Unfortunately testing corrupted UTF8 characters is very hard (how do you insert them into the database in the first place?), so I have only included a simple test. Anyone is welcome to add a "full" test if figures out how to insert corrupted characters into Oracle.

### Checklist

This pull request is:

- [ ] A short code fix
- [ x ] A new feature implementation
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/oracle/cx_oracle.py
+++ b/lib/sqlalchemy/dialects/oracle/cx_oracle.py
@@ -78,6 +78,11 @@ The parameters accepted by the cx_oracle dialect are as follows:
 
 * ``coerce_to_decimal`` - see :ref:`cx_oracle_numeric` for detail.
 
+* ``text_encoding_errors`` - set the text encoding to use in case of a
+  text encoding error, see
+  https://docs.python.org/3/library/codecs.html#error-handlers, defaults
+  to None. This option is only available above cx_Oracle version 6.4.
+
 .. _cx_oracle_unicode:
 
 Unicode
@@ -754,6 +759,7 @@ class OracleDialect_cx_oracle(OracleDialect):
         coerce_to_decimal=True,
         arraysize=50,
         threaded=None,
+        text_encoding_errors=None,
         **kwargs
     ):
 
@@ -764,6 +770,7 @@ class OracleDialect_cx_oracle(OracleDialect):
         self.auto_convert_lobs = auto_convert_lobs
         self.coerce_to_unicode = coerce_to_unicode
         self.coerce_to_decimal = coerce_to_decimal
+        self.text_encoding_errors = text_encoding_errors
         if self._use_nchar_for_unicode:
             self.colspecs = self.colspecs.copy()
             self.colspecs[sqltypes.Unicode] = _OracleUnicodeStringNCHAR
@@ -814,6 +821,12 @@ class OracleDialect_cx_oracle(OracleDialect):
                 self._returningval = self._paramval
 
         self._is_cx_oracle_6 = self.cx_oracle_ver >= (6,)
+        
+        if self.text_encoding_errors and self.cx_oracle_ver < (6, 4):
+            raise exc.InvalidRequestError(
+                "only cx_Oracle version 6.4 and above supports "
+                "the 'text_encoding_errors' parameter"
+            )
 
     def _parse_cx_oracle_ver(self, version):
         m = re.match(r"(\d+)\.(\d+)(?:\.(\d+))?", version)
@@ -919,9 +932,11 @@ class OracleDialect_cx_oracle(OracleDialect):
                         size,
                         cursor.arraysize,
                         outconverter=outconverter,
+                        encodingErrors=dialect.text_encoding_errors
                     )
                 else:
-                    return cursor.var(util.text_type, size, cursor.arraysize)
+                    return cursor.var(util.text_type, size, cursor.arraysize,
+                                      encodingErrors=dialect.text_encoding_errors)
 
             elif dialect.auto_convert_lobs and default_type in (
                 cx_Oracle.CLOB,
@@ -936,6 +951,7 @@ class OracleDialect_cx_oracle(OracleDialect):
                         size,
                         cursor.arraysize,
                         outconverter=lambda value: outconverter(value.read()),
+                        encodingErrors=dialect.text_encoding_errors
                     )
                 else:
                     return cursor.var(
@@ -943,6 +959,7 @@ class OracleDialect_cx_oracle(OracleDialect):
                         size,
                         cursor.arraysize,
                         outconverter=lambda value: value.read(),
+                        encodingErrors=dialect.text_encoding_errors
                     )
 
             elif dialect.auto_convert_lobs and default_type in (
--- a/test/dialect/oracle/test_types.py
+++ b/test/dialect/oracle/test_types.py
@@ -731,6 +731,13 @@ class TypesTest(fixtures.TestBase):
         value = testing.db.scalar("SELECT 'hello' FROM DUAL")
         assert isinstance(value, util.text_type)
 
+    @testing.only_on("oracle+cx_oracle", "cx_oracle-specific feature")
+    @testing.provide_metadata
+    def test_text_encoding_errors(self):
+        engine = testing_engine(options=dict(text_encoding_errors="replace"))
+        value = engine.scalar("SELECT 'Hello' FROM DUAL")
+        assert value == 'Hello'
+
     @testing.provide_metadata
     def test_reflect_dates(self):
         metadata = self.metadata


zzzeek:
hi there- 

this is kind of a thorny issue for a few reasons.

first off, for testing, I would imagine you put some non-valid utf-8 data into a binary field or plain VARCHAR, then select a CAST of it as a unicode string.    

I think testing is important because this PR does not seem like it will work in any case, because we aren't using cx_Oracle's faciliies for decoding UTF-8, you'll note the code is sending in an "outconverter" referencing SQLAlchemy's own unicode converters, which have been observed to be much more performant than that of cx_Oracle, so the "errors" parameter would go there, instead.

But also, the whole concept of handling "unicode_errors" has been deprecated in SQLAlchemy 1.3, as you can see in 4c2c2c40fde17c85013e00a6f3303, however this is largely because DBAPIs tend to do all the "unicode" for us now, and SQLAlchemy wants to get out of this business.  But!  for cx_oracle we are still doing it, so i can still consider having a flag for cx_Oracle directly.

Overall though, another reason this parameter was removed is that it is a very odd case and one in which you would imagine someone would use direct DBAPI access or at least custom types to resolve.    I would assume you have this situation in production for you to have interest in this feature, can you provide more detail about your use case?

zzzeek:
just looked more closely (sorry very short on time today), the SQLAlchemy unicode handlers are only used under Python 2, not 3.  So we'd need to add this flag all over the place.   Definitely needs a way to be tested though.

lib/sqlalchemy/dialects/oracle/cx_oracle.py
```diff
@@ -919,9 +932,11 @@ def output_type_handler(
                         size,
                         cursor.arraysize,
                         outconverter=outconverter,
+                        encodingErrors=dialect.text_encoding_errors
```
zzzeek: this won't have an effect here because we have the outconverter above.

lib/sqlalchemy/dialects/oracle/cx_oracle.py
```diff
@@ -919,9 +932,11 @@ def output_type_handler(
...
                     )
                 else:
-                    return cursor.var(util.text_type, size, cursor.arraysize)
+                    return cursor.var(util.text_type, size, cursor.arraysize,
+                                      encodingErrors=dialect.text_encoding_errors)
```
zzzeek: it will work here though.

lib/sqlalchemy/dialects/oracle/cx_oracle.py
```diff
@@ -936,13 +951,15 @@ def output_type_handler(
                         size,
                         cursor.arraysize,
                         outconverter=lambda value: outconverter(value.read()),
+                        encodingErrors=dialect.text_encoding_errors
```
zzzeek: wont work here...

lib/sqlalchemy/dialects/oracle/cx_oracle.py
```diff
@@ -936,13 +951,15 @@ def output_type_handler(
...
                     )
                 else:
                     return cursor.var(
                         default_type,
                         size,
                         cursor.arraysize,
                         outconverter=lambda value: value.read(),
+                        encodingErrors=dialect.text_encoding_errors
```
zzzeek: might not work here, not sure.

lib/sqlalchemy/dialects/oracle/cx_oracle.py
```diff
@@ -78,6 +78,11 @@
 
 * ``coerce_to_decimal`` - see :ref:`cx_oracle_numeric` for detail.
 
+* ``text_encoding_errors`` - set the text encoding to use in case of a
```
zzzeek: let's call it "encoding_errors" to more closely match cx_oracle's name, and also link to cx_Oracle's documentation here since it's their option.    Also needs a note that under Python 2 we are using it in our own decoder.

lib/sqlalchemy/dialects/oracle/cx_oracle.py
```diff
@@ -814,6 +821,12 @@ def _returningval(value):
                 self._returningval = self._paramval
 
         self._is_cx_oracle_6 = self.cx_oracle_ver >= (6,)
+        
+        if self.text_encoding_errors and self.cx_oracle_ver < (6, 4):
```
zzzeek: not under Python 2, we do the decode ourselves and can still do this.    I'd skip the check here and let it fail at the cx_Oracle level.

lib/sqlalchemy/dialects/oracle/cx_oracle.py
```diff
@@ -919,9 +932,11 @@ def output_type_handler(
                         size,
                         cursor.arraysize,
                         outconverter=outconverter,
+                        encodingErrors=dialect.text_encoding_errors
                     )
                 else:
-                    return cursor.var(util.text_type, size, cursor.arraysize)
+                    return cursor.var(util.text_type, size, cursor.arraysize,
```
zzzeek: this will fail for cx_Oracle < 6.4 right?  we can't send the flag at all if it isn't in the API.

zoltan-fedor:
Hi there,
Thanks for reviewing the PR.
I definitely agree that it will need better tests. If you think that this PR has a chance to get approved, then I will work on the tests and will try to somehow get some corrupted data in Oracle to test.

You asked about the use case. We have a data replicator built to load data from Oracle and load it into another DB. This moves billions of rows of data across dozes on tables and it works fine for UTF8, but at a single table it failed with the error you can see at https://github.com/oracle/python-cx_Oracle/issues/343#issuecomment-520006960
It turns out that it fails because it has faced some corrupted data in that table - one it cannot encode to UTF8.

In Python text encoding errors can be handled multiple ways - like (1) throwing an error or (2) replacing the unknown character with a questionmark, etc (see https://docs.python.org/3/library/codecs.html#error-handlers). Each of these are a reasonable strategy depending on the needs of the given application, so even if DBAPI is handling the "unicode" for SQLAlchemy, somewhere the user of SQLAlchemy needs to be able to tell DBAPI which of these strategies he would like to be used.
This is what cx_Oracle allows with the encodingErrors parameter (see https://cx-oracle.readthedocs.io/en/latest/user_guide/sql_execution.html#querying-corrupt-data).
I think this option should also be made available via SQLAlchemy, otherwise the programmer has no option to pick which strategy he would like to be used for encoding errors.

But you are right, this is scenario (corrupted data), which should be relatively rare, so I fully understand if the SQLAlchemy library decides to ignore it.

In my case I have already monkey-patched SQLAlchemy by adding the encodingErrors parameter to https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/dialects/oracle/cx_oracle.py#L924 and that has fixed the issue we were seeing (we are using Python 3.7 with Oracle 11.2 and there that monkey patch has worked).

zzzeek:
OK so if Oracle in fact is capable of storing invalid data, then the test should make that happen and then select it back.   as I said, SQLAlchemy has "unicode_errors" as part of it's API for this purpose, originally because a MySQL user wanted it, another database (now with the same parent company!) capable of storing broken strings.   But SQLAlchemy does not want to be in the unicode decoding business as the DBAPIs should be doing that, and for cx_Oracle I would have preferred if this flag was made at the connect() level, not for cursor.var().    But that is likely because the whole cursor.var() thing overall is insane compared to any other DBAPI where this highly intricate process for every datatype exposed to the calling application is entirely unnecessary.

Anyway, since SQLAlchemy is sending unicode decode directives to cx_Oracle and for Python 2 at least we are still in the business of decoding ourselves, we can have the flag for the cx_Oracle dialect.    but we have four different cursor.var() calls where this flag has to be used differently, so we really need a test suite that tests it and also the whole thing has to use the flag conditionally with cursor.var() because older cx_Oracle versions don't support it and we seem to still be supporting back to cx_Oracle 5 for the moment.

zzzeek:
I got a test case to work for VARCHAR.  It doesn't work for CLOB.  But do you have the resources to integrate the following patch ?  it includes the implementation and the tests.   otherwise I can put it in gerrit and just have you review / test it on your end.

```
diff --git a/lib/sqlalchemy/dialects/oracle/cx_oracle.py b/lib/sqlalchemy/dialects/oracle/cx_oracle.py
index 90d049aa08..a791f6a219 100644
--- a/lib/sqlalchemy/dialects/oracle/cx_oracle.py
+++ b/lib/sqlalchemy/dialects/oracle/cx_oracle.py
@@ -753,12 +753,14 @@ class OracleDialect_cx_oracle(OracleDialect):
         coerce_to_unicode=True,
         coerce_to_decimal=True,
         arraysize=50,
+        encoding_errors=None,
         threaded=None,
         **kwargs
     ):
 
         OracleDialect.__init__(self, **kwargs)
         self.arraysize = arraysize
+        self.encoding_errors = encoding_errors
         if threaded is not None:
             self._cx_oracle_threaded = threaded
         self.auto_convert_lobs = auto_convert_lobs
@@ -815,6 +817,14 @@ class OracleDialect_cx_oracle(OracleDialect):
 
         self._is_cx_oracle_6 = self.cx_oracle_ver >= (6,)
 
+    @property
+    def _cursor_var_unicode_kwargs(self):
+        if self.encoding_errors and self.cx_oracle_ver >= (6, 4):
+            # TODO: if paramter is set and we aren't on 6, 4, emit a warning
+            return {"encodingErrors": self.encoding_errors}
+        else:
+            return {}
+
     def _parse_cx_oracle_ver(self, version):
         m = re.match(r"(\d+)\.(\d+)(?:\.(\d+))?", version)
         if m:
@@ -912,7 +922,7 @@ class OracleDialect_cx_oracle(OracleDialect):
             ):
                 if compat.py2k:
                     outconverter = processors.to_unicode_processor_factory(
-                        dialect.encoding, None
+                        dialect.encoding, errors=dialect.encoding_errors
                     )
                     return cursor.var(
                         cx_Oracle.STRING,
@@ -921,7 +931,12 @@ class OracleDialect_cx_oracle(OracleDialect):
                         outconverter=outconverter,
                     )
                 else:
-                    return cursor.var(util.text_type, size, cursor.arraysize)
+                    return cursor.var(
+                        util.text_type,
+                        size,
+                        cursor.arraysize,
+                        **dialect._cursor_var_unicode_kwargs
+                    )
 
             elif dialect.auto_convert_lobs and default_type in (
                 cx_Oracle.CLOB,
@@ -929,7 +944,7 @@ class OracleDialect_cx_oracle(OracleDialect):
             ):
                 if compat.py2k:
                     outconverter = processors.to_unicode_processor_factory(
-                        dialect.encoding, None
+                        dialect.encoding, errors=dialect.encoding_errors
                     )
                     return cursor.var(
                         default_type,
@@ -943,6 +958,7 @@ class OracleDialect_cx_oracle(OracleDialect):
                         size,
                         cursor.arraysize,
                         outconverter=lambda value: value.read(),
+                        **dialect._cursor_var_unicode_kwargs
                     )
 
             elif dialect.auto_convert_lobs and default_type in (
diff --git a/test/dialect/oracle/test_types.py b/test/dialect/oracle/test_types.py
index af9719130a..9a98662bdf 100644
--- a/test/dialect/oracle/test_types.py
+++ b/test/dialect/oracle/test_types.py
@@ -883,6 +883,31 @@ class TypesTest(fixtures.TestBase):
         finally:
             testing.db.execute("DROP TABLE Z_TEST")
 
+    @testing.provide_metadata
+    def test_broken_unicode_fetch(self):
+        e = testing_engine(options={"encoding_errors": "ignore"})
+
+        from sqlalchemy.util import ue
+
+        data = ue("\uee2c\u9a66")  # this is u"\uee2c\u9a66"
+
+        utf8_w_errors = data.encode('utf-16').decode('utf-8', 'ignore')
+
+        with e.connect() as conn:
+            result = conn.execute(
+                "SELECT convert(:param, 'AL16UTF16') FROM DUAL",
+                {"param": data},
+            ).scalar()
+        eq_(result, utf8_w_errors)
+
+        # this one not really working
+        with e.connect() as conn:
+            result = conn.execute(
+                "SELECT TO_CLOB(convert(:param, 'AL16UTF16')) FROM DUAL",
+                {"param": data},
+            ).scalar()
+        # eq_(result, data) #utf8_w_errors)
+
 
 class LOBFetchTest(fixtures.TablesTest):
     __only_on__ = "oracle"

```

zoltan-fedor:
Thanks.
I have incorporated this into my PR, but the test fails. Are you sure about those test corrupt characters?

```
NLS_LANG=American_America.UTF8 py.test --dburi=oracle+cx_oracle://SERVER:1521/?service_name=myservice test/dialect/oracle/test_types.py::TypesTest::test_broken_unicode_fetch -s
=============================================================================== test session starts ================================================================================
platform linux -- Python 3.7.4, pytest-5.0.1, py-1.8.0, pluggy-0.12.0 -- /home/user/.local/share/virtualenvs/sqlalchemy-TRLVSdTR/bin/python3.7m
cachedir: .pytest_cache
rootdir: /home/user/Documents/Development/Personal/sqlalchemy, inifile: setup.cfg
collected 1 item                                                                                                                                                                   

test/dialect/oracle/test_types.py::TypesTest_oracle+cx_oracle_11_2_0_4_0::test_broken_unicode_fetch FAILED

===================================================================================== FAILURES =====================================================================================
_________________________________________________________ TypesTest_oracle+cx_oracle_11_2_0_4_0.test_broken_unicode_fetch __________________________________________________________
  File "/home/user/Documents/Development/Personal/sqlalchemy/test/dialect/oracle/test_types.py", line 901, in test_broken_unicode_fetch
    eq_(result, utf8_w_errors)
  File "/home/user/Documents/Development/Personal/sqlalchemy/test/../lib/sqlalchemy/testing/assertions.py", line 237, in eq_
    assert a == b, msg or "%r != %r" % (a, b)
AssertionError: '�f' != ',f'
assert '�f' == ',f'
  - �f
  + ,f
============================================================================= short test summary info ==============================================================================
FAILED test/dialect/oracle/test_types.py::TypesTest_oracle+cx_oracle_11_2_0_4_0::test_broken_unicode_fetch - AssertionError: '�f' != ',f'
============================================================================= 1 failed in 0.89 seconds =============================================================================

```

zzzeek:
there's two tests in there.  the first one works for me, the second one does not (has a comment that says "this one not really working").      The second test for me returns that '�f' symbol but the first returns ",f".   I tried with cx_Oracle 6.4.1 and 7.2.1 and get the same result.

if you run the tests with --log-debug=sqlalchemy.engine -s , it will log SQL.  heres what I get, what do you get?

```
test/dialect/oracle/test_types.py::TypesTest_oracle+cx_oracle_11_2_0_2_0::test_broken_unicode_fetch INFO:sqlalchemy.engine.base.Engine:SELECT USER FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
DEBUG:sqlalchemy.engine.base.Engine:Col ('USER',)
DEBUG:sqlalchemy.engine.base.Engine:Row ('SCOTT',)
INFO:sqlalchemy.engine.base.Engine:SELECT CAST('test plain returns' AS VARCHAR(60 CHAR)) AS anon_1 FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
INFO:sqlalchemy.engine.base.Engine:SELECT CAST('test unicode returns' AS VARCHAR2(60 CHAR)) AS anon_1 FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
INFO:sqlalchemy.engine.base.Engine:SELECT CAST('test nvarchar2 returns' AS NVARCHAR2(60)) AS anon_1 FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
INFO:sqlalchemy.engine.base.Engine:select value from nls_session_parameters where parameter = 'NLS_NUMERIC_CHARACTERS'
INFO:sqlalchemy.engine.base.Engine:{}
DEBUG:sqlalchemy.engine.base.Engine:Col ('VALUE',)
DEBUG:sqlalchemy.engine.base.Engine:Row ('.,',)
INFO:sqlalchemy.engine.base.Engine:SELECT convert(:param, 'AL16UTF16') FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{'param': '\uee2c驦'}
DEBUG:sqlalchemy.engine.base.Engine:Col ("CONVERT(:PARAM,'AL16UTF16')",)
DEBUG:sqlalchemy.engine.base.Engine:Row (',f',)
INFO:sqlalchemy.engine.base.Engine:SELECT TO_CLOB(convert(:param, 'AL16UTF16')) FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{'param': '\uee2c驦'}
DEBUG:sqlalchemy.engine.base.Engine:Col ("TO_CLOB(CONVERT(:PARAM,'AL16UTF16'))",)
DEBUG:sqlalchemy.engine.base.Engine:Row ('�f',)
PASSED

```

zoltan-fedor:
I do get different results at
```
DEBUG:sqlalchemy.engine.base.Engine:Col ("CONVERT(:PARAM,'AL16UTF16')",)
DEBUG:sqlalchemy.engine.base.Engine:Row ('�f',)
```

See the whole log:
```
$ NLS_LANG=American_America.UTF8 py.test --dburi=oracle+cx_oracle://SERVER:1521/?service_name=myservice test/dialect/oracle/test_types.py::TypesTest::test_broken_unicode_fetch -s --log-debug=sqlalchemy.engine
INFO:sqlalchemy.engine.base.Engine:SELECT USER FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
DEBUG:sqlalchemy.engine.base.Engine:Col ('USER',)
DEBUG:sqlalchemy.engine.base.Engine:Row ('BIAPP',)
INFO:sqlalchemy.engine.base.Engine:SELECT CAST('test plain returns' AS VARCHAR(60 CHAR)) AS anon_1 FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
INFO:sqlalchemy.engine.base.Engine:SELECT CAST('test unicode returns' AS VARCHAR2(60 CHAR)) AS anon_1 FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
INFO:sqlalchemy.engine.base.Engine:SELECT CAST('test nvarchar2 returns' AS NVARCHAR2(60)) AS anon_1 FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
INFO:sqlalchemy.engine.base.Engine:select value from nls_session_parameters where parameter = 'NLS_NUMERIC_CHARACTERS'
INFO:sqlalchemy.engine.base.Engine:{}
DEBUG:sqlalchemy.engine.base.Engine:Col ('VALUE',)
DEBUG:sqlalchemy.engine.base.Engine:Row ('.,',)
=============================================================================== test session starts ================================================================================
platform linux -- Python 3.7.4, pytest-5.0.1, py-1.8.0, pluggy-0.12.0 -- /home/user/.local/share/virtualenvs/sqlalchemy-TRLVSdTR/bin/python3.7m
cachedir: .pytest_cache
rootdir: /home/user/Documents/Development/Personal/sqlalchemy, inifile: setup.cfg
collected 1 item                                                                                                                                                                   

test/dialect/oracle/test_types.py::TypesTest_oracle+cx_oracle_11_2_0_4_0::test_broken_unicode_fetch INFO:sqlalchemy.engine.base.Engine:SELECT USER FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
DEBUG:sqlalchemy.engine.base.Engine:Col ('USER',)
DEBUG:sqlalchemy.engine.base.Engine:Row ('BIAPP',)
INFO:sqlalchemy.engine.base.Engine:SELECT CAST('test plain returns' AS VARCHAR(60 CHAR)) AS anon_1 FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
INFO:sqlalchemy.engine.base.Engine:SELECT CAST('test unicode returns' AS VARCHAR2(60 CHAR)) AS anon_1 FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
INFO:sqlalchemy.engine.base.Engine:SELECT CAST('test nvarchar2 returns' AS NVARCHAR2(60)) AS anon_1 FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
INFO:sqlalchemy.engine.base.Engine:select value from nls_session_parameters where parameter = 'NLS_NUMERIC_CHARACTERS'
INFO:sqlalchemy.engine.base.Engine:{}
DEBUG:sqlalchemy.engine.base.Engine:Col ('VALUE',)
DEBUG:sqlalchemy.engine.base.Engine:Row ('.,',)
INFO:sqlalchemy.engine.base.Engine:SELECT convert(:param, 'AL16UTF16') FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{'param': '\uee2c驦'}
DEBUG:sqlalchemy.engine.base.Engine:Col ("CONVERT(:PARAM,'AL16UTF16')",)
DEBUG:sqlalchemy.engine.base.Engine:Row ('�f',)
FAILED

===================================================================================== FAILURES =====================================================================================
_________________________________________________________ TypesTest_oracle+cx_oracle_11_2_0_4_0.test_broken_unicode_fetch __________________________________________________________
  File "/home/user/Documents/Development/Personal/sqlalchemy/test/dialect/oracle/test_types.py", line 901, in test_broken_unicode_fetch
    eq_(result, utf8_w_errors)
  File "/home/user/Documents/Development/Personal/sqlalchemy/test/../lib/sqlalchemy/testing/assertions.py", line 237, in eq_
    assert a == b, msg or "%r != %r" % (a, b)
AssertionError: '�f' != ',f'
assert '�f' == ',f'
  - �f
  + ,f
============================================================================= short test summary info ==============================================================================
FAILED test/dialect/oracle/test_types.py::TypesTest_oracle+cx_oracle_11_2_0_4_0::test_broken_unicode_fetch - AssertionError: '�f' != ',f'
============================================================================= 1 failed in 0.94 seconds =============================================================================
```

zzzeek:
OK.  This is not actually that critical to the test - here's what is.  If you take out "encoding_errors": "ignore", you should get the UnicodeDecodeError for the first one.  do you get *that* ?    I have a feeling you won't, because it looks like some error thing is already happening there.

the challenge here is to find some characters that are valid in both utf-8 as well as <some other encoding cx_oracle supports>, but it fails if you try to decode the <other encoding> version in utf-8.

What does your Python intepreter say if you do this:

```
Python 3.7.3 (default, May 11 2019, 00:38:04) 
[GCC 9.1.1 20190503 (Red Hat 9.1.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> u"\uee2c\u9a66".encode('utf-16').decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
>>> 

```

zoltan-fedor:
Unfortunately no unicode error even after taking out the `encoding_errors`.

This is now the modified test without `encoding_errors`:
```
    @testing.provide_metadata
    def test_broken_unicode_fetch(self):
        # e = testing_engine(options={"encoding_errors": "ignore"})
        e = testing_engine()

        from sqlalchemy.util import ue

        data = ue("\uee2c\u9a66")  # this is u"\uee2c\u9a66"

        utf8_w_errors = data.encode('utf-16').decode('utf-8', 'ignore')

        with e.connect() as conn:
            result = conn.execute(
                "SELECT convert(:param, 'AL16UTF16') FROM DUAL",
                {"param": data},
            ).scalar()
        eq_(result, utf8_w_errors)

        # this one not really working
        with e.connect() as conn:
            result = conn.execute(
                "SELECT TO_CLOB(convert(:param, 'AL16UTF16')) FROM DUAL",
                {"param": data},
            ).scalar()
        # eq_(result, data) #utf8_w_errors)
```

The error is still the same:
```
$ NLS_LANG=American_America.UTF8 py.test --dburi=oracle+cx_oracle://SERVER:1521/?service_name=myservice test/dialect/oracle
INFO:sqlalchemy.engine.base.Engine:SELECT USER FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
DEBUG:sqlalchemy.engine.base.Engine:Col ('USER',)
DEBUG:sqlalchemy.engine.base.Engine:Row ('BIAPP',)
INFO:sqlalchemy.engine.base.Engine:SELECT CAST('test plain returns' AS VARCHAR(60 CHAR)) AS anon_1 FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
INFO:sqlalchemy.engine.base.Engine:SELECT CAST('test unicode returns' AS VARCHAR2(60 CHAR)) AS anon_1 FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
INFO:sqlalchemy.engine.base.Engine:SELECT CAST('test nvarchar2 returns' AS NVARCHAR2(60)) AS anon_1 FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
INFO:sqlalchemy.engine.base.Engine:select value from nls_session_parameters where parameter = 'NLS_NUMERIC_CHARACTERS'
INFO:sqlalchemy.engine.base.Engine:{}
DEBUG:sqlalchemy.engine.base.Engine:Col ('VALUE',)
DEBUG:sqlalchemy.engine.base.Engine:Row ('.,',)
=============================================================================== test session starts ================================================================================
platform linux -- Python 3.7.4, pytest-5.0.1, py-1.8.0, pluggy-0.12.0 -- /home/user/.local/share/virtualenvs/sqlalchemy-TRLVSdTR/bin/python3.7m
cachedir: .pytest_cache
rootdir: /home/user/Documents/Development/Personal/sqlalchemy, inifile: setup.cfg
collected 1 item                                                                                                                                                                   

test/dialect/oracle/test_types.py::TypesTest_oracle+cx_oracle_11_2_0_4_0::test_broken_unicode_fetch INFO:sqlalchemy.engine.base.Engine:SELECT USER FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
DEBUG:sqlalchemy.engine.base.Engine:Col ('USER',)
DEBUG:sqlalchemy.engine.base.Engine:Row ('BIAPP',)
INFO:sqlalchemy.engine.base.Engine:SELECT CAST('test plain returns' AS VARCHAR(60 CHAR)) AS anon_1 FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
INFO:sqlalchemy.engine.base.Engine:SELECT CAST('test unicode returns' AS VARCHAR2(60 CHAR)) AS anon_1 FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
INFO:sqlalchemy.engine.base.Engine:SELECT CAST('test nvarchar2 returns' AS NVARCHAR2(60)) AS anon_1 FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{}
INFO:sqlalchemy.engine.base.Engine:select value from nls_session_parameters where parameter = 'NLS_NUMERIC_CHARACTERS'
INFO:sqlalchemy.engine.base.Engine:{}
DEBUG:sqlalchemy.engine.base.Engine:Col ('VALUE',)
DEBUG:sqlalchemy.engine.base.Engine:Row ('.,',)
INFO:sqlalchemy.engine.base.Engine:SELECT convert(:param, 'AL16UTF16') FROM DUAL
INFO:sqlalchemy.engine.base.Engine:{'param': '\uee2c驦'}
DEBUG:sqlalchemy.engine.base.Engine:Col ("CONVERT(:PARAM,'AL16UTF16')",)
DEBUG:sqlalchemy.engine.base.Engine:Row ('�f',)
FAILED

===================================================================================== FAILURES =====================================================================================
_________________________________________________________ TypesTest_oracle+cx_oracle_11_2_0_4_0.test_broken_unicode_fetch __________________________________________________________
  File "/home/user/Documents/Development/Personal/sqlalchemy/test/dialect/oracle/test_types.py", line 901, in test_broken_unicode_fetch
    eq_(result, utf8_w_errors)
  File "/home/user/Documents/Development/Personal/sqlalchemy/test/../lib/sqlalchemy/testing/assertions.py", line 237, in eq_
    assert a == b, msg or "%r != %r" % (a, b)
AssertionError: '�f' != ',f'
assert '�f' == ',f'
  - �f
  + ,f
============================================================================= short test summary info ==============================================================================
FAILED test/dialect/oracle/test_types.py::TypesTest_oracle+cx_oracle_11_2_0_4_0::test_broken_unicode_fetch - AssertionError: '�f' != ',f'
============================================================================= 1 failed in 0.90 seconds =============================================================================
```

zzzeek:
can you confirm your python does raise an error?

    u"\uee2c\u9a66".encode('utf-16').decode('utf-8')

if so, isn't that strange?   oracle or cx_oracle are not being very consistent.   codecs are not random.

zoltan-fedor:
Yes, my python (3.7.4) does raise an error as it should.

```
(sqlalchemy) user@userPC:~/Documents/Development/Personal/sqlalchemy$ python
Python 3.7.4 (default, Jul  9 2019, 15:11:16) 
[GCC 7.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> u"\uee2c\u9a66".encode('utf-16').decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```

Well, I am not surprised about Oracle, it wouldn't be the first time when it delivers something "odd"... :-)

zzzeek:
ding, OK, my NLS_LANG is

AMERICAN_AMERICA.AL32UTF8

if I use American_America.UTF8, then it acts like you see it.

now how to control this progammatically...

zoltan-fedor:
Settings the NLS_LANG is tricky, because it must be done before the cx_Oracle library is loaded.
But otherwise it can be done programatically (as long as you do it before the library would be imported) with just the usual 
`os.environ["NLS_LANG"] =`

zzzeek:
yeah I'm not that thrilled about that.    the problem is if I commit a fix that  can silently change without breaking tests, it will silently change someday.  guaranteed.   we can use mocks here but they are open towards new codepaths for this not being tested.

zzzeek:
but yes I can set NLS_LANG in tox.ini and just make that a conditon of this test

zzzeek:
I wonder if, on your "corrupt" database, what happens if you SELECT those bad columns using to_nclob() ?  I can't get any encoding issue to occur when this is used as it seems like Oracle might already be "ignoring" errors, not sure.

sqla-tester:
**Mike Bayer** has proposed a fix for this issue in the **master** branch:

**Implment encoding_errors for cx_oracle** https://gerrit.sqlalchemy.org/1411

zoltan-fedor:
@zzzeek 
"I wonder if, on your "corrupt" database, what happens if you SELECT those bad columns using to_nclob() ? I can't get any encoding issue to occur when this is used as it seems like Oracle might already be "ignoring" errors, not sure."
>> Unfortunately I can't, as I don't even know which are the bad columns. I read the whole table row by row when the error is thrown, so don't know which column.
But yes, clobs are implemented very differently in Oracle, wouldn't be surprised if corrupted texts would not show up the same way.

zzzeek:
I think I want to write a mock-style test for this, to see if the right flags get passed through to cx_oracle, and that will be it. it doesnt seem possible to really live-test this right now as we cannot reproduce a broken database situation.

PR #4816, opened 2019-08-22
Implement primary key custom sorting.

cans:
Implements the ability to use unsortable types as primary key, by injection of a custom order relation in the key column definition.

The goal being to retain unsortability of the python type while still being able to update rows in deterministic and predictable order across transactions. An example of such types are Enum.

### Description

This adds a sort key extraction attributes to SQL types, which is by default `None` (meaning a "natural" order exists for the type). If set, it must be a function that receives the primary key value and extracts a sortable value from it.

Alternatively the order can be provided with the type definition (e.g. for composite types, or anything inheriting from `TypeDecorator`). This is done by overidding the `sort_key_function()` method that should return a function with the same contract as above.

This PR implements the feature and provides tests
- On the new [`TypeDecorator.sort_key_function()`](https://github.com/sqlalchemy/sqlalchemy/pull/4816/files#diff-c6663807cd4a586af0b59ad165fcfb62R1771);
- On the new [`Column.sort_key_function`](https://github.com/sqlalchemy/sqlalchemy/pull/4816/files#diff-96d0a8b3b5254a6a2ddae85d53b75884R3428) argument;

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

### DIFF
--- a/test/orm/test_unitofwork.py
+++ b/test/orm/test_unitofwork.py
@@ -15,12 +15,14 @@ from sqlalchemy import literal_column
 from sqlalchemy import select
 from sqlalchemy import String
 from sqlalchemy import testing
+from sqlalchemy.inspection import inspect
 from sqlalchemy.orm import column_property
 from sqlalchemy.orm import create_session
 from sqlalchemy.orm import exc as orm_exc
 from sqlalchemy.orm import mapper
 from sqlalchemy.orm import relationship
 from sqlalchemy.orm import Session
+from sqlalchemy.orm.persistence import _sort_states
 from sqlalchemy.testing import assert_raises
 from sqlalchemy.testing import assert_raises_message
 from sqlalchemy.testing import eq_
@@ -3398,6 +3400,7 @@ class EnsurePKSortableTest(fixtures.MappedTest):
     two = MySortableEnum("two", 2)
     three = MyNotSortableEnum("three", 3)
     four = MyNotSortableEnum("four", 4)
+    five = MyNotSortableEnum("five", 5)
 
     @classmethod
     def define_tables(cls, metadata):
@@ -3415,6 +3418,24 @@ class EnsurePKSortableTest(fixtures.MappedTest):
             Column("data", String(10)),
         )
 
+        Table(
+            "t3",
+            metadata,
+            Column(
+                "id",
+                Enum(
+                    cls.MyNotSortableEnum,
+                    sort_key_function=cls.sort_enum_key_value,
+                ),
+                primary_key=True,
+            ),
+            Column("value", Integer),
+        )
+
+    @staticmethod
+    def sort_enum_key_value(value):
+        return value.value
+
     @classmethod
     def setup_classes(cls):
         class T1(cls.Basic):
@@ -3423,10 +3444,15 @@ class EnsurePKSortableTest(fixtures.MappedTest):
         class T2(cls.Basic):
             pass
 
+        class T3(cls.Basic):
+            def __str__(self):
+                return 'T3(id={})'.format(self.id)
+
     @classmethod
     def setup_mappers(cls):
         mapper(cls.classes.T1, cls.tables.t1)
         mapper(cls.classes.T2, cls.tables.t2)
+        mapper(cls.classes.T3, cls.tables.t3)
 
     def test_exception_persistent_flush_py3k(self):
         s = Session()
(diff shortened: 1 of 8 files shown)


test/orm/test_unitofwork.py
```diff
@@ -3459,3 +3485,21 @@ def test_persistent_flush_sortable(self):
         a.data = "bar"
         b.data = "foo"
         s.commit()
+
+    def test_pep435_custom_sort_key(self):
```
cans: I did not implement this test where [you suggested](https://github.com/cans/sqlalchemy/blob/implement-type-engine-value-sortkey/test/sql/test_types.py#L1660) I do. The, maybe wrong, reasons are that:

- the name of the test suite class seemed relevant;
- it simplified the test implementation;
- my understanding is that this happens at the ORM level, I am not sure how to implement this at the core level (though, now that I dug a bit deeper I might have an idea);

zzzeek: the test/sql/test_types.py test would be a simple test that you can create an Enum with a "sort_key_function" argument and that you can get it back from the sort_key_function method.   I wonder why I have that as a method.   it should likely be an attribute.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 6b77ba85a0acf25b0d2267e67f67d114d50c8d3b of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 6b77ba85a0acf25b0d2267e67f67d114d50c8d3b: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1428

test/orm/test_unitofwork.py
cans: Is [this](https://github.com/sqlalchemy/sqlalchemy/pull/4816/files#diff-f02c6a58d783b77b65f72fd0adc49333R1675) what you meant ?

I added a second commit to make an "attribute" (or property) of `sort_key_function`. It's more of a method which default implementation can be `None` (! works because it is used as a callback), than really an attribute, I think. But it does remove a level of indirection and avoids an [unnecessary nested function](https://github.com/sqlalchemy/sqlalchemy/pull/4816/files#diff-c6663807cd4a586af0b59ad165fcfb62R1781). Since the default implementation for the TypeDecorator is to proxy I don't think you can get rid of the `@property` decorator there. We could remove the one in the [TypeEngine](https://github.com/sqlalchemy/sqlalchemy/pull/4816/files#diff-f21ffccc93edfe1d096c18ff1546cedbR374) class though.

Happy to either squash or drop that commit.

zzzeek: yeah i guess just a one liner test is what we can do there.

the commits here all get squashed when i pull them into gerrit so you can either squash here or not, I only look at the full product as one patch.

cans: Ok. Since I had to make a change to not unnecessarily hit the DB, I squashed the commits.

While running the tests I also hit a tiny issue with `mock`. I managed to fix it with this patch, though I did not include it in this PR as it is off-topic. 
```patch
diff --git a/lib/sqlalchemy/testing/suite/test_types.py b/lib/sqlalchemy/testing/suite/test_types.py
index 82a842102..b8ebcb7ae 100644
--- a/lib/sqlalchemy/testing/suite/test_types.py
+++ b/lib/sqlalchemy/testing/suite/test_types.py
@@ -4,7 +4,7 @@ import datetime
 import decimal
 import json
 
-import mock
+from .. import mock
 
 from .. import config
 from .. import engines
diff --git a/tox.ini b/tox.ini
index ea520f9fe..7eb809bbe 100644
--- a/tox.ini
+++ b/tox.ini
@@ -17,7 +17,7 @@ usedevelop=
 
 deps=pytest!=3.9.1,!=3.9.2
      pytest-xdist
-     mock
+     mock; python_version < '3.3'
      # needed only for correct profiling results
      # due to speed improvements in psycopg2 as of 2.7
      postgresql: psycopg2>=2.7
```
I guess it didn't show before because tox installs mock whatever the python version. But I wanted to be able to just run pytest to get me started so...

zzzeek: oh this should be a separate commit I can make and also 1.3 / master is only supporting 3.4 and above, so ill just do that, thanks

zzzeek: I guess we need it for 2.7.  sigh

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 42266b766c1e462d5b8a409cda05d33dea13bd34 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 42266b766c1e462d5b8a409cda05d33dea13bd34 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1428

PR #4859, opened 2019-09-12
For oracle dialect initializing max_identifier_length based on server version

amandeep-sharma:
Instead for setting max_identifier_length to 30, it is initialized based on server version


zzzeek (changes requested):
hi there -

can this have dialect-level tests added please?   see https://github.com/sqlalchemy/sqlalchemy/blob/master/test/dialect/oracle/test_dialect.py for a likely place.   additionally this value interacts with the label_length engine parameter so need to make sure that continues to work OK if it was set.

zzzeek:
thanks for this but I have a more elaborate set of changes moving through so that this is adjusted incrementally.

PR #4896, opened 2019-10-07
Support for generated columns

### LINKED ISSUE #4894: Support for generated columns in postgres and mysql
Add support for specifying a column as generated by the database with `GENERATED ALWAYS`

Currently postgres and mysql support it (the list may not be complete)
https://www.postgresql.org/docs/12/ddl-generated-columns.html
https://dev.mysql.com/doc/refman/8.0/en/create-table-generated-columns.html

I've tried to search in the documentation and it does not seems to be a way to specify a generated column in sqlalchemy. 
I've found a stack overflow answer https://stackoverflow.com/questions/57596603/sqlalchemy-mysql-column-ddl-with-generated-always-as but it does not seem to be the cleanest way to deal with it.

maybe a Generated construct used in the column with the code would be appropriate?
Something like
```py
sa.Table('foo', meta,
  sa.Column('quantity', sa.Integer, nullable=False),
  sa.Column('price', sa.Float, nullable=False),
  sa.Column('revenue', sa.Float, Computed('quantity * price', always=True))
)
```

edit: updated example following https://github.com/sqlalchemy/sqlalchemy/issues/4894#issuecomment-539050154

Current support and syntax seems to be:

- Postgress: supports only stored 
`GENERATED ALWAYS AS ( generation_expr ) STORED`
  https://www.postgresql.org/docs/12/ddl-generated-columns.html
  https://www.postgresql.org/docs/12/sql-createtable.html

- mysql | mariad db:  supports both stored and virtual
  `col_name data_type [GENERATED ALWAYS] AS (expr) [VIRTUAL | STORED]`
  (maridadb supports also `PERSISTENT`, but is an alias to `STORED` 
https://dev.mysql.com/doc/refman/5.7/en/create-table-generated-columns.html
https://mariadb.com/kb/en/library/generated-columns/
- mssql: support both stored and virtual
`<computed_column_definition> ::=
column_name AS computed_column_expression
[ PERSISTED [ NOT NULL ] ]`
https://docs.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql?view=sql-server-2017
- oracle: support only virtual 
`column_name [datatype] [GENERATED ALWAYS] AS (expression) [VIRTUAL]`
https://oracle-base.com/articles/11g/virtual-columns-11gr1

CaselIT:
Support for generated column

Closes https://github.com/sqlalchemy/sqlalchemy/issues/4894

This is still wip. I've opened this now to get feedback on the initial implementation

Still TODO:
- [x] dialect support
- [ ] inspect support (is this required?)
- [ ] tests
  - [x] compile tests
  - [x] update/default orm tests
  - [ ] insert-update row in core with and without computed column
  - [ ] insert-update row in orm with and without computed column

### DIFF
--- a/lib/sqlalchemy/__init__.py
+++ b/lib/sqlalchemy/__init__.py
@@ -17,7 +17,7 @@ from .schema import DefaultClause  # noqa
 from .schema import FetchedValue  # noqa
 from .schema import ForeignKey  # noqa
 from .schema import ForeignKeyConstraint  # noqa
-from .schema import Generated  # noqa
+from .schema import Computed  # noqa
 from .schema import Index  # noqa
 from .schema import MetaData  # noqa
 from .schema import PrimaryKeyConstraint  # noqa
--- a/lib/sqlalchemy/__init__.py
+++ b/lib/sqlalchemy/__init__.py
@@ -17,6 +17,7 @@ from .schema import DefaultClause  # noqa
 from .schema import FetchedValue  # noqa
 from .schema import ForeignKey  # noqa
 from .schema import ForeignKeyConstraint  # noqa
+from .schema import Generated  # noqa
 from .schema import Index  # noqa
 from .schema import MetaData  # noqa
 from .schema import PrimaryKeyConstraint  # noqa
--- a/lib/sqlalchemy/dialects/firebird/base.py
+++ b/lib/sqlalchemy/dialects/firebird/base.py
@@ -582,6 +582,17 @@ class FBDDLCompiler(sql.compiler.DDLCompiler):
                 drop.element
             )
 
+    def visit_computed_column(self, generated):
+        if generated.persisted is not None:
+            raise exc.CompileError(
+                "firebird does not support specifying a persistence method. "
+                "Remove 'persisted' or set it to None"
+            )
+        text = " GENERATED ALWAYS AS (%s)" % self.sql_compiler.process(
+            generated.sqltext, include_table=False, literal_binds=True
+        )
+        return text
+
 
 class FBIdentifierPreparer(sql.compiler.IdentifierPreparer):
     """Install Firebird specific reserved words."""
--- a/lib/sqlalchemy/dialects/mssql/base.py
+++ b/lib/sqlalchemy/dialects/mssql/base.py
@@ -1942,13 +1942,15 @@ class MSSQLStrictCompiler(MSSQLCompiler):
 
 class MSDDLCompiler(compiler.DDLCompiler):
     def get_column_specification(self, column, **kwargs):
-        colspec = (
-            self.preparer.format_column(column)
-            + " "
-            + self.dialect.type_compiler.process(
+        colspec = self.preparer.format_column(column)
+
+        # type is not accepted in a computed column
+        if column.computed is not None:
+            colspec += self.process(column.computed)
+        else:
+            colspec += " " + self.dialect.type_compiler.process(
                 column.type, type_expression=column
             )
-        )
 
         if column.nullable is not None:
             if (
@@ -2108,6 +2110,15 @@ class MSDDLCompiler(compiler.DDLCompiler):
             self.preparer.quote(c.name) for c in constraint
         )
         text += self.define_constraint_deferrability(constraint)
+        return text
+
+    def visit_computed_column(self, generated):
+        text = " AS (%s)" % self.sql_compiler.process(
+            generated.sqltext, include_table=False, literal_binds=True
+        )
+        # explicitly check for True|False since None means server default
+        if generated.persisted is True:
+            text += " PERSISTED"
         return text
 
 
--- a/lib/sqlalchemy/dialects/mssql/base.py
+++ b/lib/sqlalchemy/dialects/mssql/base.py
@@ -1960,7 +1960,9 @@ class MSDDLCompiler(compiler.DDLCompiler):
                 or column.autoincrement is True
             ):
                 colspec += " NOT NULL"
-            else:
+            elif column.computed is None:
+                # nullable is True by default and computed column do not
+                # support NULL
                 colspec += " NULL"
 
         if column.table is None:
--- a/lib/sqlalchemy/dialects/mysql/base.py
+++ b/lib/sqlalchemy/dialects/mysql/base.py
@@ -1505,7 +1505,7 @@ class MySQLDDLCompiler(compiler.DDLCompiler):
 
         if column.computed is not None:
             # the default visit_computed_column is compatible
-            colspec.append(self.process(column.computed))
+            colspec.append(self.process(column.computed).strip())
 
         is_timestamp = isinstance(
             column.type._unwrapped_dialect_impl(self.dialect),
--- a/lib/sqlalchemy/dialects/mysql/base.py
+++ b/lib/sqlalchemy/dialects/mysql/base.py
@@ -1503,6 +1503,10 @@ class MySQLDDLCompiler(compiler.DDLCompiler):
             ),
         ]
 
+        if column.computed is not None:
+            # the default visit_computed_column is compatible
+            colspec.append(self.process(column.computed))
+
         is_timestamp = isinstance(
             column.type._unwrapped_dialect_impl(self.dialect),
             sqltypes.TIMESTAMP,
--- a/lib/sqlalchemy/dialects/oracle/base.py
+++ b/lib/sqlalchemy/dialects/oracle/base.py
@@ -1187,6 +1187,20 @@ class OracleDDLCompiler(compiler.DDLCompiler):
 
         return "".join(table_opts)
 
+    def visit_computed_column(self, generated):
+        text = " GENERATED ALWAYS AS (%s)" % self.sql_compiler.process(
+            generated.sqltext, include_table=False, literal_binds=True
+        )
+        # explicitly check for True|False since None means server default
+        if generated.persisted is True:
+            raise exc.CompileError(
+                "oracle does not support persited (stored) computed "
+                "columns. Remove 'persisted' or set it to False"
+            )
+        elif generated.persisted is False:
+            text += " VIRTUAL"
+        return text
+
 
 class OracleIdentifierPreparer(compiler.IdentifierPreparer):
 
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -1876,6 +1876,10 @@ class PGDDLCompiler(compiler.DDLCompiler):
             if default is not None:
                 colspec += " DEFAULT " + default
 
+        # computed can be before or after null / not null
+        if column.computed is not None:
+            colspec += self.process(column.computed)
+
         if not column.nullable:
             colspec += " NOT NULL"
         return colspec
@@ -2045,6 +2049,19 @@ class PGDDLCompiler(compiler.DDLCompiler):
             )
 
         return "".join(table_opts)
+
+    def visit_computed_column(self, generated):
+        # explicitly check for True|False since None means server default
+        if generated.persisted is False:
+            raise exc.CompileError(
+                "postgresql does not support virtual computed "
+                "columns. Remove 'persisted' or set it to True"
+            )
+
+        text = " GENERATED ALWAYS AS (%s) STORED" % self.sql_compiler.process(
+            generated.sqltext, include_table=False, literal_binds=True
+        )
+        return text
 
 
 class PGTypeCompiler(compiler.GenericTypeCompiler):
--- a/lib/sqlalchemy/dialects/sqlite/base.py
+++ b/lib/sqlalchemy/dialects/sqlite/base.py
@@ -1032,6 +1032,9 @@ class SQLiteCompiler(compiler.SQLCompiler):
 
 class SQLiteDDLCompiler(compiler.DDLCompiler):
     def get_column_specification(self, column, **kwargs):
+        if column.computed is not None:
+            raise exc.CompileError("sqlite does not support computed columns")
+
         coltype = self.dialect.type_compiler.process(
             column.type, type_expression=column
         )
--- a/lib/sqlalchemy/orm/__init__.py
+++ b/lib/sqlalchemy/orm/__init__.py
@@ -30,6 +30,7 @@ from .mapper import Mapper  # noqa
 from .mapper import reconstructor  # noqa
 from .mapper import validates  # noqa
 from .properties import ColumnProperty  # noqa
+from .properties import _ColumnPropertyProxy  # noqa
 from .query import AliasOption  # noqa
 from .query import Bundle  # noqa
 from .query import Query  # noqa
@@ -123,7 +124,7 @@ def dynamic_loader(argument, **kw):
     return relationship(argument, **kw)
 
 
-column_property = public_factory(ColumnProperty, ".orm.column_property")
+column_property = public_factory(_ColumnPropertyProxy, ".orm.column_property")
 composite = public_factory(CompositeProperty, ".orm.composite")
 
 
--- a/lib/sqlalchemy/orm/attributes.py
+++ b/lib/sqlalchemy/orm/attributes.py
@@ -260,6 +260,8 @@ class InstrumentedAttribute(QueryableAttribute):
     """
 
     def __set__(self, instance, value):
+        if self.prop.raise_on_set is True:
+            raise AttributeError("can't set attribute %s" % self.key)
         self.impl.set(
             instance_state(instance), instance_dict(instance), value, None
         )
--- a/lib/sqlalchemy/orm/mapper.py
+++ b/lib/sqlalchemy/orm/mapper.py
@@ -1840,6 +1840,7 @@ class Mapper(InspectionAttr):
             prop, properties.ConcreteInheritedProperty
         ):
             mapped_column = []
+            kwargs = {}
             for c in columns:
                 mc = self.persist_selectable.corresponding_column(c)
                 if mc is None:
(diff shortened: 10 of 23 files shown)


CaselIT:
I've currently named the class `Generated`, but `Computed` is what we settled in the issue, so I'll rename it

CaselIT:
@zzzeek I've done a first implementation. I'll start adding tests. Can you check if it seems ok?

There are some question regarding the implementation
- Some dialects do not support null or not null with the computed columns.
  Currently I have not added checks regarding this, since the combination can be quite a lot. (eg mssql     accepts NOT NULL only on stored generated columns. And never NULL. https://dbfiddle.uk/?rdbms=sqlserver_2017l&fiddle=cabf00c157d0c95a351061ae2f57b56c)

- some db support computed columns that depend on computed columns. Some don't. I have not checked this. I mean this:
  ```py
    t = sa.Table('foo', meta,
            sa.Column('side', sa.Float),
            sa.Column('area', sa.Float, sa.Computed('side * side')),
            sa.Column('volume', sa.Float, sa.Computed('area * side'(),
        )
  ```

CaselIT:
I've added the tests.

The fixtures or `pytest.mark.parametrize` seem not to work in pytest, so I had to use a for in some cases, instead of the pytest way for parametrizing fixtures.

I've never worked with the inspect package, so I'm not sure how to update it.

zzzeek:
we don't use any pytest fixtures in the tests themselves.   we have our own assert_raises and assert_raises_message functions.   they aren't context managers right now but we've gotten by with the callable style, e.g.
```

assert_raises_message(
   exc.CompileError,
   "couldn't frob the nizzle",
   stmt.compile, dialect=some_dialect, foo=bar...
)
```

zzzeek:
for the for loop thing, the pattern we use now is this (you'll also find older tests that are a total mess but this is the way the newer tests should do it):

```
def _test_column_computed(self, persisted, text):
    # ...


def test_column_computed_obj(self):
   self._test_column_computed(flag, "")


def test_column_computed(self):
   self._test_column_computed(None, "")


def test_column_computed_persisted(self):
   self._test_column_computed(True, " STORED")


def test_column_computed_virtual(self):
   self._test_column_computed(False, " VIRTUAL")

```

CaselIT:
I'll update the test 

Any particular reason not to use fixtures (or at least enable parametrize?). Is it to keep the test compatible with other test frameworks?

CaselIT:
I just thought that we should test that the orm updates the value of a computed column after an update / insert.
Is there a similar test I can look at?

zzzeek:
for the ORM thing we would be looking at how it handles FetchedValue defaults and there is a lot of work there already.     there's a series of tests for the "eager" version of it in test/orm/test_unitofworkv2.py -> EagerDefaultsTest, and more general tests in test/orm/test_defaults.py.   The "computed" value here behaves like an "on insert / on update default" .   you can see in test_defaults we wrote triggers to test it.   That's a really old test suite and it seems a little minimal, it would be better to have more, smaller tests that test just one thing each.    The EagerDefaultsTest should probably be in there as well but this is beyond the scope of what you are working on.

CaselIT:
I'll have a look at the test and try to do something similar for the computed.

> I'm certainly all for fixtures, it's just historically we don't want "pytest" hardcoded throughout the test suite. Because a long time ago, it was all "nose" harcoded throughout the test suite. and then suddenly nose stopped being developed and everyone started using something else. So I can't really approach that kind of nightmare again. py.test might not be the big testing framework in another 10 years but SQLAlchemy is hopefully still around.

Ok, makes sense. 
Maybe a wrapper for pytest could be useful. Something like `from sqlalchemy.test import parametrize`, so at worst just the wrapper needs to be updated

zzzeek:
pytest is wrapped as part of what https://github.com/sqlalchemy/sqlalchemy/tree/master/lib/sqlalchemy/testing accomplishes.    nose support was removed only in the past year, which means there is more leverage to do things.   however for simple things like fixtures it's likely better to just add new features in this library directly.     but also if we were to create a fixture pattern there should be some strategy to rework the hundreds or maybe even thousands of tests that use other patterns right now to standardize on one pattern.

that is, there's already many different patterns that are here over the years and i wouldn't want to just keep adding new patterns without removing old ones.

zzzeek:
18 files...ok!    do you want me to run this through the full CI to see if things are passing?   note that I haven't reviewed this code yet, it will be a big one.

CaselIT:
I've added tests to the orm for default and updated test.
I've not tested, but we need to manage the case where the user sets a computed column in the class.

I think that depending on the db, some can fail the insertion/update of the row if they obtain a value for a computed column (postgres is among them, I think).

CaselIT:
I possible (easy) solution is to raise attribute error if the user tries to set the column, and ignore it when generating the insert/update sql.

Otherwise there is the case that the user has set a value, but it is different from the one set by the db

zzzeek:
oh that is interesting, I think maybe Core should at the very least reject this.  what do you think?    Also, if the DB rejects it anyway, then, that is already what we want.    I usually let the DB complain about things that are illegal if they require the user really does something to make them happen in the first place (that is, they wont normally happen).

CaselIT:
Regarding core, I guess it should reject an update/insert with a computed column set.

I hope that all the db accept this insert syntax. But I guess some will complain, wanting `DEFAULT` or similar, in paticular in combination with Not null.

CaselIT:
> if the DB rejects it anyway, then, that is already what we want.

 I think that it will not work very well for the orm. A rest this example:
```py
class Thing:
  id= Column(Integer, primary_key=True)
  foo= Column(Integer),
  bar = Column(Integer, Computed("foo + 42")

def makeThing():
  t = Thing(id=1, foo=1)
  s.add(t)
  # send to the network
  send(t.toJson())

... # later

def updateThing():
  # update from network
  updated = recieve()
  t = Thing(**updated)
  # here t has bar. what should we do?
  s.merge(t)  
```

regarding the CI, at the moment the only tests that use a backend are the last two I added to check the orm auto update. The other should pass

zzzeek:
A read-only attribute is a read-only attribute and silently ignoring things should not be the default.    I don't think Thing(**json_data) without explicit converters to receive the JSON form is a real-world pattern, because it won't work for all kinds of things like related objects, Decimal, dates, booleans, etc.

CaselIT:
I agree regarding the read-only attribute. Otherwise there is no good solution on how to manage a value set by the user. Accepting it if correct and raising an error if not may be complex for non trivial types.

I guess a computed column would behave similarly to a `column_property`. I'm not sure if an user can set a column_property or if it raises an error.

Thinking about the example, I guess that adding a in the documentation that a computed column cannot be set is enough. The example would then raise an error when executing  `Thing(**updated)`

zzzeek:
right so for column_property() it's currently just silently ignored :)     So, more specifically, at the moment my take on it is, "do nothing, assume nobody is setting it".   because, ideally, we should have a more consistent story for both kinds of attributes, where ignoring it could be an option, or raising could be an option.    so, ideally all column_property() that isn't persisted would raise on set by default, but im not sure I can just roll that out without giving folks some way to make it behave the other way, for architectures that are relying heavily on writing these and having them be ignored.

zzzeek:
put more simply, all column_property() against a SQL expression or against Generated should raise on set by default, and then some kind of option on column_property() called "ignore" would have the value be ignored on set, *OR* ignored on persist.     right now, it isn't ignored on set; it's only ignored on persist.  meaning you see it in `__dict__` but it doesn't wind up in the INSERT.

basically this might be a bigger issue than the DDL thing you're looking to do at the moment.

CaselIT:
Ok, so we don't want computed_column`

zzzeek:
well all the columns are mapped by column_property() ultimately, the writeable ones and the read-only expressions (well there is also query_expression() but that is very different at the usage level).   so however we make this work it should still go through column_property() unless it's really different in some way.

we might want to look into making all the explicit ORM support be a separate PR, actually, it's best if this can be broken up into merges that are as small as possible while still being a complete feature.      E.g. first support the core construct and the DDL, then support reflection, then the ORM stuff.

CaselIT:
I'll have a look on `column_property`.

>put more simply, all column_property() against a SQL expression or against Generated should raise on set by default, and then some kind of option on column_property() called "ignore" would have the value be ignored on set, OR ignored on persist. right now, it isn't ignored on set; it's only ignored on persist. meaning you see it in __dict__ but it doesn't wind up in the INSERT.

Regarding this I think we could have new flag `ignore_set_value` (or `raise_on_set` inverting the values) in `column_property` that will be set to `False` raising an exception when the user set a value in a future release. For now it will be set to 'warn', raising a warning to the user indicating this changed and to explicitly set the flag to `True` if they want to keep the current behavior (ignore the setted value)

For a computed column the flag can be set always set to `False` since it is a new functionality.

>Regarding core, I guess it should reject an update/insert with a computed column set.

I though a bit about this, and I'm not sure if it's required. What I mean is that when using core an user is specifying effectively an sql query, so it should be the db to tell him if it is doing something that is wrong, like setting a value to a computed column.

> we might want to look into making all the explicit ORM support be a separate PR, actually, it's best if this can be broken up into merges that are as small as possible while still being a complete feature. E.g. first support the core construct and the DDL, then support reflection, then the ORM stuff.

As you prefer. We can keep using this branch since it is all connected, and split it before merging, or we can decide to restrict the scope now at only the core, then improving on it on following issues.

In the meantime I'll have a look at column_property and at how it works to check if I can use it to manage the computed column in the orm

zzzeek:
Im looking at the pytest.fixtures stuff right now.   so far it's not feeling like a good fit for how our tests are written, it seems oriented around module-level functions and self-contained setup/teardown, which is all nice but that's not how our test suite is constructed, we are based on a class-inheritance approach.     For one thing it is mostly compatible with the unittest library that's included in Python.    I'm also not too thrilled with all the flags and switches pytest.fixture() accepts in order to solve all the problems that regular class-based construction solves through structure.       This doesn't apply as much to the pattern in this specific PR, it's more of a wider pattern where we have a whole bunch of mappings to be built up in some way, and then a series of tests to run on those mappings.  The setup of the mappings is the part that needs to vary.  We have a whole system that does this based on classes and once you know it, it works very well.

Doing things with class-based structure is probably not as easy for many programmers that come from JS these days and it often requires some creativity.   I will keep looking at py.test's approach but mostly I really would like there to be overall fewer ad-hoc approaches throughout the test suite first, because it's mostly the lack of consistency that is hurting rather than we have some patterns that are a little old-fashioned.   At the moment, I want to refactor a bunch of very old crufty tests that have  a really crappy parameterized fixture structure and see if I can't start visualizing some consistent answer for how we do parameterized fixtures, whether that be using pytest's approach, or something similar to it, or what.

zzzeek:
OK I see pytest.mark.parameterize is a little more easy to work with than the generalized pytest.fixture notion.   I'm going to experiment with using that but via a testing. function that will hide its name from the tests.   because i do use this particular pattern all over the place and it woul be nice for it to be automed.

sqla-tester:
**Mike Bayer** referenced this issue:

**experiment with pytest.mark.parametrize** https://gerrit.sqlalchemy.org/1523

zzzeek:
ive gotten pytest.mark.paramtrize to work (unfortunately with private pytest methods, not sure if they have a public method) and converted a test suite over, one that has lots of parameterizations.   it's not clear if this looks better to me yet, im going to ask around.  check it out in https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1523

CaselIT:
I have not yet seen the pr on gerrit. I'll take a look and let you know.


So I'll start by saing that I'm not an expert on pytest since it've used it
only on smallis projects (~1000) tests, and this is my first time
contributing anything meaningful to sqlalchemy.


From my experience fixtures are really nice to use if you prefer to
organise the test as the source code, so that each source file has its own
test file. In this case it is nice to have the ability to specify which
fixture each function should use, so that you can use a single test class
for each object in the source file and they can contain different type of
tests.


The test of sqlalchemy seem to be organised by functionality, so, from what
I have seen, the class organizations work well for it. There are a lot of
tests and fixture classes, so it may be difficult to know which one to use.
But I think the problem would be the same when using fixtures.


Regarding parametrize I think it can work in both type of tests, since I
find it quite useful to reduce the verbosity in some cases. Regarding the
output of the parametrize tests, it can be quite hard to understand if a
parameter is an object without the repr, but otherwise it is quite readable.


So my two cents is that probably its not worth it changing the style of the
tests even if there were time to spare, but it may be useful to use some of
the features of pytest, like parametrize. I asked about them mainly because
I did not know they could be disabled.

CaselIT:
It's also very nice the monkeypatch fixture of pytest to change/mock objects

sqla-tester:
**Mike Bayer** referenced this issue:

**Implement facade for pytest parametrize, fixtures** https://gerrit.sqlalchemy.org/1527

CaselIT:
I've left a could of comment on the gerrit regarding pytest. 
I'm not familiar with it, so I'm not sure if you can see them.

Maybe we could open an issue in github regarding the changes on pytest, so we avoid using this PR to discurr them?

zzzeek:
I'm going pretty nuts with parametrize in the test suite and I'm going to merge it after the next build finishes, because I want to apply it to a whole bunch of really old crufty tests that need to be reworked.  I also put a little bit of pytest.fixture in a few tests and I'll likely be using that in some cases also.        both are via a facade and in particular the parametrize facade highly modifies how it is used and im getting good results now.
  
pytest has a mock library built in ?  what's wrong with "mock" builtin?

zzzeek:
also thanks very much for pointing me to this.   once I merge, you can rebase here and make use of @testing.combinations which is the facade over parametrize.

CaselIT:
> pytest has a mock library built in

it's not really a mock libray, it just lets you setup the object with the the builtin mock library or other mock you created. From my experience it works better than the builtin path library https://docs.pytest.org/en/latest/monkeypatch.html

> I'm going pretty nuts with parametrize in the test suite and I'm going to merge it after the next build finishes

I've seen the changes, and I think it works nicely.

As I commented in gerrit, using https://docs.pytest.org/en/latest/parametrize.html#basic-pytest-generate-tests-example may avoid the need of the internal pytest call.

I've tried making an example, and it works in a normal project. In sqlalchemy it does not `runpytest_generate_tests`, I'm not sure why

```py
def pytest_generate_tests(metafunc):
    if metafunc.function is not None and hasattr(metafunc.function,
                                                 '__arg_sets'):
        argnames = ','.join(metafunc.fixturenames)
        metafunc.parametrize(argnames, metafunc.function.__arg_sets, ids=['generated ids here'])


def combinations(*arg_sets, **kwargs):
    def decorate(fn):
        fn.__arg_sets = arg_sets
        fn.__kwargs = kwargs
        return fn

    return decorate

@combinations((1, 2), (2, 3))
def test_somethin(x, y):
    print(x, y)
    assert 1
```

zzzeek:
sqlalchemy is using lots of pytest hooks already so those are interuppting pytest's normal automations

CaselIT:
it may be worth a look though, from the pytest docs it seems to be the "official" way to do a custom parametrize

zzzeek:
pytest_generate_tests is still called from inside of _getnfunctions() though .  as long as  I need to be calling pytest_pycollect_makeitem I can't get around it.   now maybe i dont need to be using that hook, looking at it now, but this is not easy stuff.  if i take it out, nothing runs.  we have a custom "want test" logic that needs to be somewhere.

zzzeek:
I put out a ping to @nicoddemus on twitter.

zzzeek:
sorry I messed up your PR with comments.    feel free to make new PRs, e.g. if we can also break this up between Core and ORM.

CaselIT:
> sorry I messed up your PR with comments. feel free to make new PRs, e.g. if we can also break this up between Core and ORM.

Don't worry, I also contributed. I may split it in core and ORM

From https://github.com/sqlalchemy/sqlalchemy/pull/4896#issuecomment-543072187
> >Regarding core, I guess it should reject an update/insert with a computed column set.

>I though a bit about this, and I'm not sure if it's required. What I mean is that when using core an user is specifying effectively an sql query, so it should be the db to tell him if it is doing something that is wrong, like setting a value to a computed column if his backend does not support it.

If you agree of leaving the error to the db, I think the core part is ready for a review then

CaselIT:
I've rebased the branch, and opened a new pr for the core part in #4928 that should be done.

In this pr I'll look ad implementing the missing stuff for the orm

CaselIT:
I've implemented the logic to prevent setting to computed column in a mapped class.

```py
class Foo(Base):
    __tablename__ = "foo"

    foo = Column(Text)
    cp1 = column_property(foo + " warn")  # warn on construction of the class
    cp2 = column_property(foo + " ok", raise_on_set=False)
    cp3 = column_property(foo + " raise", raise_on_set=True)
    comp = Column(Integer, Computed("id * 42"))


f = Foo()
f.cp1 = "bar"
f.cp2 = "baz"
f.cp3 = "raise"  # this raises Attribute error
f.comp = 33  # this raises Attribute error
```

The behavior is controlled by the keyword `raise_on_set` of `ColumnProperty`

When setting to a computed column Attribute error is raised.
When defining a `column_property` directly a warning is raised, specifying that in a future version of sqlachemy the current behavior (or ignoring the sett value) will be replaced with a raise on set, and to set `raise_on_set=False` to keep the current behavior.

I'm not very sure of my implementation, since the orm is quite complex and my experience in it is very limited. it seems to work at preventing direct sets, and it does not interfere with the loading from db logic.
In particular I don't like the proxy class `_ColumnPropertyProxy` that I have created to wrap the `column_property` and it is used to set the `raise_on_set` to `'warn'` now and to True in a future version.

Where should I place the test for this?

CaselIT:
Once #4928 is merged I'll rebase this so we can add the orm support

CaselIT:
I'll open a new pr with the raise on set code for the generated column on the orm. Do you prefer if I also open an issue?

zzzeek:
let's do a reset and start with a new issue, and define the problem fully.   this applies to any column_property() that's an expression, right?    all of these are "ignore on write".  this should be a generalized thing and should likely be packaged all up for 1.4.   I think emitting a warning for now is fine, people definitely see warnings these days.

CaselIT:
Sorry I've read it only after opening a new PR with the code I had implemented here.

I can still open an issue

> this applies to any column_property() that's an expression, right? all of these are "ignore on write".

yes.

zzzeek:
not a big deal.  I'm assuming this is not priority right now, as I'm still in the midst of a long-term escape from under a rock

CaselIT:
No definitely not. You find it in #4972. I've cherry picked the commit I had added on this pr there.

Good luck on you escape 🏃‍♂️

PR #4928, opened 2019-10-18
[core] Support for generated columns

CaselIT:
Support for generated column

this it just the core parts of #4896
part of #4894

### DIFF
--- a/lib/sqlalchemy/dialects/mysql/base.py
+++ b/lib/sqlalchemy/dialects/mysql/base.py
@@ -1505,7 +1505,7 @@ class MySQLDDLCompiler(compiler.DDLCompiler):
 
         if column.computed is not None:
             # the default visit_computed_column is compatible
-            colspec.append(self.process(column.computed))
+            colspec.append(self.process(column.computed).strip())
 
         is_timestamp = isinstance(
             column.type._unwrapped_dialect_impl(self.dialect),
--- a/lib/sqlalchemy/dialects/mysql/base.py
+++ b/lib/sqlalchemy/dialects/mysql/base.py
@@ -1503,6 +1503,10 @@ class MySQLDDLCompiler(compiler.DDLCompiler):
             ),
         ]
 
+        if column.computed is not None:
+            # the default visit_computed_column is compatible
+            colspec.append(self.process(column.computed))
+
         is_timestamp = isinstance(
             column.type._unwrapped_dialect_impl(self.dialect),
             sqltypes.TIMESTAMP,
--- a/lib/sqlalchemy/dialects/oracle/base.py
+++ b/lib/sqlalchemy/dialects/oracle/base.py
@@ -1187,6 +1187,20 @@ class OracleDDLCompiler(compiler.DDLCompiler):
 
         return "".join(table_opts)
 
+    def visit_computed_column(self, generated):
+        text = " GENERATED ALWAYS AS (%s)" % self.sql_compiler.process(
+            generated.sqltext, include_table=False, literal_binds=True
+        )
+        # explicitly check for True|False since None means server default
+        if generated.persisted is True:
+            raise exc.CompileError(
+                "oracle does not support persited (stored) computed "
+                "columns. Remove 'persisted' or set it to False"
+            )
+        elif generated.persisted is False:
+            text += " VIRTUAL"
+        return text
+
 
 class OracleIdentifierPreparer(compiler.IdentifierPreparer):
 
--- a/lib/sqlalchemy/dialects/oracle/base.py
+++ b/lib/sqlalchemy/dialects/oracle/base.py
@@ -1194,7 +1194,7 @@ class OracleDDLCompiler(compiler.DDLCompiler):
         # explicitly check for True|False since None means server default
         if generated.persisted is True:
             raise exc.CompileError(
-                "oracle does not support persited (stored) computed "
+                "oracle does not support persisted (stored) computed "
                 "columns. Remove 'persisted' or set it to False"
             )
         elif generated.persisted is False:
--- a/lib/sqlalchemy/dialects/sqlite/base.py
+++ b/lib/sqlalchemy/dialects/sqlite/base.py
@@ -1032,6 +1032,9 @@ class SQLiteCompiler(compiler.SQLCompiler):
 
 class SQLiteDDLCompiler(compiler.DDLCompiler):
     def get_column_specification(self, column, **kwargs):
+        if column.computed is not None:
+            raise exc.CompileError("sqlite does not support computed columns")
+
         coltype = self.dialect.type_compiler.process(
             column.type, type_expression=column
         )
--- a/lib/sqlalchemy/testing/suite/test_select.py
+++ b/lib/sqlalchemy/testing/suite/test_select.py
@@ -863,7 +863,7 @@ class LikeFunctionsTest(fixtures.TablesTest):
 
 class ComputedColumnTest(fixtures.TablesTest):
     __backend__ = True
-    __unsupported_on__ = ("sqlite",)
+    __unsupported_on__ = ("sqlite", "postgresql < 12")
 
     @classmethod
     def define_tables(cls, metadata):
--- a/lib/sqlalchemy/testing/suite/test_select.py
+++ b/lib/sqlalchemy/testing/suite/test_select.py
@@ -9,6 +9,7 @@ from ..schema import Column
 from ..schema import Table
 from ... import bindparam
 from ... import case
+from ... import Computed
 from ... import false
 from ... import func
 from ... import Integer
@@ -858,3 +859,46 @@ class LikeFunctionsTest(fixtures.TablesTest):
         col = self.tables.some_table.c.data
         self._test(col.contains("b%cd", autoescape=True, escape="#"), {3})
         self._test(col.contains("b#cd", autoescape=True, escape="#"), {7})
+
+
+class ComputedColumnTest(fixtures.TablesTest):
+    __backend__ = True
+    __unsupported_on__ = ("sqlite",)
+
+    @classmethod
+    def define_tables(cls, metadata):
+        Table(
+            "square",
+            metadata,
+            Column("id", Integer, primary_key=True),
+            Column("side", Integer),
+            Column("area", Integer, Computed("side * side")),
+            Column("perimeter", Integer, Computed("4 * side")),
+        )
+
+    @classmethod
+    def insert_data(cls):
+        config.db.execute(
+            cls.tables.square.insert(),
+            [{"id": 1, "side": 10}, {"id": 10, "side": 42}],
+        )
+
+    def test_select_all(self):
+        res = config.db.execute(
+            select([text("*")]).select_from(self.tables.square)
+        ).fetchall()
+        assert len(res) == 2
+        for row in res:
+            assert row.area is not None
+            assert row.perimeter is not None
+            assert row.area == row.side ** 2
+            assert row.perimeter == row.side * 4
+
+    def test_select_columns(self):
+        res = config.db.execute(
+            select([self.tables.square.c.area, self.tables.square.c.perimeter])
+        ).fetchall()
+
+        assert len(res) == 2
+        for row in res:
+            assert row.area == row.perimeter ** 2 / 16
--- a/test/dialect/mssql/test_compiler.py
+++ b/test/dialect/mssql/test_compiler.py
@@ -1195,7 +1195,7 @@ class CompileTest(fixtures.TestBase, AssertsCompiledSQL):
         )
 
     @testing.combinations(
-        ("no_persisted", "", ...),
+        ("no_persisted", "", Ellipsis),
         ("persisted_none", "", None),
         ("persisted_true", " PERSISTED", True),
         ("persisted_false", "", False),
@@ -1203,7 +1203,7 @@ class CompileTest(fixtures.TestBase, AssertsCompiledSQL):
     )
     def test_column_computed(self, text, persisted):
         m = MetaData()
-        kwargs = {"persisted": persisted} if persisted != ... else {}
+        kwargs = {"persisted": persisted} if persisted != Ellipsis else {}
         t = Table(
             "t",
             m,
--- a/test/dialect/mssql/test_compiler.py
+++ b/test/dialect/mssql/test_compiler.py
@@ -1195,7 +1195,7 @@ class CompileTest(fixtures.TestBase, AssertsCompiledSQL):
         )
 
     @testing.combinations(
-        ("no_persisted", "", Ellipsis),
+        ("no_persisted", "", "ignore"),
         ("persisted_none", "", None),
         ("persisted_true", " PERSISTED", True),
         ("persisted_false", "", False),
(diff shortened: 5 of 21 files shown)


CaselIT:
if https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1523 lands before this is done, I think we could refactor the test using `combinations`

zzzeek:
OK I've merged `@testing.combinations` which ultimately calls pytest.parametrize.  want to give that a shot?  also do we want to close #4896 ?

CaselIT:
> want to give that a shot?

Nice, I'll convert the test later then

> also do we want to close #4896 ?

I'm using that pr to implement the orm part of the generated column. It continues after this pr.
I've done a first rough implementation, but I'm not sure if that is the best way to do it.

I can open a different pr if you prefer.

CaselIT:
I'm looking at combinations. What does `a` stands for?

CaselIT:
I've updated the tests.
I think you can review

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 19a1b032683a6ee600df9e1f65ef0d3cec88cbeb of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 19a1b032683a6ee600df9e1f65ef0d3cec88cbeb: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1539

zzzeek:
this is looking great.  the next thing that should be within this commit are backend round trip tests in lib/sqlalchemy/testing/suite/ ....maybe test_select.py for now.   it would be a new ComputedColumnTest and would do a create / drop and also round trip a row from a SELECT to see that it gets something back, this is more to test that the DDL actually does the right thing on the backend and that there is no driver weirdness when fetching it or anything like that.

I can do it if you'd like if you want to be done, or you can keep on going with it, thanks for your efforts!

CaselIT:
Nice, thanks for the review.

I'll check pep8 and why py2.7 is not working. (Also do you plan to drop support for it in the near future?)

I'll try adding the select test later today.

zzzeek:
the python 2 plan is that "SQLAlchemy 2.0" will not support Python 2.

SQLAlchemy 2 is tentatively planned for the major version after 1.4 (note SQLAlchemy "major versions" have up til now been the "minor version" number via semver), however 1.4 is getting quite full with a lot of big changes that precede the 2.0 changes, so there might even be a 1.5.     I can see *maybe* dropping py2k for 1.5 if it goes that way.   

anyway, we still have at least a year to go.

zzzeek:
for the pep8 stuff I use pre_commit with black / zimports / flake8, if you look in .pre-commit-config.yaml.   that will do everything except for formatting issues in docstrings.  the flake8 check, with plugins instlaled, will show if there's anything that wont pass the extended flake8 checks in the commit.

CaselIT:
> the python 2 plan is ..

seems clear

ok regarding pep8

test/dialect/mssql/test_compiler.py
```diff
@@ -1195,15 +1195,15 @@ def test_try_cast(self):
         )
 
     @testing.combinations(
-        ("no_persisted", "", ...),
+        ("no_persisted", "", Ellipsis),
```
zzzeek: I had never heard of this object before.  what does it mean in this context?

CaselIT: Is is used sometimes in the indexes, like numpy uses it sometimes. It is not very used otherwise.

Here it is just a flag that I use to check it `persisted` should be added to the attributes.
I guess `skip_arg` or a `flag=object()` would also work here

CaselIT:
I've fixed python 2.7 (I think). It does not like `...` for `Ellipsis`. The pep8 stuff seems only regarding the import order.

I've added a couple of tests in test_select
I'm not sure if a test without sqlalchemy should be also done. I mean with the raw connection, to verify that it is not sqlalchemy that generates the columns

I've thought at what other test could be done:

- insert is covered by the select one.
- update maybe could be tested, but we are checking that the db is working correctly here, so not sure useful

test/dialect/mssql/test_compiler.py
CaselIT: I guess it has also been repurposed a bit in the typing library to generate empty functions/classes and specify some types

zzzeek:
absolutely, I haven't had time to look at it.   had some time to push through small things yesterday but I'm really having a shortage of time, you are on the list

CaselIT:
Sorry, I did not mean to pressure you.

zzzeek:
no worries you've done a great job on this pr

sqla-tester:

recheck

lib/sqlalchemy/dialects/mysql/base.py
```diff
@@ -1503,6 +1503,10 @@ def get_column_specification(self, column, **kw):
             ),
         ]
 
+        if column.computed is not None:
+            # the default visit_computed_column is compatible
+            colspec.append(self.process(column.computed).strip())
```
sqla-tester: **mike bayer** (zzzeek) wrote:

why do we need the strip() here?

lib/sqlalchemy/dialects/oracle/base.py
```diff
@@ -1187,6 +1187,20 @@ def post_create_table(self, table):
...
         return "".join(table_opts)
 
+    def visit_computed_column(self, generated):
+        text = " GENERATED ALWAYS AS (%s)" % self.sql_compiler.process(
+            generated.sqltext, include_table=False, literal_binds=True
+        )
+        # explicitly check for True|False since None means server default
+        if generated.persisted is True:
+            raise exc.CompileError(
+                "oracle does not support persited (stored) computed "
```
sqla-tester: **mike bayer** (zzzeek) wrote:

typo 'persited'

lib/sqlalchemy/dialects/sqlite/base.py
```diff
@@ -1032,6 +1032,9 @@ def visit_empty_set_expr(self, element_types):
 
 class SQLiteDDLCompiler(compiler.DDLCompiler):
     def get_column_specification(self, column, **kwargs):
+        if column.computed is not None:
```
sqla-tester: **mike bayer** (zzzeek) wrote:

maybe just let it get passed through to SQLite and let it raise?   because SQLite might add this feature.

sqla-tester (commented):

(5 comments)

going to rerun pep8 in a moment...some changes so far, mostly we have to figure out the "..." thing.   i haven't looked to see what this does yet can that just be None?

test/sql/test_computed.py
```diff
@@ -0,0 +1,54 @@
...
+    fixtures,
+    assert_raises_message,
+)
+
+
+class DDLComputedTest(fixtures.TestBase, AssertsCompiledSQL):
+    __dialect__ = "default"
+
+    @combinations(
+        ("no_persisted", "", Ellipsis),
```
sqla-tester: **mike bayer** (zzzeek) wrote:

we're still doing Python 2 so this has to be something that works in py2k also.

test/dialect/mssql/test_compiler.py
sqla-tester: **mike bayer** (zzzeek) wrote:

python 2 is not accepting this.

sqla-tester:

pep8-recheck

zzzeek:
oh crap did i not pull the most recent version in?

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 9348b2d6396dea0b56c5750ab042601b810f0be0 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 9348b2d6396dea0b56c5750ab042601b810f0be0 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1539

zzzeek:
ok ignore the ellipsis thing , sorry

sqla-tester:

Code-Review-1

the pep8 failures are import order stuff, use the zimports tool to clean them up

./test/dialect/oracle/test_compiler.py:5:1: I100 Import statements are in the wrong order. 'from sqlalchemy import Computed' should be before 'from sqlalchemy import except_'
from sqlalchemy import Computed
^
./test/dialect/oracle/test_compiler.py:33:1: I100 Import statements are in the wrong order. 'from sqlalchemy.testing import assert_raises_message' should be before 'from sqlalchemy.testing import AssertsCompiledSQL'
from sqlalchemy.testing import assert_raises_message
^
./test/sql/test_computed.py:5:1: I101 Imported names are in the wrong order. Should be assert_raises_message, AssertsCompiledSQL, combinations, fixtures
from sqlalchemy.testing import (
^

sqla-tester:

database failures are currently postgresql:

psycopg2.errors.SyntaxError: syntax error at or near "("
LINE 5:  area INTEGER GENERATED ALWAYS AS (side * side) STORED,

lib/sqlalchemy/dialects/mysql/base.py
CaselIT: The computed compiler leaves a space at the end. This dialect is a bit different from the other because it uses a list then joins it a space at the end.

I'm not sure if the additional space can be left or if it will cause problems.

lib/sqlalchemy/dialects/sqlite/base.py
CaselIT: I raise the error so it is consistent with the other implementations that raise an error in case of an unsupported option. Like in oracle it raises an error if persisted is true.
If we do not raise an error here I think we also should not raise an error in case of an unsupported option in the other dialects

test/sql/test_computed.py
CaselIT: I'll change this.
The logic of the test is that we have 4 cases:
- Persisted=true
- Persisted=false
- Persisted=None
- Persisted not provided

I use Ellipsis for the last case, to ignore the value passed to the function, and call only `Computed`

CaselIT:
> ok ignore the ellipsis thing , sorry

No it makes sense clarifing the test. I'll change it

> database failures are currently postgresql:

Strange, i've run the test with postgress locally. Which version of postgress is it testing against? It must be the 12+

sqla-tester:

(1 comment)

> CaselIT@github wrote:
 > 
 > > ok ignore the ellipsis thing , sorry
 > 
 > No it makes sense clarifing the test. I'll change it
 > 
 > > database failures are currently postgresql:
 > 
 > Strange, i've run the test with postgress locally. Which version of
 > postgress is it testing against? It must be the 12+

the tests run against PG 9.6, 10, 11 and 12.   So only the 12 is passing, so you need to have a @testing.requires.computed_columns requirement added to test/requirements.py.    Also the computed column thing should be something that 3rd party dialects can support also so we should have some tests in lib/sqlalchemy/testing/suite/ as well as a closed @testing.requires.computed_columns in lib/sqlalchemy/testing/requirements.py

lib/sqlalchemy/dialects/sqlite/base.py
sqla-tester: **mike bayer** (zzzeek) wrote:

well when we see an option for which we don't know what SQL to generate, that's harder to ignore because we can't have it emit an error from the DB.   it's fine to leave this as is however.

CaselIT: Ok

CaselIT:
Thanks for the review.

I'll have a look on it and add them.

CaselIT:
I've fixed the typo and replaced the use of ellipsis in the test, with a cleared string 'ignore'. Also the pep8 stuff should be fixed

I've tried to add the requirements as you suggested, and it should work now event on previous postgres versions

lib/sqlalchemy/testing/suite/test_select.py
```diff
@@ -858,3 +859,46 @@ def test_contains_autoescape_escape(self):
         col = self.tables.some_table.c.data
         self._test(col.contains("b%cd", autoescape=True, escape="#"), {3})
         self._test(col.contains("b#cd", autoescape=True, escape="#"), {7})
+
+
+class ComputedColumnTest(fixtures.TablesTest):
+    __backend__ = True
+    __unsupported_on__ = ("sqlite", "postgresql < 12")
```
zzzeek: this should say:

    __requires__ = ('computed_columns', )

zzzeek:
> I've fixed the typo and replaced the use of ellipsis in the test, with a cleared string 'ignore'. Also the pep8 stuff should be fixed
> 
> I've tried to add the requirements as you suggested, and it should work now event on previous postgres versions

thanks!   i''ll run this through now.   most everything is here so I can take this over whenever you are burnt out on it thanks for all the efforts

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision d39c521d5ac6ebfb4fb5b53846451de79752e64c of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset d39c521d5ac6ebfb4fb5b53846451de79752e64c added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1539

sqla-tester:

MySQL reports:

(1064, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'GENERATED ALWAYS AS (side * side), \n\tperimeter INTEGER GENERATED ALWAYS AS (4 * ' at line 4")

I'm going to guess this might be because engine="MyISAM"?   try adding mysql_engine="InnoDB"

CaselIT:
> thanks! i''ll run this through now. most everything is here so I can take this over whenever you are burnt out on it thanks for all the efforts

It will provably be quicker if you can finish this up. I cannot run all the tests so it's hard to cover all the cases. Thanks

Regarding the mysql error I don't know. It may also be the version, if I'm not mistaken it's not too long that it supports the generated columns

zzzeek:
oh it's just mysql 5.6 doesn't have computed columns.  might be time to drop 5.6 test support.

CaselIT:
Thanks for making the test pass

sqla-tester:

Code-Review+1

Nice

sqla-tester:

I've added documentation as this is a major new feature and also adjusted the docs for Computed.   feel free to give it a quick proof.  there is also a backport for 1.3.    there's a lot here this is very big.

CaselIT:
I've given the documentation a look and it seems clear.
Thanks for finalizing this feature.

CaselIT:
I thought a bit more about the documentation. I don't know if we should add a note specifing that expression supported in the computed column depends on the database. Like some databases support computed columns that reference other computed columns, some don't.

zzzeek:
All of these DDL things have huge lists of caveats, that's one where they'll get an immedaite error raise if that's the case, so they'll figure it out :)

CaselIT:
Yes, that was my guess also

sqla-tester:

Code-Review+2 Workflow+1

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1539 has been **merged**. Congratulations! :)

sqla-tester:

Code-Review+2 Workflow+1

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1559 has been **merged**. Congratulations! :)

CaselIT:
Nice

PR #4929, opened 2019-10-19
Remove return statement from __init__

heckad:


zzzeek:
hm did you notice that just reading the code ?

i wonder why linters dont find that

zzzeek:
alrgihty.. no F code for that one

zzzeek:
thanks!

Issue #4959, opened 2019-11-02
chunk iterator support for results


sumau:
It would be helpful to be able to 'automatically' iterate through fetchmany, similar to pd.read_sql(chunksize) or as suggested here:
http://code.activestate.com/recipes/137270-use-generators-for-fetching-large-db-record-sets/

So for example instead of writing the following in [this](https://docs.sqlalchemy.org/en/13/_modules/examples/performance/large_resultsets.html) example:
```python
    with engine.connect() as conn:
        result = conn.execution_options(stream_results=True).execute(
            Customer.__table__.select().limit(n)
        )
        while True:
            chunk = result.fetchmany(10000)
            if not chunk:
                break
            for row in chunk:
                row["id"], row["name"], row["description"]
```
you could write something like:
```python
    with engine.connect() as conn:
        result = conn.execution_options(stream_results=True, arraysize=10000).execute(
            Customer.__table__.select().limit(n)
        )
   for chunk in result:
          for row in chunk:
                row["id"], row["name"], row["description"]
```
I'm using arraysize to be in accordance with pep249 dbapi cursor.arraysize but perhaps chunksize might be better

zzzeek:
hi there -

unfortunately that would entirely change how the resultproxy works both from an internal as well as an external point of view.   at most, I would consider adding a chunks() call to result however I'm not ready to do that right now as I am seeking to unify the API of ResultProxy and Query.   For that work I will attempt to consider this use case as well.

zzzeek:
the activestate recipe is wrong.  with the vast majority of DBAPIs, it doesn't matter if you call fetchall() or fetchmany(), the rows are already fetched into memory fully after execute() is called.   so I think this feature is kind of misleading since fetchmany() is already misleading.    The option that should be considered here is stream_results: https://docs.sqlalchemy.org/en/13/core/connections.html?highlight=stream_results#sqlalchemy.engine.Connection.execution_options.params.stream_results   only works with three DBAPis right now - without it, there are no memory savings to fetchmany().

sumau:
Makes sense - I'll add a generator where relevant in my code base as a workaround for now :-)

zzzeek:
thanks!  sorry, I do want to do this, but ive been spending months just thinking about the result set and how to unify Core / ORM the best way.

zzzeek:
[...]

PR #4960, opened 2019-11-02
Fixes: #4959 Ability to Iterate through fetchmany

sumau:
### Checklist

This pull request is:

- [x ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/engine/result.py
+++ b/lib/sqlalchemy/engine/result.py
@@ -863,6 +863,9 @@ class ResultProxy(object):
             self.connection._echo and context.engine._should_log_debug()
         )
         self._init_metadata()
+        self._arraysize = self.context.execution_options.get(
+            "arraysize", 1
+        )
 
     def _getter(self, key, raiseerr=True):
         try:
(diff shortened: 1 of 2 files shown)


zzzeek (commented):
hi -

this is not a small change and as discussed on the parent issue, there is a lot of consdieration to be made for this before we implement.    in particular, the term "arraysize" does not imply any of this behavior in pep-249 and it is also not a good idea to have the iteration of ResultProxy iterate through an entirely different kind of object (chunks) rather than rows based on an option that was set far away.    the "arraysize" concept is not implemented by most DBAPIs anyway, DBAPIs usually buffer rows fully in any case unless special server side cursor APIs are used.  The use case being requested here will be considered but the API will likely not look like this, it would likely be within result itself.

zzzeek:
prefer to close this for now as I am planning to design a completely new "result" API that will unify ResultProxy and that of Query, additionally "chunking" at the DBAPI level is normally only a convenience feature, not something that saves on any performance since most DBAPIs buffer result sets entirely unless they provide for special streaming options which SQLAlchemy also supports (cx_Oracle might be the exception to this).

Issue #4964, opened 2019-11-05
SQLAlchemy 1.3.10 issue with python3


smuzaffar:
While testing SQLAlchemy 1.3.10 with python 3.6.4 we are getting error [a]. Patching it with patch https://github.com/cms-sw/cmsdist/pull/5334/commits/e7322933f3fd4240d094a638d6382564c6c40295 allowed us to run our tests.

[a]
```
  File "sqlalchemy/1.3.10/lib/python3.6/site-packages/sqlalchemy/event/attr.py", line 285, in _exec_once_impl
    self(*args, **kw)
  File "sqlalchemy/1.3.10/lib/python3.6/site-packages/sqlalchemy/event/attr.py", line 322, in __call__
    fn(*args, **kw)
  File "sqlalchemy/1.3.10/lib/python3.6/site-packages/sqlalchemy/util/langhelpers.py", line 1485, in go
    return once_fn(*arg, **kw)
  File "sqlalchemy/1.3.10/lib/python3.6/site-packages/sqlalchemy/engine/strategies.py", line 199, in first_connect
    dialect.initialize(c)
  File "sqlalchemy/1.3.10/lib/python3.6/site-packages/sqlalchemy/dialects/oracle/base.py", line 1247, in initialize
    super(OracleDialect, self).initialize(connection)
  File "sqlalchemy/1.3.10/lib/python3.6/site-packages/sqlalchemy/engine/default.py", line 319, in initialize
    max_ident_length = self._check_max_identifier_length(connection)
  File "sqlalchemy/1.3.10/lib/python3.6/site-packages/sqlalchemy/dialects/oracle/base.py", line 1300, in _check_max_identifier_length
    if self._get_effective_compat_server_version_info(connection) &gt;= (
  File "sqlalchemy/1.3.10/lib/python3.6/site-packages/sqlalchemy/dialects/oracle/base.py", line 1262, in _get_effective_compat_server_version_info
    if self.server_version_info < (12, 2):
TypeError: '<' not supported between instances of 'NoneType' and 'tuple'
```

zzzeek:
hi there -

how is it that your server version info is None when you have an active connection?

zzzeek:
Hi there -

Unfortunately I'm not able to act upon this without the information I requested.  Please reopen when you are available to discuss the rationale.   Once that happens, you can resubmit your PR.

smuzaffar:
sorry I was away and could not reply.
[...]

PR #4965, opened 2019-11-05
python3 does not support comparison between NoneType and tuple

smuzaffar:
### Description
Python3 does not allow comparison between `NoneType` and `tuple`. Without this change we get error mentioned in the issue https://github.com/sqlalchemy/sqlalchemy/issues/4964

### Checklist

This pull request is:

- [X] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/oracle/base.py
+++ b/lib/sqlalchemy/dialects/oracle/base.py
@@ -1290,7 +1290,7 @@ class OracleDialect(default.DefaultDialect):
         super(OracleDialect, self).initialize(connection)
 
         self.implicit_returning = self.__dict__.get(
-            "implicit_returning", self.server_version_info > (10,)
+            "implicit_returning", self.server_version_info and self.server_version_info > (10,)
         )
 
         if self._is_oracle_8:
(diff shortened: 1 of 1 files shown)


lib/sqlalchemy/dialects/oracle/base.py
```diff
@@ -1290,7 +1290,7 @@ def initialize(self, connection):
         super(OracleDialect, self).initialize(connection)
 
         self.implicit_returning = self.__dict__.get(
-            "implicit_returning", self.server_version_info > (10,)
+            "implicit_returning", self.server_version_info and self.server_version_info > (10,)
```
zzzeek: server_version_info should never be None here since there is a cx_Oracle database connection.

zzzeek:
hi there -

this PR has no tests and I need to know what the use case is here, as when initialize() is called, that's only after the DBAPI connection is made, which is guaranteed to have a server version.

if OTOH the scenario here is some mock connection type of test, a mock connection passed into a SQLAlchemy dialect needs to act enough like a real connection so that errors do not occur.  otherwise it's not mocking adequately.     there's lots of mock connection tests in SQLAlchemy's test suite, and those tests have to ensure that things like server version info and such are also patched out.

zzzeek:
the issue is closed without reproducing information so this PR will need to be re-submitted.

PR #4970, opened 2019-11-09
Add note to hybrid.py setter documentation

AllanDaemon:
### Description
Almost the same note of the defining expression. Although it's already said implicitly in the note of the session before, one searching about getters and setters for hybrid properties may skip it, so it's better redundancy and beeing explicitly about it to avoid headaches.

### Motivation

I had a lot of pain with this issue, and I rechecked the documentation many times trying to solve it but I missed the note on the defining expression because I wasn't interested in using it. I was kind of expecting something like this note.

The redaction here is a suggestion based on the previous note.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/ext/hybrid.py
+++ b/lib/sqlalchemy/ext/hybrid.py
@@ -207,6 +207,24 @@ The ``length(self, value)`` method is now called upon set::
     >>> i1.end
     17
 
+.. note:: When defining an setter for a hybrid property or method, the
+   expression method **must** retain the name of the original hybrid, else
+   the new hybrid with the additional state will be attached to the object
+   with the non-matching name. To use the example above::
+
+    class Interval(object):
+        # ...
+
+        @hybrid_property
+        def radius(self):
+            return self.length / 2
+
+        # WRONG - the non-matching name will cause this function to be
+        # ignored
+        @radius.setter
+        def radius_setter(self, radius):
+            self.length = radius * 2
+
 .. _hybrid_bulk_update:
 
 Allowing Bulk ORM Update


lib/sqlalchemy/ext/hybrid.py
```diff
@@ -207,6 +207,24 @@ def length(self, value):
     >>> i1.end
     17
 
+.. note:: When defining an setter for a hybrid property or method, the
```
zzzeek: "defining a setter"

lib/sqlalchemy/ext/hybrid.py
```diff
@@ -207,6 +207,24 @@ def length(self, value):
...
+   expression method **must** retain the name of the original hybrid, else
```
zzzeek: setter method **must**

lib/sqlalchemy/ext/hybrid.py
zzzeek: "for a hybrid property"  - there are no setters for hybrid methods

lib/sqlalchemy/ext/hybrid.py
```diff
@@ -207,6 +207,24 @@ def length(self, value):
...
+   the new hybrid with the additional state will be attached to the object
+   with the non-matching name. To use the example above::
+
+    class Interval(object):
+        # ...
+
+        @hybrid_property
```
zzzeek: please use the "def length" method from the "Defining Setters" example so that it can be directly contrasted to the correct form.

zzzeek (changes requested):
OK, I didn't remember writing the other note that way but there it is, so please make these changes so that the new section is correct.  thanks!

zzzeek:
please reopen if you can make the requested changes.  thanks!

PR #4972, opened 2019-11-09
Support raising on set for read only attributes

CaselIT:
As part of the implementation for the Computed columns the discussion in #4896 pointed out that currently the values set to read only attributes, like the Computed column or column_property, are silently ignored by the orm.

This PR implements a check that prevents setting these read only attributes, by raising an `AttributeError` similar to the default python behavior when setting a read only property 

This feature is controlled by a new attribute of `ColumnProperty`, `raise_on_set`, that defaults to False on normal columns.
For Computed columns this attribute is set to True. Since this is a breaking change on the current `column_property` behavior, the default value for them is 'warn', to raise a warning to the user regarding the future change in behavior. After the warning `raise_on_set` is set to False, keeping the current behaviour. It is left to a future release to change the default behavior to raise.

Example

```py
class Foo(Base):
    __tablename__ = "foo"

    foo = Column(Text)
    cp1 = column_property(foo + " warn")  # warn on construction of the class
    cp2 = column_property(foo + " ok", raise_on_set=False)
    cp3 = column_property(foo + " raise", raise_on_set=True)
    comp = Column(Integer, Computed("id * 42"))


f = Foo()
f.cp1 = "bar" # this value is ignored
f.cp2 = "baz" # this value is ignored
f.cp3 = "raise"  # this raises Attribute error
f.comp = 33  # this raises Attribute error
```

Ref to initial implementation https://github.com/sqlalchemy/sqlalchemy/pull/4896#issuecomment-544130268

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**


lib/sqlalchemy/orm/mapper.py
```diff
@@ -1863,8 +1864,10 @@ def _property_from_column(self, key, prop):
                             "force this column to be mapped as a read-only "
                             "attribute." % (key, self, c)
                         )
+                if mc.computed is not None:
+                    kwargs["raise_on_set"] = True
```
CaselIT: This should be generalized, maybe by adding a method in the colum class, so it's easier in the future to specify if a column is read only, without a list of `if ... or ... or ..`

lib/sqlalchemy/orm/properties.py
```diff
@@ -335,3 +358,14 @@ def reverse_operate(self, op, other, **kwargs):
...
+
+class _ColumnPropertyProxy(ColumnProperty):
+    def __init__(self, *columns, **kwargs):
+        # will be changed to True in a future release
+        kwargs.setdefault("raise_on_set", "warn")
+        super().__init__(*columns, **kwargs)
+
+
+_ColumnPropertyProxy.__doc__ = ColumnProperty.__doc__
+_ColumnPropertyProxy.__init__.__doc__ = ColumnProperty.__init__.__doc__
```
CaselIT: I'm not particularly happy by this implementation to change the default value in the `column_property` while keeping the default to False in the `ColumnProperty`.

I think that `public_factory` should handle this. What do you think?

CaselIT:
The current change is just a first implementation on this to get some feedback.
I've added a couple of question I have on the code

Where should I add the test for this?

lib/sqlalchemy/orm/properties.py
```diff
@@ -335,3 +358,14 @@ def reverse_operate(self, op, other, **kwargs):
 
     def __str__(self):
         return str(self.parent.class_.__name__) + "." + self.key
+
+
+class _ColumnPropertyProxy(ColumnProperty):
```
zzzeek: why do we need a new class?

lib/sqlalchemy/orm/attributes.py
```diff
@@ -270,6 +270,8 @@ class InstrumentedAttribute(QueryableAttribute):
     """
 
     def __set__(self, instance, value):
+        if self.prop.raise_on_set is True:
```
zzzeek: tests for this kind of thing start in test/orm/test_attributes.py, then just a quick test for the columnproperty aspect likely in test/orm/test_mapper.py.  these are old test suites so you'd have to step over all the crufty old testing patterns.

lib/sqlalchemy/orm/properties.py
CaselIT: This new class is used by the `public_factory` function of `column_property` to set the default value of the parameter `raise_on_set` to `"warn"` to raise a warning regarding the change in default behaviour of a computed attribute. This class would then be used to change the default value to True to enable the new behaviour by default.
ColumnProperty should always keep its default to False.

Ideally the ability to set a default attribute should probably be on `public_factory`, but it does not support it currently.

As I wrote here https://github.com/sqlalchemy/sqlalchemy/pull/4972#discussion_r344456212 I don't like this implementation, but I don't have another alternative without changing public_factory

zzzeek: public_factory can refer to any classmethod or callable so if we wanted to have two different defaults for "raise_on_set", that would be achieved by putting a `@classmethod` factory on `ColumnProperty`.

However, whether or not a `column_property()` is able to receive a set event is not actually in the user's control.   If the property refers to one of the columns that is mapped to the table, as you've observed is how it is used internally, then the attribute should allow set events to occur.     But the way `column_property()` is used internally, that is, as a wrapper around a `Column` that is mapped, is also [public api](https://docs.sqlalchemy.org/en/13/orm/mapping_columns.html#using-column-property-for-column-level-options).

so while the flag here should be present at the attribute level, from the `ColumnProperty` perspective, it likely needs to inspect the expression given in terms of the parent `Mapper` and determine this value dynamically.   This implies some level of reverse-implementing some of what happens in https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/orm/persistence.py#L569, which is determining if the expression mapped is directly the same as `mapper._propkey_to_col[table]`  where 'table' is also present in `mapper._pks_by_table`.  this could be encapsulated in a method on `Mapper` such as `_expression_is_persistent(expr)` which tests that the expression is one of the mapped columns on one of the persistent mapped tables.

CaselIT: > public_factory can refer to any classmethod or callable so if we wanted to have two different defaults for "raise_on_set", that would be achieved by putting a @classmethod factory on ColumnProperty.

that makes more sense

> But the way column_property() is used internally, that is, as a wrapper around a Column that is mapped, is also public api.

I had not realized this. I thought its use was just to map an sql expression, similar to a computed column. My implementation does not work for other use cases.

I'm not sure I can complete this one in a satisfactory way, since there are a lot more cases that I'm not familiar with. 😞

CaselIT: I can close this and open an issue instead, since this implementation looks like a dead end

zzzeek: it's not too difficult but we should likely have an issue to track the "warn for writing to non-persisted columns" behavior.    it's known at mapping time which attributes will be persistable and which will not.

CaselIT: >  it's known at mapping time which attributes will be persistable and which will not.

is there a property with this information that I could use to set the behavior in column attribute?

zzzeek: no, but im proposing we add one like this to mapper.py.  

looking at where this happens, I think we should just have the mapper set a flag directly on the property, as we have determined if an incoming column_property() is readonly right here: https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/orm/mapper.py#L1718 and also here: https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/orm/mapper.py#L1403.

So there's a little bit of chicken/egg stuff happening with this as these column properties can be added to existing mappers or new mappers and things might happen in different orders, but overall, once the ColumnProperty is all set up, if it were to ask the mapper, "self in self.mapper._readonly_props", that's the correct answer.

CaselIT: will try to work something with it.
are computed column already treated as readonly?

zzzeek: that seems to be the case, yes.   this collection is used to determine properties that we're going to refresh automatically after a flush but it's also the properties for which we aren't going to persist anything.

CaselIT: A computed column seem to not go in `_readonly_props` 

```py
Base = declarative_base()
class Foo(Base):
    side= sa.Column(sa.Float, primary_key=True)
    area = sa.Column(sa.Float, sa.Computed(side*side))
    test = sa.orm.column_property(area + 33)
    __tablename__='foo'
Foo.area.prop in Foo.area.parent._readonly_props # False
Foo.test.prop in Foo.test.parent._readonly_props # True
```

CaselIT:
I've updated the implementation to leverage the mapper to determine if a column is read-only or not as you suggested. It's a lot cleaner this way.

The computed columns are not set as read only by the mapper. If I understood it correctly `_readonly_props` refers to properties that do not exist in the table. A computed column exists in the table, so I think that the current behavior of the mapper makes sense.
I've added instead an attribute of the `ColumnProperty` to indicate if it is read-only, where I check the mapper and if its column is computed. Maybe we could come up with a different name than `read_only`, to avoid confusion the the read-only meaning used in the mapper? (something like `allow_setter`?)

Let me know it this was what you meant by determining the value dynamically

lib/sqlalchemy/util/langhelpers.py
```diff
@@ -1470,14 +1470,14 @@ def __eq__(self, other):
...
-def warn(msg):
+def warn(msg, stacklevel=2):
     """Issue a warning.
 
     If msg is a string, :class:`.exc.SAWarning` is used as
     the category.
 
     """
-    warnings.warn(msg, exc.SAWarning, stacklevel=2)
+    warnings.warn(msg, exc.SAWarning, stacklevel=stacklevel)
```
CaselIT: I'm not using this now so it can be reverted.

I should have probably started from zero, instead of updating the previous implementation

lib/sqlalchemy/orm/properties.py
zzzeek: ah well I think it should.  I can see why it doesn't, because it looks like a regular column.  The mapper should add a test to see if the column acts as "computed" when it determines this collection.   we might want to put a flag on Column that makes this simple, like "_is_readonly" or something.  note the underscore is so we don't have to worry about end users interpreting this as something else.

CaselIT: I guess it depends on the meaning of `_readonly_props` like I wrote here https://github.com/sqlalchemy/sqlalchemy/pull/4972#issuecomment-565815472 

I agree that a property in the column would be useful, so the it can be generalized.

CaselIT:
Let's continue here, since the old conversation was of an outdated file.

Should I make the mapper `_readonly_props` include the computed columns in this issue? The current implementation does not require it since it uses `ColumnProperty.is_readonly`, but it could be removed in favor of checking only `_readonly_props`

zzzeek:
I'm not looking at code yet but the mapper._readonly_props, whether or not it's named well, the purpose of that is to indicate properties that we should try to refresh after a unit of work UPDATE operation.   that is, they have some kind of "computed" value that we can only view by SELECTing the value.   the comment currently says this:

        # determine cols that aren't expressed within our tables; mark these
        # as "read only" properties which are refreshed upon INSERT/UPDATE

This comment should be amended to read:

        # determine cols that are either table-derived expressions, or special table 
        # columns that generate a value and are never written towards.  mark these
        # as "read only" properties which are refreshed upon INSERT/UPDATE and
        # which are also candidates to block if a user-side value is set for them

currently, the only place _readonly_props is used is in [persistence.py](https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/orm/persistence.py#L1380) where the columns and expressions listed are "expired" on the object after it has been UPDATEd, so that they will load their new value on next access.

In the case of the computed column, we haven't noticed the fact that they are not in this readonly collection making a difference, because they *also* have a server_onupdate attribute, and they are still regular columns, so they are getting refreshed usually via the [postfetch](https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/orm/persistence.py#L1491) routine.

one thing that does happen if we make this change, is that these columns are now going to be hit twice within the current unit of work code,that is, expired once in the _postfetch routine and then again in the _finalize_insert_update_commands routine.  this is probably not an issue, but if it is, we can refine these processes to be sensitive to this overlap.

but overall, the main thing I'm always trying to do in order to maintain sanity is to have as few sources of truth as possible.   these computed columns look a whole lot like a column property that's against an expression so i think they should travel in that pack , and we can add extra sensitivities in dealing with _readonly_props if necessary.   if there were instead columnproperty._is_readonly *and* _readonly_props, now you have properties that can have one and not the other, and what does that mean exactly, that's how you drive yourself nuts.

CaselIT:
I agree with your comment. I just wanted to make sure, since I did not know the exact meaning of `_readonly_props`.
I think that the logic when `Colum._is_readonly` is `True` then it is in `Mapper._readonly_props` makes sense, since `Colum._is_readonly` could indicate that the column is in the db, but should behave like a column property

lib/sqlalchemy/orm/mapper.py
```diff
@@ -1398,16 +1398,24 @@ def _configure_pks(self):
...
-        # as "read only" properties which are refreshed upon INSERT/UPDATE
+        # determine cols that are either table-derived expressions, or special
+        # table columns that generate a value and are never written towards.
+        # Mark these as "read only" properties which are refreshed upon
+        # INSERT/UPDATE and which are also candidates to block if a user-side
+        # value is set for them
         self._readonly_props = set(
             self._columntoproperty[col]
             for col in self._columntoproperty
             if self._columntoproperty[col] not in self._identity_key_props
```
CaselIT: I'm not sure if we should manage the case of a computed column that is also primary key

CaselIT:
Ok, the implementation should be done if you want to take a look. I'll start adding some tests.

CaselIT:
I've added some tests. I'm not sure I've covered all cases.

I have not added anything in test/orm/test_attributes.py because I'm not sure what test should go there, since for the test to work it requries a mapper

lib/sqlalchemy/orm/mapper.py
CaselIT: if we keep it this way, should there be a test for it? (ie a primary key will not warn even if computed)

lib/sqlalchemy/orm/attributes.py
```diff
@@ -270,6 +270,13 @@ class InstrumentedAttribute(QueryableAttribute):
     """
 
     def __set__(self, instance, value):
+        if self.prop in self.parent._readonly_props:
```
zzzeek: the AttributeImpl should have a flag right up front called ".readonly", checks that flag.   the flag would be passed through when the impl is set up at https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/orm/strategies.py#L100 .

lib/sqlalchemy/util/langhelpers.py
zzzeek: i've had only bad experiences trying to get stacklevel to be useful

lib/sqlalchemy/sql/schema.py
```diff
@@ -1599,6 +1599,15 @@ def get_children(self, schema_visitor=False, **kwargs):
         else:
             return ColumnClause.get_children(self, **kwargs)
 
+    @property
+    def _is_readonly(self):
```
zzzeek: we would save a lot of method calls when mappers are set up if it were done this way:

```
class Column(...):
    _is_readonly = False

....

class Computed(...):
   def _set_parent(self, parent):
       # ...
       parent._is_readonly = True


```

test/sql/test_metadata.py
```diff
@@ -3902,6 +3903,9 @@ def test_column_info(self):
             c.info["bar"] = "zip"
             assert c.info["bar"] == "zip"
 
+    def test_readonly(self):
+        assert Column(String)._is_readonly is False
```
zzzeek: I know my test suite sucks in this regard but I'm trying to not use "assert", so  is_false(Column(String)._is_readonly)  is_true(Column(String, Computed(...))._is_readonly)

is_false and is_true are in sqlalchemy.testing

test/orm/test_mapper.py
```diff
@@ -1996,6 +1997,38 @@ class B(OldStyle, NewStyle):
...
+        a = A()
+        a.id = 3
+
+        def setComputed():
+            a.comp = 33
+
+        def setProp():
+            a.comp = 33
+
+        assert_warnings(
```
zzzeek: I prefer we use the context manager version here
```

with testing.expect_warnings("the value set to readonly..."):
    a.comp = 33


```

zzzeek (changes requested):
neat!    I left some review comments, good stuff.

lib/sqlalchemy/util/langhelpers.py
CaselIT: i'll remove the change

test/sql/test_metadata.py
CaselIT: I've used assert since that file uses assert in every test. I'll change it

test/orm/test_mapper.py
CaselIT: definitely better

lib/sqlalchemy/sql/schema.py
CaselIT: could't we use `memoized_property` here?

CaselIT:
I've rebased and addressed the comments.
Thanks for the review.

lib/sqlalchemy/sql/schema.py
zzzeek: sure but it wouldn't change the typical number of method calls, that is, create Table + Columns -> pass to mapper, calls the method on every Column.   "computed" is going to be a vanishingly small minority of mapped columns so I like the idea of it leaving no footprint if not used

test/sql/test_metadata.py
zzzeek: yeah sorry, assert is all over the place and now pytest instruments it even but long term id rather have a function doing this instead of a statement.

lib/sqlalchemy/orm/attributes.py
```diff
@@ -270,6 +270,13 @@ class InstrumentedAttribute(QueryableAttribute):
...
 
     def __set__(self, instance, value):
+        if self.impl.readonly:
+            msg = (
+                "the value set to read only attribute '%s.%s' will be ignored."
+                " A future version of sqlalchemy will change this warning to "
+                "raise an exception"
+            ) % (self.class_.__name__, self.key)
+            util.warn(msg)
         self.impl.set(
```
zzzeek: one more change that I can make if you want me to do it on the gerrit side, the warning here could come from the set() method of the impl, that way there's no change here and fewer attribute lookups

lib/sqlalchemy/orm/mapper.py
zzzeek: that would be very strange and I wonder if all the backends even allow the "GENERATED AS" on a primary key.  however, primary key cols not being "writeable" is a little more plausible at least, on SQL Server you can't easily write to an "identity" column in fact, the dialect does some kind of silly magic for that use case that if it were today I might not have it do.

that the identity_key_props are excluded here might indicate there's still some distinction between what "computed" is and what other kinds of column_property() attributes are, but we can always improve it later.

zzzeek (changes requested):
one more thing added, though this is basically ready for being finalized

lib/sqlalchemy/orm/mapper.py
CaselIT: I agree that this is an edge case. I think we could leave it this way and revisit it later it there is a need to.

lib/sqlalchemy/orm/attributes.py
CaselIT: I originally added it here because I don't know  of `impl`can be an instance of different classes, so moving it in the `impl.set` may duplicate the code.

I'm not sure if an attribute lookup will change much in the grand scheme of things. I guess we could do

```py
impl = self.impl
if impl.readlony:
  ...
impl.set()
```

So we are doing the same number of lookups as if the code was moved in` set`

zzzeek: OK, so, here is something, this behavior actually only applies to the ScalarAttributeImpl.    I think we might be better off for now making it local to that.   It definitely doesn't apply correctly to CollectionAttributeImpl since we aren't blocking collection operations.    then we can set "readonly" only in the case that we are dealing with a column loader, like this:

```diff
diff --git a/lib/sqlalchemy/orm/strategies.py b/lib/sqlalchemy/orm/strategies.py
index 79e36026c..8fc8a63a8 100644
--- a/lib/sqlalchemy/orm/strategies.py
+++ b/lib/sqlalchemy/orm/strategies.py
@@ -207,6 +207,7 @@ class ColumnLoader(LoaderStrategy):
             useobject=False,
             compare_function=coltype.compare_values,
             active_history=active_history,
+            readonly=self in mapper._readonly_props
         )
 
     def create_row_processor(
@@ -283,6 +284,7 @@ class ExpressionColumnLoader(ColumnLoader):
             useobject=False,
             compare_function=self.columns[0].type.compare_values,
             accepts_scalar_loader=False,
+            readonly=True   #  because this is ExpressionColumnLoader
         )
 
 
@@ -342,6 +344,7 @@ class DeferredColumnLoader(LoaderStrategy):
             compare_function=self.columns[0].type.compare_values,
             callable_=self._load_for_state,
             load_on_unexpire=False,
+            readonly=self in mapper._readonly_props
         )
 
     def setup_query(

```

sorry this keeps changing but I think that would be better.  As always I am willing to pick this up since I have led you around a bunch already :)

CaselIT: Please feel free to continue this. I'm really not familiar with this part, and is not very easy to follow the code, so at this point I'm just guessing.

zzzeek: OK, no problem, thanks for your efforts on this!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 9aa1f1becef4158c9ca23c7087a6a1ad6eba1e6e of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 9aa1f1becef4158c9ca23c7087a6a1ad6eba1e6e: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1659

zzzeek:
about 18 tests will need adjustment, some of them are doing things like setting attributes to see that they get overwritten so they'd need to assert that the warning is emitted, there are probably some fixtures in there that are just setting every attribute to None, that kind of thing.

CaselIT:
I can take a look.
Should we catch the warning or change the test?

zzzeek:
> I can take a look.
> Should we catch the warning or change the test?

if the test is explciitly about setting the attribute then we'd assert the warning is emitted using "with expect_warnings()".   if it's a side effect of a fixture somehwere we'd probably want to fix that but i am surprised that many tests had a problem, haven't looked to see exactly why.

CaselIT:
So it seems some of the problems is because the orm automagically transforms some expressions in column properties, like here
https://github.com/sqlalchemy/sqlalchemy/blob/190e0139e834e4271268652e058c280787ae69eb/test/orm/test_assorted_eager.py#L1150-L1173

zzzeek:
> So it seems some of the problems is because the orm automagically transforms some expressions in column properties, like here
> 
> https://github.com/sqlalchemy/sqlalchemy/blob/190e0139e834e4271268652e058c280787ae69eb/test/orm/test_assorted_eager.py#L1150-L1173
> 
> that transforms the count `props_cnt` in column property

oh and then the warning is in the one that we want to compare towards.  right?   OK.  see.   that's sort of a valid use case.   you're off the hook for more work here but do you think maybe warning on *flush* is the better option than doing it on set ?

zzzeek:
or otherwise how does one do this, they have their ORM object and they just want to create a sample one that isnt to be persisted.

CaselIT:
yes the warning is on line 1173

CaselIT:
I think the warning on flush actually makes sense

I mean, I'm not sure if it makes sense to set a property that will be computed by the db, but I guess in some cases, like in test it may make sense

But I have no idea how to do it

zzzeek:
> yes the warning is on line 1173

ahha.   OK.  So.  what do we think here?   this is probably why I was a little uncertain about this.    I'm sure a lot of folks out there are assigning to their server side column attributes for various reasons.    

certainly whatever we do , the efforts are not wasted because computed columns need to be in this "readonly" category regardless.

zzzeek:
also im sure you've noticed sqlalchemy is old and big enough that I dont have its full behavioral contract memorized anymore...

CaselIT:
> also im sure you've noticed sqlalchemy is old and big enough that I dont have its full behavioral contract memorized anymore...

every time I start looking for something I find something new, that I did not know, like [with_expression](https://docs.sqlalchemy.org/en/13/orm/loading_columns.html#sqlalchemy.orm.with_expression) just now, so it must be hard to keep it all memorized :)

CaselIT:
I've had a look at the errors, and most of them are the same as the one above.

There is some regarding the computed tests that I'll need to check

CaselIT:
I think I've broken something

A computed column now does not take advantage `testing.db.dialect.implicit_returning` when inserting a row.
https://github.com/sqlalchemy/sqlalchemy/blob/bbe754784ae4630dd0ebf30d3bc2be566f8a8fef/test/orm/test_defaults.py#L234-L239
It now also executes 
```sql
SELECT test.bar AS test_bar
FROM test
WHERE test.id = 1
```

CaselIT:
I think that the problem is that now the computed column is in `mapper._readonly_props`

CaselIT:
Yes commenting line
https://github.com/sqlalchemy/sqlalchemy/blob/a8c74fc9683a35b79508820c274a2ea066450719/lib/sqlalchemy/orm/mapper.py#L1415
resolves the error in the pg test. But it's not what we want.

CaselIT:
So to recap there are now two changes to do:
- warn only when flushing the object, not when setting it
- a computed column does not take advantage of `implicit_returning` in when using `eager_defaults`

I'm don't really know where to start on these changes 😞

zzzeek:
I probably need to change how _readonly_props is interpreted in persistence.py or add a new collection of some kind

zzzeek:
> yes the warning is on line 1173

what file is this referring towards?

CaselIT:
The file I linked, `sqlalchemy/test/orm/test_assorted_eager.py`

lib/sqlalchemy/orm/mapper.py
CaselIT: It seems that mssql allows some computed columns in the primary key. I'm not sure about the other backends. https://docs.microsoft.com/en-us/sql/relational-databases/tables/specify-computed-columns-in-a-table?view=sql-server-ver15

CaselIT:
Say I wanted to give this another go, is there some high level documentation on how the orm suff works in sqlalchemy? Currently it appears as kind a black box that does magic stuff, so it's quite hard to make changes :)

zzzeek:
> Say I wanted to give this another go, is there some high level documentation on how the orm suff works in sqlalchemy? Currently it appears as a kind of black box that does magic stuff, so it's quite hard to make changes :)

the good news is that the internals are vastly better than they were in 2007 or so.  back then, when there was a unit of work bug, which there were obviously many, I had to spend hours just understanding the string of spaghetti at that time.  It was intentionally a string of spaghetti because I couldn't get my head around the whole thing at once and I wrote it "in the dark".   Finally after three years I understood what i wrote enough to do a refactor of the unit of work, so that while it's still a lot of code it is organized and consistent.   as it turns out it basically never breaks anymore at a fundamental level so you can see there are virtually no commits at all to it anymore: https://github.com/sqlalchemy/sqlalchemy/commits/master/lib/sqlalchemy/orm/unitofwork.py  

Aaannyway that was an old man rant that had no real purpose.  The most "high level" thing there is, which still should hold mostly true today, is the architecture chapter I wrote at http://aosabook.org/en/sqlalchemy.html.     Additionally, I did a really focused talk on what the Session is all about at: https://www.sqlalchemy.org/library.html#thesqlalchemysessionindepth .      high level stuff.    then there's basically just the code and the tests :)   I'll update if I think of other resources I'm not remembering.

CaselIT:
thanks. I'll start on that and try to understand it.

I should maybe open an issue with this stuff and close this. The patch on gerrit is very different from this so it's probably best to just reset and restart later :)

CaselIT:
Closing this.

sqla-tester:

Code-Review-1

I've just identified a valid use case for writing to a read only property, where we are using an attribute event to use the "write" to set other values:

https://github.com/sqlalchemy/sqlalchemy/wiki/SymmetricEncryptionServerSide

the above uses column_property() to decrypt a value, however we can set this value locally, and we actually want to use that "written" value, as well as set some other values.

I'm not sure we really should be trying to raise for mutations to viewonly attributes across the board.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1659

CaselIT:
Makes sense. That use case would not work even if the raising would occur when saving the data.

I guess that at the end of the day this feature does provide any particularly useful feature

sqla-tester:

> Patch Set 2: Code-Review-1
> 
> I've just identified a valid use case for writing to a read only property, where we are using an attribute event to use the "write" to set other values:
> 
> https://github.com/sqlalchemy/sqlalchemy/wiki/SymmetricEncryptionServerSide
> 
> the above uses column_property() to decrypt a value, however we can set this value locally, and we actually want to use that "written" value, as well as set some other values.
> 
> I'm not sure we really should be trying to raise for mutations to viewonly attributes across the board.

do we want to salvage only the is_readonly flag, without any one change, or should we abandon this?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1659

sqla-tester:

not useful at the moment

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1659

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1659 has been **abandoned**.  That means that at least for the moment I need to close this pull request. Sorry it didn't work out :(

Issue #4900, opened 2019-10-08
py.test: error: unrecognized arguments: --max-worker-restart=5


gordthompson:
After setting up a testing database for MSSQL according to README.unittests.rst and installing pytest 5.2.1, running `py.test --db mssql` resulted in

```
(venv) C:\Users\Gord\git\sqlalchemy>py.test --db mssql
ERROR: usage: py.test [options] [file_or_dir] [file_or_dir] [...]
py.test: error: unrecognized arguments: --max-worker-restart=5
  inifile: C:\Users\Gord\git\sqlalchemy\setup.cfg
  rootdir: C:\Users\Gord\git\sqlalchemy
```

In fact, I even get that if I just try doing `py.test --version`

```
(venv) C:\Users\Gord\git\sqlalchemy>py.test --version
This is pytest version 5.2.1, imported from c:\users\gord\git\sqlalchemy\venv\lib\site-packages\pytest.py
ERROR: usage: py.test [options] [file_or_dir] [file_or_dir] [...]
py.test: error: unrecognized arguments: --max-worker-restart=5
  inifile: C:\Users\Gord\git\sqlalchemy\setup.cfg
  rootdir: C:\Users\Gord\git\sqlalchemy
```

I needed to remove `--max-worker-restart=5` from setup.cfg:

https://github.com/sqlalchemy/sqlalchemy/blob/89da96a99c12e9deeb583504563224c7156bbcc8/setup.cfg#L8

I'm running Python 3.7.5rc1 x64 on Windows_7. pytest had issues with Python 3.7.4 on that platform (wouldn't run at all), but so far other tests on external dialects (sqlalchemy-access, mysql-pyodbc) have been working fine. However, those environments did not have `--max-worker-restart=5` in their setup.cfg.

Has something changed in pytest that may require a tweak to the default setup.cfg in sqlalchemy itself?

zzzeek:
pytest-xdist must also be installed.  if you use tox , this happens automatically.

gordthompson:
Okay, thanks. That got things moving, but perhaps tox is doing another bit of magic because when I try adding the `-n4` switch to `py.test` command I get

```
(venv) C:\Users\Gord\git\sqlalchemy>py.test -n4 -k TimeTest --db mssql
======================================================= test session starts =======================================================
[...]

PR #4989, opened 2019-11-18
Show how to install pytest in the testing README

nchammas:
### Description

This PR adds a brief note to the testing README showing how to install pytest. Without it, new contributors like myself will hit the `unrecognized arguments: --max-worker-restart=5` error reported in #4900.

### Checklist

This pull request is:

### DIFF
--- a/README.unittests.rst
+++ b/README.unittests.rst
@@ -46,7 +46,9 @@ connectivity, multi process support, as well as lots of skip / database
 selection rules.
 
 Running tests with py.test directly grants more immediate control over
-database options and test selection.
+database options and test selection. To run py.test directly, install it::
+
+    pip install pytest pytest-xdist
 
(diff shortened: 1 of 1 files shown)


zzzeek:
hi -

any chance you can figure out from the py.test side how py.test can gracefully ignore errors for optional plugins that aren't installed ?   this is more of a nuisance from their end.

nchammas:
Though `argparse` can ignore unknown options (and pytest uses it under the hood) there doesn't seem to be a runtime configuration that instructs pytest to ignore unrecognized options.

zzzeek:
OK, --max-worker-restart is not required here because there is no -n option.   will move it to tox.ini along with the -n option where pytest-xdist is definitely installed.  thanks for the PR anyway!

PR #5017, opened 2019-12-01
Treat ODBC connection string keywords as case-insensitive.

### LINKED ISSUE #4576: Erroneous "No driver name specified; this is expected by PyODBC" warning.
I'm getting the following warning a lot:

> ...\lib\site-packages\sqlalchemy\connectors\pyodbc.py:79: SAWarning: No driver name specified; this is expected by PyODBC when using DSN-less connections
>   "No driver name specified; "

It happens when I use this connection url:

`url = mssql+pyodbc://*****:*****@(localdb)\ProjectsV13/ETL?DRIVER=%7BODBC+DRIVER+17+FOR+SQL+SERVER%7D&initialCatalog=ETL&trusted_Connection=no&connectTimeout=30&encrypt=no&trustServerCertificate=no&applicationIntent=ReadWrite&multiSubnetFailover=no`

so, in the above i'm passing the connection attribute "DRIVER" and alchemy is asking me to pass "driver"
looking at the code in that sqlalchemy module on lines 76-82 we get:

```
driver = keys.pop("driver", self.pyodbc_driver_name)
                if driver is None:
                    util.warn(
                        "No driver name specified; "
                        "this is expected by PyODBC when using "
                        "DSN-less connections"
                )
                else:
                    connectors.append("DRIVER={%s}" % driver)
```

So according to that if we pass the connection attribute 'DRIVER' as 'driver' we should get a connection and no warning.  That url would like this:

`mssql+pyodbc://*****:*****@(localdb)\ProjectsV13/ETL?driver=ODBC+DRIVER+17+FOR+SQL+SERVER&initialCatalog=ETL&trusted_Connection=no&connectTimeout=30&encrypt=no&trustServerCertificate=no&applicationIntent=ReadWrite&multiSubnetFailover=no`

Note in the above i also had to drop the `%7B` ('{') and `%7D` ('}') from the driver name.

It seems to me that the warning should be silenced if we are passing 'DRIVER' explicitly.  Adding:

```and not 'DRIVER' in keys```

to the code would avoid this warning being unnecessarily triggered.

after suggested changes
```
driver = keys.pop("driver", self.pyodbc_driver_name)
                if driver is None and not 'DRIVER' in keys":
                    util.warn(
                        "No driver name specified; "
                        "this is expected by PyODBC when using "
                        "DSN-less connections"
                    )
```

### PR DESCRIPTION
**Pull request by gordthompson**
Fixes: #4576

### Description
Treat ODBC connection string keywords as case-insensitive (for mssql+pyodbc).

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/test/dialect/mssql/test_engine.py
+++ b/test/dialect/mssql/test_engine.py
@@ -47,7 +47,7 @@ class ParseConnectTest(fixtures.TestBase):
         )
         connection = dialect.create_connect_args(u)
         dsn_string = connection[0][0]
-        assert ";LANGUAGE=us_english" in dsn_string
+        assert ";language=us_english" in dsn_string
         assert ";foo=bar" in dsn_string
 
     def test_pyodbc_hostname(self):
@@ -85,6 +85,27 @@ class ParseConnectTest(fixtures.TestBase):
         eq_(
             [
                 [
+                    "Server=hostspec;Database=database;UI"
+                    "D=username;PWD=password"
+                ],
+                {},
+            ],
+            connection,
+        )
+
+    def test_pyodbc_uppercase_driver_keyword_no_warn(self):
+        dialect = pyodbc.dialect()
+        u = url.make_url("mssql://username:password@hostspec/database"
+                         "?DRIVER=ODBC Driver 17 for SQL Server")
+
+        # without assert_warnings (as in test_pyodbc_host_no_driver, above)
+        #   this test will fail here if a warning is issued
+        connection = dialect.create_connect_args(u)
+
+        eq_(
+            [
+                [
+                    "DRIVER={ODBC Driver 17 for SQL Server};"
                     "Server=hostspec;Database=database;UI"
                     "D=username;PWD=password"
                 ],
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/dialect/mssql/test_engine.py [hunk 1]
```diff
@@ -93,6 +93,27 @@ def go():
             connection,
         )
 
+    def test_pyodbc_uppercase_driver_keyword_no_warn(self):
+        dialect = pyodbc.dialect()
+        u = url.make_url("mssql://username:password@hostspec/database"
+                         "?DRIVER=ODBC Driver 17 for SQL Server")
```
zzzeek: I'm -1 on this because at https://docs.sqlalchemy.org/en/13/dialects/mssql.html#hostname-connections it is documented as "driver".  As discussed in #4576, this is not a straight pass-through, as we are adding brackets before sending to pyodbc, so the user can't treat this keyword like a pyodbc key.

zzzeek (commented):
I'm -1 on this as the keywords here are documented at https://docs.sqlalchemy.org/en/13/dialects/mssql.html#hostname-connections as being lower case.  if the user wants to send ODBC strings directly they should use odbc_connect.    There is no reason someone should be passing UPPERCASE words here unless they didn't read the documentation.

gordthompson:
Fair enough. Thanks for the clarification.

PR #5062, opened 2019-12-29
or_() and and_() should emit a warning no arguments

### LINKED ISSUE #5054: or_ and and_() should emit a warning / someday raise an error with no arguments
This is using `SQLAlchemy-1.3.12`.

I would expect `or_()` to produce kind of `false`, but it has truthy behaviour.
Note, `false` is neutral element for disjunction (and `true` for conjunction), so I would expect adding `false()` to it didn't change result, that's why I would expect following statements producing the same result:

    print(session.query(literal(42)).filter(or_()).first())
    # (42,)
    print(session.query(literal(42)).filter(or_(false())).first())
    # None

Note, that for `and_` everything works as expected:

    print(session.query(literal(42)).filter(and_()).first())
    # (42,)
    print(session.query(literal(42)).filter(and_(true())).first())
    # (42,)

There might some rationale behind current behaviour I am missing.

CaselIT:
Add a warning when `and_` and `or_` are called without any warning
### Description

Raises a deprecation warning when no clause is passed to `and_` and `or_`

Fix #5054

@zzzeek currently in the test suite an empty `and_` or `or_` is used in multiple places. I don't think adding `expect_deprecated` everywhere it the best options, since someday it will raise an exception.
what dot you think is the best options?

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**


CaselIT:
In the end most tests where specifically to test what happens in case of an empty `or_` and `and_` so I used `expect_deprecated` since they will be removed (updated with an expect exception) in the future

CaselIT:
there are some uses of `and_(*<some list>)` in the code:
Output of [`rg "(and_|or_)\(\W*\*" -U`](https://github.com/BurntSushi/ripgrep) is 
```
test\sql\test_insert_exec.py
105:                criterion = and_(
106:                    *[

doc\build\orm\composites.rst
128:            return sql.and_(*[a>b for a, b in

lib\sqlalchemy\orm\descriptor_props.py
457:            return sql.and_(*comparisons)

lib\sqlalchemy\orm\mapper.py
2458:            sql.and_(*[k == v for (k, v) in params]),
2866:        cond = sql.and_(*allconds)

lib\sqlalchemy\orm\interfaces.py
321:                return sql.and_(*[a>b for a, b in

lib\sqlalchemy\orm\persistence.py
881:        stmt = table.update(sql.and_(*clauses) if len(clauses) > 0 else None)
1195:        stmt = table.update(sql.and_(*clauses) if len(clauses) > 0 else None)
1306:        return table.delete(sql.and_(*clauses) if len(clauses) > 0 else None)

lib\sqlalchemy\orm\relationships.py
1399:                    return sql.and_(
1400:                        *[
1410:            criterion = sql.and_(
1411:                *[

lib\sqlalchemy\sql\base.py
1003:        return elements.and_(*l)

lib\sqlalchemy\sql\elements.py
2199:            * Calls to `and_(*args)` where `args` may be empty, should be

lib\sqlalchemy\sql\selectable.py
895:            return and_(*crit)

lib\sqlalchemy\dialects\oracle\base.py
887:            return sql.and_(*clauses)

lib\sqlalchemy\orm\dependency.py
1188:                sql.and_(
1189:                    *[
1214:                sql.and_(
1215:                    *[
```

I think these should probably prefixed with True.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 5ca2f27281977d74e390148c0fb8deaa0e0e4ad9 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 5ca2f27281977d74e390148c0fb8deaa0e0e4ad9: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1682

.gitignore
```diff
@@ -32,3 +32,7 @@ test/test_schema.db
 *test_schema.db
 .idea
 /Pipfile*
+/.pytest_cache/
```
sqla-tester: **mike bayer** (zzzeek) wrote:

pet peeve, can we have various .gitignores be in a separate commit ?

lib/sqlalchemy/orm/persistence.py
```diff
@@ -864,23 +864,21 @@ def _emit_update_statements(
     )
 
     def update_stmt():
-        clause = sql.and_()
+        clauses = []
```
sqla-tester: **mike bayer** (zzzeek) wrote:

well here's exactly why I have and_() and or_() working the way *I* liked them, I see it as a plain list.  I'd like to make this sql.and_(True), but then it coerces it to a single boolean and I can't append the clauses.   So we will have to do something else.

I *really like* having empty and_() and or_() behave as empty collections.  It remains intuitive for me.   So we will make it explicit by adding a special symbol that just makes the pattern *I* like work without anyone getting confused:

   # behaves exactly as it does now, boolean_empty
   # is a public symbol but nobody has to know :)
   clauses = sql.and_(boolean_empty)


This patch illustrates:

diff --git a/lib/sqlalchemy/orm/persistence.py b/lib/sqlalchemy/orm/persistence.py
index 79e7c89c3..c0945599e 100644
--- a/lib/sqlalchemy/orm/persistence.py
+++ b/lib/sqlalchemy/orm/persistence.py
@@ -862,7 +862,7 @@ def _emit_update_statements(
     )
 
     def update_stmt():
-        clause = sql.and_()
+        clause = sql.and_(sql.boolean_empty)
 
         for col in mapper._pks_by_table[table]:
             clause.clauses.append(
diff --git a/lib/sqlalchemy/sql/__init__.py b/lib/sqlalchemy/sql/__init__.py
index e0ad8d372..0974d3b90 100644
--- a/lib/sqlalchemy/sql/__init__.py
+++ b/lib/sqlalchemy/sql/__init__.py
@@ -13,6 +13,7 @@ from .expression import any_  # noqa
 from .expression import asc  # noqa
 from .expression import between  # noqa
 from .expression import bindparam  # noqa
+from .expression import boolean_empty  # noqa
 from .expression import case  # noqa
 from .expression import cast  # noqa
 from .expression import ClauseElement  # noqa
diff --git a/lib/sqlalchemy/sql/elements.py b/lib/sqlalchemy/sql/elements.py
index c1fc581ef..fa1b3bee0 100644
--- a/lib/sqlalchemy/sql/elements.py
+++ b/lib/sqlalchemy/sql/elements.py
@@ -2011,6 +2011,11 @@ class ClauseList(ClauseElement):
             return False
 
 
+boolean_empty = util.symbol(
+    "empty", doc="""construct an empty BooleanClauseList"""
+)
+
+
 class BooleanClauseList(ClauseList, ColumnElement):
     __visit_name__ = "clauselist"
 
@@ -2028,6 +2033,7 @@ class BooleanClauseList(ClauseList, ColumnElement):
         clauses = [
             _expression_literal_as_text(clause)
             for clause in util.coerce_generator_arg(clauses)
+            if clause is not boolean_empty
         ]
         for clause in clauses:
 
diff --git a/lib/sqlalchemy/sql/expression.py b/lib/sqlalchemy/sql/expression.py
index 2c55e860a..b914c8105 100644
--- a/lib/sqlalchemy/sql/expression.py
+++ b/lib/sqlalchemy/sql/expression.py
@@ -110,6 +110,7 @@ from .elements import _select_iterables  # noqa
 from .elements import _string_or_unprintable  # noqa
 from .elements import _truncated_label  # noqa
 from .elements import between  # noqa
+from .elements import boolean_empty  # noqa
 from .elements import BinaryExpression  # noqa
 from .elements import BindParameter  # noqa
 from .elements import BooleanClauseList  # noqa

lib/sqlalchemy/orm/persistence.py
```diff
@@ -1180,23 +1178,21 @@ def _emit_post_update_statements(
...
```
sqla-tester: **mike bayer** (zzzeek) wrote:

clauses = sql.and_(sql.boolean_empty)

lib/sqlalchemy/orm/persistence.py
```diff
@@ -1295,21 +1291,19 @@ def _emit_delete_statements(
     )
 
     def delete_stmt():
-        clause = sql.and_()
+        clauses = []
```
sqla-tester: **mike bayer** (zzzeek) wrote:

clauses = sql.and_(boolean_empty)

test/engine/test_execute.py
```diff
@@ -391,9 +391,10 @@ def test_not_an_executable(self):
         for obj in (
             Table("foo", MetaData(), Column("x", Integer)),
             Column("x", Integer),
-            tsa.and_(),
+            # a single clause in and_ is optimized away
+            tsa.and_(tsa.text("1=1"), tsa.text("1=1")),
+            tsa.and_(tsa.text("1=1"), tsa.text("1=1")).compile(),
             column("foo"),
```
sqla-tester: **mike bayer** (zzzeek) wrote:

tsa.and_(boolean_empty).compile()

test/sql/test_compiler.py
```diff
@@ -1396,12 +1397,24 @@ def test_nested_conjunctions_short_circuit(self):
                 and_(
                     or_(
                         or_(t.c.x == 12),
-                        and_(or_(), or_(and_(t.c.x == 8)), and_()),
```
sqla-tester: **mike bayer** (zzzeek) wrote:

this is likely testing that empty or_() and and_() do what's expected so we should add boolean_empty here.

sqla-tester (commented):

(12 comments)

Here's a quick summary of the changes:

1. propose a new symbol sqlalchemy.sql.boolean_empty, which you can put inside and_() / or_() to maintain the "empty list" behavior I find useful.    While using and_(True) / or_(False) achieves almost the same thing, for the case where I'm appending to an empty clauselist, this symbol will allow me to keep that pattern, as and_(True) / or_(False) don't produce a ClauseList.

2. all of the existing tests can move to use the "boolean_empty" symbol so that they don't have to be rewritten.

3. new tests to be added to test/sql/test_deprecations.py that test the public-facing deprecation of completely empty and_() / or_().

4. we generally don't put expect_deprecated() in test suites outside of test/XYZ/test_deprecations.py.     There's some exceptions but in general anytime you are writing an expect_deprecated() it should go into test_deprecations.py somewhere.   That way the work of converting all the test suites to the "future" version is done up front and when the deprecated behavior is removed fully, it's a clean deletion of code.

* /COMMIT_MSG (line 15): "Fixes: #5054"
* /COMMIT_MSG (line 17): I'd like to keep this capability, see my later comment with the proposal of a new symbol to explicitly allow the pattern i like to use so that we don't have to wreck the codebase
* lib/sqlalchemy/sql/elements.py (line 2138): clauses = [
            _expression_literal_as_text(clause)
            for clause in util.coerce_generator_arg(clauses)
            if clause is not boolean_empty
        ]
* lib/sqlalchemy/sql/elements.py (line 2141): move this above the list comprehension so that the length check is before the conversion of elements, also don't use "len()".   e.g.:

if not clauses:
   util.warn_deprecated(...)

test/sql/test_compiler.py
```diff
@@ -1396,12 +1397,24 @@ def test_nested_conjunctions_short_circuit(self):
...
+                        and_(or_(and_(t.c.x == 8))),
                     )
                 )
             ),
             "SELECT t.x FROM t WHERE t.x = :x_1 OR t.x = :x_2",
         )
+        with expect_deprecated():
```
sqla-tester: **mike bayer** (zzzeek) wrote:

all deprecation tests need to be in a test_deprecated.py suite so that they can be easily removed, so don't put expect_deprecated() stuff here.

test/sql/test_delete.py
```diff
@@ -77,12 +78,14 @@ def test_delete(self):
     def test_where_empty(self):
         table1 = self.tables.mytable
 
-        self.assert_compile(
-            table1.delete().where(and_()), "DELETE FROM mytable"
-        )
-        self.assert_compile(
-            table1.delete().where(or_()), "DELETE FROM mytable"
-        )
+        with expect_deprecated():
```
sqla-tester: **mike bayer** (zzzeek) wrote:

use boolean_empty here, but also add new tests for the actual "empty" version in test/sql/test_deprecations.py

.gitignore
CaselIT: ok l'll move it to a new pr

CaselIT:
The bot that copies the comment to and from gerrit seems not to like the formatting very much.

What you propose make sense, but I think a kw-argument would work better to preserve the current use case. Something like `sa.and_(allow_empty=True)` without any new symbol.

What do you think?

sqla-tester (commented):

(1 comment)

* lib/sqlalchemy/sql/elements.py (line 2141): I usually like to be explicit when testing for an empty list, but I've seen that sqlalchemy does not seem to use the `len` a lot

sqla-tester (commented):

(1 comment)

* lib/sqlalchemy/sql/elements.py (line 2141): it shows up in callcounts and slows the code down.

	>>> a = []
	>>> b = [1]
	>>> import timeit
	>>> timeit.timeit("3 if a else 2", "from __main__ import a", number=10000000)
	0.16772469799616374
	>>> timeit.timeit("3 if len(a) else 2", "from __main__ import a", number=10000000)
	0.4166480970015982
	>>> timeit.timeit("3 if b else 2", "from __main__ import b", number=10000000)
	0.18806552200112492
	>>> timeit.timeit("3 if len(b) else 2", "from __main__ import b", number=10000000)
	0.4235368380032014

sqla-tester (commented):

(1 comment)

* lib/sqlalchemy/sql/elements.py (line 2141): yes I suspected it would be a bit slower, however its ~16ns vs ~41ns, so I usually prefer being explicit :)

also I often work with pandas or numpy object that do not like the `if obj` or `if not obj`

zzzeek:
> The bot that copies the comment to and from gerrit seems not to like the formatting very much.

since you're writing a lot of code what do you think of working straight in gerrit?   I'm not sure if I gave you access already but the PR sync thing is mostly so new / occasional contributors can come in, but I find gerrits a lot easier to navigate.

> 
> What you propose make sense, but I think a kw-argument would work better to preserve the current use case. Something like `sa.and_(allow_empty=True)` without any new symbol.

Yeah I would want that also but then we need to add ugly ``**kw`` to and_() and or_().  I don't feel comfortable with "allow_empty" because I still want the option to change my mind on this API, and I also don't want it to confuse people. basically it looks ugly and for 2.0 I'm trying to have a way more minimal API.   the symbol is a way to have the old behavior without committing too heavily to it.   I mostly want a new kind of construct that's like an "optional and_()" collection thing.   That's basically what you had there with the list and such :)   


> 
> What do you think?

CaselIT:
`optional_and()` why not making it a thing? 
I actually think it's the cleanest of the approaches. "If there are value place them in and|or". 

> working straight in gerrit

will try, I'm not super familiar with it.

zzzeek:
> `optional_and()` why not making it a thing?

this is really more like, I want a list of things that can convert itself into an "AND" if it's not empty.

Basically, BooleanClauseList([], operator=operators.and_), or something like that

again "optional_and()", sure, but it's seems like a weird API for which people won't know what it's for and I really don't want to do anything very conspicuious right now.   really the tests and ORM should just change at some point to something else but I want to get the public-facing behavior in now without having to rethink all that yet.



> I actually think it's the cleanest of the approaches. "If there are value place them in and|or".
> 
> > working straight in gerrit
> 
> will try, I'm not super familiar with it.

CaselIT:
> this is really more like, I want a list of things that can convert itself into an "AND" if it's not empty.

As a private api to use in sqlalchemy, or as the exposed behaviour?

Maybe it could just warn/raise on compile. But it would not be as useful for the user, since the construct that raised the warning/error could be hard to track down

zzzeek:
> > `optional_and()` why not making it a thing?
> 

OK, a decision has been made, we will go with the approach of "optional_and()" but we will call it `filter_criteria()`.   it is an empty collection upon which SQL boolean expressions can be appended.  if nothing is appended, it renders nothing, if things are appended, they are joined via AND.

I can see if we build this out, I can perhaps also add an "immutable" appender to it that I can then use for all the stuff I'm doing with Core / ORM SELECT queries.   

Looking at what persistence.py is doing it is actually bypassing all the validation that ClauseList does (again, improving performance, always having to improve performance).   this is also interesting and I have a bunch of thoughts on how to make this more of a first class thing.   

do you want me to work out some of that initial API stuff?  in the ORM part of this it would look like:

    clause = sql.filter_criteria()
    clause.append_raw(foo == 'bar')   # doesnt' validate or self_group or anything

in places like select() and Query(), perhaps, it can work like this:


    class Query():
        criterion = sql.filter_criteria(immutable=True)

        def filter(...):
            self.criterion = self.criterion.build_raw(foo == 'bar')

I've already noticed that the current implementation of Query.filter() and select.where() is doing the validation on incoming objects twice, wasting performance.

CaselIT:
Llooks good. I think you could implement the api for it, since at the moments it's a bit fuzzy.

I guess the only thing to salvage from this is the warning and the documentation update.

what would happen to the tests that use an empty `and_` or `or_` ?

zzzeek:
> Llooks good. I think you could implement the api for it, since at the moments it's a bit fuzzy.
> 
> I guess the only thing to salvage from this is the warning and the documentation update.
> 
> what would happen to the tests that use an empty `and_` or `or_` ?

still thinking here, as there is a lot changing with how SELECT statements are built up in any case and I am now leaning towards a special constructor for the internal stuff and a lot of other changes, so I'll just finish the gerrit here.  those tests are just going to change to whatever.

sqla-tester (commented):

(2 comments)

Looks good

* lib/sqlalchemy/sql/elements.py (line 2180): Shouldn't _construct call this one?

test/engine/test_execute.py
```diff
@@ -391,9 +391,10 @@ def test_not_an_executable(self):
         for obj in (
             Table("foo", MetaData(), Column("x", Integer)),
             Column("x", Integer),
-            tsa.and_(),
+            # a single clause in and_ is optimized away
```
sqla-tester: **Federico Caselli** (CaselIT) wrote:

Won't this short circuit the and_ and test something else?

sqla-tester (commented):

(2 comments)

* lib/sqlalchemy/sql/elements.py (line 2180): I want to skip all the unnecessary logic in _construct() as it is redundant / unnecessary for how this is used.

i also want to try to reduce the complexity of _construct() anyway.

test/engine/test_execute.py
sqla-tester: **mike bayer** (zzzeek) wrote:

ah you're correct in this case, because and_(True) produces an AsBoolean object and I missed that.   But I think what's being tested here is not really specific to ClauseList, it's just testing that a variety of random ClauseElement things are not going to try to invoke a statement.  this is kind of an old test, it's not that critical.

sqla-tester (commented):

(1 comment)

* lib/sqlalchemy/sql/elements.py (line 2180): I meant the opposite, _construct should call this after its checks

sqla-tester (commented):

(1 comment)

* lib/sqlalchemy/sql/elements.py (line 2180): oh at the bottom, i see.   OK.   im just not sure what im doing with this construct_raw thing.   i want to reduce some complexity in _construct so let me see where it goes.

sqla-tester (commented):

(1 comment)

* lib/sqlalchemy/sql/elements.py (line 2180): Done

sqla-tester:

(1 comment)

test/engine/test_execute.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

Done

sqla-tester (commented):

(1 comment)

* lib/sqlalchemy/sql/elements.py (line 2169): Looks nice!
Just a nitpick: `not lcc` is always true here

sqla-tester (commented):

(1 comment)

* lib/sqlalchemy/sql/elements.py (line 2180): uh oh, does this mean you updated the PR ?   once we start adding gerrit changesets it goes out of sync w the PR, sorry ...I have it calling down to construct_raw now, thanks for the suggestion.

sqla-tester (commented):

(1 comment)

* lib/sqlalchemy/sql/elements.py (line 2169): ooh yup

sqla-tester (commented):

(1 comment)

* lib/sqlalchemy/sql/elements.py (line 2180): no I was trying to resolve it. I thought that the done did that...

sqla-tester:

Code-Review+2

OK so, nit notwithstanding, I want to put this in.  give it a code-review +1 when you agree (Im assuming you don't have a +2 button).

sqla-tester:

Code-Review+1

> Patch Set 4: Code-Review+2
> 
> OK so, nit notwithstanding, I want to put this in.  give it a code-review +1 when you agree (Im assuming you don't have a +2 button).

Nice, thanks for polishing it! I do have a +1 only

sqla-tester:

Workflow+1

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1682 has been **merged**. Congratulations! :)

sqla-tester:

I think I want to see if I can get my gerrit->github tool to include the gerrit link to a comment when the comment originated from gerrit, because I get the github emails in my inbox but not the gerrits because there are too many of the latter.

Issue #4379, opened 2018-11-27
Metadata reflection slow with large schemas


flewellyn:
Using SQLAlchemy 1.2.14, I attempted to reflect a schema from a customer which contained several thousand tables, and noticed this was very slow.  Using echo="debug" as an option on the engine, I noticed that the SQLAlchem reflection code is iteratively querying the database for table metadata, instead of issuing one query.

I believe this is sub-optimal, and performance would be greatly enhanced if the SQLAlchemy reflection code issued one query to the backend instead of iterating over the table list.  

Example times: MS SQL Server database with 3,300 tables: 15 minutes.
PostgreSQL database with 694 tables: 4 minutes.
PostgreSQL database with 18,000+ tables: 45 minutes.

zzzeek:
also note that the MySQL database relies upon "SHOW CREATE TABLE", which cannot be batched amongst many tables.  So a performance enhancement here would also need to be conditional based on whether or not a dialect can provide it.   Being able to pass all the desired table names to the dialect from the inspector would also require significant changes to the entire architecture between dialect and inspector, and would need to work exactly as it does now for dialects that can't support such a performance feature including third party.

zzzeek:
note that part of my skepticism over the "one query" claim is that table reflection loads all kinds of objects, not just columns, but also indexes, constraints, for Postgresql it has to load enums and domains, all of these things have their own individual queries that themselves often have hard-to-optimize elements such as LEFT OUTER JOINs and things like that.  If all of this were to be put into *one* query, it is very possible that "one" query would be very un-optimized and slow, not to mention extremely difficult to maintain.

My mistake here is that the issue here states two separate things, and I failed to separate those things immediately. One is, "it's too slow to reflect 18000 tables".  The other is, "please write everything into one giant query".   

More realistically, an optimization here would be first off that if we do want to reflect N tables, we at least do batching, that is, methods like dialect.get_columns() would gain the ability to do their work for a list of table/schemaname combinations.    The reflection of N tables would involve M calls to get_columns(), get_indexes(), get_foreign_keys(), etc.  where M is a portion of N.  We would use batching, and not one huge result, so that results can be streamed out before the full set of tablenames is available, which also gives the chance to preserve memory within the process.   

[...]

PR #5064, opened 2019-12-29
Support inspection of computed column

### LINKED ISSUE #4051: sqlalchemy.engine.reflection.Inspector.get_columns : documentation error for SQLAlchemy 1.2.0b1
**Migrated issue, originally created by Rudolf Cardinal ([@RudolfCardinal](https://github.com/RudolfCardinal))**

In SQLAlchemy 1.2.0b1, it's possible for `sqlalchemy.engine.reflection.Inspector.get_columns` to return the following dict as one of its results: 

    {
        'type': VARCHAR(length=255),
        'nullable': False,
        'default': None,
        'name': 'device_id',
        'comment': 'Source tablet device ID'
    }

The docs at http://docs.sqlalchemy.org/en/latest/core/reflection.html#sqlalchemy.engine.reflection.Inspector.get_columns seem to be out of date for this in that (a) `attrs` is documented as part of the returned dictionary, but is absent in this real example; (b) `comment` is new and undocumented.

### LINKED ISSUE #5063: Support inspection of computed column
The current version of slqalchemy has no support for inspecting computed column, that were recently added.

This is problematic since in some backend ( oracle and postgresql among them ) a computed column is inspected as a server_default value.

While adding support for computed columns in alembic at https://github.com/sqlalchemy/alembic/pull/631 some workaround where needed depending on the backend used

CaselIT:
This pr normalizes the inspect behavior of sqlalchemy on the various backends in case of computed columns and adds support for introspecting computed column, by introducing a new key in the dict returned by `get_columns` with the computed argument

### Description

fix #5063. See also sqlalchemy/alembic#631
fix #4051
### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**


lib/sqlalchemy/dialects/oracle/base.py
```diff
@@ -1658,8 +1658,8 @@ def get_columns(self, connection, table_name, schema=None, **kw):
         text = """
             SELECT col.column_name, col.data_type, col.%(char_length_col)s,
               col.data_precision, col.data_scale, col.nullable,
-              col.data_default, com.comments\
-            FROM all_tab_columns%(dblink)s col
+              col.data_default, com.comments, col.virtual_column\
+            FROM all_tab_cols%(dblink)s col
```
CaselIT: I'm not sure if this has unintended consequences. 

From the documentation it should be ok:

> This view differs from "ALL_TAB_COLUMNS" in that hidden columns are not filtered out.

https://docs.oracle.com/cd/B19306_01/server.102/b14237/statviews_2093.htm#I1020276

lib/sqlalchemy/testing/suite/test_reflection.py
```diff
@@ -1148,6 +1148,38 @@ def test_autoincrement_col(self):
             assert id_.get("autoincrement", True)
 
 
+class ComputedReflectionTest(fixtures.TablesTest):
```
CaselIT: I've added a new class here, instead of modifying `ComponentReflectionTest` because a lot of tests needed to be changed if a new table was added. I think it is ok, since there will be other tests on the computed detection once it is complete

zzzeek:
so it seems like it would be wise to just take this a step further, when "computed" is flagged, add an element to the column dictionary called "computed" with the expression,then hook it up in reflection.py to send that expression into a Computed object.  then computed reflection is done for PG + Oracle, the test suite will be there, then we can add in MySQL  + MSSQL later based on whatever more complicatred approach we need for them.

CaselIT:
I thought it could be a different pr.
This was meant just to normalize the behaviour of the different backends.

I guess I could also add the logic to manage computed for pg and oracle.

I've some questions (that were the reason for the split pr approach):
What about the persisted keyword? Maybe we should return it also? So the return value could be:
```py
{
  ...,
  "computed": {"sqltext": "value here", "persisted": True/False/None}
}
```

Maybe persisted could be optional if a db does not return information about it?
Maybe we could also return an instance of `Computed` directlty?

What do you think?

zzzeek:
sure, I like that idea, whatever works to start with.

CaselIT:
> sure, I like that idea, whatever works to start with.

sorry which one?

CaselIT:
I'm trying to work a bit more on this.
I was reading the documentation, but it seems that it is not up to date https://docs.sqlalchemy.org/en/13/core/reflection.html?highlight=reflection#sqlalchemy.engine.reflection.Inspector.get_columns 
The `attrs` keywords seems not to be used anywhere, while `autoincrement` and `comment` are not documented.

Am I missing something or should the documentation be updated?

zzzeek:
> > sure, I like that idea, whatever works to start with.
> 
> sorry which one?

the part where the "persistent" flag is returned.  Not the part where a Computed object is inside the raw inspection dictionary, the inspection dictionary keeps things raw except for datatypes.

zzzeek:
> I'm trying to work a bit more on this.
> I was reading the documentation, but it seems that it is not up to date https://docs.sqlalchemy.org/en/13/core/reflection.html?highlight=reflection#sqlalchemy.engine.reflection.Inspector.get_columns
> The `attrs` keywords seems not to be used anywhere, 

extremely surprisingly, nobody is using attrs, I'm not really sure where to go with this because if you look at https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/dialects/mysql/reflection.py#L358 there are a crap-ton of special MySQL keywords that apply to columns that we are parsing, that I thought were supposed to be in there and...we're just losing them?   this is surprising to me.     we should take it out of the docs for now but i'd like to figure out what's up with that.

> while `autoincrement` and `comment` are not documented.

comment and autoincrement were only added recently and I have not confirmed they are always available but it looks like at least for autoincrement it's now required.

> 
> Am I missing something or should the documentation be updated?

likely the latter

CaselIT:
> we're just losing them

I think it uses them to instantiate the type here 
https://github.com/sqlalchemy/sqlalchemy/blob/2734439fff953a7bb8aecdedb5f851441b5122e9/lib/sqlalchemy/dialects/mysql/reflection.py#L206-L224 

>likely the latter

I can try to improve them a bit

CaselIT:
I've started implementing the persistence reflection in https://github.com/CaselIT/sqlalchemy/tree/reflection-of-computed-columns, currently in pg, oracle and mysql

I can add a new pr for that branch or merge in this one, if you prefer to have a single pr for this.

zzzeek:
> I've started implementing the computed reflection in https://github.com/CaselIT/sqlalchemy/tree/reflection-of-computed-columns, currently in pg, oracle and mysql
> 
> I can add a new pr for that branch or merge in this one, if you prefer to have a single pr for this.

a single PR to get computed reflection done is best if possible.

CaselIT:
I've started adding support for the backends.
Currently I've implemented
- postgres
- oracle
- mysql

I have to check for mssql and I've seen that firebird has a new external dialect, here https://github.com/pauldex/sqlalchemy-firebird, so I guess no change should be added here

README.unittests.rst
```diff
@@ -161,7 +161,7 @@ and MySQL they are schemas.   The requirement applies to all backends
 except SQLite and Firebird.  The names are::
 
     test_schema
-    test_schema_2 (only used on PostgreSQL)
+    test_schema_2 (only used on PostgreSQL and mssql)
```
CaselIT: `test/dialect/mssql/test_reflection.py` uses `test_schema_2` also in mssql

lib/sqlalchemy/testing/suite/test_reflection.py
```diff
@@ -1148,6 +1148,72 @@ def test_autoincrement_col(self):
...
+        cols = insp.get_columns("computed_column_table")
+        for col in cols:
+            if col["name"] == "with_default":
+                is_true("42" in col["default"])
+            elif not col["autoincrement"]:
+                is_(col["default"], None)
+
+    @testing.requires.computed_columns
+    @testing.requires.table_reflection
+    def test_get_column_returns_computed(self):
```
CaselIT: this if filled with `if`, `esif` for each dialect. Maybe we should just move the dialect dependent part of this test in the test_reflection of each dialect

CaselIT:
I've added support do mssql

CaselIT:
This should be done. I think you can check it @zzzeek 

I've added a new fixture for the reflection of the computed columns, since there are some differences but the code is the same. It is in `lib/sqlalchemy/testing/reflection_fixture.py`

For firebird i've open an issue in here https://github.com/pauldex/sqlalchemy-firebird/issues/4 so for not I've not added support for it

lib/sqlalchemy/dialects/mssql/base.py
```diff
@@ -2590,12 +2591,19 @@ def get_columns(self, connection, tablename, dbname, owner, schema, **kw):
...
             )
+            table_fullname = "%s.%s" % (owner, tablename)
         else:
             whereclause = columns.c.table_name == tablename
+            table_fullname = tablename
         s = sql.select(
             [columns], whereclause, order_by=[columns.c.ordinal_position]
         )
 
+        cc = ischema.computed_columns
```
zzzeek: do we need to be concerned about this table not being present in some versions of SQL server, like older versions, or azure / cloud versions?

lib/sqlalchemy/dialects/mssql/base.py
```diff
@@ -2590,12 +2591,19 @@ def get_columns(self, connection, tablename, dbname, owner, schema, **kw):
...
+        computedSql = sql.select([cc]).where(
```
zzzeek: camelcase here.    if you install the pre-commit hooks, I would *assume* that flake8 would reject these....but maybe im wrong

lib/sqlalchemy/dialects/oracle/base.py
```diff
@@ -1684,6 +1684,7 @@ def get_columns(self, connection, table_name, schema=None, **kw):
             nullable = row[5] == "Y"
             default = row[6]
             comment = row[7]
+            genetared = row[8]
```
zzzeek: typo, "generated"

lib/sqlalchemy/dialects/oracle/base.py
zzzeek: that might affect things, sure.      By "filtered out" they mean the previous view wont return as many rows, right?   it looks like this new view has a "HIDDEN_COLUMN" column which we could filter on by adding "hidden_column='NO'", which per this documentation would give us the same result rows.

is it the case that if we create a table in Oracle with a "Computed" on it, the *whole column* won't come back in reflection right now?

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2827,7 +2827,14 @@ def get_columns(self, connection, table_name, schema=None, **kw):
         table_oid = self.get_table_oid(
             connection, table_name, schema, info_cache=kw.get("info_cache")
         )
-        SQL_COLS = """
+
+        generated = (
+            "a.attgenerated as generated"
+            if self.server_version_info >= (12,)
+            else "'' as generated"
```
zzzeek: probably have this "NULL AS generated"

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -3010,6 +3023,15 @@ def _handle_array_type(attype):
                 "Did not recognize type '%s' of column '%s'" % (attype, name)
             )
             coltype = sqltypes.NULLTYPE
+
+        # If a zero byte (''), then not a generated column.
+        #  Otherwise, s = stored. (Other values might be added in the future.)
+        if generated != "":
```
zzzeek: OK it comes back as blank....OK None / blank would both test here for just "if generated"

lib/sqlalchemy/testing/reflection_fixture.py
```diff
@@ -0,0 +1,201 @@
...
+                "computed_stored",
+                self.to_sqltext("normal", "*", "42"),
+                True,
+            )
+
+    def check_table_column(self, table, name, text, persisted):
+        is_true(name in table.columns)
+        col = table.columns[name]
+        is_not_(col.computed, None)
+        is_true(isinstance(col.computed, Computed))
```
zzzeek: don't do this now but we should really have an "is_instance()" asserter in testing

lib/sqlalchemy/testing/reflection_fixture.py
```diff
@@ -0,0 +1,201 @@
...
+from .fixtures import TablesTest
+from .. import inspect
+from .. import Integer
+from ..schema import Column
+from ..schema import Computed
+from ..schema import MetaData
+from ..schema import Table
+
+
+class ComputedReflectionFixtureTest(TablesTest):
```
zzzeek: OK this is very nice but I want to see if we can get this to work out with the existing multi-dialect testing system, because otherwise we're inventing a new system, and also I would want this suite to automatically come online for any third party dialect that opts into it using the standard means.

so this would mean:

1. it's in lib/sqlalchemy/testing/suite/test_reflection.py as another "componentreflection" suite, it can be its own class since this is pretty different.

2. The flags that are set, like support_stored, support_virtual, etc., this kind of thing usually comes either from the dialect itself, e.g. there are flags on dialects like "supports_returning", or they come from the exclusion rules in lib/sqlalchemy/testing/requirements.py  test/requirements.py.   The exclusion rules are where third party dialects can also set up a matrix of what capabilities and expectations those dialects have; while you've likely seen the exclusion rules used as decorators on tests, they can also be used as boolean flags inside of tests.     so we can add exclusion rules for now like computed_columns_stored, computed_columns_virtual, im not sure if I like that but it will get us closer.

3. for a quick way to deal with the to_sqltext() thing for now, see below.

lib/sqlalchemy/testing/reflection_fixture.py
```diff
@@ -0,0 +1,201 @@
...
+
+    __backend__ = True
+    __requires__ = ("computed_columns", "table_reflection")
+
+    return_persisted = None
+    default_persisted = None
+    support_stored = None
+    support_virtual = None
+
+    def to_sqltext(self, column, op, value):
```
zzzeek: I think for now we can likely just use a regex that filters out brackets and parenthesis, and just set this up once for all the current dialects, right?  that should be good enough.

lib/sqlalchemy/testing/reflection_fixture.py
```diff
@@ -0,0 +1,201 @@
...
+            Column("with_default", Integer, server_default="42"),
+        )
+
+        Table(
+            "computed_column_table",
+            metadata,
+            Column("id", Integer, primary_key=True),
+            Column("normal", Integer),
+            Column("computed_no_flag", Integer, Computed("normal + 42")),
+            *cls._col("computed_virtual", "normal + 2", False),
```
zzzeek: python 2 won't accept the multiple *, so I think the approach should be, "if testing.requirements.computed_columns_stored.enabled(): table.append_column(Column(...))"

lib/sqlalchemy/testing/reflection_fixture.py
```diff
@@ -0,0 +1,201 @@
...
+        is_not_(col.computed, None)
+        is_true(isinstance(col.computed, Computed))
+
+        eq_(str(col.computed.sqltext), text)
+        if self.return_persisted:
+            eq_(col.computed.persisted, persisted)
+        else:
+            is_(col.computed.persisted, None)
+
+    def test_table_reflection(self):
```
zzzeek: technically I think I'm trying to have all the reflection tests in suite/ only use the inspector, and then testing that the inspector can work these things into a Table with Computed objects is in test/engine/test_reflection.py.

zzzeek (changes requested):
great effort!    I have a bunch of things I'd like to change, if possible, let me know if you want me to do any of it for expediency.

lib/sqlalchemy/dialects/postgresql/base.py
CaselIT: it returns an empty string when a column is not generated. So I made it generate an empty string for consistency. It can be changed, changing also the check below

lib/sqlalchemy/testing/reflection_fixture.py
CaselIT: I have not checked to see if it exists.

lib/sqlalchemy/dialects/mssql/base.py
CaselIT: it seems so regarding azure-cloud. I'm not sure regarding older versions.
https://docs.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-computed-columns-transact-sql?view=sql-server-ver15

lib/sqlalchemy/dialects/oracle/base.py
CaselIT: no, by filtered out they mean columns. the column `virtual_column` is not present in `all_tab_columns`. The row with the computed column appears, since oracle returns the computed expression as the default, but there is not the column that indicates if it computed or not

https://docs.oracle.com/cd/B19306_01/server.102/b14237/statviews_2093.htm
> This view differs from "ALL_TAB_COLUMNS" in that hidden columns are not filtered out.

From the documentation the change should have no side effect

lib/sqlalchemy/dialects/postgresql/base.py
CaselIT: ok i'll change it

lib/sqlalchemy/testing/reflection_fixture.py
zzzeek: it doesn't.   this is more a note for me to do this when i ever rememebr

lib/sqlalchemy/dialects/postgresql/base.py
zzzeek: yeah I like NULL when the DB doesn't support it at all.  this is all nitpicking I know.

lib/sqlalchemy/testing/reflection_fixture.py
CaselIT: there is `is_instance_of`

lib/sqlalchemy/testing/reflection_fixture.py
CaselIT: I'll try to refactor it a bit. The problem was that the sqltext is all over the place for the various backends. I'll try to see if a regexp can fix it

lib/sqlalchemy/testing/reflection_fixture.py
CaselIT: I placed it here since the I needed stored/virtual support. I'll try to refactor them in the requirements

CaselIT:
So regarding the `reflection_fixture`
Should the flag come from the dialect or be placed in requirements?

lib/sqlalchemy/dialects/mssql/base.py
CaselIT: I've found this comment, from 2009 regarding `computed_columns`, so it should be fairly old

https://stackoverflow.com/questions/1484147/get-list-of-computed-columns-in-database-table-sql-server

CaselIT: here another mssql 2005+ 
https://stackoverflow.com/questions/5047099/sql-server-computed-column-defintions-retrievable-from-database

lib/sqlalchemy/testing/reflection_fixture.py
zzzeek: yeah, im not sure that is going to look much better but I want to try to keep the number of approaches in the test suite to a minimum, even though there are many at this point within the "suite" part of things, that's what we export to third party libs so it should try to be very consistent.

lib/sqlalchemy/testing/reflection_fixture.py
zzzeek: > So regarding the `reflection_fixture`
> Should the flag come from the dialect or be placed in requirements?

so it's nice when the flags come from the dialect, but usually the dialect has such flags only if they are needed elsewhere, like if the compiler needed to know about it.  if we're talking about two flags, then I think requirements is OK for now

lib/sqlalchemy/dialects/mssql/base.py
zzzeek: ok then let's keep it

lib/sqlalchemy/dialects/oracle/base.py
zzzeek: OK just as long as there aren't rows coming back with "hidden=yes" that were rows we weren't getting before, it should be fine.

zzzeek: per https://stackoverflow.com/questions/26790333/difference-between-sys-all-tab-columns-and-sys-all-tab-cols-in-oracle-12c the number of rows changes  "ALL_TAB_COLUMNS lists only user-defined columns whereas ALL_TAB_COLS also seems to contain some system columns in each user table that are not normally visible."   this will be backwards incompatible if we don't adjust by adding "hidden='NO'".  it seems like "hidden" and "invisible" are the two terms we need to deal with.

lib/sqlalchemy/dialects/mssql/base.py
CaselIT: We could wrap it in a try catch just to be sure

lib/sqlalchemy/dialects/oracle/base.py
CaselIT: I missed that part.. it does not seem to be listed in the documentation..
Should we add a where clause or join the two tables so we are sure that the returned rows arr the dame?

zzzeek: I think if we add "where hidden='NO'" that should be enough.   the test suite will let us know.

lib/sqlalchemy/dialects/mssql/base.py
zzzeek: those try/catches are themselves a lot of work to get right and test and if this has been there since SQL server 2005 I have a feeling we dont need to worry about it

zzzeek: if I go to MS's docs and look at "sys.tables", which is definitely everywhere, it lists the same four green checkboxes as sys.computed_columns, so this is good as it is.

lib/sqlalchemy/dialects/oracle/base.py
CaselIT: Will do. Sadly I cannot test against oracle

zzzeek: no worries it'll happen on this end

CaselIT:
I've updated the fixture to use the requirements. Please check again

lib/sqlalchemy/testing/reflection_fixture.py
```diff
@@ -0,0 +1,92 @@
...
+from . import config
+from .fixtures import TablesTest
+from .. import Integer
+from .. import testing
+from ..schema import Column
+from ..schema import Computed
+from ..schema import Table
+
+
+class ComputedReflectionFixtureTest(TablesTest):
```
CaselIT: This is here just to avoid code duplication in `suite/test_reflection.py` and `engine/test_reflection.py`

maybe it can be removed or moved to fixture.py

lib/sqlalchemy/testing/requirements.py
```diff
@@ -1096,4 +1096,27 @@ def _has_cextensions(self):
...
+
+    @property
+    def computed_columns_stored(self):
+        "Supports computed columns with `persisted=True`"
+        return exclusions.closed()
+
+    @property
+    def computed_columns_virtual(self):
+        "Supports computed columns with `persisted=False`"
+        return exclusions.closed()
```
CaselIT: this must be two separate since some backend, like firebird, may not support neither of them

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 4b27493c1ae1cbabb57b471421137cb51544294a of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 4b27493c1ae1cbabb57b471421137cb51544294a: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1679

lib/sqlalchemy/testing/reflection_fixture.py
zzzeek: ok not a bad idea

CaselIT: will move it once the test finish, so I'll also try to fix any issue that pops up

CaselIT: will move it once the test finish, so I'll also try to fix any issue that pops up

CaselIT:
seems mssql in some configurations does not like other executions while the an execution is still in progress.

Do you think it's best to use another connection, or add the computed information after the first execution is done?

zzzeek:
> seems mssql in some configurations does not like other executions while an execution is still in progress.
> 
> Do you think it's best to use another connection, or add the computed information after the first execution is done?

the query for the computed column info should either be joined into the main columns query using LEFT OUTER JOIN, or likely more performant, run the computed columns query against the whole table up front, then for each row, apply the information to a dictionary keyed on column name, then as you iterate through the main column info, add what's present in that dictionary.

also, I haven't checked but none of the dialects should ever be doing a query-per-column, we are already in a big hole as far as reflection performance and #4379 if we ever get around to it will have us using much fewer queries across many tables at once.

CaselIT:
I had not realized it could take that long.
I'll rework it with an outer join.

zzzeek:
ping me  / request a review when this is at the next step.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision 017ff1ddc17d2e9385d31116247c4a783de6a248 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 017ff1ddc17d2e9385d31116247c4a783de6a248 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1679

lib/sqlalchemy/dialects/mssql/base.py
```diff
@@ -2585,15 +2586,37 @@ def get_view_definition(
...
         if owner:
             whereclause = sql.and_(
                 columns.c.table_name == tablename,
                 columns.c.table_schema == owner,
             )
+            table_fullname = "%s.%s" % (owner, tablename)
+            concat = func.concat(
+                columns.c.table_schema, ".", columns.c.table_name
+            )
+            join_on = computed_cols.c.object_id == func.object_id(concat)
```
CaselIT: I'm not sure if this is the best solution, but I have not found the object id of the table in the information_schema.column table

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision ba00fc321ce468f8885aad23b3dd33c789e50fbe of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset ba00fc321ce468f8885aad23b3dd33c789e50fbe added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1679

CaselIT:
> ping me / request a review when this is at the next step.

friendly ping

zzzeek:
> > ping me / request a review when this is at the next step.
> 
> friendly ping

OK, I will try to get to this.

CaselIT:
I think I've updated the wrong branch. Fortunately it seems that sqla-tester does nto update the gerrit once someone else pushes an update

sqla-tester:

I've fixed the merge conflict

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1679

sqla-tester:
**Gord Thompson** (gordthompson) wrote:

Code-Review+1

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1679

sqla-tester:

do we want to try this one for 1.3.x ?  can we add a changelog file ?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1679

CaselIT:
sure. I forgot to add them.

ps: i'm updating directly the gerrit changeset, but I'm still using the github branch to backup work

sqla-tester:

> Patch Set 9:
> 
> do we want to try this one for 1.3.x ?  can we add a changelog file ?

I've added the changelog files. For now it's in the 1.4 folder. Feel free to move them in the 1.3 folder if we decide to target 1.3.x.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1679

sqla-tester:
**Gord Thompson** (gordthompson) wrote:

Code-Review+1

Buona

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1679

sqla-tester:

Code-Review+1

Thanks for finishing thid

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1796

sqla-tester:

Code-Review+1

Nice :+1:

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1679

sqla-tester:

Code-Review+1

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1796

sqla-tester:

Code-Review+2 Workflow+1

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1679

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1679 has been **merged**. Congratulations! :)

sqla-tester:

Code-Review+2 Workflow+1

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1796

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1796 has been **merged**. Congratulations! :)

sqla-tester:

alembic-recheck

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/1796

PR #5069, opened 2020-01-01
Rewrite on to use `with` instead of explicit acquire

heckad:
### Description

For more simple reasons, rewrite explicit lock acquire on using `with` operator.

### DIFF
--- a/lib/sqlalchemy/orm/__init__.py
+++ b/lib/sqlalchemy/orm/__init__.py
@@ -231,17 +231,14 @@ def clear_mappers():
     upon a fixed set of classes.
 
     """
-    mapperlib._CONFIGURE_MUTEX.acquire()
-    try:
-        while _mapper_registry:
-            try:
-                # can't even reliably call list(weakdict) in jython
-                mapper, b = _mapper_registry.popitem()
-                mapper.dispose()
-            except KeyError:
-                pass
-    finally:
-        mapperlib._CONFIGURE_MUTEX.release()
+    with mapperlib._CONFIGURE_MUTEX, _mapper_registry:
+        try:
+            # can't even reliably call list(weakdict) in jython
+            mapper, b = _mapper_registry.popitem()
+            mapper.dispose()
+        except KeyError:
+            pass
+
 
 
 joinedload = strategy_options.joinedload._unbound_fn
--- a/lib/sqlalchemy/orm/__init__.py
+++ b/lib/sqlalchemy/orm/__init__.py
@@ -231,13 +231,14 @@ def clear_mappers():
     upon a fixed set of classes.
 
     """
-    with mapperlib._CONFIGURE_MUTEX, _mapper_registry:
-        try:
-            # can't even reliably call list(weakdict) in jython
-            mapper, b = _mapper_registry.popitem()
-            mapper.dispose()
-        except KeyError:
-            pass
+    with mapperlib._CONFIGURE_MUTEX:
+        while _mapper_registry:
+            try:
+                # can't even reliably call list(weakdict) in jython
+                mapper, b = _mapper_registry.popitem()
+                mapper.dispose()
+            except KeyError:
+                pass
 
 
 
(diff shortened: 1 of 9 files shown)


sqla-tester (commented):
OK, this is **sqla-tester** setting up my work to try to get revision efeac06dda5afdbe33abcf9b27c8b5b5725c8444 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change efeac06dda5afdbe33abcf9b27c8b5b5725c8444: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/sqlalchemy/+/1648

lib/sqlalchemy/orm/__init__.py
```diff
@@ -231,17 +231,14 @@ def clear_mappers():
     upon a fixed set of classes.
 
     """
-    mapperlib._CONFIGURE_MUTEX.acquire()
-    try:
-        while _mapper_registry:
```
zzzeek: this is an old jython workaround that isn't needed, however the syntax change here removes the "while"

zzzeek: yeah, this is wrong, will fix

lib/sqlalchemy/util/queue.py
```diff
@@ -63,31 +72,25 @@ def __init__(self, maxsize=0, use_lifo=False):
...
     def qsize(self):
         """Return the approximate size of the queue (not reliable!)."""
 
-        self.mutex.acquire()
-        n = self._qsize()
-        self.mutex.release()
-        return n
+        return self._qsize()
 
+    @atomic
```
zzzeek: these should look like Python 3's queue.py, will fix

lib/sqlalchemy/util/_collections.py
```diff
@@ -944,7 +944,7 @@ def size_threshold(self):
         return self.capacity + self.capacity * self.threshold
 
     def _manage_size(self):
-        if not self._mutex.acquire(False):
+        if self._mutex.locked():
```
zzzeek: this loses the aquire()!

lib/sqlalchemy/util/queue.py
heckad: It looks like python3's queue. `atomic` is the decorator define above.

zzzeek: source is from here: https://github.com/python/cpython/blob/master/Lib/queue.py#L91 .

zzzeek:
thanks for this PR!  I hadn't realized Python 2 supports context managers for mutexes.

heckad:
It supports context managers from 2.5 versions. Examples with locks were in pep.

lib/sqlalchemy/util/_collections.py
heckad: Why is this method use this pattern?

sqla-tester:

> heckad@github wrote:
 > 
 > 
 > 
 > * lib/sqlalchemy/util/_collections.py (None): Why is this method
 > use this pattern?

because the operation needs to only occur in one thread and not any others, so if a thread has already acquired the lock, there is no need to run it again.

lib/sqlalchemy/util/queue.py
heckad: I think atomic more clear and safety because not needed to think about scope. If you see atomic then this method thread-safe.

zzzeek: thanks, IMO this is too much verbiage for a small part of a single file.   if there were a concurrency framework in use throughout all of SQLAlchemy, and there was no use of "with mutex:" and instead it were all `@atomic` decorators, that would be more consistent.  however, it's inconsistent that one part of the library do a certain task in an entirely different style than the rest of the library.  Additionally these methods are ported directly from the Python 3 source code and I'm not even sure they are being used.

sqla-tester:

Code-Review+2 Workflow+1

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1648 has been **merged**. Congratulations! :)

Issue #5087, opened 2020-01-10
add scalar() , columns() etc. support to 2.0 Result object.


rgargente:
I often find that I need to transform the results of single column SQLAlchemy queries from a a list of tuples to a list of whatever type the column is. Having an option in SQLA itself would avoid us having take care of this every time.

I wrote code to implement it. The way it would be used is very simple:
```
usernames = s.query(User.name).all(scalar=True)
```
usernames would be a list of strings as opposed of a list of tuples with a single string inside each one.

zzzeek:
hi there -

this feature is already part of a longer term plan in SQLAlchemy 2 and this particular item has not been ticketed yet.     the method will apply to an as-yet-not-created "interim" object that will apply to both Core SELECT statements and similar as well as ORM queries, which by 2.0 will also be via "select()", however in 1.4 the Query object will try to feature some of these new APIs for transitional purposes.

The proposed name at the moment for this method is `.all_scalar()`, but it also needs to be accompanied by a bunch of other result-oriented features, such as being able to control reception of ORM entities alone vs. tuples.  that is, the query.all() method should be returning tuples in **all** cases, however we can't do that in 1.4 because it would be badly backwards incompatible.

basically there's a redesign of the API going on and this is just one little part of it so I'm holding off on query API changes right now until I get to that part of it.

rgargente:
Well I'm glad to see we will get this functionality one way or another!

zzzeek:
yes, it's part of the master plan to fix up all these use cases

zzzeek:
this will be part of the "futrue" API introduced at https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#sqlalchemy-1-x-to-2-0-transition

a bunch of use patterns are illustrated at https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#resultproxy-replaced-with-invocationcontext-which-has-more-refined-methods, where "scalar()" is just one of many column-organizing methods:
```

    result.scalar()  # first col of first row  (warns if additional rows remain?)
    result.scalars()  # iterator of first col of each row
    result.scalars().all()  # same, as a list

    result.columns('a', 'b').<anything>  # limit column tuples
    result.columns(table.c.a, table.c.b)  # using Column (or ORM attribute) objects

    result.columns('b', 'a')  # order is maintained

    # if the result is an ORM result, you could do:
    result.columns(User, Address)   # assuming these are available entities


    # index access and slices ?
[...]

PR #5088, opened 2020-01-10
New scalar option for query.all() method

rgargente:
I often find that I need to transform the results of single column SQLAlchemy queries from a a list of tuples to a list of whatever type the column is. Having an option in SQLA itself would avoid us having take care of this every time.

### Description
A very simple change: a new "scalar" argument in the Query.all() method that defaults to False so that it doesn't affect any existing code. If True, it transforms the result into a list of scalar values rather than returning the tuples as they are.

### Checklist
This pull request is:

- [ ] A short code fix
- [x ] A new feature implementation
	Issue #5087

**Greetings from Kopernio :) Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/orm/query.py
+++ b/lib/sqlalchemy/orm/query.py
@@ -3152,10 +3152,13 @@ class Query(Generative):
         else:
             self._suffixes = suffixes
 
-    def all(self):
+    def all(self, scalar=False):
         """Return the results represented by this :class:`.Query` as a list.
 
         This results in an execution of the underlying SQL statement.
+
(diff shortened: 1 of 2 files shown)


zzzeek:
this is going to be in SQLAlhcemy 2.0 and likely in 1.4 as well with something similar to 

```
result = session.execute(query)
result.scalars().all()

```
*maybe* a 1.x style like this:

q = session.qeury(...)

q.execute().scalars().all()

see https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#resultproxy-replaced-with-result-which-has-more-refined-methods-and-behaviors

zzzeek:
thanks for the PR , as stated above this is happening in a different way right now.

PR #5095, opened 2020-01-13
Simplify; Python 2.5 no longer supported

### LINKED ISSUE #5094: remove unused jython code
`./lib/sqlalchemy/connectors/zxJDBC.py`

```python
    supports_unicode_statements = sys.version > "2.5.0+"
```

It's best not to compare to `sys.version`, which is a string, as doing so can make assumptions that the major, minor or micro parts are single characters, which isn't guaranteed to be true. For example, Python 2.7.10 and 3.10.

It's probably safe in this case, but it Python 2.5 and 2.6 have long been dropped, and to prevent possible errors from copying and pasting, this line can be cleaned up to: 

```python
    supports_unicode_statements = True
```

hugovk:
Fixes: #5094

### Description

`./lib/sqlalchemy/connectors/zxJDBC.py`

```python
    supports_unicode_statements = sys.version > "2.5.0+"
```

It's best not to compare to `sys.version`, which is a string, as doing so can make assumptions that the major, minor or micro parts are single characters, which isn't guaranteed to be true. For example, Python 2.7.10 and 3.10.

It's probably safe in this case, but it Python 2.5 and 2.6 have long been dropped, and to prevent possible errors from copying and pasting, this line can be cleaned up to: 

```python
    supports_unicode_statements = True
```

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/connectors/zxJDBC.py
+++ b/lib/sqlalchemy/connectors/zxJDBC.py
@@ -17,7 +17,7 @@ class ZxJDBCConnector(Connector):
     supports_sane_multi_rowcount = False
 
     supports_unicode_binds = True
-    supports_unicode_statements = sys.version > "2.5.0+"
+    supports_unicode_statements = True
     description_encoding = None
     default_paramstyle = "qmark"
 


hugovk:
No longer necessary, see https://github.com/sqlalchemy/sqlalchemy/issues/5094

PR #5247, opened 2020-04-10
fix changelog

CaselIT:
Noticed a typo in the last changelog. I think this should be ported to 1_3 also
### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/doc/build/changelog/changelog_13.rst
+++ b/doc/build/changelog/changelog_13.rst
@@ -134,7 +134,7 @@
         :tickets: 5052
 
         Add ability to literal compile a :class:`DateTime`, :class:`Date`
-        or :class:"Time" when using the string dialect for debugging purposes.
+        or :class:`Time` when using the string dialect for debugging purposes.
         This change does not impact real dialect implementation that retain
         their current behavior.
 


CaselIT:
@zzzeek ping.

Do you prefer these on gerrit? we should probably setup a proper filter to the jenkins jobs to that they don't run when modifying just documentation

zzzeek:
so what's happening now is that there's a huge list of PRs and a huge list of gerrits and I don't have a single integrated system of knowing whiuch ones are ready for review.   I guess it's a matter of a single search on gerrit and a single search on PRs here.

CaselIT:
Not sure what would help.
We could use https://github.com/orgs/sqlalchemy/projects/1, but it's another place to keep in sync.

zzzeek:
when i' look at the github list of PRs what visual cue tells me the ones that are awaiting my attention?  I guess all of them are

CaselIT:
I guess we could have a label "for review". But that helps only if someone else set it, since users cannot add labels

zzzeek:
cherry-picked to rel_1_3 as well

PR #5279, opened 2020-04-23
Add issue template, other github files

CaselIT:
You can test it at https://github.com/CaselIT/sqlalchemy/issues/new/choose


CaselIT:
I was thinking that maybe it's not the best it the issue adds a label on creation, since it's harder to see the new ones?
Or maybe it could have also a new label, something like triage?

zzzeek:
yeah "requires triage" is great

CaselIT:
updated the label and added to the project

zzzeek:
this is SO GOOD.   Just needs to be in master right?

CaselIT:
yes

PR #5292, opened 2020-04-30
self.columns containing None prevent from attaching an index

ollofx:
if for instance this is declared
Index(__tablename__+'_ctry_distinct_on_idx', Country, Start, End, ScenarioType, desc('rsrc_created'))
the desc('rsrc_created') will generate a None in the self.columns which will result when evaluating difference for attaching an index
to {None} instead of set() in the _check_attach
has_string_cols = set(self._pending_colargs).difference(col_objs) 
has_string_cols = {None} will not trigger a _col_attached
Thank you

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/sql/schema.py
+++ b/lib/sqlalchemy/sql/schema.py
@@ -2908,7 +2908,8 @@ class ColumnCollectionMixin(object):
 
             # issue #3411 - don't do the per-column auto-attach if some of the
             # columns are specified as strings.
-            has_string_cols = set(self._pending_colargs).difference(col_objs)
+            has_string_cols = set([colarg for colarg in self._pending_colargs
+                                   if colarg is not None]).difference(col_objs)
             if not has_string_cols:
 
                 def _col_attached(column, table):


CaselIT:
The preferred way of contributing is to first open an issue that explains why this change in needed.

Couple you provide an example of what are you trying to do and what does this code should change?

zzzeek:
please open an issue first which indicates an [MCVE](http://stackoverflow.com/help/mcve) for your use case.    The Index already accommodates expressions like "desc('somestring')" and this does not produce None as illustrated by your code example, see below:

```
from sqlalchemy import Column
from sqlalchemy import desc
from sqlalchemy import Index
from sqlalchemy import String
from sqlalchemy.schema import CreateIndex

Base = declarative_base()


class A(Base):
    __tablename__ = "a"

    data = Column(String)
    x = Column(Integer)


idx1 = Index("some_index", A.data, A.x, desc("foo"))
idx2 = Index("some_index", desc("foo"))

assert idx1 in A.__table__.indexes

A.__table__.append_constraint(idx2)


for idx in A.__table__.indexes:
    print(CreateIndex(idx))

```

output:

```
CREATE INDEX some_index ON a (foo DESC)
CREATE INDEX some_index ON a (data, x, foo DESC)

```

ollofx:
CaselIT thanks for your fast replay, and zzzeek thanks for the MCVE, I wouldn't have tried to do more on this topic if you were not so reactive, so I also take the chance to thank you for sqlalchemy, I'm literally a fan, other languages lack such an ORM.

**synopsis:**
I had this bug yesterday, when creating my schema from scratch on deployed server which takes latest sqlalchemy version, some indexes disappeared.
my local version of sqlalquemy  was a bit out dated, and had a bug also with such a usecase but for another reason which is the usage of utils.zip_longest comparing self.columns with self._pending_colargs from diffrent size, (contained None)
this bug was fixed by you on 2019-07-23 so it was not anymore breaking but the correction has dismissed the creation of the index.

```
def sqlalchemy_1_3_16_fix_bug():
    # BUG version 1.3.16
    # https://github.com/sqlalchemy/sqlalchemy/pull/5292
    from sqlalchemy.sql.schema import ColumnCollectionMixin, Column, Table, exc

    def _check_attach(self, evt=False):
        col_objs = [c for c in self._pending_colargs if isinstance(c, Column)]

        cols_w_table = [c for c in col_objs if isinstance(c.table, Table)]

        cols_wo_table = set(col_objs).difference(cols_w_table)

        if cols_wo_table:
            # feature #3341 - place event listeners for Column objects
            # such that when all those cols are attached, we autoattach.
            assert not evt, "Should not reach here on event call"

            # issue #3411 - don't do the per-column auto-attach if some of the
            # columns are specified as strings.
            has_string_cols = set([colarg for colarg in self._pending_colargs
                                   if colarg is not None]).difference(col_objs)
            if not has_string_cols:

                def _col_attached(column, table):
                    # this isinstance() corresponds with the
                    # isinstance() above; only want to count Table-bound
                    # columns
                    if isinstance(table, Table):
                        cols_wo_table.discard(column)
                        if not cols_wo_table:
                            self._check_attach(evt=True)

                self._cols_wo_table = cols_wo_table
                for col in cols_wo_table:
                    col._on_table_attach(_col_attached)
                return

        columns = cols_w_table

        tables = {c.table for c in columns}
        if len(tables) == 1:
            self._set_parent_with_dispatch(tables.pop())
        elif len(tables) > 1 and not self._allow_multiple_tables:
            table = columns[0].table
            others = [c for c in columns[1:] if c.table is not table]
            if others:
                raise exc.ArgumentError(
                    "Column(s) %s are not part of table '%s'."
                    % (
                        ", ".join("'%s'" % c for c in others),
                        table.description,
                    )
                )

    ColumnCollectionMixin._check_attach = _check_attach
sqlalchemy_1_3_16_fix_bug()

from sqlalchemy import Column
from sqlalchemy import desc
from sqlalchemy import Index
from sqlalchemy import String
from sqlalchemy.schema import CreateIndex

Base = declarative_base()


class A(Base):
    __tablename__ = "a"

    data = Column(String)
    x = Column(Integer)
    Index("some_index_attach", data, x, desc("foo"))


idx1 = Index("some_index", A.data, A.x, desc("foo"))
idx2 = Index("some_index", desc("foo"))

assert idx1 in A.__table__.indexes

A.__table__.append_constraint(idx2)


for idx in A.__table__.indexes:
    print(CreateIndex(idx))

```
this outputs
```
CREATE INDEX some_index ON a (foo DESC)
CREATE INDEX some_index_attach ON a (data, x, foo DESC)
CREATE INDEX some_index ON a (data, x, foo DESC)
```

CaselIT:
This notation is not supported:
```py
class A(Base):
    __tablename__ = "a"

    data = Column(String)
    x = Column(Integer)
    Index("some_index_attach", data, x, desc("foo"))
```

You should define table arguments in `__table_args__`:
```py
class A(Base):
    __tablename__ = "a"

    data = Column(String)
    x = Column(Integer)
    __table_args__ = (Index("some_index_attach", data, x, desc("foo")),)
```

I think that the surprising thing is that this works:
```py
class A(Base):
    __tablename__ = "a"

    data = Column(String)
    x = Column(Integer)
    Index("some_index_attach", data, x)
```
@zzzeek maybe we should disallow this with a warn/raise, indicating that  `__table_args__` should be used.
I haven't figured out why it works

zzzeek:
so IIUC you had a workaround for an old bug in place that did something rather complicated with internal underscore methods (did I give you that workaround or your team came up with it?) , and that's what broke.

ollofx:
the patch, is your code of the latest release 1.3.16.
I debugged sqlachemy with the non creation of the index and made it functioning again.

the patch is exactly  your _check_attach function but 
```python
            has_string_cols = set([colarg for colarg in self._pending_colargs
                                   if colarg is not None]).difference(col_objs)
```
instead of
```python
            has_string_cols = set(self._pending_colargs).difference(col_objs)
```
this use case of inline_index_attach was functioning in previous versions prior to this commit
https://github.com/sqlalchemy/sqlalchemy/commit/44a8488e833675e15966c7e43e02726699e0e9e0
utils.zip_longest was matching properly self._pending_colargs and self.columns

image


```python
from sqlalchemy import Column
from sqlalchemy import desc
from sqlalchemy import Index
from sqlalchemy import String
from sqlalchemy.schema import CreateIndex

Base = declarative_base()


class A(Base):
    __tablename__ = "a"

    data = Column(String)
    x = Column(Integer)
    Index("inline_index_attach", data, x, desc("foo"))

for idx in A.__table__.indexes:
    print(CreateIndex(idx))
```

zzzeek:
oh i see that's the actual MCVE, OK.   I can make a real bug from this.  Also I apologize if this appears "reactive", it's just much faster for me to move forward with what I know is the correct MCVE and the issue rather than trying to get you to give me what is on my end pretty trivial.  MCVE coming up, will be linked here in a moment.

see #5298

zzzeek:
you're right, pending_colargs accommodates None for an expression that is not a string and not a table-attached column.   updated

ollofx:
> oh i see that's the actual MCVE, OK. I can make a real bug from this. Also I apologize if this appears "reactive", it's just much faster for me to move forward with what I know is the correct MCVE and the issue rather than trying to get you to give me what is on my end pretty trivial. MCVE coming up, will be linked here in a moment.
> 
> see #5298

I'm not native english speaker, so I apologize if reactive meant something else then react quickly to an inquery, I really liked that you guys had reacted quickly. and your first MCVE was helpful to show my case clearly, so I thanked you for that.

PR #5325, opened 2020-05-12
Folds two identical exception handlers into a single one

edelooff:
Fixes a `TODO` that searches for py2/3 compatible syntax to match multiple exception types.

### Description

Merges the two exception clauses using the syntax that exists for both Python 2 and 3 as per the exception handling tutorials ([Python 2](https://docs.python.org/2/tutorial/errors.html#handling-exceptions), [Python 3](https://docs.python.org/3/tutorial/errors.html#handling-exceptions))

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/orm/base.py
+++ b/lib/sqlalchemy/orm/base.py
@@ -313,11 +313,7 @@ def object_state(instance):
 def _inspect_mapped_object(instance):
     try:
         return instance_state(instance)
-        # TODO: whats the py-2/3 syntax to catch two
-        # different kinds of exceptions at once ?
-    except exc.UnmappedClassError:
-        return None
-    except exc.NO_STATE:
+    except (exc.UnmappedClassError,) + exc.NO_STATE:
         return None
 
 
--- a/lib/sqlalchemy/orm/exc.py
+++ b/lib/sqlalchemy/orm/exc.py
@@ -196,9 +196,7 @@ def _safe_cls_name(cls):
 def _default_unmapped(base, cls):
     try:
         mappers = base.manager_of_class(cls).mappers
-    except NO_STATE:
-        mappers = {}
-    except TypeError:
+    except (TypeError,) + NO_STATE:
         mappers = {}
     name = _safe_cls_name(cls)
 


lib/sqlalchemy/orm/base.py
```diff
@@ -319,11 +319,7 @@ def object_state(instance):
 def _inspect_mapped_object(instance):
     try:
         return instance_state(instance)
-        # TODO: whats the py-2/3 syntax to catch two
-        # different kinds of exceptions at once ?
-    except exc.UnmappedClassError:
-        return None
-    except exc.NO_STATE:
+    except (exc.UnmappedClassError, exc.NO_STATE):
```
CaselIT, line 322: NO_STATE is a tuple itself, so we cannot do this as evidenced by the test failing
https://github.com/sqlalchemy/sqlalchemy/blob/ea87d39d7a9926dc1c6bf3d70e8faf8575769cb0/lib/sqlalchemy/orm/exc.py#L15

CaselIT (changes requested):
Thanks for the pull request,
I've left a comment

lib/sqlalchemy/orm/base.py
edelooff: That's a pretty obvious and early fail indeed. I think I ran the test suite against `master` before I suggested this. I've updated the PR with a small helper function (and one additional site where a similar pattern is used), but it's less convincingly an improvement.

lib/sqlalchemy/orm/exc.py
```diff
@@ -196,9 +200,7 @@ def _default_unmapped(cls):
 
     try:
         mappers = base.manager_of_class(cls).mappers
-    except NO_STATE:
-        mappers = {}
-    except TypeError:
+    except _multi_exc(TypeError, *NO_STATE):
```
CaselIT, line 203: In python 3 we could just do `except (TypeError, *NO_STATE):` but it does not work on python 2.

I would suggest `except (TypeError,) + NO_STATE:` instead of adding a function

edelooff: I figured the function would allow the code to read as much as the Python 3 equivalent (modulo the literal function name there), but if dropping it better fits the practice within the project, I'm happy to follow that.

CaselIT: I did not think of that.  My suggestion what it was with the + it's immediately clear what it does :)

CaselIT:
can you take a look @zzzeek

CaselIT (approved):

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision d0835511e5509fb368a2193ebe4ecbbe3504b129 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change d0835511e5509fb368a2193ebe4ecbbe3504b129: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2026

sqla-tester:

recheck

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2026

sqla-tester:

Code-Review+2 Workflow+2

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2026

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2026 has been **merged**. Congratulations! :)

PR #5336, opened 2020-05-19
Add a docs link to the CockroachDB dialect

rafiss:
### Description

This is a docs only change to point to the sqlalchemy-cockroachdb dialect.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/doc/build/dialects/index.rst
+++ b/doc/build/dialects/index.rst
@@ -43,47 +43,49 @@ External Dialects
 
 Currently maintained external dialect projects for SQLAlchemy include:
 
-+---------------------------------------+-----------------------+
-| Database                              | Dialect               |
-+=======================================+=======================+
-| Amazon Redshift (via psycopg2)        | sqlalchemy-redshift_  |
-+---------------------------------------+-----------------------+
-| Apache Drill                          | sqlalchemy-drill_     |
-+---------------------------------------+-----------------------+
-| Apache Druid                          | pydruid_              |
-+---------------------------------------+-----------------------+
-| Apache Hive and Presto                | PyHive_               |
-+---------------------------------------+-----------------------+
-| Apache Solr                           | sqlalchemy-solr_      |
-+---------------------------------------+-----------------------+
-| CrateDB                               | crate-python_         |
-+---------------------------------------+-----------------------+
-| EXASolution                           | sqlalchemy_exasol_    |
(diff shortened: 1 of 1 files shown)


CaselIT (approved):

doc/build/dialects/index.rst
```diff
@@ -43,47 +43,49 @@ External Dialects
...
++---------------------------------------+-------------------------+
+| Apache Drill                          | sqlalchemy-drill_       |
++---------------------------------------+-------------------------+
+| Apache Druid                          | pydruid_                |
++---------------------------------------+-------------------------+
+| Apache Hive and Presto                | PyHive_                 |
++---------------------------------------+-------------------------+
+| Apache Solr                           | sqlalchemy-solr_        |
++---------------------------------------+-------------------------+
+| CockroachDB                           | sqlalchemy-cockroachdb_ |
```
CaselIT, line 59: I just noticed that the whole table needed to be reformatted.
Could you use a shorter placeholder here? like `sa-cockroachdb_` so we avoid 40+ changes?

CaselIT (changes requested):

doc/build/dialects/index.rst
rafiss: Thanks for the review. If I understand correctly, that's not just placeholder text -- isn't it the text that shows up in the final document? So the link name will be different that the actual name of the dialect package.

I can make this change, but I'm just checking to make sure I understand.

CaselIT: You are right, sorry. I thought that the actual text was provided below, but it seem that the text in the table is indeed what is rendered in the final document.

Since it's more appropriate using the full package name, `sqlalchemy-cockroachdb_` and it requires changing the table anyway, I think we could be a bit generous and have a few more spaces in the right columns

rafiss: That sounds good. Done.

CaselIT: thansk, sorry for the back and forth

CaselIT (approved):

gordthompson (approved):

zzzeek:
merged in 1d12bc138 with squash + cherry-pick and cherry-picked in c0ffe79c0, note the latter pulled in the firebird link also.

PR #5355, opened 2020-05-28
Annotate test_null_in_empty_set_is_false with tuple_in

rafiss:
Mark the test as requiring support for the IN operator on tuples.

This test requires the database to support `IN` on tuples. Annotate it as such
so that dialects can enable/disable this test as necessary.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/testing/suite/test_select.py
+++ b/lib/sqlalchemy/testing/suite/test_select.py
@@ -845,6 +845,7 @@ class ExpandingBoundInTest(fixtures.TablesTest):
 
         self._assert_result(stmt, [(1,), (2,), (3,), (4,)], params={"q": []})
 
+    @testing.requires.tuple_in
     def test_null_in_empty_set_is_false(self, connection):
         stmt = select(
             [


zzzeek:
hey there.

that test doesn't use tuples.    also, that's a short code change, not documentation.  can you share the actual problem you're having ?

rafiss:
Sorry, I was viewing this as a typographical fix, since I thought it was just a case of a forgotten annotation.

I had thought this test uses tuples because in PostgreSQL, a "row" (i.e. the result of `SELECT CAST(NULL AS INTEGER) WHERE 1!=1)` is represented as a tuple.

The problem I ran into is while working on testing SQLAlchemy with CockroachDB (NB: I work there). In general, CockroachDB does handle `IN <tuple>`, but it currently has a bug/missing feature that makes it fail this test case: https://github.com/cockroachdb/cockroach/issues/49651

I was planning to mark this test as `tuple_in`, then mark our CockroachDB dialect as not supporting `tuple_in` until we fix this bug.

Thanks for getting back to me!

zzzeek:
"tuple in" refers to this kind of thing:   (x, y) IN ((1, 2), (3, 4)) .    the test here specifically needs the dialect to simulate an "empty set" for an IN expression.  The oracle dialect for example does it like this:

    SELECT CASE WHEN (NULL IN (SELECT 1 FROM DUAL WHERE 1!=1)) THEN 1 ELSE 0 END AS anon_1 FROM DUAL

so basically you need some kind of SQL expression that when you put it inside of "IN (<here>)", it acts like an empty set.

rafiss:
Thanks for the details -- I agree that this PR is not the right approach. closing.

zzzeek:
OK hope you figured out your test case.

Issue #5357, opened 2020-05-28
Flexible **kwarg Definition for DeclarativeMeta base definition


ewengillies:
**Is your feature request related to a problem? Please describe.**

I am using the declarative API and trying to pass arguments to `__init_subclass__` in my "base" ORM that is generated from `declarative_base()`.  

I think code makes this clearest.  Lets first define the base that can do special things to its subclasses without redefining the `metaclass` via the `__init_subclass__` hook (introduced in [PEP 487](https://www.python.org/dev/peps/pep-0487/), exists in python 3.6 and up):
```python
from sqlalchemy.ext.declarative import declarative_base

_Base = declarative_base()
class ORMBase(_Base):
    """
    Define common methods for all tables
    """
    __abstract__ = True

    def __init_subclass__(cls, do_special_thing=False, **kwargs):
        super().__init_subclass__(**kwargs)
        if do_special_thing:
            print("i am a very special subclass")
        else:
            print("I am a normal subclass")
```

If the subclass doesn't touch `do_special_thing`, then all is fine:
```python
>>> class NormalClass(ORMBase):
>>>    pass
I am a normal subclass
```
If I touch the `do_special_thing`, then it gets passed to [`DeclarativeMeta.__init__`](https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/ext/declarative/api.py#L71) and fails. I can only assume this is because the `kwargs` that get passed to `__init_subclass__` are the same that get passed to `<metaclass>.__init__`.
```python
>>> class SpeicalClass(ORMBase, do_special_thing=True):
>>>    pass
i am a very special subclass
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-d185419c27b1> in <module>
     15             print("I am a normal subclass")
     16 
---> 17 class SpecialClass(ORMBase, do_special_thing=True):
     18     __abstract__ = True

TypeError: __init__() got an unexpected keyword argument 'do_special_thing'
```

**Describe the solution you'd like**

Solutions I can imagine:
Naively, just adding a `**kwargs` to the [`DeclarativeMeta.__init__`](https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/ext/declarative/api.py#L71) solves this.

**Describe alternatives you've considered**

Other solutions I've considered:
[...]

PR #5363, opened 2020-05-30
Add **kw support to DeclarativeMeta.__init__

ewengillies:
### Description

Add `**kw` to the signature of `DeclarativeMeta.__init__`.  This is inline with the metaclass `init()` specification in python 3.6+.  

A full discussion of this can be seen in #5357

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/test/ext/declarative/test_basic.py
+++ b/test/ext/declarative/test_basic.py
@@ -2131,6 +2131,20 @@ class DeclarativeTest(DeclarativeTestBase):
 
         assert not hasattr(Foo, "data_hybrid")
 
+    def test_kw_support_in_declarative_meta(self):
+        # This will not fail if DeclarativeMeta __init__ supports **kw
+
+        class BaseWithInitSubclass(Base):
+            __abstract__ = True
+
+            @classmethod
+            def __init_subclass__(cls, random_keyword_used_here=False, **kwargs):
+                pass
+
+        class User(BaseWithInitSubclass, random_keyword_used_here=True):
+            __tablename__ = "user"
+            id = Column(Integer, primary_key=True)
+
 
 def _produce_test(inline, stringbased):
     class ExplicitJoinTest(fixtures.MappedTest):
--- a/test/ext/declarative/test_basic.py
+++ b/test/ext/declarative/test_basic.py
@@ -2142,21 +2142,23 @@ class DeclarativeTest(DeclarativeTestBase):
                 super().__init_subclass__(**kw)
                 cls._set_random_keyword_used_here = random_keyword_used_here
 
-        # Including the kwarg in the class definition should work, i.e. not
-        # throw a TypeError
-        class User(BaseWithInitSubclass, random_keyword_used_here=True):
-            __tablename__ = "user"
-            id = Column(Integer, primary_key=True)
-
         # Omitting the kwarg in the class definition should work, i.e. not throw
         # a TypeError
         class AnotherUser(BaseWithInitSubclass):
             __tablename__ = "another_user"
             id = Column(Integer, primary_key=True)
-        
-        if util.py36:
-            eq_(User._set_random_keyword_used_here, True)
-            eq_(AnotherUser._set_random_keyword_used_here, False)
+
+        if util.py3k:
+            # Including the kwarg in the class definition should work, i.e. not
+            # throw a TypeError
+            class User(BaseWithInitSubclass, random_keyword_used_here=True):
+                __tablename__ = "user"
+                id = Column(Integer, primary_key=True)
+
+            # Check to see if __init_subclass__ works in supported versions
+            if util.py36:
+                eq_(User._set_random_keyword_used_here, True)
+                eq_(AnotherUser._set_random_keyword_used_here, False)
 
 
 def _produce_test(inline, stringbased):
--- a/test/ext/declarative/test_basic.py
+++ b/test/ext/declarative/test_basic.py
@@ -13,6 +13,7 @@ from sqlalchemy import testing
 from sqlalchemy import UniqueConstraint
 from sqlalchemy import util
 from sqlalchemy.ext import declarative as decl
+from sqlalchemy.ext.declarative import DeclarativeMeta
 from sqlalchemy.ext.declarative import declared_attr
 from sqlalchemy.ext.declarative import synonym_for
 from sqlalchemy.ext.declarative.base import _DeferredMapperConfig
@@ -2135,30 +2136,28 @@ class DeclarativeTest(DeclarativeTestBase):
     def test_kw_support_in_declarative_meta_init(self):
         # This will not fail if DeclarativeMeta __init__ supports **kw
 
-        class BaseWithInitSubclass(Base):
-            __abstract__ = True
+        class BaseUser(Base):
+            __tablename__ = "base"
+            id_ = Column(Integer, primary_key=True)
 
             @classmethod
-            def __init_subclass__(cls, random_keyword_used_here=False, **kw):
+            def __init_subclass__(cls, random_keyword=False, **kw):
                 super().__init_subclass__(**kw)
-                cls._set_random_keyword_used_here = random_keyword_used_here
+                cls._set_random_keyword_used_here = random_keyword
 
-        # Omitting the kwarg in the class definition should work, i.e. not
-        # throw a TypeError
-        class AnotherUser(BaseWithInitSubclass):
-            __tablename__ = "another_user"
-            id = Column(Integer, primary_key=True)
-
-        # Including the kwarg in the class definition should work, i.e.
-        # not throw a TypeError
-        class User(BaseWithInitSubclass, random_keyword_used_here=True):
+        class User(BaseUser):
             __tablename__ = "user"
-            id = Column(Integer, primary_key=True)
+            id_ = Column(Integer, ForeignKey("base.id_"), primary_key=True)
+
+        # Check the default option
+        eq_(User._set_random_keyword_used_here, False)
+
+        # Build the metaclass with a keyword!
+        bases = (BaseUser,)
+        UserType = DeclarativeMeta("UserType", bases, {}, random_keyword=True)
 
         # Check to see if __init_subclass__ works in supported versions
-        if util.py36:
-            eq_(User._set_random_keyword_used_here, True)
-            eq_(AnotherUser._set_random_keyword_used_here, False)
+        eq_(UserType._set_random_keyword_used_here, True)
 
 
 def _produce_test(inline, stringbased):
--- a/test/ext/declarative/test_basic.py
+++ b/test/ext/declarative/test_basic.py
@@ -2131,6 +2131,7 @@ class DeclarativeTest(DeclarativeTestBase):
 
         assert not hasattr(Foo, "data_hybrid")
 
+    @testing.requires.python3
     def test_kw_support_in_declarative_meta_init(self):
         # This will not fail if DeclarativeMeta __init__ supports **kw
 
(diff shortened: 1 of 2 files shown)


ewengillies:
This is ready, thanks sqlalchemy devs!

test/ext/declarative/test_basic.py
```diff
@@ -2131,6 +2131,28 @@ def data_hybrid(self):
...
 
+    def test_kw_support_in_declarative_meta_init(self):
+        # This will not fail if DeclarativeMeta __init__ supports **kw
+
+        class BaseWithInitSubclass(Base):
+            __abstract__ = True
+
+            @classmethod
+            def __init_subclass__(cls, random_keyword_used_here=False, **kw):
+                pass
```
CaselIT, line 2142: maybe we could set `random_keyword_used_here` to the class so that we can test that the correct value was propagated? like `cls._set_random_keyword_used_here = random_keyword_used_here` then test it like
```py
eq_(User._set_random_keyword_used_here, True)
eq_(AnotherUser._set_random_keyword_used_here, False)
```

CaselIT (changes requested):
Looks ok, just a small change in the comment

test/ext/declarative/test_basic.py
ewengillies: yeah I wasn't too sure how to check the functionality beyond "I don't want this syntax to break the ORM class".

My main concern with actually _using_ this `__init_subclass__` functionality in this test is that it only exists in python 3.6+.  As it stands, the two test cases are:
 * `__init_subclass__` exists in the python version: 
   * the class method introduced in the test overrides the `<base>.__init_subclass__` so it can participate as a hook, as intended.
   * If this function is omitted, then the `<base>.__init_subclass__` complains that it doesn't take the provided `random_keyword_used_here` kwarg.
   * The `random_keyword_used_here` kwarg is then passed to `DeclarativeMeta.__init__`, where `**kw` eats it.
 * `__init_subclass__` doesn't exist in the python version:
   * We've now defined a totally new dunder class method `__init_subclass__` that doesn't get hooked anywhere.
   * The `random_keyword_used_here` kwarg is passed directly to `DeclarativeMeta.__init__`, where `**kw` still eats it.

If I understand the CI pipeline correctly, `sqlalchemy` tests against 3.8, so we're squarely in the first test case.  If we don't move to the second test case, I can make this test a little richer.

This is my first `sqlalchemy` PR: do you require that your tests work against lower versions of python as well?

ewengillies: In addition, I'm afraid that the test, as I've written it, will not be clear on failure what went wrong.  Should I catch the `TypeError` and re-raise with a better error message?

CaselIT: oh right.
Regarding the python version, you can use `util.py36` to check that python 3.6 or later is used. So you can add the `eq_` test inside an if that checks it.
regarding the reraise I don't think it's needed. The test seems clear enough.
We test python 2.7, 3.5-3.8

ewengillies: yeah that sounds like the best approach, thanks for the feedback!  I've protected the detailed test in a `util.py36` check and used `cls._set_random_keyword_used_here ` as you suggested.

Please let me know if there is anything else.

CaselIT (approved):

CaselIT:
looks good not. I'll run it in the ci on gerrit

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 21b8702f5622b31193089d5cd016058de29c09ff of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 21b8702f5622b31193089d5cd016058de29c09ff: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2019

ewengillies:
Ahh okay, my ignorance towards python 2.7 is showing:
```python
.tox/py27-cext-sqlite-postgresql/lib/python2.7/site-packages/_pytest/python.py:507: in _importtestmodule
    mod = self.fspath.pyimport(ensuresyspath=importmode)
.tox/py27-cext-sqlite-postgresql/lib/python2.7/site-packages/py/_path/local.py:701: in pyimport
    __import__(modname)
E     File "/home/jenkins/workspace/sqlalchemy_gerrit/34da86df/test/ext/declarative/test_basic.py", line 2147
E       class User(BaseWithInitSubclass, random_keyword_used_here=True):
E                                                                ^
E   SyntaxError: invalid syntax
```
Should I just move the whole test into a `util.py3x` check?  i'll go ahead and do that, let me know if you think something else makes sense.

CaselIT:
You can just use the decorator `@testing.requires.python3` on the test function to skip the test in python2.

These are also a couple of pep errors
```
./test/ext/declarative/test_basic.py:2151:80: E501 line too long (80 > 79 characters)
./test/ext/declarative/test_basic.py:2156:1: W293 blank line contains whitespace
```

ewengillies:
its even worse than I thought, it throws a `SyntaxError` even with all the guards.  It seems to happen on `import`.

```python
/home/util/miniconda3/envs/sqlalchemy_2.7/lib/python2.7/site-packages/_pytest/python.py:507: in _importtestmodule
    mod = self.fspath.pyimport(ensuresyspath=importmode)
/home/util/miniconda3/envs/sqlalchemy_2.7/lib/python2.7/site-packages/py/_path/local.py:701: in pyimport
    __import__(modname)
E     File "/home/egillies/sqlalchemy/test/ext/declarative/test_basic.py", line 2157
E       class User(BaseWithInitSubclass, random_keyword_used_here=True):
E                                                                ^
E   SyntaxError: invalid syntax
```

Python 2.X doesn't know anything about the `class SomeClass(Base, argument=value)`, so pytest breaks hard on this.

I'm using this to test:
```terminal
pytest test/ext/declarative/test_basic.py
```

The syntax error breaks while its "gathering" the tests.  Is there a better command for testing?

CaselIT:
I'm not sure how to solve this. Syntax error happens before evaluating the code, so even if False would return the error.
@zzzeek ideas?

zzzeek:
> its even worse than I thought, it throws a `SyntaxError` even with all the guards. It seems to happen on `import`.
> 
> ```python
> /home/util/miniconda3/envs/sqlalchemy_2.7/lib/python2.7/site-packages/_pytest/python.py:507: in _importtestmodule
>     mod = self.fspath.pyimport(ensuresyspath=importmode)
> /home/util/miniconda3/envs/sqlalchemy_2.7/lib/python2.7/site-packages/py/_path/local.py:701: in pyimport
>     __import__(modname)
> E     File "/home/egillies/sqlalchemy/test/ext/declarative/test_basic.py", line 2157
> E       class User(BaseWithInitSubclass, random_keyword_used_here=True):
> E                                                                ^
> E   SyntaxError: invalid syntax
> ```
> 
> Python 2.X doesn't know anything about the `class SomeClass(Base, argument=value)`, so pytest breaks hard on this.
> 
> I'm using this to test:
> 
> ```
> pytest test/ext/declarative/test_basic.py
> ```
> 
> The syntax error breaks while its "gathering" the tests. Is there a better command for testing?

we would have to skip the new style syntax and invoke the metaclass directly, since we cant use standard metaclass syntax in any case we use the util.with_metaclass helper.   I can get part of it working with a patch but i can't get it to send the kw into the metaclass, so just test the metaclass directly.  the patch, which may not be useful, is:


```
diff --git a/lib/sqlalchemy/util/compat.py b/lib/sqlalchemy/util/compat.py
index 79de3c978..247dbc13c 100644
--- a/lib/sqlalchemy/util/compat.py
+++ b/lib/sqlalchemy/util/compat.py
@@ -399,7 +399,7 @@ def reraise(tp, value, tb=None, cause=None):
     raise_(value, with_traceback=tb, from_=cause)
 
 
-def with_metaclass(meta, *bases):
+def with_metaclass(meta, *bases, **kw):
     """Create a base class with a metaclass.
 
     Drops the middle class upon creation.
@@ -414,8 +414,15 @@ def with_metaclass(meta, *bases):
 
         def __new__(cls, name, this_bases, d):
             if this_bases is None:
-                return type.__new__(cls, name, (), d)
-            return meta(name, bases, d)
+                cls = type.__new__(cls, name, (), d)
+            else:
+                cls = meta(name, bases, d)
+
+            if hasattr(cls, "__init_subclass__") and hasattr(
+                cls.__init_subclass__, "__func__"
+            ):
+                cls.__init_subclass__.__func__(cls, **kw)
+            return cls
 
     return metaclass("temporary_class", None, {})

```

then you use it like this:

```
class TypeWKW(type):
    def __init__(cls, classname, bases, dict_, **kw):
        # can't get kw to be passed here though
        type.__init__(cls, classname, bases, dict_)


class MyClass(with_metaclass(TypeWKW, object, foo="bar")):
    def __init_subclass__(cls, **kw):
        print(kw)

```

ewengillies:
@zzzeek , thanks for the insight.  Its getting late in my part of the world, I'll resume tomorrow to finish this up.  Thanks both!

CaselIT:
As an alternative we can use a string and eval it only in python. But it's less than ideal

zzzeek:
is this stalled? there's not an urgent need to have a real test that runs the metaclass, just that **kw is accepted by the object.   this is a no-risk change basically so the bar is low.

ewengillies:
Hi Mike:

> is this stalled? 

Yes, stalled for personal (work-related) reasons, but not abandoned!  Apologies for the delay.

I will put some effort into incorporating your initial comments from [here](https://github.com/sqlalchemy/sqlalchemy/pull/5363#issuecomment-636366094) this weekend.  If nothing happens, then I'll stick a "shallower" check where I'll just call the metaclass directly.

If you need it sooner than that, I'll do my best to look at it tomorrow.  Thanks!

zzzeek:
no worries, just went through the PR list and just wanted to check.

ewengillies:
> we would have to skip the new style syntax and invoke the metaclass directly, since we cant use standard metaclass syntax in any case we use the util.with_metaclass helper. I can get part of it working with a patch but i can't get it to send the kw into the metaclass, so just test the metaclass directly. the patch, which may not be useful, is:

Right, so if I understand correctly, the patch uses [the code mentioned here in 'Metaclass Syntax Changes'](https://lucumr.pocoo.org/2013/5/21/porting-to-python-3-redux/), which mentions the use of both a dummy class and dummy metaclass, which would actually skirt around calling the metaclass we've modified in this PR, but simulate how `**kw` is used in `type` in python 3.6+.  Did I read that diff. correctly?

>there's not an urgent need to have a real test that runs the metaclass, just that **kw is accepted by the object. this is a no-risk change basically so the bar is low.

The solution I've arrived at works in both python `2.7` and `3.6` in that `2.7` no longer throws a syntax error.  Its a bit hacky passing a fully-formed ORM, `BaseUser`, to as a base of `UserType` via `DeclarativeMeta`.  Motivation:
  * This let me call DeclarativeMeta directly, since `BaseUser` is built from `Base` which was built from `declarative_base()`, meaning its been "prepared" as `DeclarativeMeta` and `sqlalchemy` at large expects (i.e. with class registry, metadata, etc.).
 * `User` tests the default kwarg of `__init_subclass__`.
 * `UserType` demonstrates that the `kwarg` is passed through okay.

What do you all think?

zzzeek:
OK great, so if we remove the **kw then the test fails right?  that's all it needs.   will run in cI now

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 0ad05a768316cba03a4d312ab39d3e8fbca7ac54 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 0ad05a768316cba03a4d312ab39d3e8fbca7ac54 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2019

ewengillies:
> OK great, so if we remove the **kw then the test fails right? that's all it needs. will run in cI now

yes, i tested this myself just to be sure!

zzzeek:
wow that is great it is actually using the pep 487 hook.  I couldn't get it to do that.

sqla-tester:

Code-Review+2 Workflow+2

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2019

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2019 has been **merged**. Congratulations! :)

sqla-tester:

Code-Review+2 Workflow+2

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2073

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2073 has been **merged**. Congratulations! :)

PR #5439, opened 2020-07-04
resolves #3757

### LINKED ISSUE #3757: Document that `load_only()` implies `undefer()`
**Migrated issue, originally created by Adrian ([@thiefmaster](https://github.com/thiefmaster))**

I just noticed that my column property which is defined with `deferred=True` is automatically undeferred if I list the column in the `load_only()` option. While this makes sense (I wouldn't specify it if I didn't want it to be loaded) I just noticed it by coincidence - the name "load *only*" sounds like it just restricts what's loaded but doesn't enable additional loading.

An extra sentence in the `load_only` docs could clarify this...

RamonWill:
I have included an additional note for the [load_only()](https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/orm/strategy_options.py#L1050) function

As per issue #3757 a bit more clarity was asked to be put into the documentation for the load_only() function.
If a column property is defined as deferred=True and then that same column's name attribute is used as a parameter in the load_only() function, the function will still load this column. This implies that this function will undefer (or override) Columns with the deferred=True property, so that the column can still be loaded as the user specified.

### Checklist

This pull request is:

- [x ] A documentation / typographical error fix
- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/orm/strategy_options.py
+++ b/lib/sqlalchemy/orm/strategy_options.py
@@ -1077,6 +1077,9 @@ def load_only(loadopt, *attrs):
                     Load(Address).load_only("email_address")
                 )
 
+     .. note:: This method will still load a :class:`_schema.Column` even
+        if the column property is defined with ``deferred=True``
+        for the :func:`.column_property` function.
 
     .. versionadded:: 0.9.0
 


zzzeek:
LGTM

zzzeek:
thank you!

PR #5451, opened 2020-07-12
Update dialect for pg8000 version 1.16.0

tlocke:
### Description

Note: This is the pull request that I wrongly sent to zzzeek/sqlalchemy first of all. Also, to answer your question, this unfortunately doesn't allow the full pg8000 suite to pass. It does bring it a bit closer though :-)

This proposed change just affects the pg8000 driver and tests. The only changes to the tests are to decorators where pg8000 used to fail but is working now. There are still ~11 tests that fail, but this is an improvement over the ~25 that currently fail. I've just released pg8000 version 1.16.0 and it's a fairly big change. The changes to pg8000 are:

* All data types are now sent as text rather than binary.

* Using adapters, custom types can be plugged in to pg8000.

* Previously, named prepared statements were used for all statements.
Now unnamed prepared statements are used by default, and named prepared
statements can be used explicitly by calling the Connection.prepare()
method, which returns a PreparedStatement object.

### Checklist

This pull request is:

- [x] A short code fix

I haven't opened an issue in the tracker because the tests were already failing, hope that's okay.

**Have a splendid day!**

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/pg8000.py
+++ b/lib/sqlalchemy/dialects/postgresql/pg8000.py
@@ -74,12 +74,16 @@ from uuid import UUID as _python_UUID
 from .base import _DECIMAL_TYPES
 from .base import _FLOAT_TYPES
 from .base import _INT_TYPES
+from .base import ENUM
+from .base import INTERVAL
 from .base import PGCompiler
 from .base import PGDialect
 from .base import PGExecutionContext
 from .base import PGIdentifierPreparer
 from .base import UUID
 from .json import JSON
+from .json import JSONB
+from .json import JSONPathType
 from ... import exc
 from ... import processors
 from ... import types as sqltypes
@@ -125,6 +129,24 @@ class _PGJSON(JSON):
         else:
             return super(_PGJSON, self).result_processor(dialect, coltype)
 
+    def get_dbapi_type(self, dbapi):
+        return dbapi.JSON
+
+
+class _PGJSONB(JSONB):
+    def get_dbapi_type(self, dbapi):
+        return dbapi.JSONB
+
+
+class _PGJSONIndexType(sqltypes.JSON.JSONIndexType):
+    def get_dbapi_type(self, dbapi):
+        return dbapi.UNKNOWN
+
+
+class _PGJSONPathType(JSONPathType):
+    def get_dbapi_type(self, dbapi):
+        return dbapi.UNKNOWN
+
 
 class _PGUUID(UUID):
     def bind_processor(self, dialect):
@@ -148,8 +170,48 @@ class _PGUUID(UUID):
             return process
 
 
+class _PGEnum(ENUM):
+    def get_dbapi_type(self, dbapi):
+        return dbapi.UNKNOWN
+
+
+class _PGInterval(INTERVAL):
+    def get_dbapi_type(self, dbapi):
+        return dbapi.INTERVAL
+
+
+class _PGTime(sqltypes.Time):
+    def get_dbapi_type(self, dbapi):
+        return dbapi.TIME
+
+
+class _PGInteger(sqltypes.Integer):
+    def get_dbapi_type(self, dbapi):
+        return dbapi.INTEGER
+
+
+class _PGNullType(sqltypes.NullType):
+    def get_dbapi_type(self, dbapi):
+        return dbapi.NULLTYPE
+
+
+class _PGBigInteger(sqltypes.BigInteger):
+    def get_dbapi_type(self, dbapi):
+        return dbapi.BIGINTEGER
+
+
+class _PGBoolean(sqltypes.Boolean):
+    def get_dbapi_type(self, dbapi):
+        return dbapi.BOOLEAN
+
+
 class PGExecutionContext_pg8000(PGExecutionContext):
-    pass
+    def pre_exec(self):
+        if not self.compiled:
+            return
+
+        if self.dialect._dbapi_version > (1, 16, 0):
+            self.set_input_sizes()
 
 
 class PGCompiler_pg8000(PGCompiler):
@@ -160,20 +222,11 @@ class PGCompiler_pg8000(PGCompiler):
             + self.process(binary.right, **kw)
         )
 
-    def post_process_text(self, text):
-        if "%%" in text:
-            util.warn(
-                "The SQLAlchemy postgresql dialect "
-                "now automatically escapes '%' in text() "
-                "expressions to '%%'."
-            )
-        return text.replace("%", "%%")
-
 
 class PGIdentifierPreparer_pg8000(PGIdentifierPreparer):
-    def _escape_identifier(self, value):
-        value = value.replace(self.escape_quote, self.escape_to_quote)
-        return value.replace("%", "%%")
+    def __init__(self, *args, **kwargs):
+        PGIdentifierPreparer.__init__(self, *args, **kwargs)
+        self._double_percents = False
 
 
 class PGDialect_pg8000(PGDialect):
@@ -195,9 +248,19 @@ class PGDialect_pg8000(PGDialect):
         {
             sqltypes.Numeric: _PGNumericNoBind,
             sqltypes.Float: _PGNumeric,
-            JSON: _PGJSON,
             sqltypes.JSON: _PGJSON,
+            sqltypes.Boolean: _PGBoolean,
+            sqltypes.NullType: _PGNullType,
+            JSONB: _PGJSONB,
+            sqltypes.JSON.JSONPathType: _PGJSONPathType,
+            sqltypes.JSON.JSONIndexType: _PGJSONIndexType,
             UUID: _PGUUID,
+            sqltypes.Interval: _PGInterval,
+            INTERVAL: _PGInterval,
+            sqltypes.Time: _PGTime,
+            sqltypes.Integer: _PGInteger,
+            sqltypes.BigInteger: _PGBigInteger,
+            sqltypes.Enum: _PGEnum,
         },
     )
 
@@ -310,6 +373,17 @@ class PGDialect_pg8000(PGDialect):
 
             def on_connect(conn):
                 self.set_isolation_level(conn, self.isolation_level)
+
+            fns.append(on_connect)
+
+        if self._dbapi_version > (1, 16, 0) and self._json_deserializer:
+
+            def on_connect(conn):
+                # json
+                conn.register_in_adapter(114, self._json_deserializer)
+
+                # jsonb
+                conn.register_in_adapter(3802, self._json_deserializer)
 
             fns.append(on_connect)
 
--- a/test/engine/test_deprecations.py
+++ b/test/engine/test_deprecations.py
@@ -579,10 +579,6 @@ class RawExecuteTest(fixtures.TablesTest):
             Column("user_name", VARCHAR(20)),
         )
 
-    @testing.fails_on(
-        "postgresql+pg8000",
-        "pg8000 still doesn't allow single paren without params",
-    )
     def test_no_params_option(self, connection):
         stmt = (
             "SELECT '%'"
--- a/test/engine/test_execute.py
+++ b/test/engine/test_execute.py
@@ -76,10 +76,6 @@ class ExecuteTest(fixtures.TablesTest):
             Column("user_name", VARCHAR(20)),
         )
 
-    @testing.fails_on(
-        "postgresql+pg8000",
-        "pg8000 still doesn't allow single paren without params",
-    )
     def test_no_params_option(self):
         stmt = (
             "SELECT '%'"
--- a/test/requirements.py
+++ b/test/requirements.py
@@ -981,7 +981,7 @@ class DefaultRequirements(SuiteRequirements):
 
     @property
     def json_array_indexes(self):
-        return self.json_type + fails_if("+pg8000")
+        return self.json_type
 
     @property
     def datetime_literals(self):
@@ -1177,13 +1177,6 @@ class DefaultRequirements(SuiteRequirements):
                     "only four decimal places ",
                 ),
                 (
-                    "postgresql+pg8000",
-                    None,
-                    None,
-                    "postgresql+pg8000 has FP inaccuracy even with "
-                    "only four decimal places ",
-                ),
-                (
                     "postgresql+psycopg2cffi",
                     None,
                     None,
@@ -1237,7 +1230,7 @@ class DefaultRequirements(SuiteRequirements):
 
     @property
     def duplicate_key_raises_integrity_error(self):
-        return fails_on("postgresql+pg8000")
+        return exclusions.open()
 
     def _has_pg_extension(self, name):
         def check(config):


zzzeek:
OK I'll do some testing locally , no worries :)

I will pull it into CI nonetheless for style checks and all that.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision c5a544f16f4dd62b5c55beaa910e87bef90567a1 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change c5a544f16f4dd62b5c55beaa910e87bef90567a1: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2085

lib/sqlalchemy/dialects/postgresql/pg8000.py
```diff
@@ -153,6 +153,40 @@ def process(value):
...
+            return proc
+        else:
+            pg_text = dialect.dbapi.PGText
+
+            def new_proc(value):
+                res = proc(value)
+                if res is None:
+                    return res
+                else:
+                    return pg_text(value)
```
zzzeek, line 186: I would assume this is a wrapper that allows pg8000 to know more about the type for strings coming in?   is this needed for python 3 or is this a compatibility thing for python 2?    also is there any difference between text/varchar, my understanding is that postgresql itself sees no difference between these two datatypes.

sqla-tester:
**Tony Locke** (tlocke) wrote:

> Patch Set 1:
> 
> (1 comment)
> 
> zzzeek@github wrote:

> I would assume this is a wrapper that allows pg8000 to know more about the type for strings coming in?

Yes. To go into the detail, pg8000 uses prepared statements exclusively (unnamed by default, but named prepared statement are available using a pg8000 extension). With PostgreSQL, to prepare a statement the types of the parameters are sent along with the SQL statement. Then the values are sent separately on each execution. For a `str` type the type sent to the server is `unknown` (705). This is so that a casual user of pg8000 can write various types as a string, and the server will attempt to cast it to the correct type. For more formal usage such as SQLAlchemy, pg8000 provides the PGText and PGVarchar types that send the `text` (25) and `varchar` (1043) types to the server.

>   is this needed for python 3 or is this a compatibility thing for python 2?

pg8000 is no longer compatible with Python 2.

>    also is there any difference between text/varchar, my understanding is that postgresql itself sees no difference between these two datatypes.

I think that deep down in PostgreSQL they're the same, although since they have a different type id in the type system pg8000 allows people to use make a distinction if they want to.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2085

sqla-tester:

> Patch Set 1:
> 
> > Patch Set 1:
> > 
> > (1 comment)
> > 
> > zzzeek@github wrote:
> 
> > I would assume this is a wrapper that allows pg8000 to know more about the type for strings coming in?
> 
> Yes. To go into the detail, pg8000 uses prepared statements exclusively (unnamed by default, but named prepared statement are available using a pg8000 extension). With PostgreSQL, to prepare a statement the types of the parameters are sent along with the SQL statement. Then the values are sent separately on each execution. For a `str` type the type sent to the server is `unknown` (705). This is so that a casual user of pg8000 can write various types as a string, and the server will attempt to cast it to the correct type. For more formal usage such as SQLAlchemy, pg8000 provides the PGText and PGVarchar types that send the `text` (25) and `varchar` (1043) types to the server.


OK.  did you consider using setinputsizes for this?  SQLAlchemy supports this method which can be called in an automated basis based on the TypeEngine objects associated with the bound parameters: https://www.python.org/dev/peps/pep-0249/#id26   I'm using this for example in order to adapt SQLAlchemy to the non-DBAPI asyncpg, which also uses prepared statements, as seen at:

https://gerrit.sqlalchemy.org/plugins/gitiles/sqlalchemy/sqlalchemy/+/refs/changes/71/2071/16/lib/sqlalchemy/dialects/postgresql/asyncpg.py#170

https://gerrit.sqlalchemy.org/plugins/gitiles/sqlalchemy/sqlalchemy/+/refs/changes/71/2071/16/lib/sqlalchemy/dialects/postgresql/asyncpg.py#219


also are there any negative impacts if we didn't wrap our string objects in these types?   they should be implicitly cast-able by Postgresql regardless.   I only ask because of all the datatypes we have to wrap for "binds", strings are typically the least controversial especially if this is py3k-only.


> 
> >   is this needed for python 3 or is this a compatibility thing for python 2?
> 
> pg8000 is no longer compatible with Python 2.

SQLAlchemy is doing python 2 throughout version 1.4, so are these pg8000 APIs compatible with the prior py2k-supporting versions of pg8000?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2085

tlocke:
Hi, I've taken up your suggestion of using `Cursor.setinputsizes` I hadn't thought of doing this before, and hadn't realized that SQLAlchemy supported it, so thanks for mentioning it :-) I've released pg8000 1.16.1 which adds an implementation of setinputsizes, and this pull request now includes a new version of the pg8000 dialect which uses setinputsizes. There are now 16 failures of the tests (using PostgreSQL 12). One source of test failures is where the setinputsizes parameter type is `DATETIME`, but the parameter sent is of type `INTERVAL`, for example `FAILED test/dialect/postgresql/test_query.py::ExtractTest_postgresql+pg8000_12_2::test_three` . The output from this test is:

```
sqlalchemy.exc.ProgrammingError: (pg8000.exceptions.ProgrammingError) {'S': 'ERROR', 'V': 'ERROR', 'C': '22007', 'M': 'invalid input syntax for type timestamp: "5 days 0 seconds 0 microseconds"', 'F': 'datetime.c', 'L': '3771', 'R': 'DateTimeParseError'}
[SQL: SELECT EXTRACT(hour FROM CURRENT_TIMESTAMP - %s) AS anon_1 
FROM t]
[parameters: (datetime.timedelta(days=5),)]

```
I'm not sure what to do to resolve this one.

As a result of using `setinputsizes` I no longer have to use special wrappers in the dialect, so PGText, PGVarchar are no longer there.

There remains your question about Python 2, and also backward compatibility. So pg8000 stopped supporting Python 2 in release 1.13.0 (2019-02-01) following the sunsetting of Python 2 on 2019-01-01. I'm not sure how best to handle that with SQLAlchemy supporting Python 2 until the end of 1.4.

On the backward compatibility of pg8000 itself between versions, this version of the dialect will only support version 1.16.1 and later. Perhaps I should put some `if` statements in to check the version and see if I can produce a backwardly compatible dialect?

zzzeek:
We usually try to use conditionals so prior versions of a dialect can be handled.

the INTERVAL thing, for the moment that actually needs some additional changes to work as I had quite a hard time getting it to work for asyncpg.   Parts of the patch at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2071  would need to merge in order for that to work with setinputsizes().   I'm not working on the asyncio patch at the moment but I can re-port the "INTERVAL" parts of it to this patch if we do this one first.

so basically, if you can get everything except INTERVAL going, and optionally if you want to achieve some backwards compatibility for prior pg8000 versions, you can then hand it off to me and I'll see if I can finish that part.

it's also probably feasible if we released pg8000 for Python 3 only in SQLAlchemy 1.4, 1.4 is still targeting Python 2 but the userbase is mostly on Python 3 now in any case so it's not that critical; Python 2 + pg8000 users can stay on SQLAlchemy 1.3 perhaps.

tlocke:
Sounds like a good approach to me. I've updated the dialect to be backwardly compatible with previous versions of pg8000. There are now 16 failures in the tests which I'm having trouble solving.  There are seven that are caused by the INTERVAL problem that we've talked about, but there remain nine that I can't work out. Here's the full list:

#### Not sure about

- FAILED test/sql/test_defaults.py::CTEDefaultTest_postgresql+pg8000_12_2::test_a_in_b[insert-select-_exclusions2]
- FAILED test/sql/test_defaults.py::CTEDefaultTest_postgresql+pg8000_12_2::test_a_in_b[update-select-_exclusions0]


#### Seems to require psycopg2

- FAILED test/dialect/postgresql/test_dialect.py::MiscBackendTest_postgresql+pg8000_12_2::test_serial_integer


#### The test didn't like the way the exception was raised, but I can't work out why

- FAILED test/engine/test_reconnect.py::PrePingRealTest_postgresql+pg8000_12_2::test_pre_ping_db_is_restarted


#### Server having trouble interpreting JSON value

- FAILED test/dialect/test_suite.py::JSONStringCastIndexTest_postgresql+pg8000_12_2::test_crit_against_int_coerce_type
- FAILED test/dialect/test_suite.py::JSONStringCastIndexTest_postgresql+pg8000_12_2::test_crit_against_string_coerce_type
- FAILED test/dialect/test_suite.py::JSONStringCastIndexTest_postgresql+pg8000_12_2::test_string_cast_crit_simple_int


#### INTERVAL problem

- FAILED test/dialect/postgresql/test_dialect.py::MiscBackendTest_postgresql+pg8000_12_2::test_extract
- FAILED test/dialect/postgresql/test_query.py::ExtractTest_postgresql+pg8000_12_2::test_eight
- FAILED test/dialect/postgresql/test_query.py::ExtractTest_postgresql+pg8000_12_2::test_fifteen
- FAILED test/dialect/postgresql/test_query.py::ExtractTest_postgresql+pg8000_12_2::test_four
- FAILED test/dialect/postgresql/test_query.py::ExtractTest_postgresql+pg8000_12_2::test_seven
- FAILED test/dialect/postgresql/test_query.py::ExtractTest_postgresql+pg8000_12_2::test_six
- FAILED test/dialect/postgresql/test_query.py::ExtractTest_postgresql+pg8000_12_2::test_three


#### Setinputsizes incorrect

- FAILED test/dialect/test_suite.py::ExpandingBoundInTest_postgresql+pg8000_12_2::test_bound_in_heterogeneous_two_tuple
- FAILED test/sql/test_query.py::QueryTest_postgresql+pg8000_12_2::test_expanding_in_composite

zzzeek:
OK, these can be hard to diagnose.  if you are not in a hurry I can see if I can find some time to go through these.   note I'm away for a week starting next friday, maybe I can take a look before then.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 7c66c1f9a9027c47bd36f1b17bcab9e6ee07819b of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 7c66c1f9a9027c47bd36f1b17bcab9e6ee07819b added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2085

tlocke:
That sounds great, thanks 🙂

zzzeek:
I've just merged the parts of a different patch that is intended to resolve the issues you are having with INTERVAL.   The patch at 83b3ce6 has merged in master and may fix the issue, but if not, it will definitely be closer to what it has to be.  in any case i think you need to rebase your patch for current master.

tlocke:
I've rebased my patch to master, but running the tests fails straight away at:

I```
NTERNALERROR>   File "/home/tlocke/sqlalchemy/test/../lib/sqlalchemy/engine/default.py", line 1412, in set_input_sizes
INTERNALERROR>     inputsizes = self.compiled._get_set_input_sizes_lookup(
INTERNALERROR> AttributeError: 'NoneType' object has no attribute '_get_set_input_sizes_lookup'
```
Not sure if there's something I need to add to the dialect?

**Comment by zzzeek** [comment]
make sure you only call the self.set_input_sizes() method on DefaultExecutionContext if self.compiled is not None, as seen at https://gerrit.sqlalchemy.org/plugins/gitiles/sqlalchemy/sqlalchemy/+/refs/changes/71/2071/16/lib/sqlalchemy/dialects/postgresql/asyncpg.py#162

**Comment by tlocke** [comment]
Thanks, that solved that problem. There are now only 3 failing tests in the dialect test suite:

```
FAILED test/dialect/test_suite.py::JSONStringCastIndexTest_postgresql+pg8000_12_2::test_crit_against_int_coerce_type
FAILED test/dialect/test_suite.py::JSONStringCastIndexTest_postgresql+pg8000_12_2::test_crit_against_string_coerce_type
FAILED test/dialect/test_suite.py::JSONStringCastIndexTest_postgresql+pg8000_12_2::test_string_cast_crit_simple_int
```
Looking into the first one `test/dialect/test_suite.py::JSONStringCastIndexTest_postgresql+pg8000_12_2::test_crit_against_int_coerce_type` the error message is:

```
sqlalchemy.exc.ProgrammingError: (pg8000.exceptions.ProgrammingError) {'S': 'ERROR', 'V': 'ERROR', 'C': '42883', 'M': 'operator does not exist: character varying = json', 'H': 'No operator matches the given name and argument types. You might need to add explicit type casts.', 'P': '114', 'F': 'parse_oper.c', 'L': '722', 'R': 'op_error'}
[SQL: SELECT data_table.name 
FROM data_table 
WHERE data_table.name = %s AND CAST((data_table.data -> %s) AS VARCHAR) = %s]
[parameters: ('r6', 'a', '5')]
```
The SQLA types of the parameters are:

`sqlalchemy.sql.sqltypes.String, sqlalchemy.sql.sqltypes.JSON.JSONIndexType, sqlalchemy.sql.sqltypes.JSON`

So my guess is that PostgreSQL isn't clever enough to know that the final parameter `'5'` is a string, even though it has the complex type JSON. Perhaps a cast is needed to help PostgreSQL?

Finally, I noticed that `setinputsizes` was sometimes being called with (what I think may be) invalid parameters. For example `test/dialect/test_suite.py::InsertBehaviorTest_postgresql+pg8000_12_2::test_insert_from_select_autoinc_no_rows` causes `setinputsizes` to be called with `('data_1_1', 'data_1_2')`. The test doesn't fail with pg8000 because it uses the type `UNKNOWN` if it's confused by a parameter from `setinputsizes`.

**Review by sqla-tester — COMMENTED** [review]
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 288df5c7be752cb42a8265b9782e40518e56b628 of this pull request into gerrit so we can run tests and reviews and stuff

**Comment by sqla-tester** [comment]
Patchset 288df5c7be752cb42a8265b9782e40518e56b628 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2085

**Comment by tlocke** [comment]
I noticed that a flake8-import-order test had failed so I've fixed that now.

**Review by sqla-tester — COMMENTED** [review]
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 639751ca9c7544801b9ede02e6cbe15a16c59c82 of this pull request into gerrit so we can run tests and reviews and stuff

**Comment by sqla-tester** [comment]
Patchset 639751ca9c7544801b9ede02e6cbe15a16c59c82 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2085

**Comment by zzzeek** [comment]
OK I think maybe you arent running as many tests as you could, here's how you can run the suite tests as well as the PG tests:

    py.test test/dialect/   --dburi postgresql+pg8000://scott:tiger@localhost/test

this will also run a suite of JSONB tests.   These will fail unless you also do the result processor thing:

```diff
diff --git a/lib/sqlalchemy/dialects/postgresql/pg8000.py b/lib/sqlalchemy/dialects/postgresql/pg8000.py
index a05661846..38e70a11a 100644
--- a/lib/sqlalchemy/dialects/postgresql/pg8000.py
+++ b/lib/sqlalchemy/dialects/postgresql/pg8000.py
@@ -134,6 +134,12 @@ class _PGJSON(JSON):
 
 
 class _PGJSONB(JSONB):
+    def result_processor(self, dialect, coltype):
+        if dialect._dbapi_version > (1, 10, 1):
+            return None  # Has native JSON
+        else:
+            return super(_PGJSON, self).result_processor(dialect, coltype)
+
     def get_dbapi_type(self, dbapi):
         return dbapi.JSONB
 

```

with that added, I get about 13 failures within that suite related to some JSON(B) expression tests, as well as a few other things.    that said, the dialect is pretty close with that above change, do you want me to go forward w/ a merge as is or do you want to try some of the remaining failures?    The failures involving JSON expressions are probably tricky to solve and likely related to the use of prepared statements, and I may have to do this work anyway for the asyncpg dialect as it also uses prepared statements.

zzzeek:
two of those tests wont pass for asyncpg either since they are doing cast(value, String) == type_coerce(value, JSON), im not sure why those are like that and that may be some legacy use case

zzzeek:
let me just go with what we have here since I want to make changes in gerrit..thanks!

tlocke:
Yes, sounds good!

zzzeek:
what's the correct setinputsies for TEXT [] ?   it seems to be 1009 but there's no fixed constant for this in pg8000.   this is what the "PGJSONPathType" is.    it's the argument to the `#>` operator here: https://www.postgresql.org/docs/9.5/functions-json.html    1009 seems to work

zzzeek:
similar need for TIMESTAMP WITH TIMEZONE - to support set_input_sizes() fullly we need all of these.  I'm hardcoding for now based on what I find in pg8000 and psycopg2 source code, I don't know where the canonical list of all these codes is (really would be handy)

```
class _PGTimeStamp(sqltypes.DateTime):
    def get_dbapi_type(self, dbapi):
        if self.timezone:
            # TIMESTAMPTZOID
            return 1184
        else:
            # TIMESTAMPOID
            return 1114

```

zzzeek:
another thing that is coming up a few times is that if I don't have a DBAPI type for a particular value, I necessarily have to pass it as "None" to setinputsizes(), because we are using positional parameters.   However if I pass None that seems to change the behavior of the cursor vs. if I didn't call setinputsizes() at all.   if a None value is sent to setinputsizes() for a certain parameter I think the cursor should behave as though nothing was passed.

zzzeek:
OK so I am tring to enable pg8000 on CI and we'll see if this gets all the way through, I had to make a few more changes to the type system to support the setinputsizes() level of accuracy.

tlocke:
I've added some more type constants to pg8000 (in the newly released 1.16.4), so now we've got:

```
BIGINTEGER = 20
BOOLEAN = 16
BOOLEAN_ARRAY = 1000
BYTES = 17
CHAR = 1042
CHAR_ARRAY = 1014
DATE = 1082
DATETIME = 1114
DECIMAL = 1700 
DECIMAL_ARRAY = 1231 
FLOAT = 701
FLOAT_ARRAY = 1022
INET = 869
INT2VECTOR = 22
INTEGER = 23 
INTEGER_ARRAY = 1016
INTERVAL = 1186
MACADDR = 829
NAME = 19
NAME_ARRAY = 1003
OID = 26
JSON = 114
JSONB = 3802
NULLTYPE = -1
NUMBER = 1700
ROWID = 26
STRING = 1043
TEXT = 25
TEXT_ARRAY = 1009
TIME = 1083
TIMEDELTA = 1186
TIMESTAMP = 1114
TIMESTAMPTZ = 1184
UNKNOWN = 705
UUID_TYPE = 2950
VARCHAR = 1043
VARCHAR_ARRAY = 1015
XID = 28
```

The canonical list can by found by running the query `select oid, typname from pg_type;` in PostgreSQL.

Regarding setting `None` as a parameter in `setinputsizes`, at the moment pg8000 interprets `None` as UNKNOWN. But I see what you mean, really it should behave for that parameter as if it hadn't been specified. I'll have a look at doing that.

zzzeek:
OK so a full suite ran for pg8000, and 11 failures which means this is production quality more or less , congrats!   most failures seem to be due to a mis-aligned tpc_prepare() method.  We absolutely don't need TPC transactions to work but since they are enabled a stack trace looks like:
```

File "/home/jenkins/workspace/sqlalchemy_gerrit/2743fa8a/.tox/py38-cext-sqlite-postgresql/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 895, in _prepare_twophase_impl
    self.engine.dialect.do_prepare_twophase(self, xid)
  File "/home/jenkins/workspace/sqlalchemy_gerrit/2743fa8a/.tox/py38-cext-sqlite-postgresql/lib/python3.8/site-packages/sqlalchemy/dialects/postgresql/pg8000.py", line 381, in do_prepare_twophase
    connection.connection.tpc_prepare()
  File "/home/jenkins/workspace/sqlalchemy_gerrit/2743fa8a/.tox/py38-cext-sqlite-postgresql/lib/python3.8/site-packages/pg8000/core.py", line 1558, in tpc_prepare
    self.execute(self._cursor, q, None)
AttributeError: 'Connection' object has no attribute 'execute'
```


then a few things went wrong with some CTE related things, we have a query that uses a CTE with RETURNING like this:

```
WITH c AS 
(UPDATE q SET x=%(x)s, y=%(y)s WHERE q.z = %(z_1)s RETURNING q.z)
 SELECT p.s, c.z 
FROM p, c 
WHERE p.s = c.z
INFO:sqlalchemy.engine.Engine:[generated in 0.00043s] {'x': 7, 'y': 5, 'z_1': 1}
DEBUG:sqlalchemy.engine.Engine:Col ('s', 'z')
DEBUG:sqlalchemy.engine.Engine:Row (1, 1)

```

and the pg8000 version seems to return no rows when one is expected.  it does not seem to be related to the setinputsizes thing.

zzzeek:
CTE thing is my fault, it's the positional parameters not working w the CTE

tlocke:
Thanks for pointing out the broken `tpc_prepare()`, I've fixed that now in the latest release 1.16.5.

sqla-tester:

Code-Review+2 Workflow+2

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2085

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2085 has been **merged**. Congratulations! :)

zzzeek:
this is fully merged now, thanks for your efforts Tony !

PR #5489, opened 2020-08-01
Fix typo #5486

### LINKED ISSUE #5486: Documentation typo in comment
Hi,
I found small typo on the following comments, in lib/sqlalchemy/sql/schema.py 1068 line

        :param type\_: The column's type, indicated using an instance which
should be
        :param type: The column's type, indicated using an instance which

Can i pr this project?

thanks!

inspire12:
This PR is for corrections to these typos. resolves #5486

### DIFF
--- a/lib/sqlalchemy/sql/schema.py
+++ b/lib/sqlalchemy/sql/schema.py
@@ -1065,7 +1065,7 @@ class Column(DialectKWArgs, SchemaItem, ColumnClause):
           :class:`_schema.Table`.  This is to support convenient
           usage within the :mod:`~sqlalchemy.ext.declarative` extension.
 
-        :param type\_: The column's type, indicated using an instance which
+        :param type_: The column's type, indicated using an instance which
           subclasses :class:`~sqlalchemy.types.TypeEngine`.  If no arguments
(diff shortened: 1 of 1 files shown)


lib/sqlalchemy/sql/schema.py
```diff
@@ -1065,7 +1065,7 @@ def __init__(self, *args, **kwargs):
           :class:`_schema.Table`.  This is to support convenient
           usage within the :mod:`~sqlalchemy.ext.declarative` extension.
 
-        :param type\_: The column's type, indicated using an instance which
```
zzzeek, line 1068: the \_ is an escape so that "type_" is not seen as an internal hyperlink target.

see https://docutils.sourceforge.io/docs/user/rst/quickref.html#internal-hyperlink-targets

and https://docutils.sourceforge.io/docs/user/rst/quickref.html#escaping

an exact example of `name\_` is at: https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#hyperlink-targets

jvanasco: I edited the comment above to put the `\_` in backticks so it would render correctly.

Issue #5429, opened 2020-06-29
proposal: Standardize `isnot` & `notin_` operator names for SqlAlchemy2.0 (1.4?)


jvanasco:
A handful of sql operators, as implemented in Python, are not consistent in SQLAlchemy.

The majority of items use a schema of `snake_case_with_trailing_underscore_` **when the sql command mimics Python reserved keywords**

For example, some of the commands adopting this scheme use:

| Python  | Sql | SQLAlchemy |
| ----- | ----- | ----- |
| and  | AND  | and_ |
| or  | OR  | or_ |
| is  | IS  |  is_ |
| in  | IS  |  in_ |

but compound operators either do not follow the same scheme, do not have an internal underscore, or both

| Python  | Sql | SQLAlchemy |
| ----- | ----- | ----- |
| is not  | IS NOT  | isnot |
| not in  | NOT IN  | notin_ |

This ticket will list them, and serve as reference to a PR if we proceed with any re-namings.

| SQLAchemy 1.3 | Proposed A | Proposed B (*rejected*) |
| ----- | ----- | ----- |
| isnot | is_not | is_not_ |
| notin_ | not_in | not_in_ |

If a change were to happen:
* the core function will be renamed
* the old function will be an alias for backwards compatibility; it will not be deprecated or warnings
* docs will be updated to use the new terms
* API docs for the old terms will state they are kept for backwards compatibility, and promote usage of the new terms

CaselIT:
I'm nor particularly fond of `is_not_` and `not_in_` vs `is_not` and `not_in`, but I understand the consistency of having the tailing undersoce

zzzeek:
just FTR I was looking to go from Python's example:

```
>>> import operator
>>> operator.is_
<built-in function is_>
>>> operator.is_not
<built-in function is_not>

```

jvanasco:
@zzzeek @CaselIT I marked my tailing underscore proposal as rejected.

zzzeek:
we can still ask twitter I often get a lot of insight from that.   either way

xarg:
The convention could be: trailing underscore should be reserved only for python single-word keywords.

Ex:

- `is` -> `is_`
- `is not` -> `is_not`

Also compatible with the `operator` module -> less surprises.

CaselIT:
I've pinned this issue and 5435, so maybe we gather some other comment from the community

jvanasco:
PR forthcoming...

[...]

PR #5501, opened 2020-08-07
First attempt at #5429: standardize operators

jvanasco:
# Description

This is a first attempt at addressing #5429 to standardize operator names.

It is very likely this PR changes too much. It is being presented for discussion.

### Checklist
This pull request is:

- [ ] A short code fix

I will immediately follow up this PR with a first comment diving into my initial concerns.


jvanasco:
I had more concerns as generating this PR, but forget them. These two stood out:

# Concern 1: `*_op`

I am unfamiliar with the `*_op` usage.  I fear I am touching some fragile elements.  

`notin_op` has been renamed to `not_in_op`

# Concern 2: testing 

These changes were made:

    - from .assertions import is_not_  # noqa	
    + from .assertions import is_not   # noqa	

    - from .assertions import not_in_  # noqa	
    + from .assertions import not_in  # noqa

jvanasco:
Requested @CaselIT for general feedback.
Requested @gordthompson for feedback concerning how these changes could effect dialects and dialect testing.

lib/sqlalchemy/sql/operators.py
```diff
@@ -670,7 +674,11 @@ def isnot(self, other):
         .. seealso:: :meth:`.ColumnOperators.is_`
 
         """
-        return self.operate(isnot, other)
+        return self.operate(is_not, other)
+
+    def isnot(self, other):
+        # SqlAlchemy 1.4 deprecates legacy terms for new standardizations (see Issue#5429)
+        return self.is_not(other)
```
CaselIT, line 681: Since at the moment we are not deprecating them with a warning, I would just do `isnot = is_not` so we avoid a function call. Same also for `notin_`

test/sql/test_operators.py
```diff
@@ -98,15 +98,15 @@ def test_scalar(self, operator, compare_to):
...
         (operators.is_, False),
-        (operators.isnot, False),
+        (operators.is_not, False),
         (operators.like_op, right_column),
         (operators.notlike_op, right_column),
         (operators.ilike_op, right_column),
         (operators.notilike_op, right_column),
         (operators.is_, right_column),
-        (operators.isnot, right_column),
+        (operators.is_not, right_column),
```
CaselIT, line 109: I think here we could keep both the new operators and the old ones, at least until we do not actively warn on the old version

CaselIT (commented):
Just a couple of minor things. Looks nice otherwise

lib/sqlalchemy/sql/operators.py
jvanasco: 100% agree. I tried to use that pattern elsewhere, not sure how this one slipped through.

test/sql/test_operators.py
jvanasco: you are absolutely right. GREAT CATCH.

there are some tests that should be adjusted to test for the legacy versions too

jvanasco:
I think I'm actually happy with the current state of the PR. After digging into a lot of the underlying code, I settled on a changeset that uses the current terms in the core library and test-suite, and silently supports the old terms in both.  A new unit test is used to ensure the legacy support options are there, and I import all of the old terms into the test-suite alongside the new terms, just in case third-party libraries leverage it for their dialect tests.

zzzeek:
I want to pull this in but my big concern is that SQL Core and testing were addressed at the same time.  For testing, changing these names will have an impact on being able to cherry-pick fixes back to 1.3.     I'd rather have the "testing" part of this, which is largely private and of very low impact on the outside world, in both master and 1.3.   Can we split this out ?

jvanasco:
Closing this in favor of #5523 and #5524

PR #5531, opened 2020-08-23
Make AsyncConnection._sync_connection() async for flexiblility

### LINKED ISSUE #5529: `create_async_engine(server_side_cursors=True)` raises wrong exception
**Describe the bug**

This is a typo in `create_async_engine()` where it should be `async_exc` instead of `exc`.

```python
In [1]: from sqlalchemy.ext.asyncio import create_async_engine

In [2]: await create_async_engine(server_side_cursors=True)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-2-2bcff6ed9bdc> in <module>
----> 1 await create_async_engine(server_side_cursors=True)

lib/sqlalchemy/ext/asyncio/engine.py in create_async_engine(*arg, **kw)
     32
     33     if kw.get("server_side_cursors", False):
---> 34         raise exc.AsyncMethodRequired(
     35             "Can't set server_side_cursors for async engine globally; "
     36             "use the connection.stream() method for an async "

AttributeError: module 'sqlalchemy.exc' has no attribute 'AsyncMethodRequired'
```

**Expected behavior**

```python
In [1]: from sqlalchemy.ext.asyncio import create_async_engine

In [2]: await create_async_engine(server_side_cursors=True)
---------------------------------------------------------------------------
AsyncMethodRequired                       Traceback (most recent call last)
<ipython-input-2-2bcff6ed9bdc> in <module>
----> 1 await create_async_engine(server_side_cursors=True)

lib/sqlalchemy/ext/asyncio/engine.py in create_async_engine(*arg, **kw)
     32
     33     if kw.get("server_side_cursors", False):
---> 34         raise async_exc.AsyncMethodRequired(
     35             "Can't set server_side_cursors for async engine globally; "
     36             "use the connection.stream() method for an async "

AsyncMethodRequired: Can't set server_side_cursors for async engine globally; use the connection.stream() method for an async streaming result set
```

**To Reproduce**

```py
import asyncio

asyncio.run(create_async_engine(server_side_cursors=True))
```

**Error**

```
  File "reproduce.py", line 4, in <module>
    asyncio.run(create_async_engine(server_side_cursors=True))
  File "..../lib/sqlalchemy/ext/asyncio/engine.py", line 34, in create_async_engine
    raise exc.AsyncMethodRequired(
AttributeError: module 'sqlalchemy.exc' has no attribute 'AsyncMethodRequired'
```

**Versions.**
 - OS: Darwin 19.5.0
 - Python: Python 3.8.2
 - SQLAlchemy: 1.4.0b1 317f2e1be2b06cdc12bc84510eb743d9752763dd
 - Database: PostgreSQL 12.4
 - DBAPI: asyncpg 0.21.0

**Additional context**

**Have a nice day!**

### LINKED ISSUE #5530: Make `AsyncConnection._sync_connection` async
**Is your feature request related to a problem? Please describe.**

In the next major version of [GINO](https://python-gino.org) which will be built on SQLAlchemy 1.4, `AsyncConnection` is subclassed to be lazy (and together with many other contextual features, to be compatible with the current GINO version). It would be neat to put the lazy-connection initialization code in the overridden `AsyncConnection._sync_connection()` method to avoid a lot of code duplicates.

**Describe the solution you'd like**

As demonstrated in https://github.com/fantix/sqlalchemy/commit/async-sync-connection, making the `_sync_connection()` method sync would allow subclasses to blend in async logic to get the actual connection.

**Describe alternatives you've considered**

Otherwise, the subclasses would have to override and copy over all `AsyncConnection` methods to add async logic.

**Additional context**

https://github.com/python-gino/gino/blob/501840be96a875490fda39f43ed5139fdee56d5d/src/gino/engine.py#L154-L166

**Have a nice day!**

fantix:
### Description

This PR makes `AsyncConnection._sync_connection()` async, even though itself in SQLAlchemy doesn't have to be async, but this allows subclasses to be able to easily blend in async code before the actual database interactions.

### Checklist

This pull request is:

- [x] A short code fix
    * Fixes #5530

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/ext/asyncio/engine.py
+++ b/lib/sqlalchemy/ext/asyncio/engine.py
@@ -79,7 +79,10 @@ class AsyncConnection(StartableContext):
         self.sync_connection = await (greenlet_spawn(self.sync_engine.connect))
         return self
 
-    def _sync_connection(self):
+    async def _sync_connection(self):
+        # Even though this method itself doesn't have to be async, it's still
+        # defined as an async method to allow subclasses to be able to add
+        # different async logic here, e.g. lazy connections.
         if not self.sync_connection:
             self._raise_for_not_started()
         return self.sync_connection
@@ -88,14 +91,12 @@ class AsyncConnection(StartableContext):
         """Begin a transaction prior to autobegin occurring.
 
         """
-        self._sync_connection()
         return AsyncTransaction(self)
 
     def begin_nested(self) -> "AsyncTransaction":
         """Begin a nested transaction and return a transaction handle.
 
         """
-        self._sync_connection()
         return AsyncTransaction(self, nested=True)
 
     async def commit(self):
@@ -110,7 +111,7 @@ class AsyncConnection(StartableContext):
         :meth:`_future.Connection.begin` method is called.
 
         """
-        conn = self._sync_connection()
+        conn = await self._sync_connection()
         await greenlet_spawn(conn.commit)
 
     async def rollback(self):
@@ -127,7 +128,7 @@ class AsyncConnection(StartableContext):
 
 
         """
-        conn = self._sync_connection()
+        conn = await self._sync_connection()
         await greenlet_spawn(conn.rollback)
 
     async def close(self):
@@ -137,7 +138,7 @@ class AsyncConnection(StartableContext):
         is in place.
 
         """
-        conn = self._sync_connection()
+        conn = await self._sync_connection()
         await greenlet_spawn(conn.close)
 
     async def exec_driver_sql(
@@ -151,7 +152,7 @@ class AsyncConnection(StartableContext):
 
         """
 
-        conn = self._sync_connection()
+        conn = await self._sync_connection()
 
         result = await greenlet_spawn(
             conn.exec_driver_sql, statement, parameters, execution_options,
@@ -175,7 +176,7 @@ class AsyncConnection(StartableContext):
         """Execute a statement and return a streaming
         :class:`_asyncio.AsyncResult` object."""
 
-        conn = self._sync_connection()
+        conn = await self._sync_connection()
 
         result = await greenlet_spawn(
             conn._execute_20,
@@ -227,7 +228,7 @@ class AsyncConnection(StartableContext):
         :return: a :class:`_engine.Result` object.
 
         """
-        conn = self._sync_connection()
+        conn = await self._sync_connection()
 
         result = await greenlet_spawn(
             conn._execute_20, statement, parameters, execution_options,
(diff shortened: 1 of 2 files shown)


lib/sqlalchemy/ext/asyncio/engine.py
```diff
@@ -88,14 +91,12 @@ def begin(self) -> "AsyncTransaction":
         """Begin a transaction prior to autobegin occurring.
 
         """
-        self._sync_connection()
```
zzzeek, line 91: this is here basically to do an error check, "if not self.sync_connection: raise.. " it doesn't need to be the same as _sync_connection() but it should be something

zzzeek (commented):
I'm not sure what to do here, because changes like these tend to make long-term code maintenance difficult.   There will be some odd API pattern that I did in a very specific way for someone, it's years later, nobody uses it anymore, and then I'm not sure if someone is using some odd pattern somewhere that is now in the way of my being able to refactor.  examples of this include the sqlalchemy.ext.instrumentation package and the Session._enable_transaction_accounting flag.  these were things that were there for individual people who needed them for something and then I was stuck with them years after anybody was using them.

I'm not saying this is a bad idea or anything it's just that I can see this placing itself as kind of an internal rigidness to how AsyncConnection works which makes me a little nervous based on past experience.

as I mentioned in the issue, the AsyncConnection itself already needs to be awaited in some way to actually connect.  GINO needs a *second* await before it actually connects?

lib/sqlalchemy/ext/asyncio/engine.py
fantix: Oh I thought this error check is done in `AsyncTransaction.start()` anyway, so I removed this early check. This also makes `await conn.begin()` consistent with `await engine.connect()` where the error is raised on `await`, not the awaitable-generating part, pls see also https://github.com/sqlalchemy/sqlalchemy/issues/5530#issuecomment-678897947.

CaselIT:
I think we should split this PR, since the fix for #5529 is straight forward, while the other change may need a bit more back and forth

zzzeek:
just an update, havent forgotten about this, when I get through some other 1.4 big things I want to circle back to async and review the API that I have in the first place, inlcuding looking at my begin() method to determine if that should be async or not.   then in order to do the thing you want here, there would have to be some first-class design feature of Connection that allows for a "async get the connection" hook that is tested in the test suite against a sample "lazy connecting" connection, and it has to be something I can live with forever, so I just want to get the time to have other 1.4 things done first.

fantix:
Oh sure, thanks for the update! At the same time, I'm also having more understanding of the 1.4 async details now by porting GINO to 1.4, which is mostly done, and the implementation is much easier now with 1.4 in place 👍

zzzeek:
per #5530 we are tabling this for now.

PR #5568, opened 2020-09-09
Remove whitespace before colon

pje1740:
Removed whitespace placed before the colons.
This was fixed in accordance with PEP8.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/test/sql/test_operators.py
+++ b/test/sql/test_operators.py
@@ -848,7 +848,7 @@ class ArrayIndexOpTest(fixtures.TestBase, testing.AssertsCompiledSQL):
         col2 = Column("y", Integer())
 
         self.assert_compile(
-            col[col2 : col2 + 5], "x[y:y + :y_1]", checkparams={"y_1": 5}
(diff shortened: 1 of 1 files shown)


zzzeek:
hey there -

there seem to be a whole bunch of PRs coming from Korea in quick succession, I guess you guys are sprinting or something, thanks!

unfortunately this particular fix is not correct as the colon in question is inside of a list slice,  per https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements :   

> However, in a slice the colon acts like a binary operator, and should have equal amounts on either side (treating it as the operator with the lowest priority). In an extended slice, both colons must have the same amount of spacing applied.

Beyond that, all of SQLAlchemy's actual code formatting is automated by the Python [black](https://github.com/psf/black) tool, so it's not even possible for adjustments to be made in most code whitespace outside of string literals as black will always set it back.

thanks for the contribution!  sorry this one didnt work out but keep on plugging.

PR #5570, opened 2020-09-09
Refactor comparison in test

yeongseon:
### Description

Fix anti-pattern related to E712 such as Comparison to true should be 'if cond is true:' or 'if cond:'

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/test/orm/test_scoping.py
+++ b/test/orm/test_scoping.py
@@ -111,11 +111,11 @@ class ScopedSessionTest(fixtures.MappedTest):
         Session = scoped_session(sa.orm.sessionmaker(), mock_scope_func)
 
         s0 = SessionMaker()
-        assert s0.autocommit == False
+        assert not s0.autocommit
 
         mock_scope_func.return_value = 0
         s1 = Session()
-        assert s1.autocommit == False
+        assert not s1.autocommit
 
         assert_raises_message(
             sa.exc.InvalidRequestError,
@@ -126,4 +126,4 @@ class ScopedSessionTest(fixtures.MappedTest):
 
         mock_scope_func.return_value = 1
         s2 = Session(autocommit=True)
-        assert s2.autocommit == True
+        assert s2.autocommit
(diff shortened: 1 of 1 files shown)


test/orm/test_scoping.py
```diff
@@ -111,11 +111,11 @@ def test_call_with_kwargs(self):
         Session = scoped_session(sa.orm.sessionmaker(), mock_scope_func)
 
         s0 = SessionMaker()
-        assert s0.autocommit == False
+        assert not s0.autocommit
```
zzzeek, line 114: So!  yes this is good.  it would be better, if you are so inspired, to use SQLAlchemy's is_true and is_false assertions, can you force-push a new change here?   the imports are:

```
from sqlalchemy.testing import is_false
from sqlalchemy.testing import is_true

```
then:
```

is_false(s0.autocommit)
is_false(s1.autocommit)
is_true(s2.autocommit)
```

zzzeek (changes requested):
Hi there!    this is a good change however small, and I have suggested a further improvement.

As the change adds some imports, SQLAlchemy makes use of [flake8-import-order](https://github.com/PyCQA/flake8-import-order) to ensure a deterministic ordering of imports.

What you can do in order to automate this change on your end is to install [pre-commit](https://pre-commit.com/) for your SQLAlchemy checkout:

```
$ cd /path/to/sqlalchemy_repo
$ pre-commit install
```

when you next do a "git commit", pre-commit will automatically apply black / zimports formatting and run flake8 on your changes before completing the commit.

thanks for your contribution!

test/orm/test_scoping.py
yeongseon: Thank you for your suggestion. I'll push new change. :)

zzzeek:
im sorry this is too trivial for a merge right now.     we have hundreds if not thousands of assertions that can ultimately be cleaned up.

PR #5616, opened 2020-09-27
Support PEP-567 context variables

### LINKED ISSUE #5615: PEP-567 context variable is not always set with async engine
**Describe the bug**

After greenlet 0.4.17 supports PEP-567 in python-greenlet/greenlet#172, SQLAlchemy async engine is not setting the context correctly in `greenlet_spawn()`, causing bugs in retrieving context variables in sync methods like event handlers.

**Expected behavior**

The context is passed down and context variables are working properly, please see example below.

**To Reproduce**

```py
import asyncio
from contextvars import ContextVar

from sqlalchemy import event

var = ContextVar("var", default="default")


def on_connect(dbapi_connection, connection_record):
    print(var.get())


async def main():
    e = create_async_engine("postgresql+asyncpg:///")
    event.listen(e.sync_engine, "connect", on_connect)

    var.set("expected")
    async with e.connect():
        pass


asyncio.run(main())
```

**Expected**: `expected`

**Actual**: `default`

**Versions.**

 - OS: Darwin Kernel Version 19.5.0
 - Python: Python 3.8.2
 - SQLAlchemy: 1.4.0b1 @ ffafbd4b9657a4ee8bec57c0861414144f37bdc5
 - Database: PostgreSQL 12.4
 - DBAPI: asyncpg 0.21.0
 - greenlet: 0.4.17

**Additional context**

Async methods called from such sync methods by `await_only()` is not affected, because the async method is actually called within the parent greenlet, where the PEP-567 context is correct. For example, this would generate `expected`:

```python
async def async_on_connect():
    print(var.get())

def on_connect(dbapi_connection, connection_record):
    await_only(async_on_connect())
```

**Have a nice day!**

fantix:
Fixes: #5615

### Description

Invoke the given function within a copy of the current PEP-567 context in `greenlet_spawn()`, so that the following sync methods could retrieve the correct context variable.

This PR doesn't cover Python < 3.7 or greenlet < 0.4.17.

However, this PR cannot fix another issue when the context variable is set in the sync method (new greenlet created by `greenlet_spawn()`), the direct async method cannot retrieve the correct context variable, for the same reason why it worked in "Additional Context" of #5615 - the direct async method is actually called from the parent greenlet, whose PEP-567 context doesn't have the variable set:

```python
import asyncio
from contextvars import ContextVar

from sqlalchemy import event
from sqlalchemy.util import await_only

var = ContextVar("var", default="default")


async def async_on_connect():
    print(var.get())


def on_connect(dbapi_connection, connection_record):
    var.set("expected")
    await_only(async_on_connect())


async def main():
    e = create_async_engine("postgresql+asyncpg:///")
    event.listen(e.sync_engine, "connect", on_connect)

    async with e.connect():
        pass


asyncio.run(main())
```

(mind where the context variable is set)

**Expected**: `expected`, **Actual**:  `default`.

A workaround is to make the async method indirect, for example, make it a new task:

```python
def on_connect(dbapi_connection, connection_record):
    var.set("expected")
    await_only(asyncio.ensure_future(async_on_connect()))
```

Expected: `expected`, Actual: `expected`.

### Checklist

This pull request is:

- [x] A short code fix

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/util/_concurrency_py3k.py
+++ b/lib/sqlalchemy/util/_concurrency_py3k.py
@@ -8,6 +8,16 @@ import greenlet
 
 from .. import exc
 
+try:
+    from contextvars import copy_context as _copy_context
+
+    # If greenlet.gr_context is present in current version of greenlet,
+    # it will be set with a copy of the current context on creation.
+    # Refs: https://github.com/python-greenlet/greenlet/pull/198
+    getattr(greenlet.greenlet, "gr_context")
+except (ImportError, AttributeError):
+    _copy_context = None
+
 
 # implementation based on snaury gist at
 # https://gist.github.com/snaury/202bf4f22c41ca34e56297bae5f33fef
@@ -18,6 +28,8 @@ class _AsyncIoGreenlet(greenlet.greenlet):
     def __init__(self, fn, driver):
         greenlet.greenlet.__init__(self, fn, driver)
         self.driver = driver
+        if _copy_context is not None:
+            self.gr_context = _copy_context()
 
 
 def await_only(awaitable: Coroutine) -> Any:
--- a/test/base/test_concurrency_py3k.py
+++ b/test/base/test_concurrency_py3k.py
@@ -1,4 +1,5 @@
 from sqlalchemy import exc
+from sqlalchemy import testing
 from sqlalchemy.testing import async_test
 from sqlalchemy.testing import eq_
 from sqlalchemy.testing import expect_raises_message
@@ -6,6 +7,11 @@ from sqlalchemy.testing import fixtures
 from sqlalchemy.util import await_fallback
 from sqlalchemy.util import await_only
 from sqlalchemy.util import greenlet_spawn
+
+try:
+    from greenlet import greenlet
+except ImportError:
+    greenlet = None
 
 
 async def run1():
@@ -101,3 +107,36 @@ class TestAsyncioCompat(fixtures.TestBase):
             await greenlet_spawn(go)
 
         await to_await
+
+    @async_test
+    @testing.requires.python37
+    @testing.skip_if(lambda: not hasattr(greenlet, "gr_context"))
+    async def test_contextvars(self):
+        import asyncio
+        import contextvars
+
+        var = contextvars.ContextVar("var")
+        event = asyncio.Event()
+        counter = [0]
+        concurrency = 5
+
+        async def async_inner(val):
+            eq_(val, var.get())
+
+        def inner(val):
+            await_only(async_inner(val))
+            eq_(val, var.get())
+
+        async def task(val):
+            var.set(val)
+            counter[0] += 1
+            if counter[0] == concurrency:
+                event.set()
+            await event.wait()
+            await greenlet_spawn(inner, val)
+
+        done, _ = await asyncio.wait(
+            [asyncio.ensure_future(task(i)) for i in range(concurrency)]
+        )
+        for fut in done:
+            await fut
--- a/test/base/test_concurrency_py3k.py
+++ b/test/base/test_concurrency_py3k.py
@@ -110,7 +110,6 @@ class TestAsyncioCompat(fixtures.TestBase):
 
     @async_test
     @testing.requires.python37
-    @testing.skip_if(lambda: not hasattr(greenlet, "gr_context"))
     async def test_contextvars(self):
         import asyncio
         import contextvars
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,7 +39,7 @@ package_dir =
     =lib
 install_requires =
     importlib-metadata;python_version<"3.8"
-    greenlet;python_version>="3"
+    greenlet != 0.4.17;python_version>="3"
 
 [options.extras_require]
 asyncio =
--- a/tox.ini
+++ b/tox.ini
@@ -17,7 +17,7 @@ usedevelop=
 
 deps=pytest>=4.6.11 # this can be 6.x once we are on python 3 only
      pytest-xdist
-     greenlet
+     greenlet != 0.4.17
      mock; python_version < '3.3'
      importlib_metadata; python_version < '3.8'
      postgresql: .[postgresql]


lib/sqlalchemy/util/_concurrency_py3k.py
```diff
@@ -14,9 +20,26 @@
     # Issue for context: https://github.com/python-greenlet/greenlet/issues/173
 
     class _AsyncIoGreenlet(greenlet.greenlet):
-        def __init__(self, fn, driver):
-            greenlet.greenlet.__init__(self, fn, driver)
-            self.driver = driver
+        if copy_context is None or greenlet.__version__ < "0.4.17":
```
fantix, line 23: If we could enforce the required version of greenlet to be 1.0 or above, we could remove the branches here and keep the last branch only.

lib/sqlalchemy/util/_concurrency_py3k.py
```diff
@@ -14,9 +20,26 @@
...
+
+            def __init__(self, fn, driver):
+                greenlet.greenlet.__init__(self, fn, driver)
+                self.driver = driver
+
+        elif greenlet.__version__ == "0.4.17":
```
zzzeek, line 29: can you fill me in on what is specific about 0.4.17 here?   I would think that we would disregard using new behavior for this specific version.   also version comparison would need to be done more robustly, however I would hope that we can use a simple duck-typing approach:

```
    if hasattr(self, "gr_context"):
        self.gr_context = copy_context()

```

fantix: Oh sure - greenlet 0.4.17 included python-greenlet/greenlet#172, which starts new greenlets with empty contexts. #5615 is only caused by greenlet 0.4.17.

I think the duck-typing approach would work fine. I'll update this PR as a draft proposal.

CaselIT: Personally I think it would be fine to constraint the install to `gr_context != 0.4.17`

auvipy (changes requested):
could you please fix the conflicts?

fantix:
Sure thing!

test/base/test_concurrency_py3k.py
```diff
@@ -101,3 +107,36 @@ def go():
             await greenlet_spawn(go)
 
         await to_await
+
+    @async_test
+    @testing.requires.python37
+    @testing.skip_if(lambda: not hasattr(greenlet, "gr_context"))
```
CaselIT, line 113: shouldn't this work in both cases, since we are blocking the 0.4.17 release?

fantix: Ah, this is because we removed the code to support greenlet < 0.4.17, and this test didn't cover the part when it's not working with <0.4.17, so I added this skip to avoid a false "PASS".

CaselIT: well technically, looking at the test, it just verifies that the context vars are properly set when using `await_only` and `greenlet_spawn` so I think it's not a false PASS for the older greenlet. It just happened to handle it that way without any other code change.

I think we could make it run so we don't accidentally break it for some other reason

fantix: That makes sense, PR is updated with the latest master now 🎉 thanks for reviewing!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision dcf42f7b78ef3e983fda17f7190cda7b4511e3b4 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change dcf42f7b78ef3e983fda17f7190cda7b4511e3b4: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2376

sqla-tester:

Code-Review+1 Workflow+1

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2376

sqla-tester:

mike you can take a look

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2376

test/base/test_concurrency_py3k.py
```diff
@@ -101,3 +107,35 @@ def go():
...
+        import contextvars
+
+        var = contextvars.ContextVar("var")
+        event = asyncio.Event()
+        counter = [0]
+        concurrency = 5
+
+        async def async_inner(val):
+            eq_(val, var.get())
+
```
sqla-tester, line 124: **mike bayer** (zzzeek) wrote:

I tried to make this test simpler and easier to understand.  if I run the latest git master of greenlet, the test passes unless I take out the new contextvar logic in which case it fails, which suggests that this test is accurately testing the behavior.  if you can confirm i am not missing anything that would be helpful.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2376

sqla-tester (commented):

Code-Review+2 Workflow+1

(1 comment)

@caselit @fantix one more review as I tried to simplify the test.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2376

sqla-tester:
**Fantix King** (fantix) wrote:

(1 comment)

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2376

test/base/test_concurrency_py3k.py
sqla-tester: **Fantix King** (fantix) wrote:

Yes excactly, that's been my way to verify that the test is working correctly, by commenting out the "self.gr_context = _copy_context()" in the greenlet subclass and watch the test fails.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2376

sqla-tester:

Code-Review+1 Workflow+2

looks good to me

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2376

sqla-tester:

Workflow+2

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2376

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2376 has been **merged**. Congratulations! :)

Issue #4757, opened 2019-07-08
rename apply_labels() and with_labels() to use_labels(), core/ ORM


zzzeek:
this is a big inconsistency that should be fixed.   Select.use_labels flag right now should become underscored which was the only backwards compat concern.

gordthompson:
@zzzeek - On looking at the code a bit closer, it appears that `.use_labels` is implemented in both `sql/selectable.py` and `orm/query.py` as a status interrogator (i.e., "Is the 'labels' option enabled?"), while `.apply_labels` and `.with_labels` (respectively) are used to set the `self._label_style` flag. Might we want to standardize the "setter" to either `.apply_labels` or `.with_labels` and leave `.use_labels` as the "getter"?

zzzeek:
hey @gordthompson - 

I haven't done a deep think on this one.   Right now in 1.4 there are *three* labeling styles, and that is something new.   the new style is LABEL_STYLE_DISAMBIGUATE_ONLY and it's likely good as the "only" label style we might actually need going forward.    the internals should not really be checking ".use_labels" at this point because it does no accommodate for the three styles that exist now.

So i had in mind some new API that accommodates for these styles:

   stmt = stmt.use_labels(LABEL_STYLE_DISAMBIGUATE_ONLY)

or something, maybe flags  use_labels(disambiguate=True, include_tablename=True), not really sure.

the main point is it should be identical for both Query and select() because they are really two versions of the same thing now.

gordthompson:
@zzzeek - Perhaps one flag might be sufficient:

(`.use_labels()` omitted) → `LABEL_STYLE_NONE`

`.use_labels()` or `.use_labels(False)` or `.use_labels(disambiguate_only=False)` → `LABEL_STYLE_TABLENAME_PLUS_COL`

`.use_labels(True)` or `.use_labels(disambiguate_only=True)` → `LABEL_STYLE_DISAMBIGUATE_ONLY`

zzzeek:
it looks like right now Select.use_labels and Query.use_labels are boolean flags indicating the classic style of labeling.   we can perhaps go for a strict approach right now, have those "use_labels" accessors emit a 2.0 deprecation warning, and then add a simple methods set_label_style() and get_label_style() and just use the symbols.     if we were to make the default style "disambiguate only", people would basically never need to worry about this entire thing at all for new code because it always does the right thing.

gordthompson:
[...]

PR #5622, opened 2020-10-01
GH-4757 Rename apply_labels to use_labels

Saphyel:
Rename all the apply_labels to use_labels 

### Description
As part of the process of moving to 2.0 apply_labels should be replaced with use_labels


gordthompson:
Thanks for your contribution. Unfortunately, it does not fully address the scope of changes required for #4757. Feel free to re-open if you can offer a more comprehensive solution.

zzzeek:
hey there -

specifically we're also looking to change the API on select() and such.

PR #5636, opened 2020-10-08
Add parens around unmarked clause

evantayloryates:
### Description
I'm not 100% sure what this sentence is saying, but I'm pretty sure it needs either commas or parentheses to break it up a little. I think parens make the most sense in this case (assuming I am reading it correctly). Here's the original sentence and the proposed edit in plaintext:

ORIG
 "...persistent objects which were removed from a collection or in some cases a scalar attribute may also be pulled into the Session of a parent object;..."

EDIT
 "...persistent objects which were removed from a collection (or in some cases a scalar attribute) may also be pulled into the Session of a parent object;..."

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/doc/build/orm/cascades.rst
+++ b/doc/build/orm/cascades.rst
@@ -93,12 +93,11 @@ becomes part of the state of that :class:`.Session`::
     >>> address3 in sess
     >>> True
 
-``save-update`` has the possibly surprising behavior which is that
-persistent objects which were *removed* from a collection
-(or in some cases a scalar attribute)
-may also be pulled into the :class:`.Session` of a parent object; this is
+A ``save-update`` cascade can exhibit surprising behavior when removing an item from
+a collection or de-associating an object from a scalar attribute. In some cases, the
+orphaned objects may still be pulled into the ex-parent's :class:`.Session`; this is
 so that the flush process may handle that related object appropriately.
-This case can usually only arise if an object is removed from one :class:`.Session`
+This case usually only arises if an object is removed from one :class:`.Session`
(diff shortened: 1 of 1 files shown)


doc/build/orm/cascades.rst
```diff
@@ -95,7 +95,7 @@ becomes part of the state of that :class:`.Session`::
 
 ``save-update`` has the possibly surprising behavior which is that
 persistent objects which were *removed* from a collection
```
zzzeek, line 97: agree sentence is bad, here is proposed wording:


`save-update`` has the possibly surprising behavior which is that persistent objects which were either *removed* from a collection, or de-associated from a scalar attribute, may nonetheless be pulled into the :class:`_orm.Session` of the parent object from which it's being removed; this is so that...

hows that?

evantayloryates: It's definitely better, but I think it's a little long. There's a lot to say and it's hard to do it all in one sentence. I tried breaking it up a little:

A ``save-update`` cascade can exhibit surprising behavior when removing an item from a collection or de-associating an object from a scalar attribute. In some cases, the orphaned objects may still be pulled into the ex-parent's session; this is so that...

zzzeek: great, send that to press please

evantayloryates: I updated the patch-1 branch to reflect the new changes. Should be good to go

zzzeek:
thank you!

Issue #5692, opened 2020-11-11
support mysql ssl_mode flag in ssl dictionary


thereisnosun:
**Description**
An ability to force TLS in MySQL

**Expected behavior**
When ssl_mode flag is specified MySQL connection should behave accordingly. Meaning, when I want to drop connection when SSL is disabled on the server it should drop. In mysqldb this is implemented via ssl_mode flag.

**To Reproduce**
Just add ssl_mode to your connect args.

Exception: "ssl_mode is not supported"

**Versions.**
 - OS: ALL
 - Python: 2.7, 3.6(Looks like all of them)
 - SQLAlchemy: 1.3.3(But I checked other version they don't provide such functionality as well"
 - Database: MySQL
 - DBAPI: mysqldb

sqla-tester:
**Gord Thompson** has proposed a fix for this issue in the **master** branch:

**Added ssl_mode flag to mysqldb** https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2358

sqla-tester:
**Gord Thompson** has proposed a fix for this issue in the **rel_1_3** branch:

**Added ssl_mode flag to mysqldb** https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2359

zzzeek:
as mentioned above I thought a patch was needed due to the separate "ssl" dictionary but per https://dev.mysql.com/doc/c-api/8.0/en/mysql-ssl-set.html that's not where "ssl_mode" is.

Also it does not seem to be supported by the driver, so that means I need to revert this change because it is not properly documented.  latest mysqlclient:

```
>>> from sqlalchemy import create_engine
e >>> e = create_engine("mysql://scott:tiger@localhost/test?ssl_mode=required")
>>> c = e.connect()
Traceback (most recent call last):
  File "/home/classic/dev/sqlalchemy/lib/sqlalchemy/engine/base.py", line 2977, in _wrap_pool_connect
__init__
    super().__init__(*args, **kwargs2)
MySQLdb._exceptions.NotSupportedError: MySQL client library does not support ssl_mode specification

```


latest pymysql:

```
>>> from sqlalchemy import create_engine
>>> e = create_engine("mysql+pymysql://scott:tiger@localhost/test?ssl_mode=required")
>>> c = e.connect()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #5693, opened 2020-11-11
Added ssl_mode flag to mysqldb

thereisnosun:
ssl_mode flag is added to mysqldb

### Description
mysqldb driver supports "ssl_mode" flag, which can be one of following values:
"DISABLED", "PREFERRED", "REQUIRED", "VERIFY_CA", "VERIFY_IDENTITY".
Depending on these values MySQL will behave accordingly to them. 

So this flag has been added to the sqlaclhemy. So in case if TLS is not supported on the server, we can drop the connection right away.
Fixes: 
 #5692 Drop connection if ssl is not in use in sqlaclhemy

### DIFF
--- a/lib/sqlalchemy/dialects/mysql/mysqldb.py
+++ b/lib/sqlalchemy/dialects/mysql/mysqldb.py
@@ -189,6 +189,7 @@ class MySQLDialect_mysqldb(MySQLDialect):
         util.coerce_kw_type(opts, "write_timeout", int)
         util.coerce_kw_type(opts, "client_flag", int)
         util.coerce_kw_type(opts, "local_infile", int)
+        util.coerce_kw_type(opts, "ssl_mode", str)
         # Note: using either of the below will cause all strings to be
         # returned as Unicode, both in raw SQL operations and with column
         # types like String and MSString.
--- a/doc/build/changelog/unreleased_13/5692.rst
+++ b/doc/build/changelog/unreleased_13/5692.rst
@@ -1,0 +1,6 @@
+.. change::
+    :tags: mysql, usecase
+    :tickets: 5692
+
+    Added ``ssl_mode`` dialect keyword to MySQL dialect for use with mysqldb.
+    Pull request courtesy Teodor Moroz.
--- a/test/dialect/mysql/test_dialect.py
+++ b/test/dialect/mysql/test_dialect.py
@@ -141,6 +141,7 @@ class DialectTest(fixtures.TestBase):
             make_url(
                 "mysql://scott:tiger@localhost:3306/test"
                 "?ssl_ca=/ca.pem&ssl_cert=/cert.pem&ssl_key=/key.pem"
+                "&ssl_mode=REQUIRED"
             )
         )[1]
         # args that differ among mysqldb and oursql
@@ -156,6 +157,7 @@ class DialectTest(fixtures.TestBase):
                     "cert": "/cert.pem",
                     "key": "/key.pem",
                 },
+                "ssl_mode": "REQUIRED",
                 "host": "localhost",
                 "user": "scott",
                 "port": 3306,


zzzeek:
please add a test case to https://github.com/sqlalchemy/sqlalchemy/blob/master/test/dialect/mysql/test_dialect.py#L139

zzzeek (changes requested):
please add a test case to https://github.com/sqlalchemy/sqlalchemy/blob/master/test/dialect/mysql/test_dialect.py#L139

thereisnosun:
@zzzeek sorry, missed that. Will update the PR.

gordthompson:
@thereisnosun - I added the test and a changelog entry for you.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **gordthompson** to try to get revision 94aed8b17d21da9a20be4b092f6a60b12f60b761 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 94aed8b17d21da9a20be4b092f6a60b12f60b761: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2358

sqla-tester:
**Gord Thompson** (gordthompson) wrote:

Code-Review+1 Workflow+1

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2358

sqla-tester:
**Gord Thompson** (gordthompson) wrote:

Code-Review+1 Workflow+1

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2359

sqla-tester:

Code-Review+2 Workflow+2

A+ would buy again

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2358

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2358 has been **merged**. Congratulations! :)

lib/sqlalchemy/dialects/mysql/mysqldb.py
```diff
@@ -189,6 +189,7 @@ def create_connect_args(self, url):
         util.coerce_kw_type(opts, "write_timeout", int)
         util.coerce_kw_type(opts, "client_flag", int)
         util.coerce_kw_type(opts, "local_infile", int)
+        util.coerce_kw_type(opts, "ssl_mode", str)
```
zzzeek, line 192: I just merged this and I have a doubt..."ssl_mode" does not go into the ssl dictionary?    This would indicate there was no need to add this at all, as we are already transferring all of url.query into opts.  so there's nothing "wrong" with this patch but it's unnecessary.

zzzeek:
as mentioned above I thought a patch was needed due to the separate "ssl" dictionary but per https://dev.mysql.com/doc/c-api/8.0/en/mysql-ssl-set.html that's not where "ssl_mode" is.

Also it does not seem to be supported by the driver, so that means I need to revert this change because it is not properly documented.  latest mysqlclient:

```
>>> from sqlalchemy import create_engine
e >>> e = create_engine("mysql://scott:tiger@localhost/test?ssl_mode=required")
>>> c = e.connect()
  File "/home/classic/dev/sqlalchemy/lib/sqlalchemy/engine/base.py", line 2977, in _wrap_pool_connect
__init__
    super().__init__(*args, **kwargs2)
MySQLdb._exceptions.NotSupportedError: MySQL client library does not support ssl_mode specification

```


latest pymysql:

```
>>> from sqlalchemy import create_engine
>>> e = create_engine("mysql+pymysql://scott:tiger@localhost/test?ssl_mode=required")
>>> c = e.connect()
    return Connection(*args, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'ssl_mode'

```


sorry, I need to revert this and I need to know exactly what driver you are using as well as a case that illustrates the flag not working right now; any string you add to the query string is passed through in any case, the above behavior is the same without the patch as with it.

sqla-tester:

the master patch should not have been merged, my mistake.  need to start again.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2359

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2359 has been **abandoned**.  That means that at least for the moment I need to close this pull request. Sorry it didn't work out :(

lib/sqlalchemy/dialects/mysql/mysqldb.py
thereisnosun: Yeah, I intentionally did not put it into SSL dict, because all the values from that dict are then passed to the `mysql_ssl_set` method.  But this is a separate configurable option, which is used in `mysql_options` method with `MYSQL_OPT_SSL_MODE` key.

thereisnosun:
Hi @zzzeek and @sqla-tester , sorry for the late response. I am using `mysqlclient` connector with the default MySQL driver inside of `sqlalchemy`. The MySQL itself does support this option:
`https://dev.mysql.com/doc/c-api/8.0/en/mysql-options.html grep for MYSQL_OPT_SSL_MODE.`
I am also not sure about passing URL options to the options dict( at least in `1.3.3` version), because when I was doing that `sqlalchemy` was throwing an error, claiming that this option is not supported. I will double-check it, maybe in the latest `sqlalchemy` version, it is already working. So in order to push this PR, I guess, I would need to:
1)create more explanatory documentation
2) provide evidence that it is actually needed.
Is that right?

zzzeek:
> Hi @zzzeek and @sqla-tester , sorry for the late response. I am using `mysqlclient` connector with the default MySQL driver inside of `sqlalchemy`. The MySQL itself does support this option:
> `https://dev.mysql.com/doc/c-api/8.0/en/mysql-options.html grep for MYSQL_OPT_SSL_MODE.`
> I am also not sure about passing URL options to the options dict( at least in `1.3.3` version), because when I was doing that `sqlalchemy` was throwing an error, claiming that this option is not supported.

that is the driver throwing that error and it is evidence that the option is being passed along to the driver.  I mistakenly thougth this was because the option was placed incorrectly but this does not seem to be the case, as I've noted mysqlclient throws this error when the underlying SSL mechanisms don't support it.   so the message you got is evidence that the keyword *is* being passed through.

gordthompson:
@thereisnosun - Just to clarify, you may receive an error if your environment has `mysqlclient` that was installed

1. on Windows via the [wheel files](https://pypi.org/project/mysqlclient/#files) distributed by `pip`, or
2. on Linux where `mysqlclient` is built from source but the mysql[client] support files (e.g., `libmysqlclient-dev` on Ubuntu) are sufficiently old that they do not support the `ssl_mode` argument.

Also note that this has nothing to do with SQLAlchemy *per se*; it also happens when using `mysqlclient` directly. On Windows, this works

``` python
import MySQLdb
cnxn = MySQLdb.connect(
    host="192.168.0.199",
    user="scott",
    passwd="tiger",
    db="test",
)
```

but this fails

``` python
import MySQLdb
cnxn = MySQLdb.connect(
    host="192.168.0.199",
    user="scott",
    passwd="tiger",
    db="test",
    ssl_mode="DISABLED",
)
```

with the error

> MySQLdb._exceptions.NotSupportedError: MySQL client library does not support ssl_mode specification

To illustrate point (2), above, using `ssl_mode="DISABLED",` works for me with Xubuntu 18.04, but fails (same error) with CentOS 7.

schwabe:
I looks that older ssl libraries require `--ssl` and `--ssl-verify-server-cert` option instead.

The overall goal is to use TLS to a MySQL and fail if SSL is not available or the server certificate cannot be verified. Like any well configured TLS connection should be.

PR #5703, opened 2020-11-16
update contains_eager docs to reflect common gotcha

leontrolski:
### Description

`contains_eager` has behaviour that feels surprising to me - although I now see why the behaviour is as it is. Colleagues and "the internet" seemed to concur.

I hope the diff speaks for itself, this is an issue I've seen at least once in the github issues: https://github.com/sqlalchemy/sqlalchemy/issues/4021

And something I've reported [via stackoverflow](https://stackoverflow.com/questions/64823896/in-sqlalchemy-should-i-always-pair-contains-eager-with-expire-all).

### Checklist

This pull request is:

### Aside

It would be a big breaking change, but it feels like (due to the use of `contains_eager`) that `q.addresses` should return *a whole different list-like thing* to `user.addresses`. Probably one that highlights that it's not necessarily all of the `addresses` of the `user`, but just the ones from the query, and thus would eg. raise errors on `user.addresses.append(...)`.

### DIFF
--- a/doc/build/orm/loading_relationships.rst
+++ b/doc/build/orm/loading_relationships.rst
@@ -1135,20 +1135,20 @@ in fact associated with the collection.
     loaded.
 
     If we have a previous query in the session that has loaded the collection,
-    future queries will use the collection as is, it won't be reloaded
-    according to filter criteria. It may be necessary to expunge instances to
+    future queries will use the collection as is - it won't be reloaded
+    according to filter criteria. It may be necessary to use
+    :meth:`.Query.populate_existing` or :meth:`.Session.expunge` to
     get the desired behaviour::
 
         user = session.query(User)
         user.addresses
 
-        # if we don't include the following line, q.addresses will
-        # include all of the addresses - not just those we filter
-        session.expunge(user)
-
+        # if we don't include `.populate_existing()`, q.addresses will
+        # include all of the addresses, not just those we filter
         q = session.query(User).join(User.addresses).\
                     filter(Address.email.like('%ed%')).\
-                    options(contains_eager(User.addresses))
+                    options(contains_eager(User.addresses)).\
+                    populate_existing()
 
     In addition, the **collection will fully reload normally** once the
     object or attribute is expired.  This expiration occurs whenever the
--- a/doc/build/orm/loading_relationships.rst
+++ b/doc/build/orm/loading_relationships.rst
@@ -1134,6 +1134,22 @@ in fact associated with the collection.
     conflicting with entries that are already in the database but not locally
     loaded.
 
+    If we have a previous query in the session that has loaded the collection,
+    future queries will use the collection as is, it won't be reloaded
+    according to filter criteria. It may be necessary to expunge instances to
+    get the desired behaviour::
+
+        user = session.query(User)
+        user.addresses
+
+        # if we don't include the following line, q.addresses will
+        # include all of the addresses - not just those we filter
+        session.expunge(user)
+
+        q = session.query(User).join(User.addresses).\
+                    filter(Address.email.like('%ed%')).\
+                    options(contains_eager(User.addresses))
+
     In addition, the **collection will fully reload normally** once the
     object or attribute is expired.  This expiration occurs whenever the
     :meth:`.Session.commit`, :meth:`.Session.rollback` methods are used


leontrolski:
I notice from the linked issue, that putting `.populate_existing()` is an option, is that preferable?

CaselIT:
> 
> 
> I notice from the linked issue, that putting `.populate_existing()` is an option, is that preferable?

Yes, I think that's the preferred option. The other is to expire the object (or just the collection). I don't think expunging is the most appropriate option here.

zzzeek:
So to correctly document this it would be easier if I just did it, I'm surprised this is not already mentioned.  There's an existing warning box at https://docs.sqlalchemy.org/en/14/orm/loading_relationships.html?highlight=contains_eager#using-contains-eager-to-load-a-custom-filtered-collection-result that  is where this should be mentioned, so that it is noticeable.   

other points to mention:

1. in 1.4 it's now execution_options(populate_existing=True), documented at https://docs.sqlalchemy.org/en/14/orm/queryguide.html#orm-queryguide-populate-existing.  this does mention contains_eager()

2. there is precedent for documenting this as it also applies to the new loader criteria features, also listed in the above section  https://docs.sqlalchemy.org/en/14/orm/loading_relationships.html?highlight=contains_eager#adding-criteria-to-loader-options  (where that doc also has to be updated to refer to the new execution_option version)

3. >It would be a big breaking change, but it feels like (due to the use of contains_eager) that q.addresses should return a whole different list-like thing to user.addresses. Probably one that highlights that it's not necessarily all of the addresses of the user, but just the ones from the query, and thus would eg. raise errors on user.addresses.append(...)

that would be extremely complicated to implement and support.   the most straightforward way would be to use events that raise on mutation, that somehow have to clear themselves when the collection is reset.  it would have to be a flag that applies to any collection where any of the above mentioned features for altering the results of collections have been used and it likely would be an integral part of the attribute system rather than using a separate event hook.   there's be a whole new class of issues like the flag being set / not being set, unexpected mutations from things like backrefs, edge cases where mutating the collection is actually feasible, etc.   the breaking change part of it would be the simplest aspect, we just emit a warning.

zzzeek:
alternatively we could simple require that any alternate criteria requires that the relationship is viewonly=True.  but detecting alternate criteria is tricky too.

zzzeek:
note that the primary purpose of contains_eager() is *not* to change what's in the collection, only to load more efficiently when joins are already present.   Having collections being loaded with arbitrary subsets of what the relationship defines seems to be popular hence we added the feature at https://docs.sqlalchemy.org/en/14/orm/loading_relationships.html#adding-criteria-to-loader-options but I think it's best that this be against collections that are dedicated as "special loading" and marked viewonly=True.

leontrolski:
Ah, I wasn't aware of "Adding Criteria to loader options" - it seems like a more intuitive way of doing the filtering anyway. (And usefully includes the warning around `.populate_existing()`). 

Maybe the fix here is to update the `contains_eager` docs to say "you probably want to [add criteria](https://docs.sqlalchemy.org/en/14/orm/loading_relationships.html#adding-criteria-to-loader-options) to your loader options instead". Or to duplicate the `.populate_existing()` warning, either seems useful!

zzzeek:
yes at this point because we have that new feature, we can advertise it as such.   but also it would be nice to get a warning about populate_existing() back into 1.3 also since those docs are going to be relevant for many months (And in those, it's still the populate_existing() method).   

so it's kind of that a wholeistic approach has to be taken here and I don't want to put that on a lone PR submitter unless you were really motivated to get into docs.

zzzeek:
the docs in 1.4 are a little more challenging to work on because of the giant split between the "old" and "new" way of doing things.

leontrolski:
Up to you, I'm happy to been shown how to contribute and put in some time (I only got as far as running `make html`), or y'all can do it at your leisure - your call.

zzzeek:
sure!    I think we should start with the 1.4 master branch and update the verbiage at https://docs.sqlalchemy.org/en/14/orm/loading_relationships.html?highlight=contains_eager#using-contains-eager-to-load-a-custom-filtered-collection-result and then also highlight the new loader criteria stuff as another way to go about doing it.    also we should refer to the new https://docs.sqlalchemy.org/en/14/orm/queryguide.html#orm-queryguide-populate-existing section when we talk about populate existing.

zzzeek:
hey there -

saw your note and it looks like I did a rework of contains_eager() to discuss populate_existing in 0d76857ee1022f41baffff4091be7 as part of #5740 .  When that issue came in, as is often the case I had a sense that "we just had this?"  but apparently I was thinking of this PR which got stalled.   

anyway take a review of what I did at 0d76857ee1022f41baffff4091be7 which I'm hoping should clear up any ambiguity here.   closing this PR.  thanks for bringing it to my attention however as it did lend urgency to #5740 even though I couldn't remember why :).

PR #5705, opened 2020-11-17
pool/imply.py: TimeoutError: better format msg

JordanP:
### Description
It's not obvious but the `timeout` argument of a Pool can be a float, because `time.time()` that is used to compute a deadline has sub-second precision. So the `TimeoutError` exception message should take into account this.

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/pool/impl.py
+++ b/lib/sqlalchemy/pool/impl.py
@@ -129,7 +129,7 @@ class QueuePool(Pool):
             else:
                 raise exc.TimeoutError(
                     "QueuePool limit of size %d overflow %d reached, "
-                    "connection timed out, timeout %d"
+                    "connection timed out, timeout %.2f"
                     % (self.size(), self.overflow(), self._timeout),
                     code="3o7r",
                 )


zzzeek:
the timeout situation is not as we'd like it to be due to the poor state of Python timer functions (none of them provide unambiguous accuracy guarantees on all platforms), and the assumption right now is that the timeout *is* an integer.   if we'd like this to support float that has to be documented and tested, such as here: https://docs.sqlalchemy.org/en/14/core/pooling.html?highlight=pool%20timeout#sqlalchemy.pool.QueuePool.params.timeout

zzzeek:
these are very old tests and are not well composed by our current standards, but in any case there would need to be some tests that assert a timeout such as .5 seconds actually works on supported platforms.  we'd have to list out those platforms too in documentation which is challenging because it varies by platform and what function you're using.     

https://github.com/sqlalchemy/sqlalchemy/blob/master/test/engine/test_pool.py#L853

JordanP:
That's valid points. It's more than I can chew right now. I also want to bring https://github.com/sqlalchemy/sqlalchemy/blob/master/test/ext/asyncio/test_engine_py3k.py#L270 to your attention.

About the platform parts: deadline are computed by adding `timeout` to `time.time()`, isn't `time.time()` "guaranteed" to return a float on all platforms ?

zzzeek:
> That's valid points. It's more than I can chew right now. I also want to bring https://github.com/sqlalchemy/sqlalchemy/blob/master/test/ext/asyncio/test_engine_py3k.py#L270 to your attention.
> 
> About the platform parts: deadline are computed by adding `timeout` to `time.time()`, isn't `time.time()` "guaranteed" to return a float on all platforms ?

they return floats but they have varying degrees of accuracy.   on windows the timer tends to be +- 16 milliseconds, but this is not defined, we all just "know" it.     I was recently frustrated that Python 3 has all new time functions and a pep: https://www.python.org/dev/peps/pep-0418/  yet *none* of these new functions document their guaranteed accuracy or give unambiguous guidance over the span of time they work for. i recently tried to use some of the new timer functions and it was a dismal failure, and we are back to time.time().   

beyond that, I can't tell you with 100% confidence that all the math we are doing with time.time() isn't rounding to an integer, so without test cases that assert we definitely are handling accuracy to say, .25 seconds, it is not yet a defined behavior.

if you could enlighten me on what the purpose of the connection pool timeout using sub-second accuracy is, that would be helpful as well.   if you wanted immediate fail on pool full, set the timeout to zero.

JordanP:
I'll see what I can do for a unittest, I am bit rusty in Python but I may be able to do something (I know this may not be enough to start considering merging this, but just as an exercise).

Our use case: small PG server with small max_connections (100). So we have small values for `pool_size` (7) and `overflow` (3). We do only "select" queries, by primary key, so queries are really fast (~5ms including network round trip). If we can't borrow from the pool, we can wait a bit (500ms) but we don't want to block more otherwise that will impact the P99 response time a lot. We can proceed/continue even without the result of these SQL queries, the logic would be different, but we handle it fine.

zzzeek:
so I think you can use the .5 ms timeout as you are for now, and if you'd like you can add a new issue "document and test subsecond pool timeout resolution".   we can say we will test down to .25 seconds for this for example.   then we'll document "this can be a float but is subject to the limitations of Python time functions which may not be reliable in the tens of milliseconds".    we can add a few tests to make sure it stays supported and that will be that.

PR #5803, opened 2020-12-29
Remove wrong default

Dreamsorcerer:
Not sure which is correct (10 or 5), but best to only mention it once to avoid this kind of mistake.


zzzeek:
the "five" refers to the pool_size setting.   We can adjust the verbiage here so that this is clearer perhaps?

Dreamsorcerer:
Ah, I see. Maybe I just read it too quickly.

Issue #5052, opened 2019-12-25
implement per-dialect datetime representations, not just strsqlcompiler


pasenor:
When I try to compile a query with datetime parameters using `literal_binds`, e.g.:

    from sqlalchemy import create_engine
    from sqlalchemy.ext.declarative import declarative_base
    from sqlalchemy import Column, DateTime
    from sqlalchemy.orm import sessionmaker
    from datetime import datetime

    engine = create_engine('sqlite:///:memory:', echo=True)
    Base = declarative_base()
    class M(Base):
        __tablename__ = 'M'
        d = Column(DateTime, primary_key=True)

    Session = sessionmaker()

    print(Session().query(M).filter(
        M.d < datetime.now()
    ).statement.compile(
        compile_kwargs={'literal_binds': True}
    ))

the engine gives me the following error

    NotImplementedError: Don't know how to literal-quote value datetime.datetime(2019, 12, 25, 14, 26, 42, 756674)

There probably should be a default representation for datetimes as an iso-formatted string in single quotes.

zzzeek:
hi there -

why aren't you passing the engine to compile?  The SQLite dialect necessarily has a datetime formatter of its own as there is no datetime support in SQlite directly.

passing engine to compile allows it to generate this form.   for StrSQLCompiler, which is not meant to be used to actually emit SQL, there is apparently not an issue for this but I would want it to generate a string using str() for any kind of object and then applies safestring quoting to it in all cases.

pasenor:
Yes, it works for SQLite, thanks. But my actual problem was with mysql: I wanted to debug an application and print out a compiled query with datetime parameters. So the example here should have been

```
from sqlalchemy import create_engine
from sqlalchemy.ext.declarative import declarative_base
from sqlalchemy import Column, DateTime
from sqlalchemy.dialects import mysql
from sqlalchemy.orm import sessionmaker
from datetime import datetime

engine = create_engine('mysql+pymysql://root:root@127.0.0.1:3306/test?charset=utf8', echo=True)
Base = declarative_base()
class M(Base):
    __tablename__ = 'M'
    d = Column(DateTime, primary_key=True)

Session = sessionmaker()

print(Session().query(M).filter(
    M.d < datetime.now()
).statement.compile(
    bind=engine,
    dialect=mysql.dialect(),
    compile_kwargs={'literal_binds': True}
))
```
[...]

PR #5869, opened 2021-01-24
Literal compile DateTime objects for PG dialect

hyperknot:
### Description
PG compiler can now literal compile datetime objects

This PR adds string handling for PG dialect. Continuation for #5052

### Checklist

Need to add test.

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -1847,6 +1847,9 @@ colspecs = {
     sqltypes.Enum: ENUM,
     sqltypes.JSON.JSONPathType: _json.JSONPathType,
     sqltypes.JSON: _json.JSON,
+    sqltypes.DateTime: default._StrDateTime,
+    sqltypes.Date: default._StrDate,
+    sqltypes.Time: default._StrTime,
 }
 
 ischema_names = {


hyperknot:
No idea how to modify this test: https://github.com/sqlalchemy/sqlalchemy/blob/2df6eb140b4feb86d7cc99c0c976d1b22eb6e25b/test/sql/test_types.py#L3751-L3771

From my understanding what we need is to have it compile on standard and PG dialect, but raise on every other dialects.

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -1847,6 +1847,9 @@ def _on_metadata_drop(self, target, bind, checkfirst=False, **kw):
     sqltypes.Enum: ENUM,
     sqltypes.JSON.JSONPathType: _json.JSONPathType,
     sqltypes.JSON: _json.JSON,
+    sqltypes.DateTime: default._StrDateTime,
+    sqltypes.Date: default._StrDate,
+    sqltypes.Time: default._StrTime,
```
CaselIT, line 1852: I don't think we would want to use those subtipes these, since they transform to string using `str` that may not be understood by pg. I think it should define new subtype that render using the iso format

CaselIT:
> 
> 
> No idea how to modify this test:
> 
> https://github.com/sqlalchemy/sqlalchemy/blob/2df6eb140b4feb86d7cc99c0c976d1b22eb6e25b/test/sql/test_types.py#L3751-L3771
> 
> From my understanding what we need is to have it compile on standard and PG dialect, but raise on every other dialect.

You would just modify the requirements indicating the pg does in fact supports datetime literals.
It's defined here https://github.com/sqlalchemy/sqlalchemy/blob/2df6eb140b4feb86d7cc99c0c976d1b22eb6e25b/test/requirements.py#L1056-L1062

lib/sqlalchemy/dialects/postgresql/base.py
hyperknot: Why are we using str for the default engine? AFAIK str() is Python specific, while ISO format is understood by most DBs. 

I mean I think it'd be better to change default.py to use isoformat().

hyperknot: I checked SQLite's reference page and it supports ISO: https://sqlite.org/lang_datefunc.html

CaselIT: > Why are we using str for the default engine

We are not. The default has no implementation. That _StrXX class with the implementation is used only by the string compiler, that is a mock compiler used by default when `str(<compiled>)` is called. It's not supposed to produce sql that is compatible with any particular db.

> I checked SQLite's reference page and it supports ISO: https://sqlite.org/lang_datefunc.html

Note that sqlite does not use that implementation, but has its own custom one:
https://github.com/sqlalchemy/sqlalchemy/blob/2df6eb140b4feb86d7cc99c0c976d1b22eb6e25b/lib/sqlalchemy/dialects/sqlite/base.py#L904-L910

I think that something similar should probably be added for the pg dialect

CaselIT: Also see this comment regarding the reason for not having a default implementation https://github.com/sqlalchemy/sqlalchemy/issues/5052#issuecomment-569091948

hyperknot: I don't quite understand what is the point of the string compiler. Who is using it and for what? Isn't this all used just for debugging purposes, when they DB is known?

Also, why can we not use ISO for "almost-all" dialects? We know that PG, MySQL and SQLite definitely supports it.

CaselIT: >  Isn't this all used just for debugging purposes, when the dialect is known?

That is its purpose yes. it gets called when `str(stmt)` or `stmt.compile()` or  `stmt.compile(compile_kwargs={'literal_binds': True})` is used without specifying a dialect.

> Also, why can we not use ISO for "almost-all" dialects? We know that PG, MySQL and SQLite definitely supports it.

My objection is purely on using the defined _StrXX classes, that are defined only for the string compiler. We could have a mixin called someting like "IsoformatMixin" that can be used to define subtypes for each dialect.

zzzeek (changes requested):
As mentioned in my comment, we absolutely need to be extremely careful in how this is implemented and tested, including:

* Remove all of _RendersLiteral , _StrDateTime, _StrDate, _StrTime, move literal_processor to the DateTime, Date, Time types in sqltypes.py
* The literal processors should use date(time).isoformat() and/or strftime() in all cases and raise TypeError if problems (see demo below)
* open up [datetime_literals](https://github.com/sqlalchemy/sqlalchemy/blob/master/test/requirements.py#L1056) for newly added backends
* add tests as illustrated in patch below
* MySQL, SQL Server, Oracle should be included, not just PostgreSQL for this patch to be worth it

This patch illustrates how to do the TypeError and how to extend the suite:

```diff
diff --git a/lib/sqlalchemy/engine/default.py b/lib/sqlalchemy/engine/default.py
index e5a6384d2..52692ba77 100644
--- a/lib/sqlalchemy/engine/default.py
+++ b/lib/sqlalchemy/engine/default.py
@@ -754,8 +754,13 @@ class _RendersLiteral(object):
 
 
 class _StrDateTime(_RendersLiteral, sqltypes.DateTime):
-    pass
-
+    def literal_processor(self, dialect):
+        def process(value):
+            try:
+                return value.isoformat()
+            except AttributeError as err:
+                util.raise_(TypeError("Object is not a datetime: %r" % value), from_=err)
+        return process
 
 class _StrDate(_RendersLiteral, sqltypes.Date):
     pass
diff --git a/lib/sqlalchemy/testing/suite/test_types.py b/lib/sqlalchemy/testing/suite/test_types.py
index ebcceaae7..eb474f29f 100644
--- a/lib/sqlalchemy/testing/suite/test_types.py
+++ b/lib/sqlalchemy/testing/suite/test_types.py
@@ -9,7 +9,7 @@ from .. import config
 from .. import engines
 from .. import fixtures
 from .. import mock
-from ..assertions import eq_
+from ..assertions import eq_, expect_raises
 from ..assertions import is_
 from ..config import requirements
 from ..schema import Column
@@ -63,7 +63,7 @@ class _LiteralRoundTripFixture(object):
             for value in input_:
                 ins = (
                     t.insert()
-                    .values(x=literal(value))
+                    .values(x=literal(value, type_=type_))
                     .compile(
                         dialect=testing.db.dialect,
                         compile_kwargs=dict(literal_binds=True),
@@ -72,7 +72,7 @@ class _LiteralRoundTripFixture(object):
                 connection.execute(ins)
 
             if self.supports_whereclause:
-                stmt = t.select().where(t.c.x == literal(value))
+                stmt = t.select().where(t.c.x == literal(value, type_=type_))
             else:
                 stmt = t.select()
 
@@ -88,6 +88,34 @@ class _LiteralRoundTripFixture(object):
 
         return run
 
+    @testing.fixture
+    def expect_type_error(self, metadata, connection):
+        def run(type_, input_, output, filter_=None):
+            t = Table("t", metadata, Column("x", type_))
+            t.create(connection)
+
+            for value in input_:
+                with expect_raises(TypeError):
+                    (
+                        t.insert()
+                        .values(x=literal(value, type_=type_))
+                        .compile(
+                            dialect=testing.db.dialect,
+                            compile_kwargs=dict(literal_binds=True),
+                        )
+                    )
+
+            if self.supports_whereclause:
+                stmt = t.select().where(t.c.x == literal(value, type_=type_))
+
+                with expect_raises(TypeError):
+                    stmt.compile(
+                        dialect=testing.db.dialect,
+                        compile_kwargs=dict(literal_binds=True),
+                    )
+
+        return run
+
 
 class _UnicodeFixture(_LiteralRoundTripFixture, fixtures.TestBase):
     __requires__ = ("unicode_data",)
@@ -333,6 +361,11 @@ class _DateFixture(_LiteralRoundTripFixture, fixtures.TestBase):
         row = connection.execute(select(date_table.c.date_data)).first()
         eq_(row, (None,))
 
+    @testing.requires.datetime_literals
+    def test_literal_rejects_strings(self, expect_type_error):
+        data = str(self.data)
+        expect_type_error(self.datatype, [data], None)
+
     @testing.requires.datetime_literals
     def test_literal(self, literal_round_trip):
         compare = self.compare or self.data
diff --git a/test/requirements.py b/test/requirements.py
index 3c9b39ac7..76b1f288d 100644
--- a/test/requirements.py
+++ b/test/requirements.py
@@ -1059,7 +1059,7 @@ class DefaultRequirements(SuiteRequirements):
 
         """
 
-        return fails_on_everything_except("sqlite")
+        return fails_on_everything_except("sqlite", "postgresql")
 
     @property
     def datetime(self):
```

hyperknot:
OK, I believe this needs to be done by a core maintainer, definitely not by me. I just wanted to cover the very common use case of "printing datetime" problems in SQLAlchemy. I totally agree that it should be isoformat() all the way, though, it seems much better than using str() for anything.

zzzeek:
sure, it's not hard at this point I think just using .isoformat() will cover it, can pull that patch in but I'm only adding it for 1.4.

PR #5943, opened 2021-02-15
Bump Sphinx

### LINKED ISSUE #5790: [docs.sqlalchemy.org] Search keeps crashing my browser
When I use the search functionality on docs.sqlalchemy.org it typically starts displaying a huge number of results which in only a couple of seconds results in using GBs of RAM, which then results in my browser crashing when the system runs out of memory.

I guess an obvious solution would be to implement pagination, rather than loading unlimited amount of results in the page.

Firefox on Linux.

Dreamsorcerer:
Hopefully fixes #5790

### DIFF
--- a/doc/build/conf.py
+++ b/doc/build/conf.py
@@ -26,7 +26,7 @@ sys.path.insert(0, os.path.abspath("."))
 # -- General configuration --------------------------------------------------
 
 # If your documentation needs a minimal Sphinx version, state it here.
-needs_sphinx = "1.6.0"
+needs_sphinx = "3.5.0"
 
 # Add any Sphinx extension module names here, as strings. They can be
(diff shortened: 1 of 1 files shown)


zzzeek:
im tying to run sphinx 1.5 on my server and getting very strange results, will have to spend some time understanding what's going on.  nonetheless the server upgrades sphinx automatically so it was already on 1.5 (and failing).

Dreamsorcerer:
OK, I guess this PR isn't needed then.

zzzeek:
we can still merge it i just need to get the build to work on the live server, it is doing something very strange at the moment.

zzzeek:
we are blocked by https://github.com/sphinx-doc/sphinx/issues/8894

zzzeek:
thanks for the PR!

PR #5972, opened 2021-02-24
ORM tutorial: As of 1.1, there are five object states

jwhitlock:
### Description
Update the ORM tutorial to refer to three of five rather than three of four object states. The ``deleted`` state was added in version 1.1.

### Checklist

This pull request is:

- A documentation / typographical error fix


zzzeek:
nice catch !

PR #6587, opened 2021-06-06
Fix pysqlcipher and issue 6586

blackm0re:
Use make_url in order to create URL object from the 'str' as the on_connect_url method expects.


zzzeek:
hi , I have a fix in https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2859.   on_connect_url [expects a URL object](https://docs.sqlalchemy.org/en/14/core/internals.html?highlight=on_connect_url#sqlalchemy.engine.default.DefaultDialect.on_connect_url.params.url) and the bug is in create.py.     thanks for reporting!

Issue #6589, opened 2021-06-07
pysqlalchemy dialect does not send the pragma params specified in the connection URL


blackm0re:
In SQLAlchemy 1.3.(24) the pragma params specified in the connection URL are not sent to the backend, hence the default values are used no matter what params are specified in the URL.

example URL: sqlite+pysqlcipher://:{passwd}@/{file}?cipher=aes-256-cfb&kdf_iter=64000

In addition to not respecting the user's security preferences, the fact that this problem does not exist in SQLAlchemy >= 1.4,
confuses the user when migrating SQLAlchemy 1.3 -> SQLAlchemy 1.4.

A small bugfix (PR) will be linked to this issue.

zzzeek:
while this is a bug, im not sure if changing it in 1.3 is going to solve anything, because it creates a backwards compatible change in a very late-stage 1.3 release.   typically when going from 1.3->1.4 is where people expect a more major behavioral change.   what's the specific scenario where putting out a release 1.3.25 helps here, for the user that's been on 1.3.x for a long time?  for the user that hasn't been on 1.3.x, they'd start with 1.4.x?

zzzeek:
that is, "confuses the user when going from 1.3 to 1.4", wouldnt this confuse the user even moreso when going from 1.3.24 -> 1.3.25?  it's much less appropriate for surprises there.    i think it's better to leave this behavior in place wtihin 1.3.x.

blackm0re:
They way I see it:
Doing nothing:
A user initiated a new database sqlite+pysqlcipher database in <= 1.3.24 and stored some data in it. The user upgrades to 1.3.25 and everything is "working" as before, although the expected security settings are not set. The user upgrades to 1.4 and 1.4 is giving "file is not a database" error when of course using the same params and password. If the bug report is not public (searchable) most users will simply fall back and "blame" 1.4
A user initiated a new database in 1.3.25 -> same problem.

Fixing in 1.3.25:
A user initiated a new database sqlite+pysqlcipher database in <= 1.3.24 and stored some data in it. The user upgrades to 1.3.25 and is now unable to unlock his/her database. The user assumes that some bug has been introduced, searches and finds this particular bug report / changelog / commit message. The same goes for 1.4. A user initiating his/her database >=1.3.25 on the other hand will have no trouble migrating to 1.4. In addition no security promises will be broken.

In any of the cases the DB settings can be changed manually as described bellow and the user should be made aware of that:
[...]

PR #6590, opened 2021-06-07
Send the pragma params specified in the connection URL to the pysqlcipher backend

blackm0re:
resolves issue #6589


CaselIT:
I think we opted for updating the documentation for 1.3 to avoid braking existing usage. Thanks for the contribution in any case!

Issue #6649, opened 2021-06-17
PG / Oracle  - 'INTERVAL' object has no attribute 'impl'


MajorDallas:
**Describe the bug**

Filtering a `query` (with eg. `filter_by()`) on an `interval` column in Postgres raises `AttributeError: 'INTERVAL' object has no attribute 'impl'`.

**To Reproduce**

```sql
CREATE TABLE "test" (
    "uid" SERIAL PRIMARY KEY NOT NULL,
    "interval_col" interval 
);

INSERT INTO test (interval_col) VALUES (interval '1 day');
```

```py
from datetime import timedelta

from sqlalchemy import MetaData, create_engine
from sqlalchemy.ext.declarative import declarative_base
from sqlalchemy.orm import sessionmaker

#engine = create_engine(<engine params>)
session = sessionmaker(bind=engine)()
metadata = MetaData(bind=engine)
metadata.reflect(engine)

class Test(declarative_base()):
    __table__ = Table(
        "test", metadata, autoload=True, autoload_with=engine
    )
    uid: int
    interval_col: timedelta

if __name__ == "__main__":
    t = session.query(Test).filter_by(interval_col=timedelta(days=1)).all()
```

**Error**

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-14-e826b49d436f> in <module>
----> 1 t = session.query(Test).filter_by(interval_col=timedelta(days=1)).all()

/usr/local/lib/python3.8/site-packages/sqlalchemy/orm/query.py in filter_by(self, **kwargs)
   1891         """
   1892
-> 1893         clauses = [
   1894             _entity_descriptor(self._joinpoint_zero(), key) == value
   1895             for key, value in kwargs.items()

/usr/local/lib/python3.8/site-packages/sqlalchemy/orm/query.py in <listcomp>(.0)
   1892
   1893         clauses = [
-> 1894             _entity_descriptor(self._joinpoint_zero(), key) == value
   1895             for key, value in kwargs.items()
   1896         ]

/usr/local/lib/python3.8/site-packages/sqlalchemy/sql/operators.py in __eq__(self,other)
    363
    364         """
--> 365         return self.operate(eq, other)
    366
    367     def __ne__(self, other):

/usr/local/lib/python3.8/site-packages/sqlalchemy/orm/attributes.py in operate(self, op, *other, **kwargs)
    217
    218     def operate(self, op, *other, **kwargs):
[...]

PR #6650, opened 2021-06-17
Add impl property to Postgres INTERVAL class

MajorDallas:
### Description

The missing `.impl` attribute causes any kind of filtering on an `interval` column to fail.

`.impl` is called in `sql.sqltypes._AbstractInterval.coerce_compared_value`, and when the `ColumnElement` is `sql.sqltypes.Interval` this will point to `sql.sqltypes.DateTime`. However, `dialects.postgres.base.INTERVAL` lacks any `.impl` class attribute.

A `.impl` property on `dialects.postgres.base.INTERVAL` is used to create a path, using `_type_affinity()`, to `DateTime` so that `INTERVAL` can be used as a filter criterion.

The fix can be tested with the same procedure as in #6649.

An alternative implementation might be to make `.impl` an attribute of `_AbstractInterval`, assuming this would not create issues in other child classes.

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/oracle/base.py
+++ b/lib/sqlalchemy/dialects/oracle/base.py
@@ -702,6 +702,9 @@ class INTERVAL(sqltypes.NativeForEmulated, sqltypes._AbstractInterval):
             day_precision=self.day_precision,
         )
 
+    def coerce_compared_value(self, op, value):
+        return self
+
 
 class ROWID(sqltypes.TypeEngine):
     """Oracle ROWID type.
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -1655,6 +1655,9 @@ class INTERVAL(sqltypes.NativeForEmulated, sqltypes._AbstractInterval):
     def python_type(self):
         return dt.timedelta
 
+    @property
+    def impl(self):
+        return self._type_affinity().impl
 
 PGInterval = INTERVAL
 
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -1655,9 +1655,8 @@ class INTERVAL(sqltypes.NativeForEmulated, sqltypes._AbstractInterval):
     def python_type(self):
         return dt.timedelta
 
-    @property
-    def impl(self):
-        return self._type_affinity().impl
+    def coerce_compared_value(self, op, value):
+        return self
 
 PGInterval = INTERVAL
 
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -1658,6 +1658,7 @@ class INTERVAL(sqltypes.NativeForEmulated, sqltypes._AbstractInterval):
     def coerce_compared_value(self, op, value):
         return self
 
+
 PGInterval = INTERVAL
 
 
--- a/test/dialect/oracle/test_types.py
+++ b/test/dialect/oracle/test_types.py
@@ -183,6 +183,10 @@ class DialectTypesTest(fixtures.TestBase, AssertsCompiledSQL):
     def test_interval(self, type_, expected):
         self.assert_compile(type_, expected)
 
+    def test_interval_coercion_literal(self):
+        expr = column("bar", oracle.INTERVAL) == datetime.timedelta(days=1)
+        eq_(expr.right.type._type_affinity, sqltypes.Interval)
+
 
 class TypesTest(fixtures.TestBase):
     __only_on__ = "oracle"
--- a/test/dialect/postgresql/test_types.py
+++ b/test/dialect/postgresql/test_types.py
@@ -2427,6 +2427,10 @@ class TimestampTest(fixtures.TestBase, AssertsExecutionResults):
         eq_(expr.type._type_affinity, types.Interval)
         assert isinstance(expr.type, postgresql.INTERVAL)
 
+    def test_interval_coercion_literal(self):
+        expr = column("bar", postgresql.INTERVAL) == datetime.timedelta(days=1)
+        eq_(expr.right.type._type_affinity, types.Interval)
+
 
 class SpecialTypesCompileTest(fixtures.TestBase, AssertsCompiledSQL):
     __dialect__ = "postgresql"


zzzeek:
hey there -

I can do this change, which involves a change to more than one dialect and it also needs tests.  the main part of the change looks like:

```diff
diff --git a/lib/sqlalchemy/dialects/oracle/base.py b/lib/sqlalchemy/dialects/oracle/base.py
index 57f55c1824..a72088a461 100644
--- a/lib/sqlalchemy/dialects/oracle/base.py
+++ b/lib/sqlalchemy/dialects/oracle/base.py
@@ -702,6 +702,9 @@ class INTERVAL(sqltypes.NativeForEmulated, sqltypes._AbstractInterval):
             day_precision=self.day_precision,
         )
 
+    def coerce_compared_value(self, op, value):
+        return self
+
 
 class ROWID(sqltypes.TypeEngine):
     """Oracle ROWID type.
diff --git a/lib/sqlalchemy/dialects/postgresql/base.py b/lib/sqlalchemy/dialects/postgresql/base.py
index deb6243296..cb7434d0b3 100644
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -1655,6 +1655,8 @@ class INTERVAL(sqltypes.NativeForEmulated, sqltypes._AbstractInterval):
     def python_type(self):
         return dt.timedelta
 
+    def coerce_compared_value(self, op, value):
+        return self
 
 PGInterval = INTERVAL
```

I would need to check other dialects for INTERVAL but I think PG/oracle are the only two.    If you are interested in working on this further we can go step by step.

MajorDallas:
Certainly!

I have two questions right off the bat:

1. You say "the main part," implying there will be other parts of the library in need of changes. This is my first time poking at sqlalchemy's internals, so I'm wondering where else to look.
2. I can think of a couple reasons that overriding `coerce_compared_value` as you do is more optimal. In fact, my _first_ attempt was also to override that method with `return super(sqltypes._AbstractInterval, self).coerce_compared_value(op, value)`. That worked, but seemed overly complicated--especially now 😄 .
Even when I went with my submitted approach, I was aware that going through `DateTime` may have unintended consequences wrt the intended differences between the native and emulated classes. I'm wondering if you spotted such a potential issue?

zzzeek:
> Certainly!
> 
> I have two questions right off the bat:
> 
>     1. You say "the main part," implying there will be other parts of the library in need of changes. This is my first time poking at sqlalchemy's internals, so I'm wondering where else to look.

the main part is the two code changes above, the other parts are the tests and the changelog file.

> 
>     2. I can think of a couple reasons that overriding `coerce_compared_value` as you do is more optimal. In fact, my _first_ attempt was also to override that method with `return super(sqltypes._AbstractInterval, self).coerce_compared_value(op, value)`. That worked, but seemed overly complicated--especially now smile .

coerce_compared_value() is called when an expression is produced to give the type object on the left hand side a chance to tell SQLAlchemy more context about what kind of datatype should be applied to the object on the right hand side, typically for boolean comparison operations (there's another hook that is used for other operators that are not boolean).    It is mostly used for custom TypeDecorator classes that might want to do something like treat a right hand side object as something different than what the left hand side is.  since INTERVAL is not a type decorator, this should just return "self", because "<interval> <some comparison op> <something else>" will generally assume "<something else>" should also be treated as INTERVAL.    

>        Even when I went with my submitted approach, I was aware that going through `DateTime` may have unintended consequences wrt the intended differences between the native and emulated classes. I'm wondering if you spotted such a potential issue?

not yet, the workings of INTERVAL in terms of its relationship to Interval / TypeDecorator / DateTime are a very thorny part of the type system.  no other type has this setup so usually we just get the code running then run the test suite.

MajorDallas:
Makes sense, thanks for the explanation.

I've just committed the changes.

zzzeek:
are you up for working on tests?  we can do that if not.

MajorDallas:
I am, but I'll need some guidance as I've somehow managed to get away with not writing tests in my own code for way too long 😓 If there's a documented protocol you prefer, I can probably figure it out.

zzzeek:
> I am, but I'll need some guidance as I've somehow managed to get away with not writing tests in my own code for way too long sweat If there's a documented protocol you prefer, I can probably figure it out.
> 
> Edit: I just found https://github.com/zzzeek/sqlalchemy/blob/master/README.unittests.rst. I'll start with that, but do let me know if there's another doc I should be familiar with.

OK the tests for this thing are going to be right around here:  https://github.com/sqlalchemy/sqlalchemy/blob/master/test/dialect/postgresql/test_types.py#L2420    There's even already a test called "test_interval_coercion" there already, which apparently is failing to test the bug we are fixing here!  looks like it is ensuring the type of the expression overall, not the right side.  The PG types tests look to be kind of disorganized right now.   But basically we can add another test, "test_interval_coercion_literal" here where you pretty much, well OK it's just easier to show it:

    def test_interval_coercion_literal(self):
        from datetime import timedelta
        expr = column("bar", postgresql.INTERVAL) == timedelta(days=1)
        eq_(expr.right.type._type_affinity, types.Interval)

there's other things about this test suite that are bad, which are on my end, such as, the above test doesn't actually need a database connection, but the "TimestampTest" fixture in which it's a part of does assume this is needed.   This is a big and old test suite so if testing is a newish thing for you, this suite is a really substantial one to take in, it's littered with 15 years of my own figuring out how to write better tests.     The thing that affected me most about testing is this old doc: https://palladion.com//home/tseaver/obzervationz/2008/unit_testing_notes-20080724  noting that there are many "rules" there that I don't follow at all actually (such as we share fixtures).    The "test just one thing" rule is one that I nowadays try to follow, which you can see in this suite is not really happening everywhere right now.   anyway this is all beyond what's necessary for this one little change.   

anyway the above test should also be somewhere in test/dialects/oracle/test_types.py since Oracle has the same issue.

MajorDallas:
I hope I put the Oracle dialect test case in a good place.

`tox` passed the new test case for Oracle, but skipped the Postgres case...

MajorDallas:
Tried again with `tox -e py39-postgresql`. The test passed this time.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 8a042249b8f2bc27767d78f541edfa0a8b7b8390 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 8a042249b8f2bc27767d78f541edfa0a8b7b8390: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2878

zzzeek:
OK so this failed due to pep8 checks.   We use automated tooling to format code here including Python black.   I can just run these in but if you want to learn how we do that, I can show you:

1. installl [pre-commit](https://pre-commit.com/)
2. in the working directory:
     
     pre-commit install
     pre-commit run --all
     git commit -a --amend
     git push -f <name of your git origin>

you can also run the pep8 test suite:

    tox -e pep8

if that passes fully, then you're good.

MajorDallas:
Heh, I thought I ran black on the files I touched, but I guess not 😛 

```
  pep8: commands succeeded
  congratulations :)
```

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision dd217a975e5f0d3157e81c731791225b6a32889f of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset dd217a975e5f0d3157e81c731791225b6a32889f added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2878

sqla-tester:

thanks for your efforts !  I hope to release tonight.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2878

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2878 has been **merged**. Congratulations! :)

Issue #6652, opened 2021-06-18
asyncpg connections not returned to pool if task is cancelled


gboor:
**Describe the bug**
Using postgresql+asyncpg, I noticed a web app eventually runs out of connections and crashes. It turns out asyncpg connections are not returned to the pool when a task is cancelled.

**To Reproduce**
Provide your [Minimal, Complete, and Verifiable](http://stackoverflow.com/help/mcve) example
here.

```pyimport logging
import asyncio

from asyncio import current_task

from sqlalchemy.ext.asyncio import create_async_engine, AsyncSession
from sqlalchemy.orm import sessionmaker


engine = create_async_engine('postgresql+asyncpg://your/db', pool_timeout=30, pool_recycle=1800,
                             pool_size=20, max_overflow=-1, future=True)
async_session_factory = sessionmaker(engine, class_=AsyncSession)

logging.basicConfig(level=logging.INFO)
logging.getLogger('sqlalchemy.engine').setLevel(logging.INFO)
logging.getLogger('sqlalchemy.pool').setLevel(logging.DEBUG)


async def get_and_return():
    async with AsyncSession(engine) as session:
        await session.execute('SELECT 1')


async def get_and_cancel():
    async with AsyncSession(engine) as session:
        await session.execute('SELECT 1')
        current_task().cancel()


async def main():
    for x in range(20):
        try:
            await get_and_return()
            logging.info(engine.pool.status())
            await get_and_cancel()
            logging.info(engine.pool.status())
        except asyncio.exceptions.CancelledError:
            pass


asyncio.run(main())
```

**Error**
Provide the complete text of any errors received **including the complete
stack trace**.   If the message is a warning, run your program with the
``-Werror`` flag:   ``python -Werror myprogram.py``

Correct behaviour from `get_and_return`
```
DEBUG:sqlalchemy.pool.impl.AsyncAdaptedQueuePool:Created new connection <sqlalchemy.dialects.postgresql.asyncpg.AsyncAdapt_asyncpg_connection object at 0x000001815DF55DC0>
INFO:sqlalchemy.engine.Engine:select version()
INFO:sqlalchemy.engine.Engine:[raw sql] ()
INFO:sqlalchemy.engine.Engine:select current_schema()
INFO:sqlalchemy.engine.Engine:[raw sql] ()
INFO:sqlalchemy.engine.Engine:show standard_conforming_strings
INFO:sqlalchemy.engine.Engine:[raw sql] ()
[...]

PR #6654, opened 2021-06-19
Explicit exception in context manager

singh-lokendra:
bare except statement are frowned upon
prefer explicit except BaseException

### Description
- session_scope context manager uses bare `except`
- This doesn't seems to be a good practice https://github.com/PyCQA/flake8-bugbear#list-of-warnings
- Instead of using `Exception` and changing the behavior prefer usage of `BaseException` 
- This should rollback the commits if `KeyboardInterrupt` or similar exceptions raise https://stackoverflow.com/questions/44655854/what-is-the-difference-between-except-and-except-baseexception/63169967#63169967

### Checklist

This pull request is:

**Have a nice day!**

### DIFF
--- a/doc/build/orm/session_basics.rst
+++ b/doc/build/orm/session_basics.rst
@@ -380,7 +380,7 @@ ntextlib.html#contextlib.contextmanager>`_::
         try:
             yield session
             session.commit()
-        except:
+        except BaseException:
             session.rollback()
             raise
         finally:


zzzeek:
OK, certainly something to look at, we often have this thing where it's not clear about catching down to Exception vs. BaseException, case in point #6652 most recently , usually though the BaseException thing is relevant when using greenlets/asyncio.   keyboardinterrupt, if we assume that's going to end the program, then the rollback() is not needed because the connection will be closed.    Are we thinking here of a program that intercepts KeyboardInterrupt but the program doesn't exit?    I feel like I have to read more about this and maybe we should suggest both forms based on what kind of program is being written (greenlets , async, interrupts?  or just plain?).    Then again what is the negative effect I'm concerned about for catching BaseException?

zzzeek:
well, "except" catches everything already.    So why does the flake8 thing you reference state that we should not do "except"?  

> Do not use bare except:, it also catches unexpected events like memory errors, interrupts, system exit, and so on. Prefer except Exception:. If you're sure what you're doing, be explicit and write except BaseException:. 

OK I'm not really sure I buy that.  Python itself has "except:" and they left this in Python 3 too.   it means a real thing and IMO it is "explicit" if you know what it means.  going to look for more citations of this one.

zzzeek:
discussion here:

https://twitter.com/zzzeek/status/1406267471087210499

in short I disagree with this style guide advise when there is a "raise" present below; the intention IMO is clearer as "except:  raise".   if the catch was not doing a straight "raise" that would be different but I think the linter in question should probably accommodate this case separately.

Vaydheesh:
Upon more research I stumbled upon Pep 8 recommendation, and I guess its better to leave it as it is. We can have a `# noqa` to silent the linter. 

https://www.python.org/dev/peps/pep-0008/#programming-recommendations

> A good rule of thumb is to limit use of bare 'except' clauses to two cases:
> 
> 2. If the code needs to do some cleanup work, but then lets the exception propagate upwards with raise. try...finally can be a better way to handle this case.

Issue #6680, opened 2021-06-26
AliasedClass object gets GC'ed leading to failure of wp subclass resolution if not strongly referenced outside of query


maparent:
**Describe the bug**
 
Regression on SQLAlchemy 1.4.x:
A simple query using an explicit with_polymorphic fails.

**To Reproduce**

```py
from sqlalchemy import Column, Integer, String, ForeignKey, create_engine
from sqlalchemy.orm import declarative_base, with_polymorphic, sessionmaker

engine = create_engine('sqlite:///:memory:', echo=True)

Base = declarative_base()

class A(Base):
  __tablename__ = 'a'
  id = Column(Integer, primary_key=True)
  poly = Column(String)
  __mapper_args__ = {
      'polymorphic_identity': 'A',
      'polymorphic_on': poly
  }


class B(A):
  __tablename__ = 'b'
  id = Column(
      Integer,
      ForeignKey(A.id),
      primary_key=True
  )
  __mapper_args__ = {
      'polymorphic_identity': 'B',
  }


Base.metadata.create_all(engine)

Session = sessionmaker(bind=engine)
session = Session()
session.query(with_polymorphic(A, "*")).all()
```

**Error**

```2021-06-26 15:40:01,165 INFO sqlalchemy.engine.Engine BEGIN (implicit)
2021-06-26 15:40:01,165 INFO sqlalchemy.engine.Engine PRAGMA main.table_info("a")
2021-06-26 15:40:01,165 INFO sqlalchemy.engine.Engine [raw sql] ()
2021-06-26 15:40:01,166 INFO sqlalchemy.engine.Engine PRAGMA temp.table_info("a")
2021-06-26 15:40:01,166 INFO sqlalchemy.engine.Engine [raw sql] ()
2021-06-26 15:40:01,166 INFO sqlalchemy.engine.Engine PRAGMA main.table_info("b")
2021-06-26 15:40:01,166 INFO sqlalchemy.engine.Engine [raw sql] ()
2021-06-26 15:40:01,166 INFO sqlalchemy.engine.Engine PRAGMA temp.table_info("b")
2021-06-26 15:40:01,166 INFO sqlalchemy.engine.Engine [raw sql] ()
2021-06-26 15:40:01,167 INFO sqlalchemy.engine.Engine
CREATE TABLE a (
	id INTEGER NOT NULL,
	poly VARCHAR,
	PRIMARY KEY (id)
)


[...]

PR #6681, opened 2021-06-26
recreate the polymorphic attributes when reconstituting the entity. Fixes: #6680

maparent:
### Description

Recreate the AliasedClass of polymorphic subclasses as attributes on the base AliasedInsp when reconstituting it.

### Checklist

This pull request is:

- [x] A short code fix for #6680

Test given as a file in #6680, not as a testcase.

### DIFF
--- a/lib/sqlalchemy/orm/util.py
+++ b/lib/sqlalchemy/orm/util.py
@@ -524,6 +524,17 @@ class AliasedClass(object):
         obj = cls.__new__(cls)
         obj.__name__ = "AliasedClass_%s" % aliased_insp.mapper.class_.__name__
         obj._aliased_insp = aliased_insp
+        for poly in aliased_insp.with_polymorphic_mappers:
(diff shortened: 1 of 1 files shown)


lib/sqlalchemy/orm/util.py
```diff
@@ -524,6 +524,17 @@ def _reconstitute_from_aliased_insp(cls, aliased_insp):
         obj = cls.__new__(cls)
         obj.__name__ = "AliasedClass_%s" % aliased_insp.mapper.class_.__name__
         obj._aliased_insp = aliased_insp
+        for poly in aliased_insp.with_polymorphic_mappers:
```
zzzeek, line 527: this is duplicating code that already exists in AliasedInsp, it should be in a method.    I want to get this in so I can release so I'll do it.

maparent: Thank you for taking it on.

PR #6709, opened 2021-07-02
Implement nesting CTE

Morikko:
### Description
 #4123 Add nesting CTE support.

Query of type:
```sql
WITH t AS (
   WITH t2 AS (
     SELECT 1
   )
   SELECT * FROM t2
)
SELECT * from t
```

### Checklist

- [X] Support `select`
- [x] Support `compound_select`
- [x] Support `insert`
- [x] Support `update` 
- [x] Support `delete`
- [X] Raise a compile error if the dialect does not support nesting

I did not call the parameter `cte_nesting` but only `nesting` as I found the `cte_` part redundant. Also `recursive` did not contain it.

The code is not clean yet, it is still a draft. I am looking for feedback if the implementation goes to the right direction.


Morikko:
By looking at the code, I believe that the compiler visits the statements as a tree. Whereas for CTE, it was normal to put it back to the top, a nesting CTE will only be visible from the current "statement" and the ones below.

As a result, the nesting CTEs are alongside the non nesting ones. But once the nesting CTE part is rendered, it is removed. This assumption makes sense as the compiler will unstack the current visitor and every local CTEs won't be available upwards any more.

**Is this assumption true and good enough to be the official implementation ?**

One problem concerns `self.ctes_recursive`. It can't stay a unique value because different nesting stages does not imply the same state (cf the unit tests). As a consequence, now it computes the value at each query rendering.

The property is still present in the codebase because I am not sure if I should remove it. I lack the knowledge if this value could be used outside of the library and so creates a breaking change. An alternative solution could be to transform `ctes_recursive` as a property that when accessed computes its value based on the current states of `self.ctes`.

**What is your opinion on this property, should it be kept ?**

zzzeek:
> By looking at the code, I believe that the compiler visits the statements as a tree. Whereas for CTE, it was normal to put it back to the top, a nesting CTE will only be visible from the current "statement" and the ones below.
> 
> As a result, the nesting CTEs are alongside the non nesting ones. But once the nesting CTE part is rendered, it is removed. This assumption makes sense as the compiler will unstack the current visitor and every local CTEs won't be available upwards any more.

seems like that's the idea, yup.   good job figuring this out the CTEs are tough.

> 
> **Is this assumption true and good enough to be the official implementation ?**

if the CTE is "nested" it's now local to the statement in which it's nested; the outer SELECT / statement can't refer to such a CTE that's nested I assume, so yes that would be the general idea.


> 
> One problem concerns `self.ctes_recursive`. It can't stay a unique value because different nesting stages does not imply the same state (cf the unit tests). As a consequence, now it computes the value at each query rendering.

ok is that because...there could be a "nested" inside there so we have to recompute?  it's probably not that big a deal.    The average CTE case is that there's just one or two CTEs at the top level.    For this change I would have to spend more time staring at it so perhaps there's a way to improve upon it, but getting the proof of concept working with the tests is always the best way to start.

> 
> The property is still present in the codebase because I am not sure if I should remove it. I lack the knowledge if this value could be used outside of the library and so creates a breaking change. An alternative solution could be to transform `ctes_recursive` as a property that when accessed computes its value based on the current states of `self.ctes`.

it looks to be only used by the compiler right now so I'm not sure why it's not an underscored attribute.  

> 
> **What is your opinion on this property, should it be kept ?**

 I'll have to look more closely but let's run the tests for now.  thanks for doing this!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 6be7744a6c93fbf5f2dcf36c5a8262034669197c of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 6be7744a6c93fbf5f2dcf36c5a8262034669197c: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2948

lib/sqlalchemy/sql/compiler.py
```diff
@@ -3323,14 +3323,37 @@ def _generate_prefixes(self, stmt, prefixes, **kw):
             clause += " "
         return clause
 
-    def _render_cte_clause(self):
+    def _render_cte_clause(
+        self,
+        nesting_only: bool = False,
```
zzzeek, line 3328: we're still supporting py2.7 for the 1.4 series :(  have to take annotations out

zzzeek (changes requested):
this was supposed to copy up from teh gerrit, for some reason the comment wasn't sent out

Morikko:
> ok is that because...there could be a "nested" inside there so we have to recompute?

Yes.

Finally I also have to keep the level information to pass `test_double_nesting_cte_in_cte`. Else the second nesting CTE SELECT would generate the first nesting CTE.

Morikko:
Should I add e2e tests on real backend ?

zzzeek:
> Should I add e2e tests on real backend ?

we dont really have that for CTEs, these are a syntactical thing on our end , should be no need

lib/sqlalchemy/sql/compiler.py
```diff
@@ -3323,14 +3335,47 @@ def _generate_prefixes(self, stmt, prefixes, **kw):
...
+                return ""
+
+            # Remove them from the visible CTEs
+            self.ctes = {
+                cte: self.ctes[cte]
+                for cte in self.ctes
+                if not cte.nesting and self.level_by_ctes[cte] == nesting_level
+            }
+
+            if ctes and not self.dialect.supports_nesting_cte:
```
zzzeek, line 3364: we tend to not do checks like this because they prohibit the construct from being usable on a database version that adds new support for a feature.  in this case MariaDB for example could support the syntax in a new version, and users should be able to use it.

as far as raising an error if the syntax isn't supported, the database does that already.   so usually we try to let the database raise for unsupported syntaxes.

lib/sqlalchemy/sql/compiler.py
```diff
@@ -3610,8 +3661,9 @@ def visit_insert(self, insert_stmt, **kw):
         if returning_clause and not self.returning_precedes_values:
             text += " " + returning_clause
 
-        if self.ctes and toplevel and not self.dialect.cte_follows_insert:
-            text = self._render_cte_clause() + text
+        if self.ctes and not self.dialect.cte_follows_insert:
```
zzzeek, line 3664: in this case, the "cte_follows_insert" flag actually has us make a different choice as to how to render, so that flag is more useful

test/sql/test_cte.py
```diff
@@ -1,8 +1,10 @@
+import functools
+import pytest
```
zzzeek, line 2: remove pytest import

test/sql/test_cte.py
```diff
@@ -1377,3 +1379,306 @@ def test_no_alias_construct(self):
...
+
+        self.assert_compile(
+            stmt,
+            "WITH RECURSIVE cte(outer_cte) AS (WITH RECURSIVE nesting(inner_cte) "
+            "AS (SELECT %(param_1)s AS inner_cte) "
+            "SELECT nesting.inner_cte AS outer_cte FROM nesting) "
+            "SELECT cte.outer_cte FROM cte",
+        )
+
+    @pytest.mark.parametrize(
```
zzzeek, line 1573: I dont think we should raise within the compiler for this condition, so IMO we don't need this test.

Also, we dont use `pytest.mark.parametrize` directly, we use our own abstraction `@testing.combinations`, see https://github.com/sqlalchemy/sqlalchemy/blob/master/test/sql/test_cte.py#L1032 in this file for example

zzzeek (changes requested):
that's a lot of tests!   there's a lot to take in here, so I just made some initial comments.

test/sql/test_cte.py
Morikko: Just for curiosity why do you use `@testing.combinations` instead of `pytest.mark.parametrize` ?

CaselIT: The idea is to avoid an hard dependency on pytest.
If I'm not mistaken sa has already outlived two other testing framework before using pytest.

Also Mike likes doing stuff it's own way 😅

zzzeek: yes originally the idea was, the test suite ran both on nose and pytest.  this is when pytest was new.      once the enormous job of migrating off of nose into a generic abstraction layer was done, we weren't going to have to do that again, so we kept the abstraction layer.

later, we started using way more pytest features.  but we built those into the abstraction layer also.

this continues to be a good thing, because:

1. it limits us to using specific pytest patterns that we have decided to adopt, rather than using the whole kitchen sink and having different approaches in different tests
2. it completely saves our bacon when we hit issues such as the older pytest 4.x series that is the newest version that runs on python 2 has some serious performance issues with certain kinds of fixtures; the abstraction layer makes sure we don't use those features
3. keeps us honest and organized about how more elaborate pytest features are used

so shorter bullets:

1. pytest is vast, our abstraction layer helps with the issue of tests from using dozens of different patterns everywhere (which they still do, but not as much as they would without the abstraction)
2. gives us a little bit of insurance for migrating to new versions of pytest and/or a new test framework if pytest goes south someday

test/sql/test_cte.py
```diff
@@ -1377,3 +1378,333 @@ def test_no_alias_construct(self):
...
+            ") SELECT cte.inner_cte_1, cte.inner_cte_2 FROM cte",
+        )
+
+        # 2 next 1
+
+        nesting_cte_2_1 = select([select_2_cte, select_1_cte]).cte("cte")
+        stmt_2_1 = select([nesting_cte_2_1])
+        self.assert_compile(
+            stmt_2_1,
+            # FIXME: nesting_1 is generated 2 times
```
Morikko, line 1501: Expected
```sql
WITH cte AS (
	WITH nesting_1 AS (
		SELECT
			%(param_1) s AS inner_cte_1
	)
	, nesting_2 AS (
		SELECT
			nesting_1.inner_cte_1 + %(inner_cte_1_1) s AS inner_cte_2
		FROM
			nesting_1
	)
	SELECT
		nesting_2.inner_cte_2 AS inner_cte_2
		, nesting_1.inner_cte_1 AS inner_cte_1
	FROM
		nesting_2
		, nesting_1
)
SELECT
	cte.inner_cte_2
	, cte.inner_cte_1
FROM
	cte
```
Actual:
```sql
WITH cte AS (
	WITH nesting_2 AS (
		WITH nesting_1 AS (
			SELECT
				%(param_1) s AS inner_cte_1
		)
		SELECT
			nesting_1.inner_cte_1 + %(inner_cte_1_1) s AS inner_cte_2
		FROM
			nesting_1
	)
	, nesting_1 AS (
		SELECT
			%(param_1) s AS inner_cte_1
	)
	SELECT
		nesting_2.inner_cte_2 AS inner_cte_2
		, nesting_1.inner_cte_1 AS inner_cte_1
	FROM
		nesting_2
		, nesting_1
)
SELECT
	cte.inner_cte_2
	, cte.inner_cte_1
FROM
	cte
```

Morikko:
Another issue with the current approach is that current nesting CTEs are eager to be rendered so some cases are not possible like to put the `WITH` statement upper in the query tree.

:heart:  https://github.com/sqlalchemy/sqlalchemy/issues/6752 just landed on time because I find I could use it or a close approach to generate a nesting CTE earlier.

test/sql/test_cte.py
Morikko: I change the approach, I let this behavior possible as it is syntactically possible so it should not be impossible. And it is hard to auto fix without deep changes for no real benefits. I prefer to let the developer the ownership of which order matter. 

As shown by the first assertion, changing the column order is a first workaround. Another one shown by the second assertion uses `add_cte`. A developer has the possibility to refer a CTE before the query does and thus define the order.

jvanasco:
I just saw some commits get pushed to this. FYI, mike is off for the next 2 weeks. If you end up needing a review before then, tag @CaselIT.

zzzeek:
One week ;). But wow theres a lot of mail great job keeping things moving everyone !

CaselIT:
> One week ;). But wow theres a lot of mail great job keeping things moving everyone !

They all waited for your to go on vacation :)

Morikko:
Added:
- Support for independent nesting CTE for insert and compound statements
- A test to reproduce a recursive query graph
- Changelog
- A bit of documentation. I have not spoken about `add_cte()` to reorganize order of nesting CTE. Eg, either the order of appearance at the same stack or the stack level (it is possible to render the nesting CTE upper than when it is called).

Side note, this code is able to generate a running query for the recursive query graph case in my company. So all good :+1:

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 0887c8899aa7424abd7762a48c0c876c7c1b5fb1 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 0887c8899aa7424abd7762a48c0c876c7c1b5fb1 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2948

zzzeek:
thanks for the update

we are quite backlogged right now so hopefully wtihin the next week i can take a look

sqla-tester:

still looking but this is good stuff, in this version i can actually skim and pretty much see how it works, which wasn't the case earlier, nice job

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2948

zzzeek:
just noticed you had further changes

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 64ab2f6ea269f2dcf37376a13ea38c48c5226fb6 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 64ab2f6ea269f2dcf37376a13ea38c48c5226fb6 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2948

zzzeek:
great job.  I've made a final pass on this and am planning to have it in the next 1.4.24 release.  thanks for the long diligent efforts here!!

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2948 has been **merged**. Congratulations! :)

PR #6712, opened 2021-07-04
Update result.all documentation

ShmuelTreiger:
Adding a warning that results.all() does not deduplicate, unlike query.all(). I raised this in #6711 

I *think* I got the method call right. I was not sure how to get the url right

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/ext/asyncio/result.py
+++ b/lib/sqlalchemy/ext/asyncio/result.py
@@ -152,6 +152,14 @@ class AsyncResult(AsyncCommon):
 
         :return: a list of :class:`.Row` objects.
 
+        .. warning::  As opposed to :meth:`_query.Query.all`,
+           this method **will not** deduplicate entries
+           based on primary key.
+
+        .. seealso::
+
+            :ref:``
+
         """
 
         return await greenlet_spawn(self._allrows)


zzzeek (changes requested):
OK we have a few result.all()'s for which this applies, additionally iterating the result, etc., also have this behavior.   Deduplicating is also not really necessary in the vast majority of cases unless joinedlad() is used, and in that case, an error is raised if unique() isn't called, so no chance of missing it there.    

Beyond that, query.all() is legacy here so this note is more of a migration type of note and going forward, new users to the SQLAlchemy ORM will be [using the new tutorial](https://docs.sqlalchemy.org/en/14/tutorial/index.html) to understand how the ORM works.  the idea that anyone would expect "all()" to deduplicate would be strictly a legacy behavior.

Where im going here is that this particular warning is very specific and arbitrary and does not really solve the problem of users migrating from 1.x style ORM use.    the impression here is that people who are trying to migrate from 1.x ORM to **asyncio** are getting hit with this, because they have not read the new tutorials etc.

So that said I think the asyncio ORM docs need just a note at the very top, e.g. here: https://docs.sqlalchemy.org/en/14/orm/extensions/asyncio.html#synopsis-orm   , beyond the "2.0 style" note, which is a note:   "As noted above, the asyncio ORM system uses 2.0 style querying.  Please read and familiraize with the [2.0 style Core/ORM tutorial](https://docs.sqlalchemy.org/en/14/tutorial/data.html) to famliarize with general behavioral differences between 1.x and 2.0 style usage, including critical elements such as results no longer de-duplicating results".

ShmuelTreiger:
I disagree somewhat. I read through the documents (at least the parts that
seemed relevant), but there's too much for an average/casual dev to keep
every single detail in mind. For example, I didn't even know we relied on
deduplication, so reading that part of the migration notes probably didn't
even phase me. Further, we use results.all() in many places and it only
failed once (though I'm now planning to check through our useage and see if
.unique() is a better fit), so I didn't even immediately suspect .all() was
the problem.

Consequently, my workflow has basically been: read migration notes --> try
out select in place of query --> check documentation when I have an issue
(including searching the migration notes) --> post here if I fail. I
imagine I'm not the only person with a similar workflow. I would argue that
for anyone doing similar to me, migration specific notes in the actual
documentation would be extremely helpful, at least until say 2.1 or some
future point when it's reasonable to assume most users who plan to migrate
have migrated.

That being said, it would constitute a lot of work for you, and your
project means your rules.

Thank you for your work and for your appropriate and timely responses. I
really appreciate your help with everything.

CaselIT:
just for reference, the query referred by @ShmuelTreiger is
https://github.com/sqlalchemy/sqlalchemy/discussions/6711#discussioncomment-962116

CaselIT:
I think a class level note may be added

PR #6713, opened 2021-07-04
#2949 Accommodate dupe col names

sonthonaxrk:


zzzeek:
hey thanks for this, I've accommodated for this case in the latest iteration of the fix which is at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/2949 and builds upon a new refactoring that isn't merged yet.

Issue #6724, opened 2021-07-06
Default DBAPIs for async engine for `backend://`


cekc:
Upon creation of an `AsyncEngine` from url that doesn't specify DBAPI, default **async driver** would be used.

For example `create_async_engine("postgresql://...")` would actually use `asyncpg` driver

Right now threre are three async diealects
+ `postgresql+asyncpg`
+ `mysql+aiomysql`
+ `sqlite+aiosqlite`

CaselIT:
Hi,

Personally I'm not sure if we want to define a default for async at the moment, given that there are only 3 dialects supported.

Like for postgresql I think it may make sense waiting for psycopg3 before deciding for that or asyncpg.

zzzeek:
as noted in the PR "default" DBAPIs were not a good idea in the first place as this is a level of opinion that has no advantage within SQLAlchemy.

francipvb:
Hello,

Some providers like heroku or dokku sets an URL without a provider to link to the database, making it difficult to use async engines.

Please consider this issue again.

zzzeek:
our understanding is that these heroku URLs are only defaults set up in some configuration file or documentation on their end.  they should change these values as they see fit to best serve their customer's needs.

zzzeek:
not to mention, the usage of any URL requires that the appropriate drivers be installed.  I dont see how it's helpful to have a URL like "postgresql" default to "postgresql+asyncpg", without the user being aware, yet they will need to know that they have to install asyncpg.   Again if this is something that runs in some heroku environment that has pre-configured packages installed, they should solve this issue as they see fit.

CaselIT:
Please note that the default dialect is legacy since https://github.com/sqlalchemy/sqlalchemy/issues/6960 so there aren't any plan to add this for async drivers.

PR #6725, opened 2021-07-06
#6724 Default DBAPIs for async engine

mishmish-dev:
For example, `create_async_engine("postgresql://...")` would actually use `asyncpg` driver


zzzeek:
the "default driver" thing is a historical artifact of a short-sighted decision I made about 16 years ago, and today I would not have included this concept; it has backwards-incompatibility built into it as the state of drivers changes, and additionally it misleads newer users who don't understand that there's a separate driver concept that is pretty critical to the behaviors and features they're going to get, so I'm pretty -1 on this change.

PR #6875, opened 2021-08-11
Fix example that uses deprecated code in docs

jpmelos:
### Description
Just a few lines above this change, it is said that creating a mapping with `mapper()` is deprecated in favor of `mapper_registry.map_imperatively()`. This PR fixes an example in the documentation that still uses the old function.

### Checklist

This pull request is:

**Have a nice day!**


CaselIT:
Thanks!

zzzeek:
thank you !

PR #6968, opened 2021-09-01
update documenation for azure sql database with pyodbc

lachhebo:
### Description

I want to update the documentation about how to connect to a Azure mssql database using sql alchemy

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/mssql/pyodbc.py
+++ b/lib/sqlalchemy/dialects/mssql/pyodbc.py
@@ -106,16 +106,23 @@ database using Azure credentials::
     import struct
     from sqlalchemy import create_engine, event
     from sqlalchemy.engine.url import URL
-    from azure import identity
+    import msal
 
     SQL_COPT_SS_ACCESS_TOKEN = 1256  # Connection option for access tokens, as defined in msodbcsql.h
     TOKEN_URL = "https://database.windows.net/"  # The token URL for any Azure SQL database
-
+    BACK_APP_ID = "fakebackendappid"
+    BACKEND_APP_CREDENTIAL = "fakebackendappcredential"
+    TENANT_ID = "faketenantid"
+    TOKEN = "token" # obtained from azure 
     connection_string = "mssql+pyodbc://@my-server.database.windows.net/myDb?driver=ODBC+Driver+17+for+SQL+Server"
 
     engine = create_engine(connection_string)
 
-    azure_credentials = identity.DefaultAzureCredential()
+    azure_credentials = msal.ConfidentialClientApplication(
+        client_id=BACK_APP_ID,
+        authority=f"https://login.microsoftonline.com/{TENANT_ID}",
+        client_credential=BACKEND_APP_CREDENTIAL,
+    )
 
     @event.listens_for(engine, "do_connect")
     def provide_token(dialect, conn_rec, cargs, cparams):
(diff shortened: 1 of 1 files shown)


zzzeek:
hi, thanks for this PR.

Unfortunately I cannot vet the code changes here and they are extensive.   I was uncomfortable accepting this code example in the original case because it was so specific and if it is indeed so brittle that it has to change this much less than a year after it was added, I would rather remove it and link people to a wiki somewhere.

cc @gordthompson  can we remove this example and somehow link people elsewhere?   these are not SQLAlchemy APIs

gordthompson (changes requested):
Since this relates directly to pyodbc it could be included in wiki there, similar to the existing entries "[Connecting to] Microsoft SQL Server from Windows", etc.. Ideally the wiki entry would include the generic pyodbc approach first and then explain how to do that for SQLAlchemy using the event listener.

Unfortunately I am also unable to test the code myself. Perhaps @v-chojas might be able to help out with that.

zzzeek:
yes we need some background why these changes are needed and why the version of the code we were given quite recently is suddenly not working

CaselIT:
> 
> 
> Hello,
> 
> To give you some background, the libraries previously used in the documentation as been deprecated by microsoft in favor of msal to implement safe authentification. It is stated on the homepage of the azure python package:
> 
> > Starting with v5.0.0, this package is deprecated. Please install the service specific packages prefixed by azure needed for your application.
> 
> So I just updated the documenation to show how we use the msal library in our project (using it on prod without any issues).

It does not seem to be deprecated. See
https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/identity/azure-identity
and
https://pypi.org/project/azure-identity/

gordthompson:
There does seem to be some confusion. The old `from azure import identity` statement appears to be referring to

https://pypi.org/project/azure/

which is where the "Starting with v5.0.0, this package is deprecated." comes from.

The list of packages at

https://azure.github.io/azure-sdk/releases/latest/all/python.html

does list azure-identity, whose last release was less two weeks ago (2021-08-19). 

MSAL has also been around for a couple of years, and its latest release was just last week (2021-08-26).

I'll open a pyodbc issue and see if we can get some (reasonably) authoritative documentation in the wiki there.

CaselIT:
> There does seem to be some confusion. The old `from azure import identity` statement appears to be referring to
> 
> https://pypi.org/project/azure/
> 
> which is where the "Starting with v5.0.0, this package is deprecated." comes from.
> 
> The list of packages at
> 
> https://azure.github.io/azure-sdk/releases/latest/all/python.html
> 
> does list azure-identity, whose last release was less two weeks ago (2021-08-19).

That's the correct import. Basically the main package is still `azure`, but each sub-package is installed separately. See the examples in `azure-identity` where it uses `from azure.identity import DefaultAzureCredential`. https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/identity/azure-identity#authenticating-with-defaultazurecredential

gordthompson:
Reply from a Microsoft employee to the pyodbc issue claims that "Azure SQL has not changed"

https://github.com/mkleehammer/pyodbc/issues/949#issuecomment-911772568

Still unclear how MSAL factors into this. More information needed.

CaselIT:
closing for now

PR #7014, opened 2021-09-10
README.unittest container edits mainly mariadb/mysql

grooverdan:
### Description

Simplify it a bit:

* test_schema2 isn't needed in MySQL/MariaDB test.
* Only permissions on test_schema are needed.
* User without qualification defaults to user@'%'.
* Use MARIADB_* names with MariaDB container that has
  been there most of this year in releases.
* We don't need to specify a database on connection
  when creating users/databases.
* On all containers - docker will always pull image
  if not available. Also postgres image didn't match
  container initialization.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

You too.

### DIFF
--- a/README.unittests.rst
+++ b/README.unittests.rst
@@ -231,11 +231,8 @@ intended for production use!
 
 **PostgreSQL configuration**::
 
-    # only needed if a local image of postgres is not already present
-    docker pull postgres:12
-
     # create the container with the proper configuration for sqlalchemy
-    docker run --rm -e POSTGRES_USER='scott' -e POSTGRES_PASSWORD='tiger' -e POSTGRES_DB='test' -p 127.0.0.1:5432:5432 -d --name postgres postgres:12-alpine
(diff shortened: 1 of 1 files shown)


CaselIT:
Regarding the images, maybe we could leave off the version, using the latest from docker?

Otherwise I think we could also update postgresql to 13 (also the default build is probably fine)

grooverdan:
`latest` maps to 13 on the postgres still - https://github.com/docker-library/official-images/pull/10713/files#diff-fa6dc4e4c26179b634d7e1f1a74fd6bab4e442e444d5625ffc293e4a385a1738R17  (not alpine, but that's not too important).

`mcr.microsoft.com/mssql/server` has a latest too. So sure I'll force push a version tag removal.

CaselIT (approved):

CaselIT:
thanks!

PR #7062, opened 2021-09-21
Fix order of warnings filters

rallyemax:
### Description
As written, the regex-based filters are completely ignored because [`warnings.filterwarnings`](https://docs.python.org/3/library/warnings.html?highlight=#warnings.filterwarnings) inserts entries at the front of the filter list, and entries closer to the front of the list override entries later in the list, if both match a particular warning.

The intended behavior is for all `RemovedIn20Warning` warnings to log and continue, except that specific warnings that match the defined regular expressions should be raised as exceptions. To accomplish this intent, the general filter must be inserted before the regular expression special cases.

The original order would work if `warnings.filterwarnings` for the default case were invoked with `append=True`, but the intent of the documentation is for this set of filters to be determinative (i.e. to override any conflicting pre-existing filters that may exist), so `append=True` should not be used.

### Checklist

This pull request is:

### DIFF
--- a/doc/build/changelog/migration_20.rst
+++ b/doc/build/changelog/migration_20.rst
@@ -260,7 +260,12 @@ the SQLAlchemy project itself, the approach taken is as follows:
 
         import warnings
         from sqlalchemy import exc
-
+        
+        # for warnings not included in regex-based filter below, just log
+        warnings.filterwarnings(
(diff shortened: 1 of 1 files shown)


zzzeek:
thanks for the contribution!

PR #7114, opened 2021-09-29
clarify that @reconstructor only applies to a single method

eseiver:
### Description

Add minor wording clarifying that like `__init__`, there can only be one reconstructor method per class.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/doc/build/orm/constructors.rst
+++ b/doc/build/orm/constructors.rst
@@ -21,7 +21,7 @@ If you need to do some setup on database-loaded instances before they're ready
 to use, there is an event hook known as :meth:`.InstanceEvents.load` which
 can achieve this; it is also available via a class-specific decorator called
 :func:`_orm.reconstructor`.   When using :func:`_orm.reconstructor`,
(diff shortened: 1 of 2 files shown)


zzzeek:
did you try placing `@reconstructor` on more than one method?    how would that work exactly, they'd be called in non-deterministic order?

eseiver:
Yes, this one was entirely from personal experience, hoping to save someone else the headache!
In terms of the actual error when 2+ decorators are used, for me, none of them worked when I loaded the objects. It would be nice if a more direct error message had surfaced as well (e.g.: 'warning: reconstructor decorator appears multiple times in class') instead of simply puzzling over why they had stopped loading, but figured the documentation was the easiest fix for now.

zzzeek:
sure.   i just would never have thought of that hence there's not a specific error checking for this

zzzeek:
thanks!

Issue #7117, opened 2021-09-30
SQLAlchemy can't handle reflections for tables with no columns - (raising "NoSuchTableError")


bohemia420:
### Describe the bug

The bug being SQLAlchemy being unable to handle/provide reflection of tables with no columns, i.e call to :

```Table(name, metadata, schema=schema, autoload_with=autoload_with, extend_existing=True)```
would result in a NoSuchTableException being raised, should the table exist but with no columns. 

SQLAlchemy's `engine/reflection.py` i.e `lib/sqlalchemy/engine/reflection.py` seems to recognise i.e "find" tables with the `table_found` variable, **only through iterating over its columns**. Since an empty table wouldn't have any, the `table_found` stays False and results in "NoSuchTableError" Exception being thrown:

file : `sqlalchemy/engine/reflection.py`
```
        found_table = False
        cols_by_orig_name = {}

        for col_d in self.get_columns(
            table_name, schema, **table.dialect_kwargs
        ):
            found_table = True

            self._reflect_column(
                table,
                col_d,
                include_columns,
                exclude_columns,
                cols_by_orig_name,
            )

        if not found_table:
            raise exc.NoSuchTableError(table.name)
```

### To Reproduce

```python
upgrading the version to `1.4.23`, even `1.4.25` doesn't work

# pip3.9 freeze | grep -i sqlalchemy
SQLAlchemy==1.4.23
```

and execute the following steps in python3 shell:

```
# python3.9
Python 3.9.7 (default, Sep  3 2021, 20:19:49) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from sqlalchemy import create_engine, Table, Column, Integer, String, MetaData
>>> engine = create_engine('postgresql://<user>:<pass>@<db>:5432/<database>')
>>> engine.table_names()
<stdin>:1: SADeprecationWarning: The Engine.table_names() method is deprecated and will be removed in a future release.  Please refer to Inspector.get_table_names(). (deprecated since: 1.4)
[...]
>>> metadata = MetaData(bind=engine)
>>> emp23 = Table('emp23', metadata, )
>>> tab23 = Table('tab23', metadata,Column('id', Integer, primary_key = True), )
>>> metadata.create_all(engine)
>>> engine.table_names()
['emp23', 'tab23', ...]
>>> Table('tab23', MetaData(bind=engine), schema='public', autoload_with=engine, extend_existing=True)
[...]

PR #7118, opened 2021-09-30
initial draft : setting found_table properly so as to also reflect ex…

bohemia420:
Corresponds to issue #7117 wherein a minor yet important change, enabling reflections for tables that exist but with no columns, by changing the way `found_table` is set. This also saves us through redundant iteration through all the columns unnecessarily setting `found_table` on and on:

```
         for col_d in self.get_columns(
             table_name, schema, **table.dialect_kwargs
         ):
-            found_table = True
-
         ...
``` 

I have tested it with existing but no column tables getting reflected well, whilst tables that don't exist, won't be, resulting in a desired "NoSuchTableError" execption being raised

Fixes: #3247

### DIFF
--- a/lib/sqlalchemy/engine/reflection.py
+++ b/lib/sqlalchemy/engine/reflection.py
@@ -768,7 +768,7 @@ class Inspector(object):
             if isinstance(table_name, str):
                 table_name = table_name.decode(dialect.encoding)
 
-        found_table = True if table.name in self.engine.table_names() else False
+        found_table = True if self.has_table(table.name) else False
         cols_by_orig_name = {}
 
         for col_d in self.get_columns(
--- a/lib/sqlalchemy/engine/reflection.py
+++ b/lib/sqlalchemy/engine/reflection.py
@@ -768,7 +768,6 @@ class Inspector(object):
             if isinstance(table_name, str):
                 table_name = table_name.decode(dialect.encoding)
 
-        found_table = True if self.has_table(table.name) else False
         cols_by_orig_name = {}
 
         for col_d in self.get_columns(
@@ -782,7 +781,7 @@ class Inspector(object):
                 cols_by_orig_name,
             )
 
-        if not found_table:
+        if not self.has_table(table.name):
             raise exc.NoSuchTableError(table.name)
 
         self._reflect_pk(
--- a/lib/sqlalchemy/engine/reflection.py
+++ b/lib/sqlalchemy/engine/reflection.py
@@ -768,14 +768,12 @@ class Inspector(object):
             if isinstance(table_name, str):
                 table_name = table_name.decode(dialect.encoding)
 
-        found_table = False
+        found_table = True if table.name in self.engine.table_names() else False
         cols_by_orig_name = {}
 
         for col_d in self.get_columns(
             table_name, schema, **table.dialect_kwargs
         ):
-            found_table = True
-
             self._reflect_column(
                 table,
                 col_d,
--- a/lib/sqlalchemy/engine/reflection.py
+++ b/lib/sqlalchemy/engine/reflection.py
@@ -781,7 +781,7 @@ class Inspector(object):
                 cols_by_orig_name,
             )
 
-        if not self.has_table(table.name):
+        if not self.has_table(table.name, schema):
             raise exc.NoSuchTableError(table.name)
 
         self._reflect_pk(


lib/sqlalchemy/engine/reflection.py
```diff
@@ -768,14 +768,12 @@ def reflect_table(
             if isinstance(table_name, str):
                 table_name = table_name.decode(dialect.encoding)
 
-        found_table = False
+        found_table = True if table.name in self.engine.table_names() else False
```
CaselIT, line 771: `eninge.table_names()` is deprecated.

Please use the inspector `has_table` method.

Also I think it makes more sense to add this control only if found_table = False after the loop, avoiding a query in most cases

CaselIT (changes requested):

lib/sqlalchemy/engine/reflection.py
bohemia420: thanks @CaselIT for reviewing, I'll look up `has_table` thanks for letting me know `.table_names()` is deprecated. 
should we be having `found_table` be set in a loop, wondering this maybe redundant albeit harmless, considering tables where cols are 1000 it is unnecessarily setting it to the very value it already is. 

Nevertheless I can rather have this `has_table` check setting `found_table` after the loop/before raising exc.

CaselIT: the reasoning is that I would think the use case for tables with no columns is really small, it's the first time it came up in 16+ years, so we can avoid the additional has_table query most of the times.

But it's probably not too important, so it's fine also to leave as is, just replacing the table_names with has_table.

> thanks for letting me know `.table_names()` is deprecated.

well it does raise a deprecation warning in your output here https://github.com/sqlalchemy/sqlalchemy/issues/7117

bohemia420: @CaselIT sorry I overlooked. Nevertheless, I've replaced the deprecated `table_names()` with `has_table(table.name)`.
Secondly,
> Also I think it makes more sense to add this control only if found_table = False after the loop, avoiding a query in most cases

I believe the `found_table` currently, is being done with, bit redundantly. Ideally, we should check for `found_table` at once using this fix, and **right away raise the NoSuchTableError exception** should it be False. Thereafter, we can do the needful on reflecting columns. Let me know 👍

CaselIT: makes sense, we could as well do `if not self.has_table(table): raise ...`

bohemia420: done. 
- so removed redundant `found_table`. 
- made referring/using `self.has_table(table.name)` more compact, as suggested above.

bohemia420: @CaselIT apparently it is failing for all python2.7 envs and cancelled operations for `3.9-cext` and `3.9-nocext` envs), while the error, more than breaking changes rather looks like a NoSuchTableError being thrown for tables like "dingalings", "foob", "a" etc:
```
    metadata._remove_table(name, schema)
  File "/home/runner/work/sqlalchemy/sqlalchemy/.tox/github-cext/lib/python2.7/site-packages/sqlalchemy/util/langhelpers.py", line 72, in __exit__
    with_traceback=exc_tb,
  File "/home/runner/work/sqlalchemy/sqlalchemy/.tox/github-cext/lib/python2.7/site-packages/sqlalchemy/sql/schema.py", line 602, in __new__
    table._init(name, metadata, *args, **kw)
  File "/home/runner/work/sqlalchemy/sqlalchemy/.tox/github-cext/lib/python2.7/site-packages/sqlalchemy/sql/schema.py", line 682, in _init
    resolve_fks=resolve_fks,
  File "/home/runner/work/sqlalchemy/sqlalchemy/.tox/github-cext/lib/python2.7/site-packages/sqlalchemy/sql/schema.py", line 717, in _autoload
    _extend_on=_extend_on,
  File "/home/runner/work/sqlalchemy/sqlalchemy/.tox/github-cext/lib/python2.7/site-packages/sqlalchemy/engine/reflection.py", line 785, in reflect_table
    raise exc.NoSuchTableError(table.name)
NoSuchTableError: dingalings
```
Are these tables present at first, in the PostgresSQL on the server, being referred to (appropriate schema)? Wondering if I could get a step closer debugging this, ***interactively***. Doesn't look like a break due to changes in this PR.

CaselIT: It seems schema is missing form the has_table call

bohemia420: ok, yes perhaps the schema being picked up was a `None` owing to the definition:
```
def has_table(self, table_name, schema=None):
        """Return True if the backend has a table of the given name.
```

have fixed it as calling has_table(table.name, schema) since the variable `schema` is set well above and being used (e.g for getting/reflecting columns). Will keep an eye as the pipeline/workflow triggers and watch of for failing envs.

CaselIT: thanks I'll make it execute

CaselIT:
it does seem happier now. I'll run it in gerrit. thanks

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision cb8ce01957e9a1453290a7c2728af8c60ef55fa1 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change cb8ce01957e9a1453290a7c2728af8c60ef55fa1: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3127

bohemia420:
I see that in the Gerrit review above, 5 checks are failing at the "ci_verification" stage:

- `openstack_gerrit/nova`: **oslo_db.sqlalchemy.enginefacade.AlreadyStartedError:** this TransactionFactory is already started
- `sqlalchemy_gerrit/cext,oracle-backendonly,py39`: **sqlalchemy.exc.NoSuchTableError:** ptable, users_v, local_table, test_a23371f2d0ac_ts1_pt, 
- `sqlalchemy_gerrit/cext,oracle-backendonly,py27`: **NoSuchTableError:** test_table_syn, users_v, ptable, local_table, test_dd9d47e87988_ts1_pt, email_addresses_v
- `sqlalchemy_gerrit/cext,mssql-backendonly,py39`: **sqlalchemy.exc.NoSuchTableError:** email_addresses_v, users_v
- `sqlalchemy_gerrit/cext,mssql-backendonly,py27`: NoSuchTableError: email_addresses_v, users_v

@sqla-tester how can we help these?

zzzeek (changes requested):
we need to still use found_table if present and not call upon a new SQL query in the normal case so let's put that back and make has_table check only if not found_table.

CaselIT:
> > Found the issue.
> > @zzzeek `has_table` seem to be not properly specified for views: postgresql (at least) returns true for views, oracle and mssql do no. I think we could expand it to support views in all cases, what do you think?
> 
> that's very strange because we have all the included DBs listed as supporting view reflection, the suite/test_reflection.py should certainly be confirming that has_table() is agreeing with get_columns() for views.

it does not. we are not testing has_table with views. I've added them in a local change.

I'll open a new issue and then close both in this change set

CaselIT:
> I'll open a new issue and then close both in this change set

Opended https://github.com/sqlalchemy/sqlalchemy/issues/7161

sqla-tester:

I'm a little nervous about changing has_table for mssql/oracle in a tiny point release and it would be easier for that part of the change to be a 2.0 thing.  for the "support table with no columns" use case, because we are going to use both booleans "found_columns" + "has_table" in any case, for the moment we can specify a dialect-level flag "has_table_is_authoritative" or something like that, or even "supports_table_no_columns" and just fix this for the databases people care about (postgresql, as usual) and limit the tests to those dialects w/ the supports flag.  i can't imagine a table with no columns in sql server /oracle.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3127

bohemia420:
> **mike bayer** (zzzeek) wrote:
> 
> I'm a little nervous about changing has_table for mssql/oracle in a tiny point release and it would be easier for that part of the change to be a 2.0 thing. for the "support table with no columns" use case, because we are going to use both booleans "found_columns" + "has_table" in any case, for the moment we can specify a dialect-level flag "has_table_is_authoritative" or something like that, or even "supports_table_no_columns" and just fix this for the databases people care about (postgresql, as usual) and limit the tests to those dialects w/ the supports flag. i can't imagine a table with no columns in sql server /oracle.
> 
> View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3127

ok, so do we recommend it as a separate `flag` to be inferred from settings and be `false` by default so that we still raise `NoSuchTableError` exception should a table with 0 columns be reflected?

PR #7133, opened 2021-10-01
Fix recursive CTE to support nesting

Morikko:
Issue: #4123

https://github.com/sqlalchemy/sqlalchemy/issues/4123#issuecomment-927314811

> This has not been caught by the tests because the nesting recursive queries there did not union against itself, eg there was only the root clause...

- Now tests are real recursive queries
- Add tests on aliased nested CTEs (recursive or not)
- Adapt the `_restates` attribute to use it as a reference
- Add some docs around to explain some variables usage

### DIFF
--- a/lib/sqlalchemy/sql/selectable.py
+++ b/lib/sqlalchemy/sql/selectable.py
@@ -2118,8 +2118,8 @@ class CTE(
             name=name,
             recursive=self.recursive,
             nesting=self.nesting,
+            # _unique_id is not need as _cte_alias is doing the link
             _cte_alias=self,
-            # _unique_id=self.unique_id,
             _prefixes=self._prefixes,
             _suffixes=self._suffixes,
         )
--- a/lib/sqlalchemy/sql/selectable.py
+++ b/lib/sqlalchemy/sql/selectable.py
@@ -2051,6 +2051,7 @@ class CTE(
             ("_cte_alias", InternalTraversal.dp_clauseelement),
             ("_restates", InternalTraversal.dp_clauseelement),
             ("recursive", InternalTraversal.dp_boolean),
+            ("nesting", InternalTraversal.dp_boolean),
         ]
         + HasPrefixes._has_prefixes_traverse_internals
         + HasSuffixes._has_suffixes_traverse_internals
--- a/lib/sqlalchemy/sql/selectable.py
+++ b/lib/sqlalchemy/sql/selectable.py
@@ -2074,6 +2074,7 @@ class CTE(
         name=None,
         recursive=False,
         nesting=False,
+        _unique_id=None,
         _cte_alias=None,
         _restates=(),
         _prefixes=None,
@@ -2083,6 +2084,9 @@ class CTE(
         self.nesting = nesting
         self._cte_alias = _cte_alias
         self._restates = _restates
+        import uuid
+
+        self.unique_id = _unique_id if _unique_id else uuid.uuid4()
         if _prefixes:
             self._prefixes = _prefixes
         if _suffixes:
@@ -2115,6 +2119,7 @@ class CTE(
             recursive=self.recursive,
             nesting=self.nesting,
             _cte_alias=self,
+            # _unique_id=self.unique_id,
             _prefixes=self._prefixes,
             _suffixes=self._suffixes,
         )
@@ -2125,6 +2130,7 @@ class CTE(
             name=self.name,
             recursive=self.recursive,
             nesting=self.nesting,
+            _unique_id=self.unique_id,
             _restates=self._restates + (self,),
             _prefixes=self._prefixes,
             _suffixes=self._suffixes,
@@ -2136,6 +2142,7 @@ class CTE(
             name=self.name,
             recursive=self.recursive,
             nesting=self.nesting,
+            _unique_id=self.unique_id,
             _restates=self._restates + (self,),
             _prefixes=self._prefixes,
             _suffixes=self._suffixes,
--- a/lib/sqlalchemy/sql/selectable.py
+++ b/lib/sqlalchemy/sql/selectable.py
@@ -2074,9 +2074,8 @@ class CTE(
         name=None,
         recursive=False,
         nesting=False,
-        _unique_id=None,
         _cte_alias=None,
-        _restates=(),
+        _restates=None,
         _prefixes=None,
         _suffixes=None,
     ):
@@ -2084,9 +2083,6 @@ class CTE(
         self.nesting = nesting
         self._cte_alias = _cte_alias
         self._restates = _restates
-        import uuid
-
-        self.unique_id = _unique_id if _unique_id else uuid.uuid4()
         if _prefixes:
             self._prefixes = _prefixes
         if _suffixes:
@@ -2118,7 +2114,6 @@ class CTE(
             name=name,
             recursive=self.recursive,
             nesting=self.nesting,
-            # _unique_id is not need as _cte_alias is doing the link
             _cte_alias=self,
             _prefixes=self._prefixes,
             _suffixes=self._suffixes,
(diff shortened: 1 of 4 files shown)


Morikko:
This is still a draft so no need for running the tests. 

But just a question: Is using uuid ok ?

I will look to simplify the code if possible next week or at least to make it more readable. Also I would like to do more tests on `alias`.

zzzeek:
hey there  -

thanks for doing this!   our CI is loaded down right now so i want to wait for it to clear out from current issues before running this through.

re: uuid, right that's too heavy for this.   to get a unique identifier local to a compiler, I would say use an integer counter that is local to the compiler and is only initialized once a CTE is encountered.

CaselIT:
we can at least make the github work run

Morikko:
> re: uuid, right that's too heavy for this. to get a unique identifier local to a compiler

Reference is linked at CTE creation time rather than compiler specific so it was not possible. But anyway, I don't use a unique id anymore, I adapt the `_restates` attribute to fulfill the role.

Morikko:
Now the merge request looks good for me.

The only thing not tested are queries like this:
```sql
WITH RECURSIVE rec(x) AS (
  SELECT 1 AS x
  UNION
  SELECT rec.x + 1 FROM rec WHERE rec.x < 10
  UNION
  SELECT rec.x + 2 FROM rec WHERE rec.x < 10
)
SELECT rec.x FROM rec;
```

But I believe sqlalchemy does not support them right ?

zzzeek:
do you mean, more than one restatement in the CTE?   im not seeing what's special about that query.

lib/sqlalchemy/sql/selectable.py
```diff
@@ -2075,13 +2075,14 @@ def _init(
         recursive=False,
         nesting=False,
         _cte_alias=None,
-        _restates=(),
+        _restates=None,
```
zzzeek, line 2078: ok, above in the cache spec:

    ("_restates", InternalTraversal.dp_clauseelement_list),

has to change to

    ("_restates", InternalTraversal.dp_clauseelement),

Morikko: Should I declare `nesting` there as well ?

Morikko:
> do you mean, more than one restatement in the CTE? im not seeing what's special about that query.

There are 2 unions (or more) referring to the same recursive query. Those are called non linear recursive queries.

zzzeek:
> > do you mean, more than one restatement in the CTE? im not seeing what's special about that query.
> 
> There are 2 unions (or more) referring to the same recursive query. Those are called non linear recursive queries.

i wouldn't think it makes a difference?  but i guess you have been deep in this implemenation for awhile so is there some reason that's not going to work?  is this because you changed "restates" to be a single CTE and not a collection?  that change seems a little scary

zzzeek:
if that was working previously and isn't now it means this introduces a regression so we would have to work that out.

lib/sqlalchemy/sql/selectable.py
zzzeek: probably.  it's not very easy for it to make a difference but yes, it should be declared with a dp_boolean

Morikko:
> > > do you mean, more than one restatement in the CTE? im not seeing what's special about that query.
> > 
> > 
> > There are 2 unions (or more) referring to the same recursive query. Those are called non linear recursive queries.
> 
> i wouldn't think it makes a difference? but i guess you have been deep in this implementation for awhile so is there some reason that's not going to work? is this because you changed "restates" to be a single CTE and not a collection? that change seems a little scary

To be honest I don't know if it worked :sweat_smile: I just know it is a valid SQL query. But I can't generate the query with what would be the way for me: 

```python
nesting_cte = select(literal(1).label("inner_cte")).cte(
    "nesting", recursive=True
)
rec_part = select(nesting_cte.c.inner_cte).where(
    nesting_cte.c.inner_cte == literal(1)
)
rec_part_2 = select(nesting_cte.c.inner_cte).where(
    nesting_cte.c.inner_cte == literal(2)
)
complete_nesting_cte = nesting_cte.union(rec_part).union(rec_part_2)

select(complete_nesting_cte.c.inner_cte)

# Raises
# AttributeError: 'CompoundSelect' object has no attribute 'union'
```

I present this to double check and prevent another bug but maybe it was not supported as well before. If you don't know how to generate it as well, I would say it was not.

zzzeek:
API missing.  would you be able to integrate this?

```diff
diff --git a/lib/sqlalchemy/sql/selectable.py b/lib/sqlalchemy/sql/selectable.py
index 8f8e6b2a7e..9c5e59274c 100644
--- a/lib/sqlalchemy/sql/selectable.py
+++ b/lib/sqlalchemy/sql/selectable.py
@@ -2119,9 +2119,9 @@ class CTE(
             _suffixes=self._suffixes,
         )
 
-    def union(self, other):
+    def union(self, *other):
         return CTE._construct(
-            self.element.union(other),
+            self.element.union(*other),
             name=self.name,
             recursive=self.recursive,
             nesting=self.nesting,
@@ -2130,9 +2130,9 @@ class CTE(
             _suffixes=self._suffixes,
         )
 
-    def union_all(self, other):
+    def union_all(self, *other):
         return CTE._construct(
-            self.element.union_all(other),
+            self.element.union_all(*other),
             name=self.name,
             recursive=self.recursive,
             nesting=self.nesting,
@@ -6233,19 +6233,19 @@ class Select(
         else:
             return SelectStatementGrouping(self)
 
-    def union(self, other, **kwargs):
+    def union(self, *other, **kwargs):
         """Return a SQL ``UNION`` of this select() construct against
         the given selectable.
 
         """
-        return CompoundSelect._create_union(self, other, **kwargs)
+        return CompoundSelect._create_union(self, *other, **kwargs)
 
-    def union_all(self, other, **kwargs):
+    def union_all(self, *other, **kwargs):
         """Return a SQL ``UNION ALL`` of this select() construct against
         the given selectable.
 
         """
-        return CompoundSelect._create_union_all(self, other, **kwargs)
+        return CompoundSelect._create_union_all(self, *other, **kwargs)
 
     def except_(self, other, **kwargs):
         """Return a SQL ``EXCEPT`` of this select() construct against

```

demo:

```py
from sqlalchemy import select, column, union
from sqlalchemy.sql.elements import literal_column

stmt = select(literal_column("1").label("x"))

cte = stmt.cte(recursive=True)

u1 = select(cte.c.x + 1).where(cte.c.x < 10)
u2 = select(cte.c.x + 2).where(cte.c.x < 20)

cte = cte.union(u1, u2)

stmt = select(cte.c.x)

print(stmt)
```

Morikko:
I will do it in another merge request. I am glad you are so open for adding changes :+1:  Let's first close this bug.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 2633f34f7f5336a4a85bd3f71d07bca33ce27a2c of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 2633f34f7f5336a4a85bd3f71d07bca33ce27a2c: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3133

sqla-tester:

Hi Eric -

I've pushed another patch locally to Gerrit.  Can you review my changes over here?  you can view either the change against current git master or you can select a diff between patchsets 2 and 3.   I've consolidated some of the variables and also changed a conditional that was coming up as ambiguous under pylance.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3133

sqla-tester:

the nova thing is a false positive, they havent updated their stuff to work w the most recent oslo.db.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3133

PR #7143, opened 2021-10-04
📚 DOCS: Update insertion performance script

chrisjsewell:
### Description

Heya, given the script was still in Python 2.7, I thought it was high time to update!

Updated to Python 3, and to use `future=True`.
Also added outputs for `return_default=True`.

One reason I did this, is that in https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session.bulk_insert_mappings, there is a warning that states:

> Please note that newer versions of SQLAlchemy are greatly improving the efficiency of the standard flush process.
> It is strongly recommended to not use the bulk methods as they represent a forking of SQLAlchemy’s functionality and are slowly being moved into legacy status

Given the performance numbers, this does not really seem to be the case 😬 (they are still about the same).
Is there any new insertion methods / tricks I am missing?

(perhaps also I should add a reference in these warnings to this performance data)

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/doc/build/faq/performance.rst
+++ b/doc/build/faq/performance.rst
@@ -313,36 +313,61 @@ a small degree of ORM-based automation.
 
 The example below illustrates time-based tests for several different
 methods of inserting rows, going from the most automated to the least.
-With cPython 2.7, runtimes observed::
+With cPython, runtimes observed::
 
-    SQLAlchemy ORM: Total time for 100000 records 6.89754080772 secs
-    SQLAlchemy ORM pk given: Total time for 100000 records 4.09481811523 secs
-    SQLAlchemy ORM bulk_save_objects(): Total time for 100000 records 1.65821218491 secs
-    SQLAlchemy ORM bulk_insert_mappings(): Total time for 100000 records 0.466513156891 secs
-    SQLAlchemy Core: Total time for 100000 records 0.21024107933 secs
-    sqlite3: Total time for 100000 records 0.137335062027 sec
+    Python: 3.8.12 | packaged by conda-forge | (default, Sep 29 2021, 19:42:05)  [Clang 11.1.0 ]
+    sqlalchemy v1.4.22 (future=True)
+    SQLA ORM:
+            Total time for 100000 records 5.722 secs
+    SQLA ORM pk given:
+            Total time for 100000 records 3.781 secs
+    SQLA ORM bulk_save_objects:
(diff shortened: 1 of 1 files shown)


zzzeek:
> ### Description
> 
> Heya, given the script was still in Python 2.7, I thought it was high time to update!
> 
> Updated to Python 3, and to use `future=True`. Also added outputs for `return_default=True`.
> 
> One reason I did this, is that in https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session.bulk_insert_mappings, there is a warning that states:
> 
> > Please note that newer versions of SQLAlchemy are greatly improving the efficiency of the standard flush process.

you would need to use the psycopg2 driver for this.  we have made a [dramatic improvement](https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#orm-batch-inserts-with-psycopg2-now-batch-statements-with-returning-in-most-cases) to how we treat that driver which we will be generalizing to other postgresql drivers.  Postgresql is where we get the most complaints about ORM insert performance due to this specific issue.



> > It is strongly recommended to not use the bulk methods as they represent a forking of SQLAlchemy’s functionality and are slowly being moved into legacy status
> 
> Given the performance numbers, this does not really seem to be the case grimacing (they are still about the same). Is there any new insertion methods / tricks I am missing?

as the suite shows, use the insert() construct directly.   There is too much confusion over what ORM features "bulk" does and does not accommodate and i have no resources to support that.   Users nowadays find these "bulk" methods first, not even knowing about add(), and report all the missing features as bugs.  huge mistake.  

the best approach we have right now is regular insert() and you can get ORM objects back from it using the approach at https://docs.sqlalchemy.org/en/14/orm/persistence_techniques.html#using-postgresql-on-conflict-with-returning-to-return-upserted-orm-objects .   you do need to batch the inserts though because sending a single VALUES call with 100k rows is likely too large.

zzzeek:
looks good as is.  thanks!

chrisjsewell:
> you would need to use the psycopg2 driver for this. we have made a dramatic improvement to how we treat that driver which we will be generalizing to other postgresql drivers. Postgresql is where we get the most complaints about ORM insert performance due to this specific issue.

Interesting, thanks for the info!

PR #7162, opened 2021-10-08
Add missing * and - operators for PostgreSQL range types

lpsinger:
### Description
Add missing range `*` and `-` operators for PostgreSQL range types.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**


CaselIT:
I believe that the reason is that the default `+` is used for concatenation and renders `||`. But I may be wrong here

CaselIT:
I think we can close this, thanks for the change in any case!

PR #7195, opened 2021-10-16
Fix inline documentation for Index.create() and Index.drop(): bind argument required in 2.0

jdavcs:
I think this documentation line is missing: it should be the same as for `Table.create` ([see this line](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/sql/schema.py#L939)) and `Table.drop`, for the same reason ([Implicit” and “Connectionless” execution, “bound metadata” removed](https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#migration-20-implicit-execution)).

I did a "sanity test" with unbound metadata and got the expected error: 
`sqlalchemy.exc.UnboundExecutionError: Index object 'ix1' is not bound to an Engine or Connection.  Execution can not proceed without a database to execute against.`

I assume this is minor and does not require an issue and a test.
(First PR here, sorry if I messed up.)

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/sql/schema.py
+++ b/lib/sqlalchemy/sql/schema.py
@@ -4195,6 +4195,9 @@ class Index(DialectKWArgs, ColumnCollectionMixin, SchemaItem):
         :class:`.Index`, using the given :class:`.Connectable`
         for connectivity.
 
+        .. note:: the "bind" argument will be required in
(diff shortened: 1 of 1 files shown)


CaselIT (approved):
Thanks!

PR #7197, opened 2021-10-16
Update reserved words for MySQL dialect

### LINKED ISSUE #7167: Add MySQL Reserved Word 'lead'
### Describe the bug

Using MariaDB 10.6.4 with a database table called `lead`.  Turns out this is a new reserved word in MariaDB.   I am using SQLAlchemy 1.3.24 and adding `lead` to your `dialects/mysql/base.py` `RESERVED_WORDS` set fixes the issue, causing the `lead` table to be escaped properly.  Looking at your latest 1.4.x branch, `lead` is not in there either.

I know 1.3.24 is old, but still used in the industry.  If you could backport that reserved word in 1.3 that would be appreciated.

### To Reproduce

```python
Run a JOIN on a table called `lead` on MariaDB 10.6.4.  An example SQL produced will be 

SELECT DISTINCT opportunities.id, opportunities.name, opportunities.lead_id, lead.id AS "lead__id", lead.name
AS "lead__name", lead.primary_contact_id AS "lead__primary_contact_id" FROM opportunities LEFT OUTER JOIN
leads AS lead ON opportunities.lead_id = lead.id;
```

Which fails with `"You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'lead ON opportunities.lead_id = lead.id'` because `lead` is not escaped properly as it is not in your set of RESERVED words.
```


### Error

```
"You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'lead ON opportunities.lead_id = lead.id'
``` 

### Versions

- OS: Linux Manjaro
- Python: 3.9.0
- SQLAlchemy: 1.3.24
- Database: MariaDB 10.6.4
- DBAPI (eg: psycopg, cx_oracle, mysqlclient): aiomysql via encode/databases


### Additional context

_No response_

### PR DESCRIPTION
**Pull request by kkirsche**
Update MySQL dialect with reserved words for MariaDB 10.6 (excluding function names in the second symbol list).

Fix: #7167

### Description
This pull request updates the list of reserved words from the MariaDB 10.6 lexer (https://github.com/MariaDB/server/blob/10.6/sql/lex.h) list of symbol names. It does not include the list of function names though.

### Checklist

This pull request is:

- [X] A short code fix

**Have a nice day!**

### CODE REVIEW

**Comment by kkirsche** [comment]
I would need to do that another time as I was doing that from an iPad while I had some free time.

**Comment by CaselIT** [comment]
Some test also need updating. I can finish this if you prefer. Thanks for getting the ball rolling :)

**Review by sqla-tester — COMMENTED** [review]
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision ce4dfa61a520d81054ad9a8caf76f13cf0f03518 of this pull request into gerrit so we can run tests and reviews and stuff

**Comment by sqla-tester** [comment]
New Gerrit review created for change ce4dfa61a520d81054ad9a8caf76f13cf0f03518: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3171

**Comment by kkirsche** [comment]
Changes made so far:

* Refactored the list of reserved words to a standalone file with a list of links to MySQL and MariaDB lexers. The list of MariaDB lexers is based on the currently supported versions (https://mariadb.com/kb/en/mariadb-server-release-dates/). MySQL doesn't have many named branches, so I simply grabbed those four.
* Fixed the use of the keyword `data` in tests
* (Hopefully) fixed the use of the keyword `name` in tests

**Comment by kkirsche** [comment]
Rough one-liner just was working on to simplify pulling the reserved words from the various lex files.

MySQL:
```
cat lex.h|grep '{SYM'|cut -d '"' -f 2 | awk '{ print "\""tolower($0)"\","}'
```

MariaDB:
```
cat lex.h|grep '{ "'|cut -d '"' -f 2 | awk '{ print "\""tolower($0)"\","}'
```

PR #7207, opened 2021-10-18
fix: Update reserved words list of MySQL / MariaDB dialect

### LINKED ISSUE #7167: Add MySQL Reserved Word 'lead'

kkirsche:
### Description
Fixes: #7167 
Supercedes: #7197 

This is two commits:
1. Move reserved words to it's own file.
2. Add missing reserved words from https://mariadb.com/kb/en/reserved-words/
    * Note: this only adds MariaDB though links to MySQL, it also does not include the reserved words for Oracle mode, as listed in the link.

### Checklist

This pull request is:

- [X] A short code fix

**Have a nice day!**

Sorry @CaselIT and @zzzeek for the new request.

Edit:
Force push was to fix a formatting issue with too many blank lines.


kkirsche:
I would kindly recommend ensuring that core members include concerns about merge requests on the actual merge request to ensure that users contributing do not unnecessarily waste time and can more quickly locate discrepancies such as the lex.h list of symbols isn't identical to the reserved words lists.

I understand that it seems it was a source of confusion, as comments, we also misdirected in the issue, but it is demoralizing to contribute and have that occur. We're all human, and miscommunications occur, so I hope that doesn't come off the wrong way. I'm sure you get a lot of emails and issues and it's hard to keep track of who is who. Thanks again for the great library, cheers

kkirsche:
All good, like I said, things happen. I've updated the MR with the comments we had originally discussed @CaselIT against the reserved words list from MariaDB. I have not updated against the commented MySQL, but have included it in the comment.

zzzeek:
one thing that might be useful is, is the current list that we have of reserved words useful?  how is it that there's a long list of words yet we can create tables and columns that use those words without quoting?   i wonder if i can create a table with a really really reserved word, like "table" as the name, that would be strange?

```
MariaDB [test]> create table table (column integer, varchar varchar(20));
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'table (column integer, varchar varchar(20))' at line 1
MariaDB [test]> create table `table` (`column` integer, `varchar` varchar(20));
Query OK, 0 rows affected (0.040 sec)

```

OK so there are some words that *are* really reserved, using those obvious names above.   So what category are those words in, compared to a word like [CURRENT_POS](https://github.com/MariaDB/server/blob/10.6/sql/lex.h#L161), [CYCLE](https://github.com/MariaDB/server/blob/10.6/sql/lex.h#L168) or [DATA](https://github.com/MariaDB/server/blob/10.6/sql/lex.h#L169) ? :+1: 

```
MariaDB [test]> create table current_pos (cycle integer, data integer);
Query OK, 0 rows affected (0.016 sec)
```

kkirsche:
https://mariadb.com/kb/en/identifier-names/ hints that they are "Identifier quotes" 

As they say:

> The following characters are valid, but identifiers using them must be quoted:
> * ASCII: U+0001 .. U+007F (full Unicode Basic Multilingual Plane (BMP) except for U+0000)
> * Extended: U+0080 .. U+FFFF
> * Identifier quotes can themselves be used as part of an identifier, as long as they are quoted.

Can't seem to find yet where that points to

zzzeek:
> https://mariadb.com/kb/en/identifier-names/ hints that they are "Identifier quotes"
> 
> As they say:
> 
> > The following characters are valid, but identifiers using them must be quoted:
> > 
> > * ASCII: U+0001 .. U+007F (full Unicode Basic Multilingual Plane (BMP) except for U+0000)
> > * Extended: U+0080 .. U+FFFF
> > * Identifier quotes can themselves be used as part of an identifier, as long as they are quoted.
> 
> Can't seem to find yet where that points to
> 
> EDIT: but that seems like a circular loop. Let me see if I can find something in their code...

it links to [this](https://mariadb.com/kb/en/reserved-words/).  this looks like "really really" reserved words.    taking two that look unfamiliar, they are really reserved:

```
MariaDB [test]> create table cross (dec integer);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'cross (dec integer)' at line 1
MariaDB [test]> create table cross (`dec` integer);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'cross (`dec` integer)' at line 1
MariaDB [test]> create table `cross` (dec integer);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'dec integer)' at line 1
MariaDB [test]> create table `cross` (`dec` integer);
Query OK, 0 rows affected (0.015 sec)

```

kkirsche:
OK, that's where I had pulled this list of new words. Potentially it would make sense to differentiate the MariaDB and MySQL lists, as there were a number here that are not in MariaDB's?

kkirsche:
Based on:
https://docs.sqlalchemy.org/en/14/dialects/

You support MySQL 5.6+, MariaDB 10.4+

If you'd like, I can re-do the list of reserved words using:

* https://dev.mysql.com/doc/refman/8.0/en/keywords.html
* https://dev.mysql.com/doc/refman/5.7/en/keywords.html
* https://dev.mysql.com/doc/refman/5.6/en/keywords.html
* https://mariadb.com/kb/en/reserved-words/

Where from the MySQL links I use the `(R)` indicated keywords, and include comments related to version differences, such as `array` (`ARRAY; added in 8.0.17 (reserved); became nonreserved in 8.0.19`).

kkirsche:
MariaDB wordlist verified

CaselIT:
the comments on where to find the list were removed by the commit 697b8c5996ec7bf1f1eb22142cd3f853b22e9749

CaselIT:
No, that's an issue that happens sometimes, particularly in py2.7, when dealing with the gc.

It's not this change that caused it, it just happens sometimes here on github

kkirsche:
MySQL verified against 8.0 — `"slow"` was the only one that wasn't in 8.0 but is in the 5.6 one as it was caught by:

```
SLOW; became nonreserved in 5.6.1
```

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 233ad5c79cab049e4f81a7448e4e47b3d83e45c1 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 233ad5c79cab049e4f81a7448e4e47b3d83e45c1: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3177

zzzeek:
OK I've looked and everything that is here is perfect.

as far as what's not yet here, which what would be nice, since we have that new conditional  here, would be a quick test to validate that a "mysql" and "mariadb" dialect have different reserved word behaviors.

there is an extremely rudimentary test at https://github.com/sqlalchemy/sqlalchemy/blob/e86c40b9254f25ca9a765622f911c6dbd4bd2f1f/test/dialect/mysql/test_compiler.py#L69 .  would propose another one of similar simplicity that looks like:

```py
    def test_reserved_words_mysql_vs_mariadb(self):
        table = Table(
            "rw_table",
            MetaData(),
            Column("lead", Integer),
            Column("master_ssl_verify_server_cert", Integer),
            Column("page_checksum", Integer),
        )
        x = select(table)

        self.assert_compile(
            x,
            "SELECT rw_table.lead, "
            "rw_table.`master_ssl_verify_server_cert`, "
            "rw_table.`page_checksum` FROM rw_table",
            dialect="mariadb"
        )

        self.assert_compile(
            x,
            "SELECT rw_table.`lead`, "
            "rw_table.`master_ssl_verify_server_cert`, "
            "rw_table.page_checksum FROM rw_table",
            dialect="mysql"
        )
```

now we can note that if either database adds "lead" or "page_checksum" then the test above will fail.   I dont think that's happening soon, but if we wanted to guard against it we could instead patch an artificial value into each of the MARIADB_RESERVED_WORDS and MYSQL_RESERVED_WORDS sets instead.  either way is fine by me.

kkirsche:
Github tests passed with the new comment-based unit test following the line-length formatting fix in 5a682e3.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 5a682e331069520ccad9e6bf3cc5e4a77a889ef0 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:

I think this is ready it just needs a changelog file, which I can add on the gerrit side.    just waiting for all +2s

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3177

PR #7315, opened 2021-11-12
add is_literal flag to TextClause

villebro:
### Description
Currently `TextClause` doesn't support a literal mode like `ColumnClause` does. While strings prefixed by colons can be escaped with a backslash, it would be convenient to be able to explicitly state that a `TextClause` doesn't include any binds, especially given the fact that colons can't be universally escaped (see added docs). This need is typical in scenarios where the input text is known not to contain any binds. See #125 and https://stackoverflow.com/questions/49902843/avoid-parameter-binding-when-executing-query-with-sqlalchemy for prior similar needs.

This PR introduces a new parameter `is_literal` to `TextClause`, which skips parsing the text for bound parameters. The PR also refines the regular expression that is used to detect bound parameters, which currently incorrectly detects the following binds:
- `:foo:bar` is detected as the bind parameter `fo` due to matching until the last non-colon word character and are added to `self._bindparams`. The same applies to `:foo::TIMESTAMP`. It is unclear to me why these don't later raise an exception during compilation, but apparently there is some additional logic downstream that removes these false positives.

### Checklist
This pull request is:

### DIFF
--- a/lib/sqlalchemy/sql/elements.py
+++ b/lib/sqlalchemy/sql/elements.py
@@ -1754,7 +1754,7 @@ class TextClause(
 
     _is_textual = True
 
-    _bind_params_regex = re.compile(r"(?<![:\w\x5c]):(\w+)(?!:)", re.UNICODE)
+    _bind_params_regex = re.compile(r"(?<![:\w\x5c]):(\w+)(:)?", re.UNICODE)
     _execution_options = Executable._execution_options.union(
         {"autocommit": PARSE_AUTOCOMMIT}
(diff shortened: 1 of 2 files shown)


CaselIT:
Hi,

Please one an issue before opening a pr, so we can discuss feasibility and provide guidance, thanks

villebro:
Thanks @CaselIT - I think I found a simpler solution to this - I'll get back to this later, either in the form of an issue for discussion or a PR to refine the documentation.

PR #7437, opened 2021-12-10
Add a note on using server_onupdate=FetchedValue()

abrookins:
Add a note on using `server_onupdate=FetchedValue()` when using SQL expressions with `onupdate`.

### Description
My team encountered an issue with using a SQL expression with `onupdate`.
Despite the dialect (PG) supporting `RETURNING`, we needed to mark the column with
`server_onupdate=FetchedValue()` in order to get the column used with `onupdate`
to appear in the `RETURNING` clause of `UPDATE` statements.

This was not clear from the documentation, so I want to make it crystal clear for other
folks defining similar columns.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/doc/build/orm/persistence_techniques.rst
+++ b/doc/build/orm/persistence_techniques.rst
@@ -306,36 +306,6 @@ above table will look like:
 
    INSERT INTO my_table DEFAULT VALUES RETURNING my_table.id, my_table.timestamp, my_table.special_identifier
 
-Usage with server-generated ``onupdate`` values
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-If you are using the :paramref:`.orm.mapper.eager_defaults` flag on a table
-definition with any columns marked with :paramref:`_schema.Column.onupdate`
-that use a server-side function to generate the column's value,
-you should mark the column with ``server_default=FetchedValue()``.
-
-Doing so ensures that SQLAlchemy includes these columns in the RETURNING clause
-on dialects that support RETURNING.
-
-This is especially useful for columns that track the "updated" time of a row
-using server-side generated timestamps, like the following example:
-
-    class MyModel(Base):
-        __tablename__ = 'my_table'
-
-        id = Column(Integer, primary_key=True)
-        created = Column(DateTime(), server_default=func.now())
-
-        # Use the dialect-appropriate ``now()`` function to generate an updated
-        # time and include the "updated" column in the RETURNING clause.
-        updated = Column(DateTime(), onupdate=func.now(), server_default=FetchedValue())
-
-        __mapper_args__ = {"eager_defaults": True}
-
-Without ``server_default=FetchedValue()``, SQLAlchemy will make a SELECT after
-an INSERT or UPDATE to get the server-generated value for the column even if
-the dialect supports RETURNING.
-
 
 Case 2: non primary key, RETURNING or equivalent is not supported or not needed
 --------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


zzzeek:
hi -

these docs aren't correct, so we need to first work out what you actually want to achieve and the correct way to do it first.     additionally, there might be a bug here in any case.

As given, the "func.now()" is not part of an INSERT, only part of an UPDATE.  server_default has nothing to do with an UPDATE, and in the example given here RETURNING is not used correctly, there's a second SELECT emitted that actually loads the value.

what does add the column to RETURNING is to use [server_onupdate](https://docs.sqlalchemy.org/en/14/core/metadata.html?highlight=server_onupdate#sqlalchemy.schema.Column.params.server_onupdate).   Try out the demonstration below.

without server_default or server_onupdate, the eager defaults in this case seems to be failing to use RETURNING or a SELECT statement.  that looks mostly like a bug and it's likely because it is assumed that a Python side onupdate is a Python function, but it would be most ideal if it just used RETURNING for all columns with any kind of default so that no directives are needed.

So steps here are 

1. new bug report that using client side "default" or "onupdate" that nonetheless is detected as being a SQL expression which is ultimately embedded into the INSERT or UPDATE should participate in eager defaults automatically, using RETURNING if available and never a SELECT afterwards as the case below illustrates actually happens right now
2. docs for the moment should indicate server_onupdate is the current signal to get these values back
3. the docs should be after all the other cases here

zzzeek:
```py
from sqlalchemy import Column, String
from sqlalchemy import DateTime
from sqlalchemy import FetchedValue
from sqlalchemy import func
from sqlalchemy.orm.base import MANYTOONE
Base = declarative_base()


class MM1(Base):
    __tablename__ = 'my_table1'

    data = Column(String)
    created = Column(DateTime(), server_default=func.now())

    # updated with the server default
    updated = Column(DateTime(), onupdate=func.now(), server_default=FetchedValue())

    __mapper_args__ = {"eager_defaults": True}

class MM2(Base):
    __tablename__ = 'my_table2'

    data = Column(String)
    created = Column(DateTime(), server_default=func.now())

    # updated without the server default
    updated = Column(DateTime(), onupdate=func.now())


    __mapper_args__ = {"eager_defaults": True}

class MM3(Base):
    __tablename__ = 'my_table3'

    data = Column(String)
    created = Column(DateTime(), server_default=func.now())

    # updated with the server onupdate
    updated = Column(DateTime(), onupdate=func.now(), server_onupdate=FetchedValue())

    __mapper_args__ = {"eager_defaults": True}


e = create_engine("postgresql://scott:tiger@localhost/test", echo='debug')
Base.metadata.drop_all(e)
Base.metadata.create_all(e)

s = Session(e)

mm1, mm2, mm3 = MM1(), MM2(), MM3()
s.add_all([mm1, mm2, mm3])
s.flush()

# all are None, no value inserted
assert mm1.updated is None
assert mm2.updated is None
assert mm3.updated is None

mm1.data = 'new data'
mm2.data = 'new data'
mm3.data = 'new data'

s.flush()

# server_default was loaded, *however* if you look in the SQL log
# you will see:
# UPDATE my_table1 SET data=%(data)s, updated=now() WHERE my_table1.id = %(my_table1_id)s
# 2021-12-10 16:08:23,138 INFO sqlalchemy.engine.Engine SELECT my_table1.updated AS my_table1_updated
# FROM my_table1 WHERE my_table1.id = %(pk_1)s
# i.e. no RETURNING
assert mm1.__dict__['updated'] is not None

# server_onupdate was loaded correctly, SQL log shows:
# UPDATE my_table3 SET data=%(data)s, updated=now() WHERE my_table3.id = %(my_table3_id)s RETURNING my_table3.updated
assert mm3.__dict__['updated'] is not None

# neither default or onupdate was loaded, fails
assert mm2.__dict__['updated'] is not None



```

abrookins:
Hey Mike! It's not every day that I get to chat on GitHub with an absolute legend. 🤩 

Uhm, thanks for reading my text closely because I banged it out without reviewing closely, it appears. You're right -- what solved this problem for us was `server_onupdate=FetchedValue()` (typo on my part). I will fix that and move this case after the others.

The actual problem we had was that a model equivalent to **MM1** caused SELECT queries immediately after any UPDATE to retrieve the server-generated `updated` timestamp. This caused lots of extra queries.

Your examples were helpful and illustrate the problem. After some investigation, I figured SQLAlchemy wasn't aware that we were using a server default and thus didn't include the column in a RETURNING clause, so I added `server_onupdate=FetchedValue()`, which fixed it. And then rushed to submit some docs. :)

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 4845fb09a90ab58f0ae882e0d335ddba09b32ca0 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 4845fb09a90ab58f0ae882e0d335ddba09b32ca0: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3859

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3859 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3860 has been **merged**. Congratulations! :)

PR #7443, opened 2021-12-12
Add reflection for Postgres expression-based indexes

### LINKED ISSUE #7442: Postgres: add support for reflection of expression-based indexes
### Describe the use case

We are using SQLAlchemy to work with Postgres DB that has full-text indexes (`to_tsvector` and so on), and whenever those need a change it is unfortunate that alembic is unable to generate the automated diff. One reason for that is that `postgres` dialect currently does not support reflecting expression-based indexes.

It is possible to live without autogenerated migrations, but we had cases when a schema change has been missed or a typo in the index expression slipped by.

Postgres (any drivers)

### Example Use

Create table with an expression-based index

```sql
create table t123123 (x varchar, y varchar);
create index t123123idx_xlowery on t123123 ((x || lower(y)));
```

Reflect the index from the live table:
```python
import sqlalchemy as sa
engine = sa.create_engine('postgresql://scott:tiger@172.17.0.1:5432/test')
meta = sa.MetaData()
user_table = sa.Table('t123123', meta)
insp = sa.inspect(engine)
insp.reflect_table(user_table, None)

# Current implementation will fail on the next line, because t123123idx_xlowery index is not reflected
xlowery_idx = [i for i in user_table.indexes if i.name == 't123123idx_xlowery'][0]
print(str(sa.schema.CreateIndex(xlowery_idx).compile()))
# CREATE INDEX t123123idx_xlowery ON t123123 (CAST(x AS TEXT) || lower(CAST(y AS TEXT)))
```

### Additional context

Adding `pg_get_expr(ix.indexprs, ix.indrelid)` to the list of index metadata requested in  `sa.dialects.postgres.base.PGDialect.get_indexes` should be straightforward. Also, there are parsers for Postgres code (https://github.com/pganalyze/libpg_query and derivatives) that should make the process of converting Postgres code into SA expression objects easier.

immerrr:
### Description

This PR is a proof-of-concept implementation of reflection for expression-based indexes in Postgres.

It is intentionally suboptimal and rough on the edges to start the discussion ASAP, and I'm fine with the first implementation being thrown away completely.

It also simplifies slightly the code for processing the indexes (avoiding a join that would generate multiple rows for a single index), but I think I'm going to have to revert that as `LATERAL` subqueries were only added in PG 9.3.

Fixes #7442 

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -1375,14 +1375,19 @@ E.g.::
 
 from collections import defaultdict
 import datetime as dt
+import decimal
 import re
 from uuid import UUID as _python_UUID
+
+import psqlparse
 
 from . import array as _array
 from . import hstore as _hstore
 from . import json as _json
 from . import ranges as _ranges
+from ... import cast
 from ... import exc
+from ... import func
 from ... import schema
 from ... import sql
 from ... import util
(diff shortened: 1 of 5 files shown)


CaselIT:
thanks for the effort on this, This will be completed on https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3969

Issue #7594, opened 2022-01-21
failure mode is premature for INSERT that returns NULL when using server generated PKs, regression from 1.3


akeeman:
### Describe the bug

When a pg before insert trigger causes an insert to be skipped (i.e. returns NULL instead of a record) there is no returning record. If a table has a primary key with server default (e.g. auto increment), and no value is set for that column then an exception will be thrown when the record insert is persisted.

From the 1.4 branch:
In the following snipped `primary_key` will be `None`. Therefore `zip` will raise a `TypeError` ("'NoneType' object is not iterable").
https://github.com/sqlalchemy/sqlalchemy/blob/ab42b0e3d98386c8a13edea3206ef43f018de3b6/lib/sqlalchemy/orm/persistence.py#L1227-L1236

From the 1.3 branch:
In 1.3 this is not a problem, as the code will not run.
https://github.com/sqlalchemy/sqlalchemy/blob/57f324deaf261400d67b07eda114ebc744d47763/lib/sqlalchemy/orm/persistence.py#L1139-L1150

### To Reproduce

```python
from sqlalchemy import DDL, Column, Integer, event
from sqlalchemy.engine import create_engine
from sqlalchemy.ext.declarative import declarative_base
from sqlalchemy.orm import Session

Base = declarative_base()

# the trigger function that always skips the insert itself
fn_ddl = DDL(
    """\
CREATE OR REPLACE FUNCTION the_trigger()
RETURNS TRIGGER AS $the_trigger$
BEGIN
RETURN NULL;
END;
$the_trigger$ LANGUAGE plpgsql
VOLATILE
SECURITY INVOKER
PARALLEL SAFE
COST 1;
"""
)
event.listen(Base.metadata, "before_create", fn_ddl)

# the entity with a pk col that has a server default value
# (auto incrementing id)
class A(Base):
    __tablename__ = "a"

    id = Column(Integer, primary_key=True)


# run the trigger function before insert
trig_ddl = DDL(
    """\
CREATE TRIGGER a_trigger BEFORE INSERT
ON a
FOR EACH ROW EXECUTE PROCEDURE
the_trigger();
"""
)
event.listen(A.__table__, "after_create")

# perform an insert action
dsn = "postgresql://user:password@localhost:5432/test"
engine = create_engine(dsn, echo=True)
Base.metadata.create_all(engine)
session = Session(engine)
try:
    session.add(A())
    session.commit()
finally:
    session.close()
```

### Error

```
[...]

PR #7595, opened 2022-01-21
don't update state when db did not really insert

akeeman:
Fixes: https://github.com/sqlalchemy/sqlalchemy/issues/7594

### Description

When a pg before insert trigger causes an insert to be skipped (i.e. returns NULL instead of a record) there is no returning record. If a table has a primary key with server default (e.g. auto increment), and no value is set for that column then an exception will be thrown when the record insert is persisted.

Solution: Don't set primary key attributes when there is none. This is what SQLAlchemy <1.4 did.

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

### No test yet
Can anyone point me in the right direction on how to write a test for this?

### DIFF
--- a/lib/sqlalchemy/orm/persistence.py
+++ b/lib/sqlalchemy/orm/persistence.py
@@ -1225,15 +1225,16 @@ def _emit_insert_statements(
                         )
 
                     primary_key = result.inserted_primary_key
-                    for pk, col in zip(
-                        primary_key, mapper._pks_by_table[table]
-                    ):
-                        prop = mapper_rec._columntoproperty[col]
-                        if (
-                            col in value_params
-                            or state_dict.get(prop.key) is None
+                    if primary_key is not None:
+                        for pk, col in zip(
+                            primary_key, mapper._pks_by_table[table]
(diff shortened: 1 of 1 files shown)


zzzeek:
hey there -

the main thing we need with PRs is tests and in this case the test is pretty tricky to write.  the fix for the issue wtihin persistence.py is just:

```diff
diff --git a/lib/sqlalchemy/orm/persistence.py b/lib/sqlalchemy/orm/persistence.py
index 6696b34d59..b1479d2c1b 100644
--- a/lib/sqlalchemy/orm/persistence.py
+++ b/lib/sqlalchemy/orm/persistence.py
@@ -1225,6 +1225,8 @@ def _emit_insert_statements(
                         )
 
                     primary_key = result.inserted_primary_key
+                    if primary_key is None:
+                        continue
                     for pk, col in zip(
                         primary_key, mapper._pks_by_table[table]
                     ):

```

as noted in the issue, this is still an exception case, just it has to raise the correct exception.  fix is coming, thanks anyway for the PR!

PR #7589, opened 2022-01-19
Fix overlapping slots, base classes without slots

ariebovenberg:
Some `__slots__` were not in order.

Fixes #7527

### Description

I'm fixing two types of slots mistakes:
- [x] remove overlapping slots (i.e. slots already defined on a base class)
- [x] fix broken inheritance (i.e. slots class inheriting from a non-slots class)
  - [x] slots added to base class `TransactionalContext`. It seemed to use two attributes, which I've added as slots.
  - [x] empty slots removed from `ORMOption`. Its base class explicitly makes use of `__dict__` so empty slots don't add anything.
  - [x] empty slots added to `PostLoader`. It doesn't appear to use any slots not already defined on its base classes.
  - [x] empty slots added to `IterateMappersMixin`. It doesn't appear to use any slots not already defined on its subclasses.
  - [x] empty slots added to `ImmutableContainer`. It doesn't use any fields.
  - [x] empty slots added to `OperatorType`. It's a protocol.
  - [x] empty slots added to `InternalTraversal`, `_HasTraversalDispatch`. They don't seem to use attributes on their own.

### Checklist

This pull request is:

- [x] A short code fix
	- please include the issue number, and create an issue if none exists, which
	  must include a complete example of the issue.  one line code fixes without an
	  issue and demonstration will not be accepted.
	- Please include: `Fixes: #<issue number>` in the commit message
	- please include tests.   one line code fixes without tests will not be accepted.

**Have a nice day!**

PR #7670, opened 2022-02-05
add slotscheck to CI

ariebovenberg:
### Description

As discussed in #7589, `slotscheck` can prevent slots-related mistakes from creeping back in.

Note that because slotscheck imports files, `additional_dependencies` was needed in the pre-commit config. The versions of these dependencies are not kept in sync with `sqlalchemy` requirements. If this is not OK, there are [alternative setups](https://slotscheck.readthedocs.io/en/latest/advanced.html#pre-commit-hook) or the check could be moved to `tox`.

### Checklist
- [x] A short code fix

**Have a nice day!**

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -151,7 +151,7 @@ commands =
     pytest -m mypy {posargs}
 
 # thanks to https://julien.danjou.info/the-best-flake8-extensions/
-[testenv:pep8]
+[testenv:lint]
 basepython = python3
 deps=
       flake8
@@ -164,9 +164,12 @@ deps=
       pydocstyle
       pygments
       black==21.12b0
+      mypy
+      slotscheck>=0.12,<0.13
 commands =
      flake8 ./lib/ ./test/ ./examples/ setup.py doc/build/conf.py {posargs}
      black --check ./lib/ ./test/ ./examples/ setup.py doc/build/conf.py
+     slotscheck -m sqlalchemy
 
 
 # command run in the github action when cext are active.
--- a/tox.ini
+++ b/tox.ini
@@ -172,6 +172,14 @@ commands =
      slotscheck -m sqlalchemy
 
 
+# "pep8" env was renamed to "lint".
+# Kept for backwards compatibility until rename is completed elsewhere.
+[testenv:pep8]
+basepython = {[testenv:lint]basepython}
+deps = {[testenv:lint]deps}
+commands = {[testenv:lint]commands}
+
+
 # command run in the github action when cext are active.
 [testenv:github-cext]
 deps = {[testenv]deps}
--- a/.github/workflows/run-on-pr.yaml
+++ b/.github/workflows/run-on-pr.yaml
@@ -84,8 +84,8 @@ jobs:
       - name: Run tests
         run: tox -e mypy ${{ matrix.pytest-args }}
 
-  run-pep8:
-    name: pep8-${{ matrix.python-version }}
+  run-lint:
+    name: lint-${{ matrix.python-version }}
     runs-on: ${{ matrix.os }}
     strategy:
       # run this job using this matrix, excluding some combinations below.
@@ -115,7 +115,7 @@ jobs:
           pip list
 
       - name: Run tests
-        run: tox -e pep8
+        run: tox -e lint
 
   # Arm emulation is quite slow (~20min) so for now just run it when merging to main
   # run-test-arm64:
--- a/.github/workflows/run-test.yaml
+++ b/.github/workflows/run-test.yaml
@@ -156,8 +156,8 @@ jobs:
       - name: Run tests
         run: tox -e mypy ${{ matrix.pytest-args }}
 
-  run-pep8:
-    name: pep8-${{ matrix.python-version }}
+  run-lint:
+    name: lint-${{ matrix.python-version }}
     runs-on: ${{ matrix.os }}
     strategy:
       # run this job using this matrix, excluding some combinations below.
@@ -187,7 +187,7 @@ jobs:
           pip list
 
       - name: Run tests
-        run: tox -e pep8
+        run: tox -e lint
 
   run-pep484:
     name: pep484-${{ matrix.python-version }}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,6 +27,13 @@ repos:
           - pydocstyle
           - pygments
 
-
-
-
+-   repo: https://github.com/ariebovenberg/slotscheck
+    rev: v0.11.0
+    hooks:
+    -   id: slotscheck
+        exclude: "^(?!lib/sqlalchemy)"
+        additional_dependencies:
+          - typing_extensions
+          - mypy
+          - greenlet
+        entry: env PYTHONPATH=lib slotscheck -v
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,13 +27,6 @@ repos:
           - pydocstyle
           - pygments
 
--   repo: https://github.com/ariebovenberg/slotscheck
-    rev: v0.11.0
-    hooks:
-    -   id: slotscheck
-        exclude: "^(?!lib/sqlalchemy)"
-        additional_dependencies:
-          - typing_extensions
-          - mypy
-          - greenlet
-        entry: env PYTHONPATH=lib slotscheck -v
+
+
+
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,6 +10,9 @@
 line-length = 79
 target-version = ['py37']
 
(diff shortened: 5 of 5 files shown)


jvanasco:
Is this ready for additional testing in gerrit?

ariebovenberg:
@zzzeek yes. I'm wondering though if pre-commit is the best place for slotscheck. Is there a specific reason some checks are in tox/pre-commit/github-actions?

zzzeek:
> @zzzeek yes. I'm wondering though if pre-commit is the best place for slotscheck. Is there a specific reason some checks are in tox/pre-commit/github-actions?

Things in tox are tested in Jenkins CI.     Things in github actions are tested there.  things in pre-commit are run every time we commit.

so in pre-commit, black and zimports are there since they actually change the code.  so those only belong in pre-commit.  but we also have flake8 there, because it catches so many quick things very fast, it just helps filter out quick issues before we push up to gerrit, which would then kick of jenkins builds which are more time / resource consuming.  

my impression of slotscheck is that it can work as a quick flake8- sort of thing.   but yes, overall it should likely be part of tox and github actions as well.    it seems right now to not fit cleanly into any of our particular targets, but it's more convenient for us if we make it part of the existing targets - the pep8 target is right now where we have generic "code quality" checks where we not only call upon flake8 (which checks way more than pep8) we also have "black check" in there.

id have to try the tool out a little bit to get a feel for it but yes I think it would first be part of the "pep8" target at the moment, how does that seem?

ariebovenberg:
@zzzeek thanks for the comprehensive answer! I initially added it to pre-commit following the suggestion in https://github.com/sqlalchemy/sqlalchemy/pull/7589#issuecomment-1030181975, but I'd say `slotscheck` fits more in the `tox`/GH-actions crowd than pre-commit.

There was a similar discussion in https://github.com/aio-libs/aiohttp/pull/6547#issuecomment-1019273829, where the conclusion was to add `slotscheck` to GH actions, not pre-commit. The reason was that slots are not frequently changed, reducing the benefits of running the check _every commit_.

zzzeek:
that's fine let's do it that way then!

ariebovenberg:
- I've moved slotscheck to `tox`
- I took the liberty of renaming `pep8` to `lint`, as this name better reflects the diversity of checks included. Even `black` is technically not fully PEP8 compliant

zzzeek:
the rename is fine but you would need to add another target pep8 that imports the "lint" target, so that our CI which is hardcoded to "pep8" for all projects continues to work, until we can make this name change for all projects (alembic, dogpile, mako)

ariebovenberg:
I've re-added `pep8` env which takes its configuration from the `lint` env

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 3e77fe5449615a3c7c61ce9a1e4e79cd6636a89a of this pull request into gerrit so we can run tests and reviews and stuff

zzzeek:
lets try it out

sqla-tester:
New Gerrit review created for change 3e77fe5449615a3c7c61ce9a1e4e79cd6636a89a: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3576

sqla-tester:

just gave it a spin removing a "__slots__()", to see how straightforward it is.

output is very clear:

ERROR: 'sqlalchemy.sql.roles:DMLTableRole' has slots but superclass does not.
ERROR: 'sqlalchemy.sql.roles:StrictFromClauseRole' has slots but superclass does not.
Oh no, found some problems!
Scanned 207 module(s), 1347 class(es).


very nice!



View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3576

sqla-tester:

I've searched the entire codebase of slotscheck to see where it's using mypy, and I see nothing; no imports of anything in mypy, no command line running of mypy.  Why is mypy a dependency?  

https://github.com/ariebovenberg/slotscheck/search?q=mypy

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3576

sqla-tester:

oh that's to import *our* mypy plugin.  OK.  let me alter that to indicate it's using the sqlalchemy[mypy] target.  I found some other slots checks that were missing because they are in our non-C ext fallback code, so I am fixing that also.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3576

sqla-tester:

Looks ok to me

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3576

ariebovenberg:
@zzzeek indeed `mypy` is required to import _sqlalchemy_'s mypy plugin. 

Regarding C-extensions, the situation here is interesting! It looks like it may be needed to have _two_ slotscheck checks (one `HAS_CYEXTENSION`, the other not). Looking at `sqlalchemy.util/_collections.py` I can see:

```python
if typing.TYPE_CHECKING or not HAS_CYEXTENSION:
    from ._py_collections import ImmutableContainer
    # ...
else:
    from sqlalchemy.cyextension.immutabledict import ImmutableContainer
    # ...
```

where both `ImmutableContainer`s were missing `__slots__` 🤔.
In PR #7589 I fixed `cyextension/immutabledict.pyx:ImmutableContainer`, but not the pure Python one.

zzzeek:
> @zzzeek indeed `mypy` is required to import _sqlalchemy_'s mypy plugin.
> 
> Regarding C-extensions, the situation here is interesting! It looks like it may be needed to have _two_ slotscheck checks (one `HAS_CYEXTENSION`, the other not). Looking at `sqlalchemy.util/_collections.py` I can see:
> 
> ```python
> if typing.TYPE_CHECKING or not HAS_CYEXTENSION:
>     from ._py_collections import ImmutableContainer
>     # ...
> else:
>     from sqlalchemy.cyextension.immutabledict import ImmutableContainer
>     # ...
> ```
> 
> where both `ImmutableContainer`s were missing `__slots__` thinking. In PR #7589 I fixed `cyextension/immutabledict.pyx:ImmutableContainer`, but not the pure Python one.

oh i wasnt even sure if cython was an option here (@CaselIT  does slots apply to cython even?  I thought we have to name all the variables anyway?) .  OK, well, if cython builds, the _py_ versions of the objects are still available, and slotscheck is probably checking them.  in this case it likely missed these two because there were no inheriting objects from them at runtime.

CaselIT:
> oh i wasnt even sure if cython was an option here (@CaselIT does slots apply to cython even? I thought we have to name all the variables anyway?)

that's on `cdef class`es. a plain `class` behaves the same as a python one

ariebovenberg:
>  OK, well, if cython builds, the py versions of the objects are still available, and slotscheck is probably checking them

`slotscheck` _does_ recurse properly into extension modules. Because it's a 'runtime' check it is also able to follow conditional imports. 

But of course, it cannot trace multiple code branches at once (`HAS_CYEXTENSION` and `not HAS_CYEXTENSION`). For full slotscheck 'coverage' you'd have to run on both types of installations.

This 'double' check is the only way to find both `ImmutableContainer` problems. If `HAS_CYEXTENSION` then`ImmutableProperties` inherits from `cyextension.pyx:ImmutableContainer`, otherwise `ImmutableProperties` inherits from `_py_collections.py:ImmutableContainer`. Both base classes need `__slots__` because `ImmutableProperties` does.

zzzeek:
yup, all makes sense, thanks

tox.ini
```diff
@@ -164,9 +164,20 @@ deps=
...
      flake8 ./lib/ ./test/ ./examples/ setup.py doc/build/conf.py {posargs}
      black --check ./lib/ ./test/ ./examples/ setup.py doc/build/conf.py
+     slotscheck -m sqlalchemy
+
+
+# "pep8" env was renamed to "lint".
+# Kept for backwards compatibility until rename is completed elsewhere.
+[testenv:pep8]
+basepython = {[testenv:lint]basepython}
+deps = {[testenv:lint]deps}
```
sqla-tester, line 179: **mike bayer** (zzzeek) wrote:

the lint action doesn't, problem solved

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3576

sqla-tester (commented):

ah, we only run lint on ubuntu, we're good

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3576

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3576 has been **merged**. Congratulations! :)

PR #7694, opened 2022-02-11
7212: Move UUID class to sqltypes

### LINKED ISSUE #7212: Add UUID data type to mariadb/mysql for 10.7
### Describe the use case

The UUID data type is being added to MariaDB 10.7.0, currently in preview: https://mariadb.com/kb/en/uuid-data-type/

This looks identical to the current PG implementation `id UUID`

There is also the function `UUID()` which generates a V1 UUID. I assume this is already callable via `func.UUID()` but haven't tested. Unfortunate that MG and Maria differ here, PG currently requires an extension and `uuid_generate_v1` or `uuid_generate_v4`.

Mariadb

### Example Use

```SQL
CREATE TABLE t1 (id UUID);
```

```python
# either:
from sqlalchemy.dialects.mariadb import UUID
# or:
from sqlalchemy import UUID 
# UUID is appearing in more dialects. Type could default to a 128 bit integer for unsupported dialects
...
Column(UUID(as_uuid=True), primary_key=True, server_default=func.uuid())
```

### Additional context

It's worth noting that this will exist in MariaDB, but not necessarily MySQL, at least as of what I could find now. Not sure how SQLA currently handles Maria-My differences.

tgross35:
### Description
This change fixes #7212 and will allow the `UUID` datatype for all dialects.

### Checklist
This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**


CaselIT:
Hi,

Thanks for the work.

Reading the issue and proposed features, I think that the as_uuid flag should be only in the postgresql implementation. The general one will always render UUID and will error if a DB does not support that type.
 Pinging to confirm @zzzeek

zzzeek:
> Hi,
> 
> Thanks for the work.
> 
> Reading the issue and proposed features, I think that the as_uuid flag should be only in the postgresql implementation. The general one will always render UUID and will error if a DB does not support that type. Pinging to confirm @zzzeek

as_uuid refers to the Python type returned or accepted, that is, uuid.UUID vs. string.   it's not related to the database.  I think as_uuid should default to True but it doesnt seem necessary to lose the option for sending/receiving as strings as there might be a performance bump for that, not to mention backwards compat ease.

zzzeek:
this PR would need suite tests added to [suite test_types.py](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/testing/suite/test_types.py) which would then need to be turned on for all Postgresql DBs as well as qualifying MariaDB versions in test/requirements.py.   What MariaDB versions are needed?  we are up to mariadb 10.6 in our CI environment.  as 10.7 seems to be out we should also add that to our CI.

zzzeek (changes requested):
as mentioned, needs test support in suite/test_types.py and turned on for PG, participating MariaDB versions

tgross35:
@zzzeek Maria 10.7+ supports UUID, I believe some MySQL versions as well

tgross35:
Working on the changes to tests now. I have something preliminary done that passes locally on mariadb and postgres (assuming I'm running them properly).

I stand corrected about support, MySQL doesn't seem to have a UUID type (surprisingly). It is supported by:
- MariaDB 10.7+ (this is latest on Docker official)
- PG 8.3+ (all supported versions)
- MS, all supported versions as `UNIQUEIDENTIFIER`

PG is marked as `>= 8.3` for a lot of tests - is there any point in keeping that designation at this point? Since 8.3 ended support in 2013.

I think this is ready for a preliminary run on your CI, at least to see if there are any regressions. I'll remove the draft designation.

lib/sqlalchemy/sql/sqltypes.py
```diff
@@ -3268,6 +3269,97 @@ class MatchType(Boolean):
...
+            return process
+        else:
+            return None
+
+    def literal_processor(self, dialect):
+        if self.as_uuid:
+
+            def process(value):
+                if value is not None:
+                    value = "'%s'::UUID" % value
```
CaselIT, line 3349: I don't think we can use this hare, since this is a postgresql specific syntax

lib/sqlalchemy/testing/suite/test_types.py
```diff
@@ -1590,6 +1592,60 @@ def test_string_cast_crit_against_string_basic(self):
         )
 
 
+class UUIDTest(fixtures.TestBase):
+    __backend__ = True
+
```
CaselIT, line 1597: this is missing a `__requires__ = ('uuid_data_type',)`

CaselIT, line 1597: this is missing a `__requires__ = ('uuid_data_type',)`

CaselIT (changes requested):

lib/sqlalchemy/sql/sqltypes.py
tgross35: Thanks Federico, this is related to what we briefly discussed on gitter

image

I guess I'm not completely clear - the purpose of `literal_processor` is to convert a python datatype to the format it will be inserted/selected as in DML? If so, I guess mariaDB doesn't need a special one, it just uses the string representation of a UUID.

CaselIT: In that reply I meant that type of `select`, not the `::uuid` cast.

> the purpose of literal_processor is to convert a python datatype to the format it will be inserted/selected as in DML

it's run when a query that specified listeral_bind=True is run. From my limited experience with mariadb/mysql I too don't think there is anything particular to do for it, it's usually PG that's particularly picky about types.

I think we could have here only the version that stringifies the uuid, while the pg specific implementation could override this method to have the cast

tgross35: That's easy enough, I'll change that

zzzeek:
there's no tracking issue for this.    hence it keeps falling off my list

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 1552e1b7fd55b00b13776f728c74ebf4bad9b74b of this pull request into gerrit so we can run tests and reviews and stuff

zzzeek:
this is now moved manually to https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3871 .

current status is a messy squash + rebase, I will salvage this into the final change.

tgross35:
For anyone landing here from Google looking for UUID support - zzzeek took care of this in 349a7c5e0e2aeeac98fad789b0043a4bdfeed837

This isn't on SQLA 1.4 (current), awaiting 2.0 release (not public as-of 2022-06-14)

PR #7702, opened 2022-02-14
Fix keyword arguments not being passed to AsyncSessionTransaction

heckad:


zzzeek:
hey there -

AsyncSessionTransation does not accept any additional keyword arguments so these `**kw` should not be passed.   The presence of `**kw` on begin() and begin_nested() instead seems to be the bug, this seems like it's a legacy carry-over of some kind but these arguments should not be present AFAICT.

zzzeek:
see #7703 which will remove these.  thanks for the heads up

PR #7893, opened 2022-04-04
[mypy plugin] Use api.defer() when PlaceholderNode is encountered

mehdigmira:
### Description

This is in the context of using mypy >=0.710 and the new semantic analyzer
> Mypy 0.710 introduced a new semantic analyzer, and the old semantic analyzer was removed in mypy 0.730

Mypy's plugin doc (https://mypy.readthedocs.io/en/stable/extending_mypy.html) states:

> When looking up symbols, you may encounter placeholder nodes that are used for names that haven’t been fully processed yet. You’ll generally want to request another semantic analysis iteration by deferring in that case.

This isn't done today by sqlalchemy and results in some types not being correctly inferred. I had the case on our code base where we have something like this

```python
# file a.py

from typing import TYPE_CHECKING:
    if TYPE_CHECKING:
        from .b import B

@declarative_mixin
class AMixin:
    @declared_attr
    def b(cls) -> Mapped["B"]:
        return relationship("B", lazy="joined", innerjoin=True)
reveal_type(AMixin.b)
```

```python
# file b.py

from .other import OtherMixin
from .base import Base

class B(Base, OtherMixin):
    ...
```

Currently this produces `note: Revealed type is "B?"`, because in unbound_to_instance, `node.node` is a `PlaceHolderNode`.

After the fix, this produces  `note: Revealed type is b.B*`. I think there may be other cases where deferring is required, but I only cover here the case that I encountered.

### Checklist

**It's a bit hard to have a simple test for this issue since it relies on the order at which mypy covers the codebase, and requires multiple files/modules. So I didn't put too much effort into building a minimal test case for this, but instead have used our codebase as a testcase for this.
I can try building a testcase for this, provided you find this interesting, and you think you can merge it into the repo.**

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**


zzzeek:
hey there -

thanks for this!   though I have to ask that yes, I'd need to at least see what the existing issue is in some way, at least what the failure is that's being experienced and what the code looks like when this failure occurs.    As you are likely aware, the mypy plugin is already deprecated and I am not spending any resources on it other than accepting PRs for simple fixes with tests or at least some concrete example of what's not working.   If this PR had a test case that reproduces the issue without the fix present, that would be very helpful.

the mypy plugin includes a comprehensive system of testing multi-file layouts that change.  Take a look at some of the examples in https://github.com/sqlalchemy/sqlalchemy/tree/main/test/ext/mypy/incremental to see what these look like.    you basically lay out py files to be type checked and an optional series of ".testpatch" files that will be applied to the files in order, with mypy being run each time.    it's a little tricky to establish the files correctly but it tests these complex mypy issues effectively (and also you can see why we are getting away from having to do this in 2.0).

mehdigmira:
Ok, I'll give it a try ! Though it's not straightforward to build the right test sequence for these UnboundType issue :/
I've also encountered other cases, and added a commit to handle more UnboundType cases.

I guess dropping this approach in 2.0 is a good idea. With our large codebase making extensive use of SQLAlchemy, our experience with typing with this version of the plugin has been a bit painful so far. 
Do you have a document describing how would typing look like in 2.0 ?

zzzeek:
it's committed to main but not added to the docs yet.    however you can see example mappings in the test suite at https://github.com/sqlalchemy/sqlalchemy/blob/main/test/ext/mypy/plain_files/ .    the two differences are:  declarative base is made by subclassing a new `DeclarativeBase` root class, and then instead of using `Column` in declarative, use `mapped_column`.  that's it!  no more plugins.

zzzeek:
so as you probably know, SQLAlchemy 2.0 has added new APIs so that your ORM models can be typed without any plugins and we expect usage of the mypy plugin to go to basically zero, since it's terrible (was a lot of work to write it too ... :( )  .   so for the moment I'd rather just leave it like it is, assuming on your end you've found a way to work around these issues until I can release 2.0's API.

PR #7917, opened 2022-04-07
Update session.py 100 flush error message and no_autoflush docstring

aathan:
### Description
See https://github.com/sqlalchemy/sqlalchemy/discussions/7916

I also added a small addition to no_autocommit()'s docstring.

I am not creating an issue for this because it is not a bug, and this is akin to a documentation or typographical error.

### Checklist

This pull request is:

- [X] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/orm/session.py
+++ b/lib/sqlalchemy/orm/session.py
@@ -889,9 +889,9 @@ class SessionTransaction(_StateChange, TransactionalContext):
                 self.session.flush()
             else:
                 raise exc.FlushError(
-                    "Over 100 subsequent flushes have occurred within "
-                    "session.commit() - is an after_flush() hook "
-                    "creating new objects?"
+                    "session.commit() (or session.prepare()) has now attempted 100 flush()es "
+                    "to bring the session into clean state - is an after_flush() "
+                    "hook creating new objects? is flush() having no effect()?"
                 )
 
         if self._parent is None and self.session.twophase:
--- a/lib/sqlalchemy/orm/session.py
+++ b/lib/sqlalchemy/orm/session.py
@@ -2135,6 +2135,9 @@ class Session(_SessionClassMethods):
         of objects which involve existing database queries,
         where the uncompleted object should not yet be flushed.
 
+        This does not disable the flush() that automatically
+        occurs within commit().
+
         """
         autoflush = self.autoflush
         self.autoflush = False


lib/sqlalchemy/orm/session.py
```diff
@@ -2135,6 +2135,9 @@ def no_autoflush(self):
         of objects which involve existing database queries,
         where the uncompleted object should not yet be flushed.
 
+        This does not disable the flush() that automatically
+        occurs within commit().
```
CaselIT, line 2139: I'm not sure this is needed, since it seems obvious. Without that flush commit would be pointless, since changes would not be persisted to the db

aathan: This context manager is unfortunately called `no_autoflush()` rather than `no_autoflush_on_query()`. 

I suggested this change because, given the function name, I had to read the code to be sure wheter the doc string itself may have been incorrectly worded.

It's possible it seems obvious because you're an expert. I see stackexchange questions exist about the interplay of flush and commit, for example.

A new user seeing the old docstring may be wondering if *automatic* flushes are also disabled during commits that may be part of complex call-trees within the context manager's lifetime (i.e., allowing this context manager to be used to temporarily cause commits to be noops). They may see this context manager and think, "oh, I can disable commits for a while with this," as misguided as that may be.

I don't feel strongly about the edit, but ... I thought it might save someone an excursion through the source.

CaselIT: I also don't feel strongly, but to me it seems to be stating an obvious thing,
It it did disable flush on commit this method could be as well be called `enable_bugs()`.

Also it's the documentation kind of states when it disable the commits:

> Operations that proceed within the ``with:`` block will not be subject to flushes occurring upon query  access

aathan: For what it's worth, it is not at all obvious to me as a non-SQLA expert, that disabling flush() and calling commit() would create bugs. I sort of expect the commit() would then just do nothing. Whether that's actually a bug depends on the application, which may instead call a non-flush-disabled-commit later.

I don't have anything more to say about it and it's clearly not my decision :)

If you'd like me to do something to help resolve this PR, just LMK what that is.

CaselIT: I'm not trying to dismiss the change, I'm just trying to understand what's not clear about the current docstring that requires this.

> it is not at all obvious to me as a non-SQLA expert, that disabling flush() and calling commit() would create bugs

I honestly find it surprising. with just basic database knowledge, I would expect that calling a function called `commit()` would persist changes to the db. If that's not the case then the behaviour would be surprising I guess.

> I sort of expect the commit() would then just do nothing

but this methog is called `no_autoflush()` not `disable_commit()`. auto-flush implies something that happens automatically while doing other things, and the documentation states that the other things are queries.
Commit is not an automatic thing and is not a query.

aathan: My little attempt to adjust a docstring may have veered into bikeshed-ing at this point. The docs are quite explicit about this exact same issue.

In other words, the person that wrote the docs thought the same thing needed to be made very clear and spent many words making it so.

That same claritfying point is simply not reflected in the *docstring*.

From the docs
----------------
To reiterate: The flush process always occurs when transactional methods such as [Session.commit()](https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session.commit) and [Session.begin_nested()](https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session.begin_nested) are called, regardless of any “autoflush” settings, when the [Session](https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session) has remaining pending changes to process.

So, it really just comes down to whether adding that sentence to the docstring is worth the clicks. No worries either way.

CaselIT: > The flush process always occurs when transactional methods such as [Session.commit()](https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session.commit) and [Session.begin_nested()](https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session.begin_nested) are called, regardless of any “autoflush” settings, when the [Session](https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session) has remaining pending changes to process.

I actually like this part. I'm ok on adding it in the no_autoflush method

aathan: I'm not sure what you mean. Do you mean you want me to replace

`This does not disable the flush() that automatically occurs within commit()`

`The flush process always occurs when transactional methods such as [Session.commit()](https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session.commit) and [Session.begin_nested()](https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session.begin_nested) are called, regardless of any “autoflush” settings, when the [Session](https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session) has remaining pending changes to process.`

in this PR?

zzzeek:
closing this because I think the newer message is a little more confusing ("is flush having no effect?")   this error occurs for a really straightforward reason assuming the SessionTransaction isn't being subclassed (which it should not).

PR #7921, opened 2022-04-08
Session transaction creation

aathan:
### Description
SessionTransaction is a class closely associated with Session. Subclassing Session likely requires changes to SessionTransaction too. Session directly creates SessionTransaction, making it impossible to change the combined
behavior of the two classes without monkey patching. This small change causes SessionTransaction creation to be
an instance method of Session.

I would be happy to provide a test for this change but I got discouraged by what seemed like a difficult learning curve to configure and run the test suite for a small change of this nature.

I'd be happy to write a quick guide for that kind of thing if you give me some guidance.

I was also just not sure what kind of test for this change would be appropriate, since it's a fairly small change unlikely to introduce anything but syntax errors in the source file itself. Do you want to see a test that actually changes the SessionTransaction class instantiated by a subclass of Session?

### Checklist

Not ignoring the request for a test. Just not sure how...

This pull request is:

- [X] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/orm/session.py
+++ b/lib/sqlalchemy/orm/session.py
@@ -1383,7 +1383,7 @@ class Session(_SessionClassMethods):
             # a subtransaction
 
             assert not nested and not _subtrans
-            trans = self._create_transaction(self)
+            trans = self._create_transaction()
             assert self._transaction is trans
 
         return self._transaction  # needed for __enter__/__exit__ hook
--- a/lib/sqlalchemy/orm/session.py
+++ b/lib/sqlalchemy/orm/session.py
@@ -1261,6 +1261,9 @@ class Session(_SessionClassMethods):
             with self.begin():
                 yield self
 
+    def _create_transaction(self, *args, **kwargs):
+        return SessionTransaction(self, *args,**kwargs)
+
     def in_transaction(self):
         """Return True if this :class:`_orm.Session` has begun a transaction.
 
@@ -1319,7 +1322,7 @@ class Session(_SessionClassMethods):
     def _autobegin(self):
         if self._transaction is None:
 
-            trans = SessionTransaction(self, autobegin=True)
+            trans = self._create_transaction(autobegin=True)
             assert self._transaction is trans
             return True
 
@@ -1343,7 +1346,8 @@ class Session(_SessionClassMethods):
          documentation on SAVEPOINT transactions, please see
          :ref:`session_begin_nested`.
 
-        :return: the :class:`.SessionTransaction` object.  Note that
+        :return: the object created by _create_transaction(), normally
+         a :class:`.SessionTransaction` object.  Note that
          :class:`.SessionTransaction`
          acts as a Python context manager, allowing :meth:`.Session.begin`
          to be used in a "with" block.  See :ref:`session_explicit_begin` for
@@ -1379,7 +1383,7 @@ class Session(_SessionClassMethods):
             # a subtransaction
 
             assert not nested and not _subtrans
-            trans = SessionTransaction(self)
+            trans = self._create_transaction(self)
             assert self._transaction is trans
 
         return self._transaction  # needed for __enter__/__exit__ hook
@@ -1393,7 +1397,8 @@ class Session(_SessionClassMethods):
         For documentation on SAVEPOINT
         transactions, please see :ref:`session_begin_nested`.
 
-        :return: the :class:`.SessionTransaction` object.  Note that
+        :return: the object created by _create_transaction(), normally a
+         :class:`.SessionTransaction` object.  Note that
          :class:`.SessionTransaction` acts as a context manager, allowing
          :meth:`.Session.begin_nested` to be used in a "with" block.
          See :ref:`session_begin_nested` for a usage example.
--- a/doc/build/orm/extensions/associationproxy.rst
+++ b/doc/build/orm/extensions/associationproxy.rst
@@ -57,13 +57,13 @@ with ``User`` requires traversal from each collection element to the ``.keyword`
 attribute, which can be awkward::
 
     >>> user = User('jek')
-    >>> user.kw.append(Keyword('cheese inspector'))
+    >>> user.kw.append(Keyword('cheese-inspector'))
     >>> print(user.kw)
     [<__main__.Keyword object at 0x12bf830>]
     >>> print(user.kw[0].keyword)
-    cheese inspector
+    cheese-inspector
     >>> print([keyword.keyword for keyword in user.kw])
-    ['cheese', 'inspector']
+    ['cheese-inspector']
 
 The ``association_proxy`` is applied to the ``User`` class to produce
 a "view" of the ``kw`` relationship, which only exposes the string
@@ -88,9 +88,9 @@ which is both readable and writable.  New ``Keyword`` objects are created
 for us transparently::
 
     >>> user = User('jek')
-    >>> user.keywords.append('cheese inspector')
+    >>> user.keywords.append('cheese-inspector')
     >>> user.keywords
-    ['cheese inspector']
+    ['cheese-inspector']
     >>> user.keywords.append('snack ninja')
     >>> user.kw
     [<__main__.Keyword object at 0x12cdd30>, <__main__.Keyword object at 0x12cde30>]
@@ -120,11 +120,11 @@ assignment event) is intercepted by the association proxy, it instantiates a
 new instance of the "intermediary" object using its constructor, passing as a
 single argument the given value. In our example above, an operation like::
 
-    user.keywords.append('cheese inspector')
+    user.keywords.append('cheese-inspector')
 
 Is translated by the association proxy into the operation::
 
-    user.kw.append(Keyword('cheese inspector'))
+    user.kw.append(Keyword('cheese-inspector'))
 
 The example works here because we have designed the constructor for ``Keyword``
 to accept a single positional argument, ``keyword``.   For those cases where a
--- a/doc/build/orm/extensions/associationproxy.rst
+++ b/doc/build/orm/extensions/associationproxy.rst
@@ -63,7 +63,7 @@ attribute, which can be awkward::
     >>> print(user.kw[0].keyword)
     cheese inspector
     >>> print([keyword.keyword for keyword in user.kw])
-    ['cheese inspector']
+    ['cheese', 'inspector']
 
 The ``association_proxy`` is applied to the ``User`` class to produce
 a "view" of the ``kw`` relationship, which only exposes the string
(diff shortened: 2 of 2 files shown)


lib/sqlalchemy/orm/session.py
```diff
@@ -1261,6 +1261,9 @@ def _maker_context_manager(self):
             with self.begin():
                 yield self
 
+    def _create_transaction(self, *args, **kwargs):
+        return SessionTransaction(self, *args,**kwargs)
```
CaselIT, line 1265: Usually for these kind of thing a class variable on the class is used.
Here it could be `transaction_cls = SessionTransaction` then it's called as `self.transaction_cls(...)`

CaselIT (changes requested):
> I would be happy to provide a test for this change but I got discouraged by what seemed like a difficult learning curve to configure and run the test suite for a small change of this nature.

Not sure what you found difficult to configure. Running `pytest test/path/to/test/file` does not work? Do you have suggestions on how can we help making test easier to start with?

> I'd be happy to write a quick guide for that kind of thing if you give me some guidance.
> I was also just not sure what kind of test for this change would be appropriate, since it's a fairly small change unlikely to introduce anything but syntax errors in the source file itself. 

without test there is no guarantee that future changes will not break this change

> Do you want to see a test that actually changes the SessionTransaction class instantiated by a subclass of Session?

I guess you could create a session transaction subclass, use that in a session subclass and then test that in the places a transaction is constructed the subclass is used. You can use something like `isinstance(session._transaction, TheSubclass)`

zzzeek:
> ### Description
> 
> SessionTransaction is a class closely associated with Session. Subclassing Session likely requires changes to SessionTransaction too. 

I would need to see a rationale for this.   `Session` subclasses themselves are extremely uncommon, but the use case for such a subclass is for things like the [horizontal sharding extension](https://docs.sqlalchemy.org/en/14/orm/extensions/horizontal_shard.html?highlight=horizontal%20sharding).    However, subclassing `SessionTransaction` should never be necessary.  In order to allow end-user code and libraries to apply behaviors to transactional events, the [SessionEvent api](https://docs.sqlalchemy.org/en/14/orm/events.html?highlight=sessionevent#sqlalchemy.orm.SessionEvents) aims to provide hooks that can accommodate all contingencies.  For example if your use case requires that custom code runs when the `SessionTransaction` begins, rolls back, or commits,  there are events both at the [session transaction level](https://docs.sqlalchemy.org/en/14/orm/events.html?highlight=sessionevent#sqlalchemy.orm.SessionEvents.after_transaction_create) as well as at the [per connection level](https://docs.sqlalchemy.org/en/14/orm/events.html?highlight=sessionevent#sqlalchemy.orm.SessionEvents.after_begin).    If you can accurately describe what it is you're trying to do, we can evaluate what hooks can suit these needs or if there is a case for new hooks to be added, such as a "before_prepare()" hook, as I see we dont yet have hooks in that way.

SQLAlchemy's event system exists so that end-users ideally never have to subclass any internal mechanism, as it is not feasible to maintain a library with such constraints on it, and the internals of `SessionTransaction` are subject to change at any time.    



> 
> I would be happy to provide a test for this change but I got discouraged by what seemed like a difficult learning curve to configure and run the test suite for a small change of this nature.

Can you elaborate on this?   SQLAlchemy's test suite does not require any "configuration" other than installing `tox`.   There is very detailed documentation on both how to quickstart run the tests (a single command, no arguments) as well as how to run specific tests and things like that in the top level directory at https://github.com/sqlalchemy/sqlalchemy/blob/main/README.unittests.rst .   We spend a good deal of effort ensuring the test suite is very easy to work with so it's unexpected to see a comment like this.


> 
> I'd be happy to write a quick guide for that kind of thing if you give me some guidance.

OK it seems like you might have missed https://github.com/sqlalchemy/sqlalchemy/blob/main/README.unittests.rst , we refer to this document also from the site from the [SQLAlchemy Development Guide](https://www.sqlalchemy.org/develop.html#testing), which I guess you might have missed also, can you give me some clues on what documentation sources you looked at so we can make sure people dont miss things?  thanks


> 
> I was also just not sure what kind of test for this change would be appropriate, since it's a fairly small change unlikely to introduce anything but syntax errors in the source file itself. 

a test for this would include a custom Session subclass w/ custom SessionTransaction subclass, create the session, run `trans = session.begin()`, assert `isinstance(trans, MySessionTransactionSubclass)`.  but as I said I'd rather ensure adequate event hooks are present here.

> Do you want to see a test that actually changes the SessionTransaction class instantiated by a subclass of Session?

simple and direct would do it, yup.

lib/sqlalchemy/orm/session.py
aathan: Sure. Can be done either way.

This seemed more flexible because you could intercept the creation without subclassing the SessionTransaction. Doesn't matter, since the PR doesn't seem likely to be accepted given the design philosphy issues in the rest of the PR response.

zzzeek: > Sure. Can be done either way.
> 
> This seemed more flexible because you could intercept the creation without subclassing the SessionTransaction. Doesn't matter, since the PR doesn't seem likely to be accepted given the design philosphy issues in the rest of the PR response.

It's too early to state that.    I would like the session event system to be able to accommodate your use case.  however if there's some compelling case for why they can't work, or why it's too much burden, I'd want to see that.   It's very difficult for us to field these kinds of requests without knowing what it is someone's trying to do; generally we never add anything at all, not even something uncontroversial, without some understanding of the use case, so that we can make sure we make the best change possible with a clear rationale.   There's lots of examples of weird API features we've had to carry along for years, making it difficult to refactor things, because someone wanted some particular thing at some point which seemed easy enough and we just accepted it.  or, someone wanted something simple, we accepted it, then we realize later the area in which they are working is actually much more complicated than what they needed, someone else needs a similar feature but we've already bolted an incompatible API in place.    so we can't really afford to make that mistake these days, hence we request that we get a clear rationale why something is needed.

aathan:
Guys, thank you so much for your attention and time, and for an amazingly flexible, capable, well engineered library and accompanying body of work. Really. It's awesome. I mean that sincerely.

My input comes down to avoiding having simple things become a big cognitive lift for users that are focused on quick end results vs deep-diving the SQLA way.

Concrete suggestions:

1) The eventing system could probably be more visible. I don't think it's mentioned in session.py.
i.e., entering SQLA via the source code (vs the docs) doesn't lead you to eventing. If you give me a sample of docstring text that would an acceptable way to link to that knowledge base from some strategic spots, I can submit a PR with some docstring edits that include your snippet.

2) Similarly, I could strategically place some admonitions that subclassing is not the SQLA way? E.g., Session's class level docstring?

3) A short section between "Basic Test Running" and "Advanced Tox Options" that explicitly shows someone how to write, where to put, and how to run a very simple test -- including sample output.

Yes, I saw both the dev guide and the testing page, and thought "nope, too much to deal with" becuase I almost immediately ran into this off-putting sentence: "Within the realm of pytest, SQLAlchemy itself is adding a large series of option and customizations to the pytest runner using plugin points, to allow for SQLAlchemy's multiple database support, database setup/teardown and connectivity, multi process support, as well as lots of skip / database selection rules."

-- do I have to install databases? how complex is this suite? where do I even put this little test and cause it to function within the suite? I have to learn what tox is and  how it interfaces with pytest? argh!

aathan:
As far as my usecase for this change:  In the early days when gevent was introduced into this complex system I'm maintaining, there were problems with commits happening at unexpected times from other greenlets. The solution then was to introduce one "root" query Session and many other "per work item" Sessions. The approach taken was to subclass Session so that add()s in the non-root-query Session are collected into a container (but the objects are still in the root Session as far as SQLA is concerned).

In those sub-sessions, on commit(), the session is added to a work-queue on a single "DB thread" and when its turn comes, the collected objects are hoisted from the root query session into the "writer" session and flushed to the DB. They are then put back into the root "query" session. This allows the querying of objects not to result in multiple object instances of the same "DB item". The subsidiary sessions can be passed around, including across greenlet boundaries, to complete complex tasks that have then collected all the things that must be persisted, before the commit() occurs. Task composition may squelch internal commits.

The whole thing needs a new look. Possibly none of this should be *in* Session vs some new separate class. Possibly a lot of this can be solved via eventing instead of overriding Session/SessionTransaction methods. Possibly the asyncio features introduced into SQLA are relevant...

zzzeek:
> As far as my usecase for this change: In the early days when gevent was introduced into this complex system I'm maintaining, there were problems with commits happening at unexpected times from other greenlets. The solution then was to introduce one "root" query Session and many other "per work item" Sessions. The approach taken was to subclass Session so that add()s in the non-root-query Session are collected into a container (but the objects are still in the root Session as far as SQLA is concerned).
> 
> In those sub-sessions, on commit(), the session is added to a work-queue on a single "DB thread" and when its turn comes, the collected objects are hoisted from the root query session into the "writer" session and flushed to the DB. They are then put back into the root "query" session. This allows the querying of objects not to result in multiple object instances of the same "DB item". The subsidiary sessions can be passed around, including across greenlet boundaries, to complete complex tasks that have then collected all the things that must be persisted, before the commit() occurs. Task composition may squelch internal commits.
> 
> The whole thing needs a new look. Possibly none of this should be _in_ Session vs some new separate class. Possibly a lot of this can be solved via eventing instead of overriding Session/SessionTransaction methods. Possibly the asyncio features introduced into SQLA are relevant...

this sounds pretty burdensome, but overall, the Session never commits itself, so if you were squleching a commit, that would be on your end in your session subclass (provided you arent using autocommit=True in any of these sessions, which you definitely should not do).   it seems like also you'd be better off with a facade rather than a subclass.  that way you dont need to mess with session at all.    finally the notion of "collecting objects for a background queue", seems better that you'd collect those into some colleciton you maintain yourself that is not a Session.  Sessions should be short lived and created per-transactional-thing-to-do.  passing them over greenlet boundaries seems like a bad idea. lists/dicts/sets of detached objects, better idea.

aathan:
The squelching commit piece is not really that relevant or important, ignore it.

I'm generally wondering though, if this is a good or bad pattern for SQLA in a greenlet environment where we don't want multiple in memory copies of the same database item.

Basically we're serializing the database operations by expunge()ing objects out of the main "query" session, into writer sessions when the collection is ready to commit, calling commit() on that session, and then moving those objects back to the query session...

I'm assuming that the calculations regarding which fields have changed aren't affected by those moves. I.e., I can change field values while the object is in session A, move it to session B, and commit() and still get the changed fields correctly persisted.

zzzeek:
as discussed, SessionTransaction should be considered a "Final" class.

PR #7959, opened 2022-04-25
Format code in documentation

notypecheck:
### Description
This PR formats code blocks and fixes minor code issues in documentation.
For example `sqlalchemy.ext.asyncio.async_sessionmaker` is used in some code blocks which doesn't exist,
and current approach is to use `sesionmaker` with `class_=AsyncSession`

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/doc/build/orm/basic_relationships.rst
+++ b/doc/build/orm/basic_relationships.rst
@@ -272,12 +272,12 @@ remote tables with which to link::
     this ensures that duplicate rows won't be persisted within the table regardless
     of issues on the application side::
 
-    association_table = Table(
-        "association",
-        Base.metadata,
-        Column("left_id", ForeignKey("left.id"), primary_key=True),
-        Column("right_id", ForeignKey("right.id"), primary_key=True),
-    )
+        association_table = Table(
+            "association",
+            Base.metadata,
+            Column("left_id", ForeignKey("left.id"), primary_key=True),
+            Column("right_id", ForeignKey("right.id"), primary_key=True),
+        )
 
 For a bidirectional relationship, both sides of the relationship contain a
 collection.  Specify using :paramref:`_orm.relationship.back_populates`, and
--- a/doc/build/orm/basic_relationships.rst
+++ b/doc/build/orm/basic_relationships.rst
@@ -8,11 +8,9 @@ A quick walkthrough of the basic relational patterns.
 The imports used for each of the following sections is as follows::
 
     from sqlalchemy import Column, ForeignKey, Integer, Table
-    from sqlalchemy.ext.declarative import declarative_base
-    from sqlalchemy.orm import relationship
+    from sqlalchemy.orm import declarative_base, relationship
 
     Base = declarative_base()
-
 
 .. _relationship_patterns_o2m:
 
--- a/doc/build/orm/basic_relationships.rst
+++ b/doc/build/orm/basic_relationships.rst
@@ -7,9 +7,9 @@ A quick walkthrough of the basic relational patterns.
 
 The imports used for each of the following sections is as follows::
 
-    from sqlalchemy import Table, Column, Integer, ForeignKey
-    from sqlalchemy.orm import relationship
+    from sqlalchemy import Column, ForeignKey, Integer, Table
     from sqlalchemy.ext.declarative import declarative_base
+    from sqlalchemy.orm import relationship
 
     Base = declarative_base()
 
@@ -24,28 +24,30 @@ the parent.  :func:`_orm.relationship` is then specified on the parent, as refer
 a collection of items represented by the child::
 
     class Parent(Base):
-        __tablename__ = 'parent'
+        __tablename__ = "parent"
         id = Column(Integer, primary_key=True)
         children = relationship("Child")
 
+
     class Child(Base):
-        __tablename__ = 'child'
+        __tablename__ = "child"
         id = Column(Integer, primary_key=True)
-        parent_id = Column(Integer, ForeignKey('parent.id'))
+        parent_id = Column(Integer, ForeignKey("parent.id"))
 
 To establish a bidirectional relationship in one-to-many, where the "reverse"
 side is a many to one, specify an additional :func:`_orm.relationship` and connect
 the two using the :paramref:`_orm.relationship.back_populates` parameter::
 
     class Parent(Base):
-        __tablename__ = 'parent'
+        __tablename__ = "parent"
         id = Column(Integer, primary_key=True)
         children = relationship("Child", back_populates="parent")
 
+
     class Child(Base):
-        __tablename__ = 'child'
+        __tablename__ = "child"
         id = Column(Integer, primary_key=True)
-        parent_id = Column(Integer, ForeignKey('parent.id'))
+        parent_id = Column(Integer, ForeignKey("parent.id"))
         parent = relationship("Parent", back_populates="children")
 
 ``Child`` will get a ``parent`` attribute with many-to-one semantics.
@@ -55,7 +57,7 @@ on a single :func:`_orm.relationship` instead of using
 :paramref:`_orm.relationship.back_populates`::
 
     class Parent(Base):
-        __tablename__ = 'parent'
+        __tablename__ = "parent"
         id = Column(Integer, primary_key=True)
         children = relationship("Child", backref="parent")
 
@@ -88,13 +90,14 @@ Many to one places a foreign key in the parent table referencing the child.
 attribute will be created::
 
     class Parent(Base):
-        __tablename__ = 'parent'
+        __tablename__ = "parent"
         id = Column(Integer, primary_key=True)
-        child_id = Column(Integer, ForeignKey('child.id'))
+        child_id = Column(Integer, ForeignKey("child.id"))
         child = relationship("Child")
 
+
     class Child(Base):
-        __tablename__ = 'child'
+        __tablename__ = "child"
         id = Column(Integer, primary_key=True)
 
 Bidirectional behavior is achieved by adding a second :func:`_orm.relationship`
@@ -102,13 +105,14 @@ and applying the :paramref:`_orm.relationship.back_populates` parameter
 in both directions::
 
     class Parent(Base):
-        __tablename__ = 'parent'
+        __tablename__ = "parent"
         id = Column(Integer, primary_key=True)
-        child_id = Column(Integer, ForeignKey('child.id'))
+        child_id = Column(Integer, ForeignKey("child.id"))
         child = relationship("Child", back_populates="parents")
 
+
     class Child(Base):
-        __tablename__ = 'child'
+        __tablename__ = "child"
         id = Column(Integer, primary_key=True)
         parents = relationship("Parent", back_populates="child")
 
@@ -116,9 +120,9 @@ Alternatively, the :paramref:`_orm.relationship.backref` parameter
 may be applied to a single :func:`_orm.relationship`, such as ``Parent.child``::
 
     class Parent(Base):
-        __tablename__ = 'parent'
+        __tablename__ = "parent"
         id = Column(Integer, primary_key=True)
-        child_id = Column(Integer, ForeignKey('child.id'))
+        child_id = Column(Integer, ForeignKey("child.id"))
         child = relationship("Child", backref="parents")
 
 .. _relationships_one_to_one:
@@ -143,16 +147,17 @@ a :ref:`many-to-one <relationship_patterns_m2o>` (``Child.parent``)
 relationships::
 
     class Parent(Base):
-        __tablename__ = 'parent'
+        __tablename__ = "parent"
         id = Column(Integer, primary_key=True)
 
         # one-to-many collection
         children = relationship("Child", back_populates="parent")
 
+
     class Child(Base):
-        __tablename__ = 'child'
+        __tablename__ = "child"
         id = Column(Integer, primary_key=True)
-        parent_id = Column(Integer, ForeignKey('parent.id'))
+        parent_id = Column(Integer, ForeignKey("parent.id"))
 
         # many-to-one scalar
         parent = relationship("Parent", back_populates="children")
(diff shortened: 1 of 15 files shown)


ThirVondukr:
Is it better to encourage using context managers with `Session` objects
 or it would be better to leave it as is?

```py
session = Session(bind=engine)
for id_ in random.sample(ids, n):
    session.query(Customer).filter(Customer.id == id_).one()
```
```py
with Session(bind=engine) as session:
    for id_ in random.sample(ids, n):
        session.query(Customer).filter(Customer.id == id_).one()
```

CaselIT:
Thanks for the great effort!

doc/build/orm/basic_relationships.rst
```diff
@@ -263,55 +272,64 @@ remote tables with which to link::
...
-        association_table = Table('association', Base.metadata,
-            Column('left_id', ForeignKey('left.id'), primary_key=True),
-            Column('right_id', ForeignKey('right.id'), primary_key=True)
-        )
+    association_table = Table(
+        "association",
+        Base.metadata,
+        Column("left_id", ForeignKey("left.id"), primary_key=True),
+        Column("right_id", ForeignKey("right.id"), primary_key=True),
+    )
```
CaselIT, line 280: I think this should have an additional indentation for rst to consider it code

doc/build/orm/declarative_mixins.rst
```diff
@@ -273,12 +282,12 @@ or alternatively, the string form (which ultimately generates a lambda)::
...
         def target_id(cls):
-            return Column('target_id', ForeignKey('target.id'))
+            return Column("target_id", ForeignKey("target.id"))
 
         @declared_attr
         def target(cls):
-            return relationship("Target",
-                primaryjoin="Target.id==%s.target_id" % cls.__name__
+            return relationship(
+                Target, primaryjoin=lambda: Target.id == cls.target_id
```
CaselIT, line 290: ok

CaselIT (commented):
Great work, just a minor thing. Thanks a lot

@zzzeek you can take a look

doc/build/orm/basic_relationships.rst
ThirVondukr: That's probably the case, i'm not really familiar with rst, i will restore original indentation

doc/build/orm/declarative_mixins.rst
ThirVondukr: I think this is "better" since we have access to current class, it's also a bit easier to read. I don't think i need `lambda` here though since it's a `declared_attr`?

CaselIT: Yes, I think you can avoid it in this case, but it's fine also with the lambda.

I left a comment here, since it was the only place I found with some changes to the code, other than formatting

ThirVondukr: I also used `f-strings` in some places instead of `%` formatting, since `.format` and `f-strings` are the recommended way as of now

doc/build/orm/basic_relationships.rst
```diff
@@ -674,23 +702,23 @@ class were available, we could also apply it afterwards::
     # first, module A, where Child has not been created yet,
     # we create a Parent class which knows nothing about Child
 
+
     class Parent(Base):
-        # ...
+        ...
```
zzzeek, line 707: this is a bit of a change (remember these docs date back in some cases to Python 2.3).....but I guess this allows code formatters to work so OK

doc/build/orm/collections.rst
```diff
@@ -203,24 +204,28 @@ to achieve a simple dictionary collection.  It produces a dictionary class that
 of the mapped class as a key.   Below we map an ``Item`` class containing
 a dictionary of ``Note`` items keyed to the ``Note.keyword`` attribute::
 
-    from sqlalchemy import Column, Integer, String, ForeignKey
+    from sqlalchemy import Column, ForeignKey, Integer, String
+    from sqlalchemy.ext.declarative import declarative_base
```
zzzeek, line 208: this should be sqlalchemy.orm.declarative_base

these should all have been migrated awhile back so not sure how this is still around (others maybe, I guess, if this one was missed)

doc/build/orm/composites.rst
```diff
@@ -26,12 +26,14 @@ A simple example represents pairs of columns as a ``Point`` object.
             return self.x, self.y
 
         def __repr__(self):
-            return "Point(x=%r, y=%r)" % (self.x, self.y)
+            return f"Point(x={self.x}, y={self.y})"
```
zzzeek, line 29: this is a good change but technically this should be:

```py
    return f"Point(x={self.x!r}, y={self.y!r})"
```

doc/build/orm/composites.rst
```diff
@@ -164,16 +177,19 @@ itself be a composite object, which is then mapped to a class ``HasVertex``::
...
```
zzzeek, line 180: same here  `{self.x!r}`

doc/build/orm/declarative_mixins.rst
zzzeek: OK but this is wrong in context, this doc example is explicitly illustrating string config, the lambda is in the previous example:

```
    The condition above is resolved using a lambda::
            # example with lambda

     or alternatively, the string form (which ultimately generates a lambda)::

          # the changed example is here, and should use as string (or we'd not have this whole paragraph)

```

so this has to change back (or be removed, change it back seems more appropriate at the moment)

doc/build/orm/declarative_mixins.rst
```diff
@@ -510,19 +534,20 @@ function should be invoked **for each class in the hierarchy**, in *almost*
         @declared_attr.cascading
         def id(cls):
             if has_inherited_table(cls):
-                return Column(ForeignKey('person.id'), primary_key=True)
-            else:
```
zzzeek, line 514: this one is a preference call, and I generally prefer the first way:

```py
if x:
    return a
else:
    return b
```

I find it makes it clear that the two conditions are equally "weighted"

doc/build/orm/extensions/asyncio.rst
```diff
@@ -140,19 +142,10 @@ illustrates a complete example including mapper and session configuration::
...
-    from sqlalchemy import DateTime
-    from sqlalchemy import ForeignKey
-    from sqlalchemy import func
-    from sqlalchemy import Integer
-    from sqlalchemy import String
-    from sqlalchemy.ext.asyncio import AsyncSession
-    from sqlalchemy.ext.asyncio import async_sessionmaker
+    from sqlalchemy import Column, DateTime, ForeignKey, Integer, String, func
     from sqlalchemy.ext.asyncio import create_async_engine
     from sqlalchemy.future import select
```
zzzeek, line 147: this should be "from sqlalchemy import select"

doc/build/orm/extensions/asyncio.rst
```diff
@@ -140,19 +142,10 @@ illustrates a complete example including mapper and session configuration::
...
     import asyncio
 
-    from sqlalchemy import Column
-    from sqlalchemy import DateTime
-    from sqlalchemy import ForeignKey
-    from sqlalchemy import func
-    from sqlalchemy import Integer
-    from sqlalchemy import String
-    from sqlalchemy.ext.asyncio import AsyncSession
-    from sqlalchemy.ext.asyncio import async_sessionmaker
```
zzzeek, line 150: async_sessionmaker is used in the example so would need to be in the imports

doc/build/orm/extensions/asyncio.rst
```diff
@@ -720,13 +724,18 @@ constructor::
 
     from asyncio import current_task
 
-    from sqlalchemy.ext.asyncio import async_sessionmaker
-    from sqlalchemy.ext.asyncio import async_scoped_session
-    from sqlalchemy.ext.asyncio import AsyncSession
-
-    async_session_factory = async_sessionmaker(some_async_engine, expire_on_commit=False)
```
zzzeek, line 727: here also, async_sessionmaker() should be used.   sessionmaker() doesn't handle typing for asyncio correctly so the pattern is async_sessionmaker().

doc/build/orm/extensions/asyncio.rst
```diff
@@ -720,13 +724,18 @@ constructor::
 
     from asyncio import current_task
 
-    from sqlalchemy.ext.asyncio import async_sessionmaker
-    from sqlalchemy.ext.asyncio import async_scoped_session
-    from sqlalchemy.ext.asyncio import AsyncSession
```
zzzeek, line 725: the AsyncSession import can stay removed though

zzzeek (changes requested):
thanks for the great work!   All the stylistic changes that don't actually change the parse tree of the code are great here.   The changes that change the choices made in specific API objects and patterns have some issues that need to be fixed.     it's best to keep formatting / style change patches separate from changes that alter the code itself as formatting changes can be skimmed much more quickly if nothing else changes.

zzzeek:
`sqlalchemy.ext.asyncio.async_sessionmaker` is used in some code blocks which doesn't exist

this is new API for 2.0 and is in source code here: https://github.com/sqlalchemy/sqlalchemy/blob/7efcda1758d1690b5974395806526808c748b65b/lib/sqlalchemy/ext/asyncio/session.py#L1244

doc/build/orm/collections.rst
ThirVondukr: I didn't really pay attention to imports, since declarative is now part of orm i will fix this

doc/build/orm/declarative_mixins.rst
ThirVondukr: Generally it's a good idea to reduce nesting by adding early returns:
```py
def function():
    if condition:
        if another_condition:
            return 42
        else:
            raise ValueError
    else:
        return 0


def function():
    if not condition:
        return 0
    if not another_condition:
        raise ValueError
    return 42
```
If you don't like that i can change it back to if-else

ThirVondukr:
> > `sqlalchemy.ext.asyncio.async_sessionmaker` is used in some code blocks which doesn't exist
> 
> this is new API for 2.0 and is in source code here:
> 
> https://github.com/sqlalchemy/sqlalchemy/blob/7efcda1758d1690b5974395806526808c748b65b/lib/sqlalchemy/ext/asyncio/session.py#L1244

I wasn't able to find it in my local installation (`SQLAlchemy = "^1.4.34"`), is it new/still in development?

doc/build/orm/declarative_mixins.rst
zzzeek: well in this case, the max level of nesting doesnt change.    if the "else" condition started garnering a lot more logic, then yes, I'd switch it.  but currently, it looks nicer to me showing these two conditions as equal branch points, since extra logic to increase nesting could just as easily be incurred in the first conditional and not the second, or it could be in both, in which case you'd refactor into two separate methods, there's all kinds of possibilities that nonetheless none are realized.  so don't make that decision early, leave the two branches as equal weight.

zzzeek:
> > > `sqlalchemy.ext.asyncio.async_sessionmaker` is used in some code blocks which doesn't exist
> > 
> > 
> > this is new API for 2.0 and is in source code here:
> > https://github.com/sqlalchemy/sqlalchemy/blob/7efcda1758d1690b5974395806526808c748b65b/lib/sqlalchemy/ext/asyncio/session.py#L1244
> 
> I wasn't able to find it in my local installation (`SQLAlchemy = "^1.4.34"`), is it new/still in development?

it's committed to main.   the git checkout you're working in here is the same branch, so it's in that set of files...I wouldn't be looking in a 1.4.34 installation....

ThirVondukr:
Is there a reason docs import `DeclarativeMeta` from `sqlalchemy.orm.decl_api` and not from `sqlalchemy.orm`?

ThirVondukr:
@zzzeek In `inheritance.rst` i formatted all `__mapper_args__` to be multiline so they're uniform, you don't mind?

doc/build/orm/extensions/asyncio.rst
```diff
@@ -720,13 +732,21 @@ constructor::
...
+    from sqlalchemy.ext.asyncio import (
+        AsyncSession,
+        async_scoped_session,
+        async_sessionmaker,
+    )
 
+    async_session_factory = async_sessionmaker(
+        some_async_engine,
+        expire_on_commit=False,
+        class_=AsyncSession,
```
CaselIT, line 744: I think we can remove `class_` here, since that's implied by the `async_sessionmaker`. 

should be removed also by the imports once updated

CaselIT (changes requested):
just a minor change.

Thanks for the effort!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision fd8f60fcfe9cda0c2ba6dc9ddd171bf85a180295 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change fd8f60fcfe9cda0c2ba6dc9ddd171bf85a180295: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3842

doc/build/orm/extensions/asyncio.rst
zzzeek: yeah this should be removed

doc/build/orm/extensions/asyncio.rst
```diff
@@ -720,13 +732,21 @@ constructor::
...
+    )
+    AsyncScopedSession = async_scoped_session(
+        async_session_factory,
+        copefunc=current_task,
```
zzzeek, line 748: scopefunc

zzzeek (commented):
still a few things, github PR reviews are very hard to read so i will pull into gerrit and fix

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision fd8f60fcfe9cda0c2ba6dc9ddd171bf85a180295 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset fd8f60fcfe9cda0c2ba6dc9ddd171bf85a180295 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3842

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3842 has been **merged**. Congratulations! :)

PR #7964, opened 2022-04-26
Add git-blame-ignore-revs file

stephenfin:
### Description

This is helpful when using git-blame on files that 'black' heavily modified back in 2019'ish. While git doesn't specify this name, there seems to be a general convention around using this. See Django [1] and Twisted [2], for example.

[1] https://github.com/django/django/blob/main/.git-blame-ignore-revs
[2] https://github.com/twisted/twisted/blob/trunk/.git-blame-ignore-revs

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.
- [x] None of the above
	- This is infra-related

**Have a nice day!**


jvanasco:
I think this would be useful **if** the hashes were each commented with the reason why.

This file format supports comments, and commenting them is highly recommended by many projects and users.

stephenfin:
> I think this would be useful **if** the hashes were each commented with the reason why.
> 
> This file format supports comments, and commenting them is highly recommended by many projects and users.

Good idea. Done.

zzzeek:
hi @stephenfin  -

this is interesting I didnt know you could do this.   in fact it would have saved me a lot of time trying to blame things in the past, there's actually probably dozens of such revs like this over the years.   

@CaselIT any comments?

stephenfin:
It's rather nifty alright. The only issues I see with this is that GitHub and other forges don't support this and the user needs to configure `git` to use this by default (`git config --global blame.ignoreRevsFile .git-blame-ignore-revs` will do the trick - you could sub `--global` for `--local`, of course). Certainly better than nothing though.

I only added the two big black-related commits I spotted in here but if there are more then we could easily extend this. I just wanted to find other projects that were using `black` so I could use that as justification to start using it in OpenStack land :smile:

jvanasco:
> Good idea. Done.

Can you confirm the format you used works? While I do know about this, and have used it before, I've only seen it implemented with the comment on a dedicated line.  IMHO your inline comments look way better and don't require the spaces for visual reading the other method uses - I've just never seen an implementation using them before.

zzzeek:
> It's rather nifty alright. The only issues I see with this is that GitHub and other forges don't support this and the user needs to configure `git` to use this by default (`git config --global blame.ignoreRevsFile .git-blame-ignore-revs` will do the trick - you could sub `--global` for `--local`, of course). Certainly better than nothing though.
> 

oh OK, so this file is not automatically consumed by git anyway, one has to configure it.  that's almost even better in my opinion because it wont be misleading to people who don't know it's happening.

However in that case is it possible we could also put a little more comments in the file indicating "to use this file, configure git as follows" so that people know how to take advantage of it.

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+1e1a38e7801f410f244e4bbb44ec795ae152e04e  # initial blackification
+1e278de4cc9a4181e0747640a960e80efcea1ca9  # follow up mass refactoring
```
CaselIT, line 2:
```suggestion
1e278de4cc9a4181e0747640a960e80efcea1ca9  # follow up mass style changes
```

CaselIT (changes requested):
Ok for me, like @zzzeek suggested an initial comment on how to set this git to use this file is useful

stephenfin:
> > Good idea. Done.
> 
> Can you confirm the format you used works? While I do know about this, and have used it before, I've only seen it implemented with the comment on a dedicated line. IMHO your inline comments look way better and don't require the spaces for visual reading the other method uses - I've just never seen an implementation using them before.

It appears to:

```
$ git blame examples/adjacency_list/adjacency_list.py | head -20 | tail -10
7f91210e94 (Mike Bayer       2010-01-18 21:29:55 +0000  11) from sqlalchemy.orm.collections import attribute_mapped_collection
7f91210e94 (Mike Bayer       2010-01-18 21:29:55 +0000  12) 
7f91210e94 (Mike Bayer       2010-01-18 21:29:55 +0000  13) 
007309015b (Mike Bayer       2012-01-28 17:22:04 -0500  14) Base = declarative_base()
7f91210e94 (Mike Bayer       2010-01-18 21:29:55 +0000  15) 
8dad0f93e8 (Paul Brown       2017-03-17 00:23:49 -0500  16) 
007309015b (Mike Bayer       2012-01-28 17:22:04 -0500  17) class TreeNode(Base):
007309015b (Mike Bayer       2012-01-28 17:22:04 -0500  18)     __tablename__ = "tree"
007309015b (Mike Bayer       2012-01-28 17:22:04 -0500  19)     id = Column(Integer, primary_key=True)
007309015b (Mike Bayer       2012-01-28 17:22:04 -0500  20)     parent_id = Column(Integer, ForeignKey(id))
```

```
$ git config --global blame.ignoreRevsFile ''
```

```
$ git blame examples/adjacency_list/adjacency_list.py | head -20 | tail -10
7f91210e94 (Mike Bayer       2010-01-18 21:29:55 +0000  11) from sqlalchemy.orm.collections import attribute_mapped_collection
7f91210e94 (Mike Bayer       2010-01-18 21:29:55 +0000  12) 
7f91210e94 (Mike Bayer       2010-01-18 21:29:55 +0000  13) 
007309015b (Mike Bayer       2012-01-28 17:22:04 -0500  14) Base = declarative_base()
7f91210e94 (Mike Bayer       2010-01-18 21:29:55 +0000  15) 
8dad0f93e8 (Paul Brown       2017-03-17 00:23:49 -0500  16) 
007309015b (Mike Bayer       2012-01-28 17:22:04 -0500  17) class TreeNode(Base):
1e1a38e780 (Mike Bayer       2019-01-06 01:14:26 -0500  18)     __tablename__ = "tree"
007309015b (Mike Bayer       2012-01-28 17:22:04 -0500  19)     id = Column(Integer, primary_key=True)
007309015b (Mike Bayer       2012-01-28 17:22:04 -0500  20)     parent_id = Column(Integer, ForeignKey(id))
```

The key difference is the change to the `__tablename__ = "tree"`, which was modified in 1e1a38e7801f410f244e4bbb44ec795ae152e04e

stephenfin:
> Ok for me, like @zzzeek suggested an initial comment on how to set this git to use this file is useful

Happy to do that. Just tell me where it needs to go. I can't find any obvious `CONTRIBUTING` doc in-tree, only a `README.unittests.rst` file, and the developer docs seem to live elsewhere? https://www.sqlalchemy.org/develop.html

stephenfin:
> > Ok for me, like @zzzeek suggested an initial comment on how to set this git to use this file is useful
> 
> Happy to do that. Just tell me where it needs to go. I can't find any obvious `CONTRIBUTING` doc in-tree, only a `README.unittests.rst` file, and the developer docs seem to live elsewhere? [sqlalchemy.org/develop.html](https://www.sqlalchemy.org/develop.html)

I put it in the file itself. Don't like how I've worded it so open to suggestions.

CaselIT:
It looks ok to me. Makes it clear what's the purpose of the file and how to use it. Thanks!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 4be7330a51ee48e360e10a7c5a6d90fea62a9515 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 4be7330a51ee48e360e10a7c5a6d90fea62a9515: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3814

stephenfin:
Funnily enough, this popped up in the GitHub UI today: [Ignore commits in the blame view](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)

sqla-tester:

this looks great, thanks for the contribution!

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3814

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3814 has been **merged**. Congratulations! :)

PR #8022, opened 2022-05-15
Fixes: 8023. Getting the original error from the exception wrapper.

dotX12:
adds functionality for: https://github.com/sqlalchemy/sqlalchemy/discussions/8023

When we need to handle an error in depth, just get its details, which constraint caused the error, what table we are working with, we need to know the original error.

In the current realities, this is impossible to do, because. ex.orig is the string that is concatenated from DETAIL + HINT.

My revision allows you to get the original error through
```
orig.obj
detail via orig.detail
message via orig.message
```

This addition partially applies to #6199.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/doc/build/changelog/unreleased_14/6199_2.rst
+++ b/doc/build/changelog/unreleased_14/6199_2.rst
@@ -1,0 +1,9 @@
+.. change::
+    :tags: usecase, asyncio, postgresql
+    :tickets: 6199
+
+    Added accessors ``.detail``, ``.message`` and original object exception ``.obj`` to the ``.orig``
+    attribute of the SQLAlchemy exception class raised by the asyncpg DBAPI
+    adapter, that is, the intermediary exception object that wraps on top of
+    that raised by the asyncpg library itself, but below the level of the
+    SQLAlchemy dialect.
--- a/doc/build/changelog/unreleased_14/8023.rst
+++ b/doc/build/changelog/unreleased_14/8023.rst
@@ -1,6 +1,6 @@
 .. change::
     :tags: usecase, asyncio, postgresql
-    :tickets: 6199
+    :tickets: 8023
 
     Added accessors ``.detail``, ``.message`` and original object exception ``.obj`` to the ``.orig``
     attribute of the SQLAlchemy exception class raised by the asyncpg DBAPI
--- a/lib/sqlalchemy/dialects/postgresql/asyncpg.py
+++ b/lib/sqlalchemy/dialects/postgresql/asyncpg.py
@@ -653,6 +653,11 @@ class AsyncAdapt_asyncpg_connection(AdaptedConnection):
                     translated_error.pgcode = (
                         translated_error.sqlstate
                     ) = getattr(error, "sqlstate", None)
+
+                    translated_error.detail = getattr(error, "detail", None)
+                    translated_error.message = getattr(error, "message", None)
+                    translated_error.obj = error
+
                     raise translated_error from error
             else:
                 raise error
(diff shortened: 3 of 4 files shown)


CaselIT:
Why don't we create a mixin class with `__getattr__` that gets properties from an `orig` attribute?
This could when be added to the error classes used by sqlalchemy to emulate the DBAPI errors, and may additionally be used by other dialects to do the same

zzzeek:
I'm a little hesitant to add `__getattr__` schemes since they make attribute access / typing / documentation / predictability more complicated (which I know because we have so many `__getattr__` schemes).

I think we should try to hit this in terms of all the async dialects all of which have to do what we are doing here, which would be that we make a common mixin / base for all the async-adapted DBAPIs.  for asyncpg it would be used here: https://github.com/sqlalchemy/sqlalchemy/blob/19f0566dc98cc1d088777a9f1e1e7e898d861c05/lib/sqlalchemy/dialects/postgresql/asyncpg.py#L760

```py
    from ..exc import EmulatedDBAPIException

    class Error(EmulatedDBAPIException):
        pass
```


that way people can look at `EmulatedDBAPIException` to see the fields, which would be like:

```py
class EmulatedDBAPIException:
    """an exception class used to adapt non-DBAPI database driver exceptions
    into DBAPI exceptions"""

    message: str
    """string message indicating the error"""

    orig: Exception
    """the original driver-level exception object"""
```


that looks kind of boring so we can also add an accessor on https://github.com/sqlalchemy/sqlalchemy/blob/19f0566dc98cc1d088777a9f1e1e7e898d861c05/lib/sqlalchemy/exc.py#L525 called "driver_exception".  This would work similarly to how we have "dbapi_connection" / "driver_connection" on Connection.   

that way someone that knows asyncpg can receive the top level SQLAlchemy DBAPIException and just refer to .driver_exception to get all the asyncpg fields.

@dotX12  can we do that here?

dotX12:
> I'm a little hesitant to add `__getattr__` schemes since they make attribute access / typing / documentation / predictability more complicated (which I know because we have so many `__getattr__` schemes).
> 
> I think we should try to hit this in terms of all the async dialects all of which have to do what we are doing here, which would be that we make a common mixin / base for all the async-adapted DBAPIs. for asyncpg it would be used here:
> 
> https://github.com/sqlalchemy/sqlalchemy/blob/19f0566dc98cc1d088777a9f1e1e7e898d861c05/lib/sqlalchemy/dialects/postgresql/asyncpg.py#L760-L761
> 
> ```python
>     from ..exc import EmulatedDBAPIException
> 
>     class Error(EmulatedDBAPIException):
>         pass
> ```
> 
> that way people can look at `EmulatedDBAPIException` to see the fields, which would be like:
> 
> ```python
> class EmulatedDBAPIException:
>     """an exception class used to adapt non-DBAPI database driver exceptions
>     into DBAPI exceptions"""
> 
>     message: str
>     """string message indicating the error"""
> 
>     orig: Exception
>     """the original driver-level exception object"""
> ```
> 
> that looks kind of boring so we can also add an accessor on
> 
> https://github.com/sqlalchemy/sqlalchemy/blob/19f0566dc98cc1d088777a9f1e1e7e898d861c05/lib/sqlalchemy/exc.py#L525
> 
> called "driver_exception". This would work similarly to how we have "dbapi_connection" / "driver_connection" on Connection.
> that way someone that knows asyncpg can receive the top level SQLAlchemy DBAPIException and just refer to .driver_exception to get all the asyncpg fields.
> 
> @dotX12 can we do that here?
Yes, your suggestion sounds cool, I'll be glad if you can finish it in this PR.
Thank you!

dotX12:
> 

Yes, your suggestion sounds cool, I'll be glad if you can finish it in this PR.
Thank you!

PR #8039, opened 2022-05-19
doc change - Link to dialect for Actian Avalanche, Vector, Actian X, and Ingres

clach04:
### Description

Added new entry in table, reformatted table due to longer name

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**


CaselIT:
from the readme it seems to be "work in progress 1.3-1.4". Do you know if it's now working?

It would not make much sense to add a dialect that supports only version lower than 1.3 :)

clach04:
> from the readme it seems to be "work in progress 1.3-1.4". Do you know if it's now working?
> 
> It would not make much sense to add a dialect that supports only version lower than 1.3 :)

@CaselIT Thanks for taking the time to dig into this! It does work :-) I'm currently using it with SQLAlchemy 1.4.36.

* pip install (from code) works cleanly with latest SQLAlchemy and demo code runs without issues (well there is a warning about cache support https://github.com/clach04/ingres_sa_dialect/issues/4)
* ipython-sql, etc. also works

The test suite is not 100% clean yet. I was originally hesitant to link to it but people are not finding it and it has been usable for a while now.

Because the test suite is not 100% I (conservatively and pessimistically) consider it work in progress.

CaselIT:
understood, I think we can add it.

cc @zzzeek to confirm

PR #8072, opened 2022-05-29
Update declarative_styles.rst: Update code example to attrs TNG usage.

do3cc:
### Description

Update docs to new style usage of attrs.
This is the default since December 2021.
While the old style still works, the newer one looks much nicer and is likely to be dominant pretty quickly. Imho.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/doc/build/orm/declarative_styles.rst
+++ b/doc/build/orm/declarative_styles.rst
@@ -229,7 +229,8 @@ with a new __slots__ based class, which is not supported. When using the old
 style annotation ``@attr.s`` or using ``define(slots=False)``, the class
 does not get replaced. Furthermore attrs removes its own class-bound attributes
 after the decorator runs, so that SQLAlchemy's mapping process takes over these
-attributes without any issue.
+attributes without any issue. Both decorators, ``@attr.s`` and ``@define(slots=False)``
+work with SQLAlchemy.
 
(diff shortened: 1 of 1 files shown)


doc/build/orm/declarative_styles.rst
```diff
@@ -450,7 +450,7 @@ A mapping using ``@attr.s``, in conjunction with imperative table::
```
CaselIT, line 450: A couple of lines above it says

> A mapping using ``@attr.s``, in conjunction with imperative table::

That should be updated too. Also I think we may mention in that like that using the old style `attr.s` and `attr.ib` is also supported

CaselIT (commented):
Thanks for the PR

The old style seems to no longer be mentioned looking quickly at https://www.attrs.org/en/stable/examples.html so I think it makes sense to update the style.

CaselIT (approved):
Looks ok for me 

@zzzeek can you take a look?

zzzeek (approved):
this is fine, make sure it's correctly backported to 1.4.

These docs are going to be also undergoing a huge reorganiztion for both main / 1.4 in the coming days / weeks.

CaselIT:
thanks!

Issue #8062, opened 2022-05-27
Leading  curly brace (%7B) character in password of a connection string leads to login failure


aagateuip:
### Describe the bug

When password in a connection string starts with "{" (url encoded to %7B) the login fails

Database: Azure SQL Database 

Tried with SQLAlchemy==1.4.36 and SQLAlchemy==1.3.24

 

### To Reproduce

```python
from sqlalchemy import create_engine
engine = create_engine('mssql+pyodbc://testuser@sqlserverhost-sql.database.windows.net:%7B4cbW123@sqlserverhost-sql.database.windows.net:1433/AppDb?driver=ODBC+Driver+17+for+SQL+Server')
engine.connect()
```

### Error

```
Traceback (most recent call last):
  File "/home/appuser/venv/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 2336, in _wrap_pool_connect
    return fn()
  File "/home/appuser/venv/lib/python3.8/site-packages/sqlalchemy/pool/base.py", line 304, in unique_connection
    return _ConnectionFairy._checkout(self)
  File "/home/appuser/venv/lib/python3.8/site-packages/sqlalchemy/pool/base.py", line 778, in _checkout
    fairy = _ConnectionRecord.checkout(pool)
  File "/home/appuser/venv/lib/python3.8/site-packages/sqlalchemy/pool/base.py", line 495, in checkout
    rec = pool._do_get()
  File "/home/appuser/venv/lib/python3.8/site-packages/sqlalchemy/pool/impl.py", line 140, in _do_get
    self._dec_overflow()
  File "/home/appuser/venv/lib/python3.8/site-packages/sqlalchemy/util/langhelpers.py", line 68, in __exit__
    compat.raise_(
  File "/home/appuser/venv/lib/python3.8/site-packages/sqlalchemy/util/compat.py", line 182, in raise_
    raise exception
  File "/home/appuser/venv/lib/python3.8/site-packages/sqlalchemy/pool/impl.py", line 137, in _do_get
    return self._create_connection()
  File "/home/appuser/venv/lib/python3.8/site-packages/sqlalchemy/pool/base.py", line 309, in _create_connection
    return _ConnectionRecord(self)
  File "/home/appuser/venv/lib/python3.8/site-packages/sqlalchemy/pool/base.py", line 440, in __init__
    self.__connect(first_connect_check=True)
[...]

PR #8124, opened 2022-06-11
Attempt to resolve edge case related to escaping of curly braces in ODBC connection strings

jayaddison:
### Description
Issue #8062 (resolved by 8ac7cb92b4972a08b8008b80b34989694510139f) relates to handling of curly braces within ODBC connection strings.

There may be a remaining [edge case](https://github.com/sqlalchemy/sqlalchemy/issues/8062#issuecomment-1152894834) if a connection string token begins with an opening curly brace and finishes with a closing curly brace (in other words, it appears to already be escaped by curly braces).

This pull request offers a solution, which is not to perform escaping on tokens that appear to already be escaped.

**Question**: should I open a follow-up issue for this?

### Checklist
- [x] A short code fix
	- please include tests.

### DIFF
--- a/lib/sqlalchemy/connectors/pyodbc.py
+++ b/lib/sqlalchemy/connectors/pyodbc.py
@@ -77,9 +77,6 @@ class PyODBCConnector(Connector):
         else:
 
             def check_quote(token: str) -> str:
-                # token is already quoted, so return it as-is
-                if str(token).startswith("{") and str(token).endswith("}"):
-                    return token
                 if ";" in str(token) or str(token).startswith("{"):
                     token = "{%s}" % token.replace("}", "}}")
                 return token
--- a/lib/sqlalchemy/connectors/pyodbc.py
+++ b/lib/sqlalchemy/connectors/pyodbc.py
@@ -76,12 +76,11 @@ class PyODBCConnector(Connector):
             connectors = [unquote_plus(keys.pop("odbc_connect"))]
         else:
 
-            def check_quote(token: Any) -> str:
-                token = str(token)
+            def check_quote(token: str) -> str:
                 # token is already quoted, so return it as-is
-                if token.startswith("{") and token.endswith("}"):
+                if str(token).startswith("{") and str(token).endswith("}"):
                     return token
-                if ";" in token or token.startswith("{"):
+                if ";" in str(token) or str(token).startswith("{"):
                     token = "{%s}" % token.replace("}", "}}")
                 return token
 
--- a/lib/sqlalchemy/connectors/pyodbc.py
+++ b/lib/sqlalchemy/connectors/pyodbc.py
@@ -77,6 +77,9 @@ class PyODBCConnector(Connector):
         else:
 
             def check_quote(token: str) -> str:
+                # token is already quoted, so return it as-is
+                if token.startswith("{") and token.endswith("}"):
+                    return token
                 if ";" in str(token) or str(token).startswith("{"):
                     token = "{%s}" % token.replace("}", "}}")
                 return token
--- a/lib/sqlalchemy/connectors/pyodbc.py
+++ b/lib/sqlalchemy/connectors/pyodbc.py
@@ -76,8 +76,7 @@ class PyODBCConnector(Connector):
             connectors = [unquote_plus(keys.pop("odbc_connect"))]
         else:
 
-            def check_quote(token: Any) -> str:
-                token = str(token)
+            def check_quote(token: str) -> str:
                 # token is already quoted, so return it as-is
                 if token.startswith("{") and token.endswith("}"):
                     return token
@@ -85,7 +84,10 @@ class PyODBCConnector(Connector):
                     token = "{%s}" % token.replace("}", "}}")
                 return token
 
-            keys = dict((k, check_quote(v)) for k, v in keys.items())
+            keys = dict(
+                (k, check_quote(v) if type(v) == str else v)
+                for k, v in keys.items()
+            )
 
             dsn_connection = "dsn" in keys or (
                 "host" in keys and "database" not in keys
--- a/lib/sqlalchemy/connectors/pyodbc.py
+++ b/lib/sqlalchemy/connectors/pyodbc.py
@@ -78,7 +78,7 @@ class PyODBCConnector(Connector):
 
             def check_quote(token: str) -> str:
                 # token is already quoted, so return it as-is
-                if token.startswith("{") and token.endswith("}"):
+                if str(token).startswith("{") and str(token).endswith("}"):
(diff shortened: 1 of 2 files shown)


lib/sqlalchemy/connectors/pyodbc.py
```diff
@@ -77,6 +77,9 @@ def create_connect_args(self, url: URL) -> ConnectArgsType:
         else:
 
             def check_quote(token: str) -> str:
+                # token is already quoted, so return it as-is
+                if token.startswith("{") and token.endswith("}"):
```
jayaddison, line 81: Note: this line does _not_ perform a type-cast (`str(token)`) unlike the existing conditional.

CaselIT:
Thanks.

Do you know if there is a spec somewhere?

jayaddison:
The documentation at https://docs.microsoft.com/en-us/openspecs/sql_server_protocols/ms-odbcstr/13b4e848-b36c-4b11-acce-d6bf199d5391 ("[MS-ODBCSTR]: ODBC Connection String Structure") seems to be the most comprehensive spec I could find; it has a few simple examples about special character escaping, but not a lot of them.

The [ODBC Connection String Format ABNF syntax](https://docs.microsoft.com/en-us/openspecs/sql_server_protocols/ms-odbcstr/55953f0e-2d30-4ad4-8e56-b4207e491409) suggests that there are generally two special characters to consider: `;` and `=`, so perhaps we'd want to add coverage for both of those.

lib/sqlalchemy/connectors/pyodbc.py
jayaddison: (now I understand that it probably should be, though, since a port number, for example, could be passed as a non-string value)

lib/sqlalchemy/connectors/pyodbc.py
```diff
@@ -77,11 +77,17 @@ def create_connect_args(self, url: URL) -> ConnectArgsType:
...
+                    return token
+                if ";" in token or token.startswith("{"):
                     token = "{%s}" % token.replace("}", "}}")
                 return token
 
-            keys = dict((k, check_quote(v)) for k, v in keys.items())
+            keys = dict(
+                (k, check_quote(v) if type(v) == str else v)
```
jayaddison, line 88: Oops; I didn't intend to push this change - it's something I'd been experimenting with locally.  I'm not certain whether an equality check against the `str` type is reliable / universal.

lib/sqlalchemy/connectors/pyodbc.py
```diff
@@ -77,6 +77,9 @@ def create_connect_args(self, url: URL) -> ConnectArgsType:
         else:
 
             def check_quote(token: str) -> str:
+                # token is already quoted, so return it as-is
```
gordthompson, line 80: But what if alice's password really is `{password}`? Should we expect her to ODBC-escape the password herself? Or, put another way, if she does "the right thing" (IMO) and uses `URL.create()` then the result is not correct:

```python
import sqlalchemy as sa
from sqlalchemy.dialects.mssql import pyodbc

connection_url = sa.engine.URL.create(
    "mssql+pyodbc",
    username="alice",
    password="{password}",
    host="localhost",
    database="mydb",
    query=dict(driver="ODBC Driver 17 for SQL Server")
)
dialect = pyodbc.dialect()
connection = dialect.create_connect_args(connection_url)
print(connection[0][0])
# DRIVER={ODBC Driver 17 for SQL Server};Server=localhost;Database=mydb;UID=alice;PWD={password}
```

It should be `PWD={{password}}}`

test/dialect/mssql/test_engine.py
```diff
@@ -256,19 +257,35 @@ def test_pyodbc_odbc_connect_ignores_other_values(self):
                 "larry",
                 "{moe",
                 "localhost",
+                3306,
                 "mydb",
             ),
             (
-                "DRIVER={foob};Server=localhost;"
+                "DRIVER={foob};Server=localhost,3306;"
```
gordthompson, line 264: Super-nit-pick: This is

test/dialect/mssql/test_engine.py

but 3306 is the default port for mysql, not mssql. Maybe a better choice would be a value that is clearly not a default port, like 43210?

gordthompson (changes requested):
Thanks for your help with this!

gordthompson:
> there are generally two special characters to consider: `;` and `=`, so perhaps we'd want to add coverage for both of those.

`;` is already covered, and `=` in a password does not appear to be a problem. I just tested this with plain pyodbc and it worked:

```python
    connection_string = (
        "DRIVER=ODBC Driver 17 for SQL Server;"
        "SERVER=192.168.0.199;"
        "UID=archie;"
        "PWD=pi=3.14;"
        "DATABASE=test;"
    )
```

(archie's password is `pi=3.14`)

test/dialect/mssql/test_engine.py
jayaddison: Agh, thank you - my memory failing me.  That's corrected, although I take your point - omitting it where it's the default probably makes more sense.  I'll update that soon.

lib/sqlalchemy/connectors/pyodbc.py
jayaddison: Ok, yep - it's the balancing of these curly braces that was puzzling me here.  I do now see that `ESCAPEDRCB` exists in the language spec (where `RCB` = right curly bracket), and that there's no corresponding LCB.

I'm taking a bit more time to internalize and grok that, but initially I suppose it's reasonable.

jayaddison: Ok, I think I've convinced myself that 8ac7cb92b4972a08b8008b80b34989694510139f is correct as-is; my apologies for the distractions here.

Something about the directional nature of the escaping (see a `{`, apply the `}}` rule on the suffix) seemed to make it tricky to mentally enumerate the possibilities and validate strings visually - but I do think the implementation is correct.  I'll likely close this pull request soon.

Thanks for the review and explanations.

jayaddison:
Closing this because the underlying issue initially claimed:

> There may be a remaining [edge case](https://github.com/sqlalchemy/sqlalchemy/issues/8062#issuecomment-1152894834) if a connection string token begins with an opening curly brace and finishes with a closing curly brace (in other words, it appears to already be escaped by curly braces).

...does not appear to be correct.

Let me know if any of the test coverage on port numbers is potentially useful (it does seem like there are some loose type hints around there), but I don't think there is a problem with the logic here.

CaselIT:
> my apologies for the distractions here.

No problem. Better an extra set of eyes that one less. Thanks for taking a look to verify the implementation

Issue #7659, opened 2022-02-02
v2 doc todos checklist


CaselIT:
SQLAlchemy has a large volume of documentation that's currently using a mix of mostly v1 and some v2 style queries especially when talking with the orm.

For v2, while there is no plan to remove `query`, most of the docs should adopt the v2 style of doing things.

- [x] session.query -> session.execute/session.scalars
- [x] update Query API docs so that every method refers to background on migrating to the 2.0 pattern, or the corresponding method on Select.  general format/ language to use is:

```
        .. seealso::

            This API is available using :term:`2.0 style` queries as well,
            documented at:

            * :ref:`queryguide_inspection`
```

- [x] Declarative mappings need to use mapped_column()
- [x] update tutorials to introduce mapped_column(), some tips/etc. regarding "what happened to Column()?" etc. describe they are equivalent Not sure to the extent that the docs will push for “annotation only” style, probably more than less.   New dataclasses docs.  The existing “dataclasses” approach with __sa_field_whatever__ it’s called becomes legacy.
- [x] new DeclarativeBase should be used everywhere instead of declarative_base() function
- [ ] update central mypy docs in ORM guide
- [ ] more links in mypy extension to new mypy docs in ORM guide
- [x] document annotation-enabled mapped_column() usage, likely in the ORM tutorial
- [x] remove dead links to removed sections, running the sphinx build shows a lot of warnings to old sections
- [x] Reorganizing mapper config docs further as it’s very hard to follow between declarative/imperative w/ or w/out dataclasses, want to make room for new dataclasses approach.
- [x] update mypy plugin docs with a full deprecation note at the top
- [x] asyncio docs, note with "uses SQLAlchemy’s future mode, which has several potentially breaking changes. " should be reworded or removed for 2.0 docs
- [ ] update the section [Building Query-Enabled Properties](https://docs.sqlalchemy.org/en/14/orm/join_conditions.html#building-query-enabled-properties) tha does not have a direct 1->2 porting, since the standalone `with_parent` requires a relationship
- [ ] update the error section [An alias is being generated automatically for raw clauseelement](https://docs.sqlalchemy.org/en/20/errors.html#error-xaj1)
- [x] is an equivalent for [`Query.only_return_tuples`](https://docs.sqlalchemy.org/en/20/orm/queryguide/query.html#sqlalchemy.orm.Query.only_return_tuples) needed?
[...]

PR #8136, opened 2022-06-15
Add tools for documentation formatting and validation

tgross35:
### Description

The goal of this PR is to add tools that help with contributing to documentation, from discussion started on #7659. Goals include:
- Automatic formatting of code blocks in .rst documents and in docstrings  (blacken-docs)
- Flake8 validation of these code blocks (flake8-rst-docstrings, flake8-docstrings)
- Addition of an optional spellcheck tool for documentation (sphinxcontrib-spelling)
- Adding documentation on documentation. Seems like there's a little blurb here https://www.sqlalchemy.org/develop.html but I'm not sure where the source lives. I'd be happy to add more to this page, or a separate page on docs.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**


tgross35:
Status as-of dddad321d61d3240dab4d6de047e897610315033

### Spelling

Currently the spelling plugin is installed and working with `make spelling` (this requires the `enchant` C lib). Note for docs building docutils 0.17.1 needs to be installed (0.18.1 is their latest). The issue is somewhere in SQLA's tools and not with spelling though, zzzeek is looking at it.

There are some quirks with it
- To actually run, the `enchant` C library needs to be installed. I don't think this is a blocker to anything since the requirements can run and 
- The tokenizer pyenchant has some weirdness with tokenization. It splits on `'` and `-` so things like `as-hoc` or `doesn't` show up as `ad, hoc, doesn, t`, which really sucks. I also wouldn't consider this a blocker for making the tool available, I've already gone through and added a lot of split words to the dictionary. There is an open issue to fix this https://github.com/pyenchant/pyenchant/issues/286

### flake8

It looks like flake8-rst-docstrings and flake8-docstrings are already on the pre-commit dependencies, just need to make sure they're running

### blacken-docs

So far I like this one a lot.

One note is that it only runs in `.. code-block:: python` and `.. code-block:: pycon`,  (same in docstrings) and not `::` ([relevant issue here[(https://github.com/asottile/blacken-docs/issues/21)). This actually kind of provides an easy way to opt documents into formatting - specify the language type whenever it's ready. Only 7 docs get caught 

It doesn't pick up the `.. sourcecode:: pycon+sql` directives, which is big. My first thought on an easy way to fix this is a pretty easy syntax change. Just update zzzeeksphinx to be the lexer for `..code::pycon` and accept a syntax like:

```rst
.. code:: pycon

    >>> Base.metadata.create_all(engine)
    """{opensql}
    ...
    CREATE TABLE user_account (
        id INTEGER NOT NULL,
        name VARCHAR(30) NOT NULL,
        fullname VARCHAR,
        PRIMARY KEY (id)
    )
    ...
    """
```

Probably pretty easy to change the identifiers, then the whole block stays as valid (formattable) python. Nice too that then rst syntax highlighters would pick it up.

Side note: a `python` and `sql` role might be useful. Would give syntax highlighting to things like

```
 :python:`session.execute(delete(obj))` will issue :sql:`DELETE FROM user WHERE...`
```

```rst
.. role:: python(code)
   :language: python

.. role:: sql(code)
   :language: sql
```

tgross35:
Turns out I mixed up the flake8 things. `flake8-rst-docstrings` and `flake8-docstrings` don't validate the python within the docstrings, just the docstrings themselves. [flake8-rst](https://pypi.org/project/flake8-rst/) does validate the docstrings, but it's unmaintained since 2020. I reached out to their maintainer and also the maintainer of pycqa to see if somebody might take it over and maintain it.

Could also add doctest as a precommit hook.

Interestingly, zzzeek was the inspiration for flake8-rst

image

zzzeek:
id have to look and see what the rationale for "pycon+sql" as a separate thing was.  it's very old and lots of things have changed since then, including that I've ditched the [SQL] button thing which seemed so clever 15 years ago.  also would need to see how blacken-docs formats python console examples.

doc/build/spelling_wordlist.txt
```diff
@@ -0,0 +1,220 @@
...
+unhashable
+unicode
+uniquing
+unmarshalling
+userland
+validator
+visitable
+walkthrough
+zxJDBC
+zxJDBC
```
CaselIT, line 220: this seems duplicate

doc/build/conf.py
```diff
@@ -467,3 +469,20 @@
...
+    "enchant.tokenize.URLFilter",
+]
+spelling_show_suggestions = True
+spelling_suggestion_limit = 2
+# All of the below are true by default
+# spelling_ignore_wiki_words = True
+# spelling_ignore_acronyms = True
+# spelling_ignore_python_builtins = True
+# spelling_ignore_importable_modules = True
+# spelling_ignore_contributor_names = True
```
CaselIT, line 488: is there a configuration that indicates the location of the wordlist file?

tgross35: Yes there's `spelling_word_list_filename`, I just included the word filter related options. Do you want it to be explicit?

Full options here https://sphinxcontrib-spelling.readthedocs.io/en/latest/customize.html

doc/build/spelling_wordlist.txt
tgross35: Good catch - I'll dedupe before marking the PR as non-draft

doc/build/conf.py
CaselIT: I think it would make sense to make it explicit yes, so searching the file name would make it obvious where it's used

tgross35:
@zzzeek tested it out, it does about what you'd expect for pycon, just looks like standard black formatting.

One note is it does remove blank lines in pycon, which are sometimes nice to have. I guess the correct solutions here would be to add a # on these blank lines, or use the plain python format, not sure what you'd find acceptable. I mean _technically_ having the blank lines there means the character for character code isn't quite right (since no trailing whitespace for a tab in console), but that's not the biggest issue since sphinx-copybutton seems to remove them.

Did a quick test on the orm quickstart page to show how it formats. _test is the original file, _formatted is post blacken-docs. Had to zip the files because gh doesn't like rst, for no good reason

[quickstart.zip](https://github.com/sqlalchemy/sqlalchemy/files/8911084/quickstart.zip)

doc/build/conf.py
CaselIT: also a comment with the link to the option list is always handy :)

CaselIT:
> Side note: a `python` and `sql` role might be useful. Would give syntax highlighting to things like
> 
> ```
>  :python:`session.execute(delete(obj))` will issue :sql:`DELETE FROM user WHERE...`
> ```
> 
> ```
> .. role:: python(code)
>    :language: python
> 
> .. role:: sql(code)
>    :language: sql
> ```

these would be nice to have

tgross35:
Heh, having fun testing the inline roles

```rst
If you want to you could :mysql:`set @@system_versioning_alter_history = true`,
or if you feel like it :tsql:`USE sqlalchemy; GO`, or if you don't feel like it
:postgres:`SELECT 'i know my a b c's'::tsvector @@ 'a & b'::tsquery;`. Maybe
even :ccode:`++i; i++; printf("don't be fooled, they're different\n");`, but at
the end of the day you just want to :python:`print([relax for relax in paradise if
relax.temp_c < 28])`

Compared to unformatted ``SELECT 'i know my a b c'::tsvector @@ 'a &
b'::tsquery;``
```

image

I didn't even know they had sql flavor-specific highlighting, sqlalchemy is probably the best project possible to make use of that

tgross35:
I'm hitting a bug running spelling on the changelogs, ([report](https://github.com/sphinx-contrib/spelling/issues/180)) so I just excluded that directory for now.

There's also something zzzeeksphinx doesn't like, it does the spellcheck fine but throws an error at the end looking for some css/scss files in the `output/spelling` directory. I'll look into that a bit more, but it doesn't need to stop anything.

That being said, it runs successfully for the most part and I have a semifinal word list done. A lot of the remaining errors are just unquoted code or variable names that use "" rather than ````. I notice that a ton of docs even backtick quote simple things like `jsonb` (postgres), I wonder if that's for this reason.

Excluding code quoting, looks like the docs are overall very good for spelling, probably only 100 or so errors.

I think it would be good to merge this PR (when ready) without any fixes, then do a separate formatting PR that can be marked as ignore in [.git-blame-ignore-revs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view). I don't think the flake8 checks will work now - I actually took over the flake8-rst repo from the old maintainer (think he just didn't have time) but it will take a bit of time before that is ready to go.

So the main thing here is probably just deciding how to handle code blocks. Could also put the SQL in a collapsable section like this I guess https://stackoverflow.com/a/25543713/5380651, or do the docstring thing, or just leave those blocks unformatted.

Additionally, I think I'd like to write some docs about writing docs, to help anyone new out (I struggled a bit getting started when I wrote the docs for the temporal thingy). Where's a good place for that?

zzzeek:
there's a lot happening here....spellcheck is nice but this seems like a lot of new tooling to deal with.

any way we can have separate PRs for individual new checkers and such so we can just do small things one at a time?   

if there's no longer interest in working on this I'll close it for now.

zzzeek:
hi -

looking through old PRs, and for this one I think we have most of this:

1. we have flake8-docstrings and flake8-rst-docstrings now
2. errors in pyproject.toml we would know via the build being broken
3. trailing whitespace/EOF fixer might be useful though we have never seen this happen in rsts in practice, it would show up in review, we could add these perhaps
4. the spelling thing seems to be the biggest thing here and it looks like a very big deal (needs `enchant` library etc) and we really dont need this these days, LLMs can proofread our rsts more effectively than a simple spellchecker and I also dont want to have to deal with that wordlist


so if we want separately to add a few small checkers to pyproject for rst files that's fine but it's not critical.  the main thing here the spelling contrib IMO not worth it.

PR #8161, opened 2022-06-23
feat: add `drop constraint if exists` to compiler

### LINKED ISSUE #8141: Support `DROP CONSTRAINT IF EXISTS` for Postgres Dialect
### Describe the use case

Recently was using `alembic.op.drop_constraint()` and found that there was no way to implement "IF EXISTS..." behavior, leading to a ProgrammingError as a result if the constraint wasn't there.

`IF EXISTS` support was added to sqlalchemy in https://github.com/sqlalchemy/sqlalchemy/issues/2843 and the notes target Constraints, but that change doesn't appear to have entered the code.

It appears that it was entertained, but not implemented, so there's an opportunity for sqlalchemy to implement the behavior, and then alembic can use it.

Postgres, psycopg2

### Example Use

Use case:

```python
alembic.op.drop_constraint("release_files_name_fkey", "release_files", type_="foreignkey")
```

```sql
ALTER TABLE release_files DROP CONSTRAINT IF EXISTS release_files_name_fkey;
```

### Additional context

Depending on whether this is a good idea, the scope could expand to "add constraint if not exists" semantics, or even potentially cover other types like sequences.

https://github.com/sqlalchemy/sqlalchemy/issues/7354
https://github.com/sqlalchemy/alembic/issues/524

I'm happy to take a stab at this, but considering there's been more than a few efforts on this over the years, I wanted to make sure this was inline with the project's scope and direction.

miketheman:
### Description

Add `DROP CONSTRAINT ... IF EXISTS` behavior to the compiler.

Fixes https://github.com/sqlalchemy/sqlalchemy/issues/8141

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/sql/compiler.py
+++ b/lib/sqlalchemy/sql/compiler.py
@@ -5210,8 +5210,9 @@ class DDLCompiler(Compiled):
                 "Can't emit DROP CONSTRAINT for constraint %r; "
                 "it has no name" % drop.element
             )
-        return "ALTER TABLE %s DROP CONSTRAINT %s%s" % (
+        return "ALTER TABLE %s DROP CONSTRAINT %s%s%s" % (
             self.preparer.format_table(drop.element.table),
+            drop.if_exists and "IF EXISTS " or "",
             formatted_name,
             drop.cascade and " CASCADE" or "",
         )
--- a/lib/sqlalchemy/sql/compiler.py
+++ b/lib/sqlalchemy/sql/compiler.py
@@ -5212,9 +5212,9 @@ class DDLCompiler(Compiled):
             )
         return "ALTER TABLE %s DROP CONSTRAINT %s%s%s" % (
             self.preparer.format_table(drop.element.table),
-            drop.if_exists and "IF EXISTS " or "",
+            "IF EXISTS " if drop.if_exists else "",
             formatted_name,
-            drop.cascade and " CASCADE" or "",
+            " CASCADE" if drop.cascade else "",
         )
 
     def get_column_specification(self, column, **kwargs):
--- a/doc/build/changelog/unreleased_20/8141.rst
+++ b/doc/build/changelog/unreleased_20/8141.rst
@@ -1,0 +1,11 @@
+.. change::
+    :tags: usecase, schema
+    :tickets: 8141
+
+    Added parameter :paramref:`_ddl.DropConstraint.if_exists` to the
+    :class:`_ddl.DropConstraint` construct which result in "IF EXISTS" DDL
+    being added to the DROP statement.
+    This phrase is not accepted by all databases and the operation will fail
+    on a database that does not support it as there is no similarly compatible
+    fallback within the scope of a single DDL statement.
+    Pull request courtesy Mike Fiedler.
--- a/test/sql/test_constraints.py
+++ b/test/sql/test_constraints.py
@@ -1261,6 +1261,16 @@ class ConstraintCompilationTest(fixtures.TestBase, AssertsCompiledSQL):
             "ALTER TABLE tbl DROP CONSTRAINT my_test_constraint CASCADE",
         )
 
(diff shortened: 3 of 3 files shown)


miketheman:
Thanks! I thought about that first, but then backed out, since it seemed dialect-specific and didn't want to break any compiler rules. Will re-work the change.

miketheman:
Thanks for the pointers - I've moved the code around to the base compiler, and rewrote the test to be similar to other tests in the module.

I did some more reading of the Postgres docs, and it seems that the only constrain operation that supports the EXISTS behavior is drop constraint, so no need to implement the ADD side.

Let me know what you think!

lib/sqlalchemy/sql/compiler.py
```diff
@@ -5210,8 +5210,9 @@ def visit_drop_constraint(self, drop, **kw):
                 "Can't emit DROP CONSTRAINT for constraint %r; "
                 "it has no name" % drop.element
             )
-        return "ALTER TABLE %s DROP CONSTRAINT %s%s" % (
+        return "ALTER TABLE %s DROP CONSTRAINT %s%s%s" % (
             self.preparer.format_table(drop.element.table),
+            drop.if_exists and "IF EXISTS " or "",
             formatted_name,
             drop.cascade and " CASCADE" or "",
```
CaselIT, line 5217: these could be replaced with inline if-else (these probably date back to a python version that did not yet support it )

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 120171a7902c80dd7f6dc93f1abadb1bf23ee194 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 120171a7902c80dd7f6dc93f1abadb1bf23ee194: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3970

lib/sqlalchemy/sql/compiler.py
miketheman: Thanks! I tried to remain consistent with the existing patterns rather than diverge for my changes.
Origin of this pattern https://github.com/sqlalchemy/sqlalchemy/commit/cfa1a42328b406d0f494672f0b268890471d6d2e
If it’s desirable to convert to newer patterns, happy to do so!

CaselIT: not really required, but I think it's safe to convert to the new syntax since we are modifying this part anyway.

If you make the change, could you also add a changelog for this pr?

miketheman: Changes updated.
Changelog added as well - wasn't really certain if the location (2.0) and tags we correct - I was basing it off the opened ticket labels.

CaselIT: look great, thanks!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 43276e29fa864fc66900c5a3fa0bf84df5f14271 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 43276e29fa864fc66900c5a3fa0bf84df5f14271 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3970

miketheman:
@zzzeek I'm not super familiar with the code review process for sqlalchemy, and [having read the docs](https://www.sqlalchemy.org/develop.html), I'm not sure if there's another action I'm meant to take?

CaselIT:
Currently just wait for a review :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3970 has been **merged**. Congratulations! :)

PR #8180, opened 2022-06-27
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,4 +3,4 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "monthly"


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
CaselIT, line 6: I think monthly is more than enough. is there an option to make it less frequent?

CaselIT (commented):
Hi, thanks for the pr.

it seems some action that we are using as outdated. I guess there is always the option "if it ain't broke, don't fix it" :)

since not all action run on a pr, every pr of dependabot would require extra effort. I'm not opposed to it, but weekly is too frequent for this, it would be great if something like quarterly were possible, if not, monthly I guess it should be fine

.github/dependabot.yml
naveensrinivasan: I have changed it to monthly.

CaselIT:
what do you think about this @zzzeek ?

CaselIT (approved):
to me it looks ok, let's wait mike opinion

naveensrinivasan:
> to me it looks ok, let's wait mike opinion

A friendly ping.

zzzeek:
my impression is that I have no idea what this does or what dependabot does so if someone can educate me on that.

zzzeek:
i guess there are links, moment

zzzeek:
OK confirm i get this:

1. our github workflows rely upon other workflows, which scanning through the files seem to be referenced by the "uses:" parameter, such as `uses: RalfG/python-wheels-manylinux-build@v0.3.4-manylinux2014_x86_64`
2. dependabot is somehow linked into the system such that if `RalfG/python-wheels-manylinux-build@v0.3.4-manylinux2014_x86_64` has a security issue that's fixed, we will get a PR sent to us automatically that provides an update for that dep.
3. dependabot has nothing to do with (or does?  im not sure) the Python deps we have, including lines in workflow files like `pip install greenlet` as well as in the actual Python dependencies we have in our setup.cfg.  That is, dependabot is specific to **github action workflow dependencies**

do i have that all correct?

seems like this should certainly be enabled and I would think github would have like a checkbox to enable this kind of thing (or just be automatic), is there a reason we have to have a file and all that?

also, "monthly" means, every month it checks versions and we get a PR ?  why not daily / hourly ?

naveensrinivasan:
> OK confirm i get this:
> 
> 1. our github workflows rely upon other workflows, which scanning through the files seem to be referenced by the "uses:" parameter, such as `uses: RalfG/python-wheels-manylinux-build@v0.3.4-manylinux2014_x86_64`
> 2. dependabot is somehow linked into the system such that if `RalfG/python-wheels-manylinux-build@v0.3.4-manylinux2014_x86_64` has a security issue that's fixed, we will get a PR sent to us automatically that provides an update for that dep.
> 3. dependabot has nothing to do with (or does?  im not sure) the Python deps we have, including lines in workflow files like `pip install greenlet` as well as in the actual Python dependencies we have in our setup.cfg.  That is, dependabot is specific to **github action workflow dependencies**
> 
> do i have that all correct?
> 
Yes 
> seems like this should certainly be enabled and I would think github would have like a checkbox to enable this kind of thing (or just be automatic), is there a reason we have to have a file and all that?

AFAIK this is how you let Dependabot know which project is interested via file. 

> also, "monthly" means, every month it checks versions and we get a PR ? why not daily / hourly ?

I changed it to Monthly based on the request :)

zzzeek:
oh.  is the concern that we'd be flooded with dependabot PRs and that by making it monthly, they'd be aggregated into one big PR ?   I bet they aren't.

naveensrinivasan:
> oh. is the concern that we'd be flooded with dependabot PRs and that by making it monthly, they'd be aggregated into one big PR ? I bet they aren't.

Yes, they aren't.

CaselIT:
> 3\. dependabot has nothing to do with (or does?  im not sure) the Python deps we have, including lines in workflow files like `pip install greenlet` as well as in the actual Python dependencies we have in our setup.cfg.  That is, dependabot is specific to **github action workflow dependencies**

I believe it supports also python dependency by configuring the dependabot file appropriately 

> also, "monthly" means, every month it checks versions and we get a PR ? why not daily / hourly ?

Because to merge a pr to a release action we need to create a test release on the test pypi, so I think it's annoying having to do it more often, given that the threat level should be quite low

CaselIT:
> > oh. is the concern that we'd be flooded with dependabot PRs and that by making it monthly, they'd be aggregated into one big PR ? I bet they aren't.
> 
> Yes, they aren't.

but you get say 3 PR that you can merge togher, test with a test release and then merge

CaselIT:
Well practically the only pipeline that has the actuon versions specified is the release one, so I think we should create a test release to verify everything is working when an update happens to it.

But as you prefer, I just don't see the need to run it more frequently, that's all

zzzeek:
we can always let the PRs pile up and attend to them monthly, just if a PR comes in for MAJOR ATTACK we can see it.

this..change does not apply to our PR thing ?   seems like we use "actions/checkout" and "actions/setup-python", those could have security issues?

CaselIT:
> this..change does not apply to our PR thing ? seems like we use "actions/checkout" and "actions/setup-python", those could have security issues?

I guess so, but again the threat vector seems very low

zzzeek:
> > this..change does not apply to our PR thing ? seems like we use "actions/checkout" and "actions/setup-python", those could have security issues?
> 
> I guess so, but again the threat vector seems very low

nevertheless, if there *was* a problem, I'd not want to wait a month.  I'd sort of want to know that day.   

if there are no threats, then it doesnt matter if the thing runs every day, we wont ever hear about it.

so i dont see any reason this shouldnt be daily.

zzzeek:
UNLESS, dependabot is sending us a PR *every time* a dependency has any update?     is that the case?  and if so can that be tuned to only receive security alerts?

CaselIT:
> UNLESS, dependabot is sending us a PR _every time_ a dependency has _any_ update

that's how it works. but it should not be much of a problem since we don't use many actions.

Again as you prefer mike regarding the frequency

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision ef796e21f30da98c96e804bcdc056a17465525ca of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change ef796e21f30da98c96e804bcdc056a17465525ca: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4013

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4013 has been **merged**. Congratulations! :)

PR #8237, opened 2022-07-10
Update async_scoped_session example

hyzyla:
Added example adding new object using `async_scoped_session` proxy object

### Description

When exploring docs about `async_scoped_session` with my colleague, I found the example with `async_scoped_session` to be a bit confusing. I thought we should create `AsyncSession` to add a new object, but after trying, I found that we shouldn't — that object can be added through `async_scoped_session` proxy. For that reason, I've added an example where an object could be added via proxy object.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/doc/build/orm/extensions/asyncio.rst
+++ b/doc/build/orm/extensions/asyncio.rst
@@ -756,6 +756,8 @@ the usual ``await`` keywords are necessary, including for the
     async def some_function(some_async_session, some_object):
        # use the AsyncSession directly
        some_async_session.add(some_object)
+       # or as alternative use the AsyncSession via the context-local proxy
+       AsyncScopedSession.add(some_object)
 
        # use the AsyncSession via the context-local proxy
        await AsyncScopedSession.commit()


doc/build/orm/extensions/asyncio.rst
```diff
@@ -756,6 +756,8 @@ the usual ``await`` keywords are necessary, including for the
     async def some_function(some_async_session, some_object):
        # use the AsyncSession directly
        some_async_session.add(some_object)
+       # or as alternative use the AsyncSession via the context-local proxy
+       AsyncScopedSession.add(some_object)
```
zzzeek, line 760: im +0 on this.  I dislike the class-level proxy thing that scoped session has and I think it confuses people.

zzzeek:
>I thought we should create `AsyncSession` to add a new object, but after trying, I found that we shouldn't

so when you call:

```
some_async_session = AsyncScopedSession()
```

that only conditionally creates a new session.  it is giving you the current scope-local session object, if already present.

Then once you have that object, that's what you work with. 

I prefer this pattern to the "proxy" thing which is a really old thing I came up with in 2007 to ease people migrating from sqlalchemy 0.1 which had an "everything is global thread local always" design that was awful.

personally I'd never use scopedsession either sync or not because I think code is easier to reason about if it includes real plumbing for explicit contexts to be passed around.

hyzyla:
I have a few use cases in web development where I find a useful local proxy

The first one is to have a cheap dependency injection for a web application where you can instantiate a repository only once at thethe  start of application or as a global dependency

rough example
```python
_session = ContextVar('_session')

def get_session_context():
    return _session.get()

@contextmanager
def new_session_context():
    token = _session.set(uuid.uuid4())
    try:
        yield
    finally:
        _session.reset(token)


session = async_scoped_session(..., scope_func=get_session_context)

class Repository:
    def __init__(session: async_scoped_session):
        self.session = session
    async def create_something(self):
        self.session.add(...)

repository = Repository(session=session)

@app.post('/some-route')
def create_something():
    with new_session_context():
           repository.create_something()
           session.commit()
```

Also pattern with globally scoped session is widely used in Flask web apps and `async_scoped_session` can be useful for projects that are migrating to FastAPI or aiohttp from Flask. 


Example of integration of FastAPI with async_scoped_session: 
https://github.com/hyzyla/fastapi-async-sqlalchemy-orm/blob/main/app/main.py

hyzyla:
It's a little bit offtopic, when I didn't know about the scoped_session, I made a very simple context session mechanism, that can be used as an alternative to `scoped_session`:


```python
# db.py
from contextlib import contextmanager
from contextvars import ContextVar
from typing import Iterator, TypeVar

from sqlalchemy.engine import Connection
from sqlalchemy.orm import sessionmaker

from app.config import config

T = TypeVar("T")

engine = create_engine(url=config.DATABASE_URL, future=True)
Session = sessionmaker(engine)

_session_ctx: ContextVar[Connection] = ContextVar("_session_ctx")


class NoDBConnection(Exception):
    ...


@contextmanager
def session()  -> Iterator[Session]:
    with Session() as s:
        with set_context_var(var=_session_ctx, value=s):
            yield s


def get_session() -> Session:
    try:
        return _session_ctx.get()
    except LookupError:
        raise NoDBConnection(
            "Database session not started. "
            "Use context manager `db.session()`"
        )
```
and then an example of using:
```python
# example.py
import db

def create_db_object():
    session = db.get_session()
    session.add(...)


def create_object_service():
    create_db_object()

def create_object_handler():
    with db.session():
        create_object_service()
        
 create_object_handler()
```

CaselIT:
@zzzeek should this be merged of you prefer not to? just to avoid leaving this PR in lombo forever

zzzeek:
yeah I'd rather not, sorry

CaselIT:
Ok. Thanks anyway @hyzyla

PR #8277, opened 2022-07-22
Add alembic `extras_require` to setup.cfg

kkirsche:
### Description
This merge request adds a new `extras_require` to setup.cfg to allow users to install Alembic to be installed with SQLAlchemy via extras requires.

While alembic currently requires SQLAlchemy to function, it does not allow the user to include the database driver or mypy at the same time. By providing this option, a user will be able to install SQLAlchemy, a database driver, Mypy, and Alembic via a single command for easy inclusion into projects.

Thank you for your consideration of this change.

Apologies for the newline at the end, that was added by Github (done via web UI) and I'm not sure how to easily remove this.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,6 +40,7 @@ install_requires =
     typing-extensions >= 4.1.0
 
 [options.extras_require]
+alembic = alembic
 asyncio =
     greenlet!=0.4.17
 mypy =


zzzeek:
just a little devils advocate here, "a single command", I can do "pip install sqlalchemy[mysql] alembic" right now and that's also "a single command".    the reason `sqlalchemy[mysql]` is helpful here is that `mysql` refers to the SQLAlchemy default mysql dialect which knows to install "mysqlclient", that is, "mysqlclient" is a dependency of the "mysql" sub-component of SQLAlchemy, "greenlet" is a dependency of the "asyncio" sub-component of SQLAlchemy.   OTOH "alembic" is not a dependency of any part of SQLAlchemy and isn't part of some group of dependencies, like "greenlet / asyncpg" in order to install sqlalchemy[asyncpg] implies.  it's just the name itself.    seems like "more than one way to do it" here as if someone wants to install alembic they'd just add it to their pip command line.

kkirsche:
That's fair. For me it's more about the relationship ship the two tools have than truly being one command. Alembic is directly related to SQLAlchemy and by including it as an extra the relationship is explicit in the user's pyproject.toml (if in use) where when using separate additions it's an implicit relationship the developer needs to be aware of. That usually primarily effects newer developers coming from other languages, which is more common in an enterprise environment than OSS community at least in my experience. 

I fully understand your point of view though, and respect whichever decision you make.

CaselIT:
For me is fine either way. As you prefer mike

CaselIT:
Hi, after talking abut it at a dev meeting we think it's kind of strange to have alembic an optional dependency of sqlalchemy since alembic depends on it already.

Thanks in any case!

kkirsche:
🤷‍♂️ no worries, worst thing that can be decided is no. Thanks for considering it, regardless

PR #8306, opened 2022-07-30
glossary: update the acronym definition

CalK16:
### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/doc/build/glossary.rst
+++ b/doc/build/glossary.rst
@@ -320,7 +320,7 @@ Glossary
        An acronym for **Data Manipulation Language**.  DML is the subset of
        SQL that relational databases use to *modify* the data in tables. DML
        typically refers to the three widely familiar statements of INSERT,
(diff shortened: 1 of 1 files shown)


CaselIT:
Hi,

It seems thaf the acronym is wrong, the R should stay for read. https://en.m.wikipedia.org/wiki/Create,_read,_update_and_delete
@zzzeek do you think we should update it?

bb-ben:
> Hi,
> 
> It seems thaf the acronym is wrong, the R should stay for read. https://en.m.wikipedia.org/wiki/Create,_read,_update_and_delete @zzzeek do you think we should update it?

Thanks for pointing that out!

zzzeek:
absolutely, if I got it wrong, it's wrong.   I never knew the "R" was for "read"

PR #8438, opened 2022-08-26
Fix doc constraints snippet: List -> Tuple

lmeilibr:
When using a composite ForeignKeyConstraint, the doc example uses a List for the columns attribute, while my IDE complains that instead of a list, it should be a tuple

### Description
Old: ['invoice_id', 'ref_num']
New: ('invoice_id', 'ref_num')

Screen Shot 2022-08-25 at 22 49 26

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

### DIFF
--- a/doc/build/core/constraints.rst
+++ b/doc/build/core/constraints.rst
@@ -78,7 +78,7 @@ And then a table ``invoice_item`` with a composite foreign key referencing
         Column('item_name', String(60), nullable=False),
         Column('invoice_id', Integer, nullable=False),
         Column('ref_num', Integer, nullable=False),
-        ForeignKeyConstraint(['invoice_id', 'ref_num'], ['invoice.invoice_id', 'invoice.ref_num'])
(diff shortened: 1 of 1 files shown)


zzzeek:
hi -

this is a problem in your IDE.    sqlalchemy2-stubs has this as an [iterable](https://github.com/sqlalchemy/sqlalchemy2-stubs/blob/7f6a489d823a54b226e4c09b00f471d8f206a237/sqlalchemy-stubs/sql/schema.pyi#L512), in SQLAlchemy 2.0 it will be a [sequence](https://github.com/sqlalchemy/sqlalchemy/blob/5c7ab1f808b20fe5e58d83a738b201a7eaa8973f/lib/sqlalchemy/sql/schema.py#L4210).  I've always written and used it as a list since 2006 on this end, so not sure why your IDE thinks it's a tuple.  We don't support other stub packages here if that's what's involved.

PR #8646, opened 2022-10-16
Add handle `nullable=Flase` case to mypy extension

heckad:
### Description
If column has nullable with the meaning of false then type should be without None variant.

### DIFF
--- a/lib/sqlalchemy/ext/mypy/infer.py
+++ b/lib/sqlalchemy/ext/mypy/infer.py
@@ -7,6 +7,8 @@
 
 from __future__ import annotations
 
+from contextlib import suppress
+
 from typing import Optional
 from typing import Sequence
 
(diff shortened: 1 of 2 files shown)


heckad:
Please tell me where to add the test, otherwise in `sqlalchemy/test/ext/mypy/plain_files/mapped_column.py` something strange happens for version 1.4.41

CaselIT:
Hi,

The mypy extension is currently legacy/deprecated, so I don't know if this will be merged.

Thanks for the effort instead

zzzeek:
we would need a test case inside of test/ext/mypy/plugin_files/ to even see what this does, and you also should [create a new bug report](https://www.sqlalchemy.org/develop.html#pullrequests) showing the issue, I don't know what this does right now.

CaselIT:
@heckad please take a moment to explain what this PR does like asked by mike above. Thanks

heckad:
Yes, of course, I'm sorry, work, so I finished it when I had a minute. We often use `nullable=False` in our project and it is very tedious to explicitly set the type that it is not None. Therefore, I made a small addition that checks that if the arguments contain exactly `nullable=Ture`, then the type is definitely not None.

Issue #8706, opened 2022-10-24
Improve new SA 2.0 PG Range


lelit:
### Describe the use case

In SA 2.0 there is a new _native_ `Range` class that exposes the underlying PostgreSQL `range` types.

The new class is kept intentionally minimal, and indeed it has been implemented using a `dataclass` carrying just the values that make up a range, the `lower` and the `upper` ends, their `bounds` and whether the range is `empty`.

However, some more methods would make the life easier, and an [exchange] identified three of them: `contains_value()`, `issubset()` and `issuperset()`.

### Databases / Backends / Drivers targeted

The new class is PostgreSQL specific, and drivers agnostic.

### Example Use

```python

from datetime import date

from sqlalchemy import create_engine, select
from sqlalchemy.dialects.postgresql import DATERANGE, Range
from sqlalchemy.orm import DeclarativeBase, Mapped, mapped_column, Session


class Base(DeclarativeBase):
    pass


class Contract(Base):
    __tablename__ = "contracts"

    id: Mapped[int] = mapped_column(primary_key=True)
    validity: Mapped[Range[date]] = mapped_column(DATERANGE)


engine = create_engine("postgresql+psycopg://scott:tiger@pg14/dbname")

Base.metadata.create_all(engine)

with Session(engine) as session:
    old_contract = Contract(validity=Range(date(2021, 1, 1), date(2021, 12, 31))
    session.add(old_contract)
    cur_contract = Contract(validity=Range(date(2022, 1, 1), date(2022, 12, 31))
    session.add(cur_contract)
    session.commit()

    for contract in session.execute(select(Contract)):
        print(contract.validity.contains_value(date.today())
```

sqla-tester:
**Lele Gaifax** has proposed a fix for this issue in the **main** branch:

**Implement contains_value(), issubset() and issuperset() on PG Range** https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4151

PR #8707, opened 2022-10-24
Implement contains_value(), issubset() and issuperset() on PG Range

lelit:
Fixes issue #8706.

This pull request is:

- [ ] A short code fix

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/ranges.py
+++ b/lib/sqlalchemy/dialects/postgresql/ranges.py
@@ -80,7 +80,7 @@ class Range(Generic[_T]):
     def __bool__(self) -> bool:
         return self.empty
 
-    def contains_value(self, value: T) -> bool:
+    def contains_value(self, value: _T) -> bool:
         "Check whether this range contains the given `value`."
 
         if self.empty:
@@ -135,7 +135,8 @@ class Range(Generic[_T]):
             if not lower_side:
                 return False
 
-        # Lower end already considered, an upper-unbound range surely contains this
+        # Lower end already considered, an upper-unbound range surely contains
+        # this
         if other.upper is None:
             return True
 
--- a/lib/sqlalchemy/dialects/postgresql/ranges.py
+++ b/lib/sqlalchemy/dialects/postgresql/ranges.py
@@ -89,19 +89,28 @@ class Range(Generic[_T]):
         if self.lower is None:
             if self.upper is None:
                 return True
-            return (self.upper is None or
-                    (value < self.upper if self.bounds[1] == ")"
-                     else value <= self.upper))
+            return self.upper is None or (
+                value < self.upper
+                if self.bounds[1] == ")"
+                else value <= self.upper
+            )
 
         if self.upper is None:
-            return (value > self.lower if self.bounds[0] == "("
-                    else value >= self.lower)
+            return (
+                value > self.lower
+                if self.bounds[0] == "("
+                else value >= self.lower
+            )
 
-        return ((value > self.lower if self.bounds[0] == "("
-                 else value >= self.lower)
-                and
-                (value < self.upper if self.bounds[1] == ")"
-                 else value <= self.upper))
+        return (
+            value > self.lower
+            if self.bounds[0] == "("
+            else value >= self.lower
+        ) and (
+            value < self.upper
+            if self.bounds[1] == ")"
+            else value <= self.upper
+        )
 
     def issubset(self, other) -> bool:
         "Determine whether this range is a contained by `other`."
@@ -130,7 +139,7 @@ class Range(Generic[_T]):
         if self.lower is not None and other.lower is not None:
             lower_side = other.lower < self.lower
             if not lower_side:
-                if self.bounds[0] == '(' or other.bounds[0] == '[':
+                if self.bounds[0] == "(" or other.bounds[0] == "[":
                     lower_side = other.lower == self.lower
             if not lower_side:
                 return False
@@ -143,7 +152,7 @@ class Range(Generic[_T]):
         # Check the upper end
         upper_side = other.upper > self.upper
         if not upper_side:
-            if self.bounds[1] == ')' or other.bounds[1] == ']':
+            if self.bounds[1] == ")" or other.bounds[1] == "]":
                 upper_side = other.upper == self.upper
         return upper_side
 
--- a/lib/sqlalchemy/dialects/postgresql/ranges.py
+++ b/lib/sqlalchemy/dialects/postgresql/ranges.py
@@ -137,13 +137,13 @@ class Range(Generic[_T]):
             return False
 
         slower = self.lower
-        slower_inc = self.bounds[0] == '['
+        slower_inc = self.bounds[0] == "["
         supper = self.upper
-        supper_inc = self.bounds[1] == ']'
+        supper_inc = self.bounds[1] == "]"
         olower = other.lower
-        olower_inc = other.bounds[0] == '['
+        olower_inc = other.bounds[0] == "["
         oupper = other.upper
-        oupper_inc = other.bounds[1] == ']'
+        oupper_inc = other.bounds[1] == "]"
 
         # A bilateral unbound range contains any other range
         if olower is oupper is None:
--- a/lib/sqlalchemy/dialects/postgresql/ranges.py
+++ b/lib/sqlalchemy/dialects/postgresql/ranges.py
@@ -136,18 +136,15 @@ class Range(Generic[_T]):
         if other.empty:
             return False
 
-        slower = self.lower
-        slower_inc = self.bounds[0] == "["
-        supper = self.upper
-        supper_inc = self.bounds[1] == "]"
         olower = other.lower
-        olower_inc = other.bounds[0] == "["
         oupper = other.upper
-        oupper_inc = other.bounds[1] == "]"
 
         # A bilateral unbound range contains any other range
         if olower is oupper is None:
             return True
+
+        slower = self.lower
+        supper = self.upper
 
         # A lower-bound range cannot contain a lower-unbound range
         if slower is None and olower is not None:
@@ -157,19 +154,27 @@ class Range(Generic[_T]):
         if supper is None and oupper is not None:
             return False
 
+        slower_inc = self.bounds[0] == "["
+        supper_inc = self.bounds[1] == "]"
+        olower_inc = other.bounds[0] == "["
+        oupper_inc = other.bounds[1] == "]"
+
         # Check the lower end
+        step = -1
         if slower is not None and olower is not None:
             lside = olower < slower
             if not lside:
                 if not slower_inc or olower_inc:
                     lside = olower == slower
             if not lside:
-                step = self._get_discrete_step()
-                if step is not None:
-                    # Cover (1,x] vs [2,x) and (0,x] vs [1,x)
-                    if not slower_inc and olower_inc and slower < olower:
+                # Cover (1,x] vs [2,x) and (0,x] vs [1,x)
+                if not slower_inc and olower_inc and slower < olower:
+                    step = self._get_discrete_step()
+                    if step is not None:
                         lside = olower == (slower + step)
-                    elif slower_inc and not olower_inc and slower > olower:
+                elif slower_inc and not olower_inc and slower > olower:
+                    step = self._get_discrete_step()
+                    if step is not None:
                         lside = (olower + step) == slower
             if not lside:
                 return False
@@ -185,12 +190,16 @@ class Range(Generic[_T]):
             if not supper_inc or oupper_inc:
                 uside = oupper == supper
             if not uside:
-                step = self._get_discrete_step()
-                if step is not None:
-                    # Cover (x,2] vs [x,3) and (x,1] vs [x,2)
-                    if supper_inc and not oupper_inc and supper < oupper:
+                # Cover (x,2] vs [x,3) and (x,1] vs [x,2)
+                if supper_inc and not oupper_inc and supper < oupper:
+                    if step == -1:
+                        step = self._get_discrete_step()
+                    if step is not None:
                         uside = oupper == (supper + step)
-                    elif not supper_inc and oupper_inc and supper > oupper:
+                elif not supper_inc and oupper_inc and supper > oupper:
+                    if step == -1:
+                        step = self._get_discrete_step()
+                    if step is not None:
                         uside = (oupper + step) == supper
         return uside
 
--- a/lib/sqlalchemy/dialects/postgresql/ranges.py
+++ b/lib/sqlalchemy/dialects/postgresql/ranges.py
@@ -8,6 +8,8 @@
 from __future__ import annotations
 
 import dataclasses
+from datetime import date
+from datetime import timedelta
 from typing import Any
 from typing import Generic
 from typing import Optional
@@ -111,6 +113,18 @@ class Range(Generic[_T]):
             else value <= self.upper
         )
 
+    def _get_discrete_step(self):
+        "Determine the “step” for this range, if it is a discrete one."
+
+        # See
+        # https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-DISCRETE
+        # for the rationale
+
+        if isinstance(self.lower, int) or isinstance(self.upper, int):
+            return 1
+        if isinstance(self.lower, date) or isinstance(self.upper, date):
+            return timedelta(days=1)
+
     def _contained_by(self, other: Range) -> bool:
         "Determine whether this range is a contained by `other`."
 
@@ -122,38 +136,63 @@ class Range(Generic[_T]):
         if other.empty:
             return False
 
+        slower = self.lower
+        slower_inc = self.bounds[0] == '['
+        supper = self.upper
+        supper_inc = self.bounds[1] == ']'
+        olower = other.lower
+        olower_inc = other.bounds[0] == '['
+        oupper = other.upper
+        oupper_inc = other.bounds[1] == ']'
+
         # A bilateral unbound range contains any other range
-        if other.lower is other.upper is None:
+        if olower is oupper is None:
             return True
 
         # A lower-bound range cannot contain a lower-unbound range
-        if self.lower is None and other.lower is not None:
+        if slower is None and olower is not None:
             return False
 
         # Likewise on the right side
-        if self.upper is None and other.upper is not None:
+        if supper is None and oupper is not None:
             return False
 
         # Check the lower end
-        if self.lower is not None and other.lower is not None:
-            lower_side = other.lower < self.lower
-            if not lower_side:
-                if self.bounds[0] == "(" or other.bounds[0] == "[":
-                    lower_side = other.lower == self.lower
-            if not lower_side:
+        if slower is not None and olower is not None:
+            lside = olower < slower
+            if not lside:
+                if not slower_inc or olower_inc:
+                    lside = olower == slower
+            if not lside:
+                step = self._get_discrete_step()
+                if step is not None:
+                    # Cover (1,x] vs [2,x) and (0,x] vs [1,x)
+                    if not slower_inc and olower_inc and slower < olower:
+                        lside = olower == (slower + step)
+                    elif slower_inc and not olower_inc and slower > olower:
+                        lside = (olower + step) == slower
+            if not lside:
                 return False
 
         # Lower end already considered, an upper-unbound range surely contains
         # this
-        if other.upper is None:
+        if oupper is None:
             return True
 
         # Check the upper end
-        upper_side = other.upper > self.upper
-        if not upper_side:
-            if self.bounds[1] == ")" or other.bounds[1] == "]":
-                upper_side = other.upper == self.upper
-        return upper_side
+        uside = oupper > supper
+        if not uside:
+            if not supper_inc or oupper_inc:
+                uside = oupper == supper
+            if not uside:
+                step = self._get_discrete_step()
+                if step is not None:
+                    # Cover (x,2] vs [x,3) and (x,1] vs [x,2)
+                    if supper_inc and not oupper_inc and supper < oupper:
+                        uside = oupper == (supper + step)
+                    elif not supper_inc and oupper_inc and supper > oupper:
+                        uside = (oupper + step) == supper
+        return uside
 
     def contains(self, value: Union[_T, Range]) -> bool:
         "Determine whether this range contains `value`."
--- a/lib/sqlalchemy/dialects/postgresql/ranges.py
+++ b/lib/sqlalchemy/dialects/postgresql/ranges.py
@@ -80,6 +80,77 @@ class Range(Generic[_T]):
     def __bool__(self) -> bool:
         return self.empty
 
+    def contains_value(self, value: T) -> bool:
+        "Check whether this range contains the given `value`."
+
+        if self.empty:
+            return False
+
+        if self.lower is None:
+            if self.upper is None:
+                return True
+            return (self.upper is None or
+                    (value < self.upper if self.bounds[1] == ")"
+                     else value <= self.upper))
+
+        if self.upper is None:
+            return (value > self.lower if self.bounds[0] == "("
+                    else value >= self.lower)
+
+        return ((value > self.lower if self.bounds[0] == "("
+                 else value >= self.lower)
+                and
+                (value < self.upper if self.bounds[1] == ")"
+                 else value <= self.upper))
+
+    def issubset(self, other) -> bool:
+        "Determine whether this range is a contained by `other`."
+
+        # Any range contains the empty one
+        if self.empty:
+            return True
+
+        # An empty range does not contain any range except the empty one
+        if other.empty:
+            return False
+
+        # A bilateral unbound range contains any other range
+        if other.lower is other.upper is None:
+            return True
+
+        # A lower-bound range cannot contain a lower-unbound range
+        if self.lower is None and other.lower is not None:
+            return False
+
+        # Likewise on the right side
+        if self.upper is None and other.upper is not None:
+            return False
+
+        # Check the lower end
+        if self.lower is not None and other.lower is not None:
+            lower_side = other.lower < self.lower
+            if not lower_side:
+                if self.bounds[0] == '(' or other.bounds[0] == '[':
+                    lower_side = other.lower == self.lower
+            if not lower_side:
+                return False
+
+        # Lower end already considered, an upper-unbound range surely contains this
+        if other.upper is None:
+            return True
+
+        # Check the upper end
+        upper_side = other.upper > self.upper
+        if not upper_side:
+            if self.bounds[1] == ')' or other.bounds[1] == ']':
+                upper_side = other.upper == self.upper
+        return upper_side
+
+    def issuperset(self, other) -> bool:
+        "Determine whether this range contains `other`."
+
+        return other.issubset(self)
+
 
 class AbstractRange(sqltypes.TypeEngine):
     """
--- a/lib/sqlalchemy/dialects/postgresql/ranges.py
+++ b/lib/sqlalchemy/dialects/postgresql/ranges.py
@@ -122,8 +122,10 @@ class Range(Generic[_T]):
 
         if isinstance(self.lower, int) or isinstance(self.upper, int):
             return 1
-        if isinstance(self.lower, date) or isinstance(self.upper, date):
+        elif isinstance(self.lower, date) or isinstance(self.upper, date):
             return timedelta(days=1)
+        else:
+            return None
 
     def _contained_by(self, other: Range) -> bool:
         "Determine whether this range is a contained by `other`."
--- a/lib/sqlalchemy/dialects/postgresql/ranges.py
+++ b/lib/sqlalchemy/dialects/postgresql/ranges.py
@@ -87,8 +87,6 @@ class Range(Generic[_T]):
             return False
 
         if self.lower is None:
-            if self.upper is None:
-                return True
             return self.upper is None or (
                 value < self.upper
                 if self.bounds[1] == ")"
--- a/lib/sqlalchemy/dialects/postgresql/ranges.py
+++ b/lib/sqlalchemy/dialects/postgresql/ranges.py
@@ -12,6 +12,7 @@ from typing import Any
 from typing import Generic
 from typing import Optional
 from typing import TypeVar
+from typing import Union
 
 from ... import types as sqltypes
 from ...util import py310
@@ -80,7 +81,7 @@ class Range(Generic[_T]):
     def __bool__(self) -> bool:
         return self.empty
 
-    def contains_value(self, value: _T) -> bool:
+    def _contains_value(self, value: _T) -> bool:
         "Check whether this range contains the given `value`."
 
         if self.empty:
@@ -110,7 +111,7 @@ class Range(Generic[_T]):
             else value <= self.upper
         )
 
-    def issubset(self, other) -> bool:
+    def _contained_by(self, other: Range) -> bool:
         "Determine whether this range is a contained by `other`."
 
         # Any range contains the empty one
@@ -154,10 +155,18 @@ class Range(Generic[_T]):
                 upper_side = other.upper == self.upper
         return upper_side
 
-    def issuperset(self, other) -> bool:
-        "Determine whether this range contains `other`."
+    def contains(self, value: Union[_T, Range]) -> bool:
+        "Determine whether this range contains `value`."
 
-        return other.issubset(self)
+        if isinstance(value, Range):
+            return value._contained_by(self)
+        else:
+            return self._contains_value(value)
+
+    def contained_by(self, other: Range) -> bool:
+        "Determine whether this range is a contained by `other`."
+
+        return self._contained_by_range(other)
 
 
 class AbstractRange(sqltypes.TypeEngine):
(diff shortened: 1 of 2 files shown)


lib/sqlalchemy/dialects/postgresql/ranges.py
```diff
@@ -80,6 +80,87 @@ def __init__(
...
+    def contains_value(self, value: _T) -> bool:
+        "Check whether this range contains the given `value`."
+
+        if self.empty:
+            return False
+
+        if self.lower is None:
+            if self.upper is None:
+                return True
+            return self.upper is None or (
```
CaselIT, line 92: `self.upper is None` is always true on this line

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 9ce067fcb446c939d6db7aa2a61cb1c985850831 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 9ce067fcb446c939d6db7aa2a61cb1c985850831: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4151

CaselIT:
Thanks!

lib/sqlalchemy/dialects/postgresql/ranges.py
lelit: Right, nice spot.
Now that this is "carried" on Gerrit, can I keep going pushing new changesets here or what?

CaselIT: Sure, continue here, we can ask the bot to update gerrit from the pr

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision b993bc06fe6536c2d08b344f77b17e5cffa93587 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset b993bc06fe6536c2d08b344f77b17e5cffa93587 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4151

lelit:
Maybe the difference with SA 1.x should be mentioned in "SA 2.0 migration" section?

zzzeek:
> Maybe the difference with SA 1.x should be mentioned in "SA 2.0 migration" section?

it's currently here: https://docs.sqlalchemy.org/en/20/changelog/whatsnew_20.html#new-range-multirange-support-and-changes-for-postgresql-backends  

there really should ideally not be much difference, I will wait and see if other people report issues but I really dont think the psycopg2 methods on Range were very widely used.

lib/sqlalchemy/dialects/postgresql/ranges.py
```diff
@@ -80,6 +80,85 @@ def __init__(
     def __bool__(self) -> bool:
         return self.empty
 
+    def contains_value(self, value: _T) -> bool:
```
zzzeek, line 83: so I forgot that we're SQLAlchemy, and we make Python operators symmetric with the SQL operators.   Assuming this is equivalent to the "@>" opeator, this op should be called [contains](https://github.com/sqlalchemy/sqlalchemy/blob/b993bc06fe6536c2d08b344f77b17e5cffa93587/lib/sqlalchemy/dialects/postgresql/ranges.py#L212).

We could in theory add Python versions of all the operators below:

https://github.com/sqlalchemy/sqlalchemy/blob/b993bc06fe6536c2d08b344f77b17e5cffa93587/lib/sqlalchemy/dialects/postgresql/ranges.py#L205-L273

test/dialect/postgresql/test_dialect.py
```diff
@@ -1210,3 +1210,56 @@ def test_get_dialect(self):
...
+        (Range(1, 4, bounds="[]"), "[1,4]"),
+        (Range(1, 4, bounds="(]"), "(1,4]"),
+        (Range(1, 4, bounds="()"), "(1,4)"),
+        (Range(-4, 1, bounds="[)"), "[-4,1)"),
+        (Range(2, 3, bounds="[)"), "[2,3)"),
+        (Range(0, 6, bounds="[)"), "[0,6)"),
+        argnames="r1,r1repr",
+    )
+    def test_is_sub_super_set(self, connection, r1, r1repr, r2, r2repr):
+        q = text(f"select '{r1repr}'::int4range <@ '{r2repr}'::int4range")
```
zzzeek, line 1262: then the tests here could be making use of those different operators and asserting equivalent behavior SQL vs. python

zzzeek (changes requested):
So I slept on this and remembered that we implement all these comparisons in SQL.  The Python versions should be equivalent.   Can we modify in that way?   then we will have a very solid change with a good foundation.

lib/sqlalchemy/dialects/postgresql/ranges.py
lelit: It seems plausible, but just to be sure I got your suggestion right: do you mean that I should _collapse_ the two new methods `contains_value()` and `issuperset()` into a single `contains()` one, and then discriminating the actual impl on the type of the first argument? And then rename `issubset()` with `contained_by()`?

test/dialect/postgresql/test_dialect.py
lelit: Can you help me understand what you mean here?

lib/sqlalchemy/dialects/postgresql/ranges.py
zzzeek: Hm I see your point, is that what PG's "@>" operator does?  That is, for a daterange, you can hand it an individual date, or another range?   Odd that PG would do that, however, yes it's usually best if Python and SQL interfaces look the same.  it allows one to write code that works in both contexts transparently.    I wrote a blog post about this many years ago at https://techspot.zzzeek.org/2011/10/21/hybrids-and-value-agnostic-types/ - see the "Amount" object, which works in both Python client-side and SQL server-side contexts.

zzzeek: cc @CaselIT if you have any reservations about this.

lelit: > is that what PG's "@>" operator does? That is, for a daterange, you can hand it an individual date, or another range?

Yes, you can see that behaviour in action in the two tests I wrote (although I used the `<@` operator actually), giving a single date for the `contains_value()` and a range for the `issubset()`.

> Odd that PG would do that

Why? Many operators are _overloaded_, like `+` that may sum two integers, or two floats, or a date and an interval...

test/dialect/postgresql/test_dialect.py
zzzeek: well you'd have a callable that says: `lambda obj, other: obj.contains(other)` then we could run it in both ways:

```py

local_data = some_data_object() # like a datetime
python_range_loaded_from_db = connection.scalar(select(table.c.some_range))
eq_(
    connection.scalar(my_callable(table.c.some_range, local_data)),
    my_callable(python_range_loaded_from_db, local_data)
)
```

lib/sqlalchemy/dialects/postgresql/ranges.py
lelit: I will wait for Federico's POV.

I think that if we go that way, we could have, say:
```
  def _contains_value(self, scalar_value):
    # current impl of contains_value()

  def _contains_range(self, range_value):
    # current impl of issuperset()

  def contains(self, value):
    if isinstance(value, Range):
      return self._contains_range(value)
    else
      return self._contains_value(value)
```

How does that sound? ... I would need some help in _typification_ though :-)

zzzeek: yeah it would work that way, with the branching aspect of it.

lelit: I know next to nothing about typing, but wouldn't mypy complain about calling either methods, depending on the type of `value`? Shall I use a `cast()` or something?

zzzeek: for typing, we can help with that part, but short answer is there would be two `@overload` versions for the method.   no need to get into that we can do it if you dont want to.

lelit:
I've implemented the _new_ approach, but with some anguish I'm afraid I hit a showstopper :sob:... I **knew** this is a tricky matter!

While the `contains_value()` is mostly straightforward and generalizable, and it works well as I implemented it, when it comes to "comparing" a range with another range things become _impossible_ to implement with what we have now: to do that properly, a range would need to know whether or not it covers [discrete](https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-DISCRETE) values, and if it does, it should carry also the concept of "minimal step" between the values. In other words, the current `Range` class would need to grow in a family of subclasses in the same way an `AbstractRange` gets specialized into `INT4RANGE`, `NUMRANGE` and so on.

Consider the following two ranges: `(1,2]::int4range` and `[2,3)::int4range`. They represent the very same set of values, and indeed `select '(1,2]'::int4range <@ '[2,3)'::int4range` returns `True`. But to do that, the logic needs to know that we are dealing with discrete integer values and consequently that the "minimal step" is 1. By contrast, `select '(1,2]'::numrange <@ '[2,3)'::numrange` returns `False`, because they are not discrete, and there isn't a valid "minimal step" between the values.

In the light of this, I'm seeking for your opinion on how to proceed.

BTW, even asyncpg fails:
```python
>>> r1 = asyncpg.Range(1, 2, lower_inc=False, upper_inc=True)
>>> r2 = asyncpg.Range(2, 3, lower_inc=True, upper_inc=False)
>>> r1.issubset(r2)
False
```

lelit:
For the history, I committed the status quo.

zzzeek:
So here are the types we have:  INT..RANGE, NUMRANGE, TSRANGE, DATERANGE, TSTZRANGE

and in https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-DISCRETE, DATE and INT are the two discrete ones.

So, why not:

```py
@util.memoized_property
def get_discrete_step(self):
    if isinstance(self.lower, int):
        return 1
    elif isinstance(self.lower, datetime.date):
        return datetime.timedelta(days=1)
    else:
        return None
```

I agree there is no point implementing this stuff if we don't do the discrete logic correctly.   So either we do the above, or we go back to simple "contains_value()" that just gets a single data value and not a range, and that's all we implement.

lelit:
> I agree there is no point implementing this stuff if we don't do the discrete logic correctly. So either we do the above, or we go back to simple "contains_value()" that just gets a single data value and not a range, and that's all we implement.

Ok, I will try to get there, and will revert back to the simple scalar containment if do not reach a good point in a reasonable time :wink: 

Thanks for your patience!

lelit:
Mike, I could not use the `memoized_property` because under py310+ the dataclass uses `__slots__`, so it has no `__dict__`.

If you think the approach is good enough, I will try to rewrite the tests as you suggested above.

CaselIT:
> Mike, I could not use the `memoized_property` because under py310+ the dataclass uses `__slots__`, so it has no `__dict__`.
> 
> If you think the approach is good enough, I will try to rewrite the tests as you suggested above.

They use slots only if we use slots=True. I guess we can remove it if we have it or we add slots explicitly and use memoized_slots

zzzeek:
> Mike, I could not use the `memoized_property` because under py310+ the dataclass uses `__slots__`, so it has no `__dict__`.
> 
> If you think the approach is good enough, I will try to rewrite the tests as you suggested above.

let's skip the memoized part of it for now.  IIUC the logic can be written so that the `get_discrete_step()` method is only called upon if we are in fact within this unusual edge case for the comparison.

lelit:
> let's skip the memoized part of it for now. IIUC the logic can be written so that the `get_discrete_step()` method is only called upon if we are in fact within this unusual edge case for the comparison.

Right, improved that aspect in last commit.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision faf358f4090572ac6373acdb7eb53d19eb5b2e9e of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset faf358f4090572ac6373acdb7eb53d19eb5b2e9e added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4151

lelit:
Sorry for these glitches... is there a recommended setup that runs all the linters locally?

CaselIT:
most things are solved by running black and zimports. To run them all you can setup pre-commit https://pre-commit.com/

lelit:
> To run them all you can setup pre-commit https://pre-commit.com/

Excellent, thanks! 

Searching for it I found that recommendation [here](https://www.sqlalchemy.org/develop.html#environment), maybe it deserves a link also in the PR template, or in a README.contribute.rst?

CaselIT:
> maybe it deserves a link also in the PR template, or in a README.contribute.rst?

makes sense. Care to update them and open a PR? Thanks

lib/sqlalchemy/dialects/postgresql/ranges.py
```diff
@@ -80,6 +83,139 @@ def __init__(
...
+    def _get_discrete_step(self):
+        "Determine the “step” for this range, if it is a discrete one."
+
+        # See
+        # https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-DISCRETE
+        # for the rationale
+
+        if isinstance(self.lower, int) or isinstance(self.upper, int):
+            return 1
+        if isinstance(self.lower, date) or isinstance(self.upper, date):
```
sqla-tester, line 125: **mike bayer** (zzzeek) wrote:

would look better as an elif

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4151

lib/sqlalchemy/dialects/postgresql/ranges.py
```diff
@@ -80,6 +83,139 @@ def __init__(
...
+            return timedelta(days=1)
+
```
sqla-tester, line 127: **mike bayer** (zzzeek) wrote:

needs an "else: return None", mypy in particular wants that

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4151

lib/sqlalchemy/dialects/postgresql/ranges.py
```diff
@@ -80,6 +83,139 @@ def __init__(
...
+        oupper = other.upper
+
+        # A bilateral unbound range contains any other range
+        if olower is oupper is None:
+            return True
+
+        slower = self.lower
+        supper = self.upper
+
+        # A lower-bound range cannot contain a lower-unbound range
```
sqla-tester, line 149: **mike bayer** (zzzeek) wrote:

all very nice

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4151

lib/sqlalchemy/dialects/postgresql/ranges.py
```diff
@@ -80,6 +83,139 @@ def __init__(
...
+            return False
+
+        slower_inc = self.bounds[0] == "["
+        supper_inc = self.bounds[1] == "]"
+        olower_inc = other.bounds[0] == "["
+        oupper_inc = other.bounds[1] == "]"
+
+        # Check the lower end
+        step = -1
+        if slower is not None and olower is not None:
```
sqla-tester, line 164: **mike bayer** (zzzeek) wrote:

wow gets intense!

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4151

test/dialect/postgresql/test_dialect.py
```diff
@@ -1210,3 +1212,67 @@ def test_get_dialect(self):
...
+        "empty",
+        "(,)",
+        "(,4]",
+        "[1,)",
+        "[1,4)",
+        "[1,4]",
+        "(1,4]",
+        "(1,4)",
+        argnames="rrepr",
+    )
```
sqla-tester, line 1232: **mike bayer** (zzzeek) wrote:

so here I want to use literal(), like this:

    connection.scalar(
        select(literal(r, INT4RANGE).contains(v))
    )

will that work?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4151

sqla-tester (commented):

looks great!!!!!!

for changes, we have:

1. elif / else change
2. can we use the datatypes for the SQL round trip in the test
3. short changelog file doc/build/changelog/unreleased_20/8706.rst

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4151

lelit:
Thank you for the review (and the kind words :-)
Do my last two commits fulfill points 1. and 2. of your requested changes?
If so, wrt the third point, can you suggest the `:tags:` for this? I'm unsure if this is a `usecase` or a `feature`...

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision ceee1951194f5014ee17416e4cef70abc2ffbfd9 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset ceee1951194f5014ee17416e4cef70abc2ffbfd9 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4151

lelit:
> I have time to finish this up if you want to be done working on it, looks great just need to fix a few tests

You're welcome, but I'm willing to do whatever you suggest!

zzzeek:
at the moment some of the tests are failing, it looks like the new expressions using contains() are maybe not organized correctly

zzzeek:
could be a false positive of some kind.  the link to the review w/ tests is in the big github box below where it says "ci_verification — Gerrit review has failed verification ".   the gerrit is https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4151, the jenkins run is https://jenkins.sqlalchemy.org/job/sqlalchemy_2x_gerrit/cext=cext,db=postgresql,pyv=py311/2580/ , and an example of the failure is https://jenkins.sqlalchemy.org/job/sqlalchemy_2x_gerrit/cext=cext,db=postgresql,pyv=py311/2580/testReport/junit/test.dialect.postgresql.test_dialect/TestRange_postgresql+asyncpg_async_15_0/test_contains_value_____0_/ :+1: 

```
sqlalchemy.exc.DBAPIError: (sqlalchemy.dialects.postgresql.asyncpg.Error) <class 'asyncpg.exceptions.DataError'>: invalid input for query argument $1: '(,)' (list, tuple or Range object expected (got type <class 'str'>))
[SQL: SELECT %s::INT4RANGE AS anon_1, %s::INT4RANGE @> %s::INTEGER AS anon_2]
[parameters: ('(,)', '(,)', 0)]
```

zzzeek:
looks like it's all asyncpg / pg8000.  so if we further enhance that test to not use any textual ranges, that should fix it.

lelit:
> looks like it's all asyncpg / pg8000. so if we further enhance that test to not use any textual ranges, that should fix it.

The fact that the failure is only about one single syntax (ie `(,)`) surprises me even more! The failure heads [here](https://github.com/MagicStack/asyncpg/blob/master/asyncpg/protocol/codecs/range.pyx#L54), and I fail to see why in all other cases that gets the right kind of value, and just with the empty set it gets a `str`...

Anyway, given that the other new test does not fail, is it really important here to use `literal()` instead of using `literal_column() like done there?

lelit:
I mean something like
```diff
diff --git a/test/dialect/postgresql/test_dialect.py b/test/dialect/postgresql/test_dialect.py
index 608843b59..7bb3740f5 100644
--- a/test/dialect/postgresql/test_dialect.py
+++ b/test/dialect/postgresql/test_dialect.py
@@ -1232,8 +1232,8 @@ class TestRange(fixtures.TestBase):
     )
     def test_contains_value(self, connection, rrepr, v):
         q = select(
-            literal(rrepr, INT4RANGE),
-            literal(rrepr, INT4RANGE).contains(v),
+            literal_column(f"'{rrepr}'::int4range", INT4RANGE),
+            literal_column(f"'{rrepr}'::int4range", INT4RANGE).contains(v),
         )
         r, expected = connection.execute(q).first()
         eq_(r.contains(v), expected)
```

zzzeek:
i'm not sure, id have to play with it.   The thing that stood out to me was where a single literal_column had the whole expression rendered inside of it:

```literal_column(f"'{r1repr}'::int4range @> '{r2repr}'::int4range")```

I assumed that was the part that was failing but I didnt check.

**Comment by lelit** [comment]
AFAICT the failures are from `test_contains_value()`, not from `test_contains_range()`.

**Comment by zzzeek** [comment]
all the tests in TestRange::test_contains_value fail for me for the asyncpg driver.  For pg8000 I dont think we are supporting ranges, so by rolling these tests into ``_RangeTypeRoundTrip`` in test_types the pg8000 part will be fixed.   doing this will require that some additional lists of data are generated for the different datatypes, which we can do in an automated fashion like this:

```py
diff --git a/test/dialect/postgresql/test_types.py b/test/dialect/postgresql/test_types.py
index 91eada9a81..5824b8e5e0 100644
--- a/test/dialect/postgresql/test_types.py
+++ b/test/dialect/postgresql/test_types.py
@@ -3850,6 +3850,46 @@ class _RangeTypeRoundTrip(fixtures.TablesTest):
     __requires__ = ("range_types",)
     __backend__ = True
 
+    def _bounds_combinations_str(self):
+        spec = self._data_str()
+        l, r = spec.strip("[]()").split(",")
+
+        return [
+            "empty",
+            "(,)",
+            f"(,{r}]",
+            f"[{l},)",
+            f"[{l},{r})",
+            f"[{l},{r}]",
+            f"({l},{r}]",
+            f"({l},{r})",
+        ]
+
+    def _bounds_combinations_obj(self):
+        obj = self._data_obj()
+        l, r =obj.lower, obj.upper
+
+        return [
+            Range(empty=True),
+            Range(bounds="()"),
+            Range(upper=r, bounds="(]"),
+            Range(lower=l, bounds="[)"),
+            Range(lower=l, upper=r, bounds="[)"),
+            Range(lower=l, upper=r, bounds="[]"),
+            Range(lower=l, upper=r, bounds="(]"),
+            Range(lower=l, upper=r, bounds="()"),
+        ]
+
+    def _values(self):
+        """subclasses should hardcode five values here,
+        i.e. leftwards, rightwards, inside, left bound, right bound.
+
+        the ints can all use the same numbers and then we'd need the dates.
+
+        """
+
+
+
     @classmethod
     def define_tables(cls, metadata):
         # no reason ranges shouldn't be primary keys,
@@ -4036,7 +4076,6 @@ class _DateRangeTests:
     def _data_obj(self):
         return Range(datetime.date(2013, 3, 23), datetime.date(2013, 3, 24))
 
-
 class _DateTimeRangeTests:
 
     _col_type = TSRANGE
```




the originating failure for asyncpg is the same in all the tests:

```
  File "asyncpg/protocol/prepared_stmt.pyx", line 168, in asyncpg.protocol.protocol.PreparedStatementState._encode_bind_msg
  File "asyncpg/protocol/codecs/base.pyx", line 206, in asyncpg.protocol.protocol.Codec.encode
  File "asyncpg/protocol/codecs/base.pyx", line 128, in asyncpg.protocol.protocol.Codec.encode_range
  File "asyncpg/protocol/codecs/range.pyx", line 55, in asyncpg.protocol.protocol.range_encode
TypeError: list, tuple or Range object expected (got type <class 'str'>)

```

So i think to fix, the tests here should just replace the range strings with Range objects.    the range string version can be maintained, just should be marked as supported only on psycopg, psycopg2.

zzzeek:
note if you want to run asyncpg tests locally, just install asyncpg with pip. if you have a database on localhost called "test" , you can run:

```
pytest test/dialect/postgresql/test_dialect.py::TestRange --db asyncpg
```

otherwise for a manual URL:

```
pytest test/dialect/postgresql/test_dialect.py::TestRange --dburi postgresql+asyncpg://user:pass@host/dbname?async_fallback=true

```

that `async_fallback` part is the magic that allows us to test asyncpg in a synchronous environment.

lelit:
I will try to understand what's going on with asyncpg, but that shall await tomorrow :-)

lelit:
Trying it locally, it seems that the change I suggested [above](https://github.com/sqlalchemy/sqlalchemy/pull/8707#issuecomment-1299005724) fixes the problem with asyncpg:
```
16 $ pytest test/dialect/postgresql/test_dialect.py::TestRange --dburi postgresql+asyncpg://localhost/sa_tests?async_fallback=true
======================================================== test session starts =========================================================
platform linux -- Python 3.10.8, pytest-7.1.3, pluggy-1.0.0 -- /home/lele/wip/sqlalchemy/env310/bin/python3
cachedir: .pytest_cache
rootdir: /home/lele/wip/sqlalchemy, configfile: pyproject.toml
collected 160 items                                                                                                                  

test/dialect/postgresql/test_dialect.py::TestRange_postgresql+asyncpg_async_11_17::test_contains_range[(0,1]-(,)] PASSED       [  0%]
test/dialect/postgresql/test_dialect.py::TestRange_postgresql+asyncpg_async_11_17::test_contains_range[(0,1]-(1,2]] PASSED     [  1%]
test/dialect/postgresql/test_dialect.py::TestRange_postgresql+asyncpg_async_11_17::test_contains_range[(0,1]-(1,4)] PASSED     [  1%]
...
test/dialect/postgresql/test_dialect.py::TestRange_postgresql+asyncpg_async_11_17::test_contains_value[empty-0] PASSED         [ 97%]
test/dialect/postgresql/test_dialect.py::TestRange_postgresql+asyncpg_async_11_17::test_contains_value[empty-1] PASSED         [ 98%]
test/dialect/postgresql/test_dialect.py::TestRange_postgresql+asyncpg_async_11_17::test_contains_value[empty-2] PASSED         [ 98%]
test/dialect/postgresql/test_dialect.py::TestRange_postgresql+asyncpg_async_11_17::test_contains_value[empty-4] PASSED         [ 99%]
test/dialect/postgresql/test_dialect.py::TestRange_postgresql+asyncpg_async_11_17::test_contains_value[empty-5] PASSED         [100%]

======================================================== 160 passed in 1.79s =========================================================
```

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 3a74a0d93e63032ebee02992977498c717a077ff of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 3a74a0d93e63032ebee02992977498c717a077ff added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4151

lelit:
I saw the failure: if I understand it correctly, it is now failing only for pg8000, right?

zzzeek:
> I saw the failure: if I understand it correctly, it is now failing only for pg8000, right?

yes, there is one seemingly unrelated psycopg2 failure which may be an artifact of these pg8000 failures, not sure

lelit:
Ok.

I'm afraid I'd need further clarification on how to proceed, as I did not understand your hint above about "rolling these tests into _RangeTypeRoundTrip".

zzzeek:
1. remove the tests from test_dialect.py and also remove the TestRange class entirely.
2. add test_contains_value() and test_contains_range() tests to test/dialect/postgresql/test_types.py -> _RangeTypeRoundTrip class:
3. in each test, where you currently have `INT4RANGE` hardcoded, you would instead use `self._col_type`, or `self._col_str` for the string representation
4. for the testing.combinations that lists out various ranges, you would use new method "_bounds_combination_str" which would be defined on _RangeTypeRoundTrip.  to achieve this we likely need to use `@staticmethod` for it, and also apply `@staticmethod` to the existing `def _data_str()` and `def _data_obj()` methods on each subclass
5. once that works I do want to get test coverage using Range objects passed into the SQL expressions as well, so that we can round trip the SQL-level comparison functions also.

So with the mirror of the code I have in Gerrit, I can go in and add these changes, it's just once I do that, the PR here will no longer be in sync, since we dont have sync in that direction.  if you wanted to sync this PR after I update Gerrit you could force push the changes.

lelit:
Thank you for the plan. I will try to implement that, but it will take some time to me to understand the _RangeTypeRoundTrip mixin logic and its concrete subclasses and I won't be able to attempt that immediately, so if you are willing to do it sooner, I'd really appreciate that.

zzzeek:
OK I am generalizing the tests and that will be in the gerrit shortly, but not every test is passing yet.  

in particular can you figure this one out for me?   I dont know why PG says false but it seems we would need an adjustment to the comparison logic.

```py
>>> from sqlalchemy import create_engine
>>> e = create_engine("postgresql://scott:tiger@localhost/test")
>>> c = e.connect()
>>> c.exec_driver_sql("SELECT '[2013-03-22 14:30:00,2013-03-31 23:30:00)'::TSRANGE <@ '(,2013-03-30 23:30]'::TSRANGE").scalar()
False
>>> from sqlalchemy.dialects.postgresql import Range
>>> import datetime
>>> r1 = Range(lower=datetime.datetime(2013, 3, 22, 14, 30), upper=datetime.datetime(2013, 3, 31, 23, 30), bounds='[)', empty=False)
>>> r2 = Range(lower=None, upper=datetime.datetime(2013, 3, 30, 23, 30), bounds='(]', empty=False)
>>> r2.contains(r1)
True

```

lelit:
Yes, will do asap.

lelit:
PG is clearly right, because r2 ends a day before r1, regardless inclusive/exclusive

lelit:
Got it! It's the usual date/datetime quirk: a `datetime` **is a** `date` subclass, and thus `_get_discrete_step()` as written returns `timedelta(days=1)`... sorry

This fixes it:
```diff
diff --git a/lib/sqlalchemy/dialects/postgresql/ranges.py b/lib/sqlalchemy/dialects/postgresql/ranges.py
index 9a05411c0..6235fdf80 100644
--- a/lib/sqlalchemy/dialects/postgresql/ranges.py
+++ b/lib/sqlalchemy/dialects/postgresql/ranges.py
@@ -9,6 +9,7 @@
 
 import dataclasses
 from datetime import date
+from datetime import datetime
 from datetime import timedelta
 from typing import Any
 from typing import Generic
@@ -122,6 +123,9 @@ def _get_discrete_step(self):
 
         if isinstance(self.lower, int) or isinstance(self.upper, int):
             return 1
+        elif isinstance(self.lower, datetime) or isinstance(self.upper, datetime):
+            # This is required, because a `isinstance(datetime.now(), date)` is True
+            return None
         elif isinstance(self.lower, date) or isinstance(self.upper, date):
             return timedelta(days=1)
         else:
```

Can I push it here?

zzzeek:
i just pushed a huge change up so let me put it in mine, and then we hopefully should be there

zzzeek:
OK it's in, the refactoring of the tests are at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4151/8/test/dialect/postgresql/test_types.py

zzzeek:
alrighty.  We can put this in with some final notes / docs I can add, but what are your thoughts on all the other comparison methods?   can these be stated in terms of the contains that was already done?  (like,    "a overlaps b" means make new ranges and use contains, like "[a_lower, b_upper] contains [a_upper, b_lower]" , something like that)

thoughts?  See below

```py
    def overlaps(self, other):
        """Boolean expression. Returns true if the column overlaps
        (has points in common with) the right hand operand.
        """
        raise NotImplementedError("not yet implemented")

    def strictly_left_of(self, other):
        """Boolean expression. Returns true if the column is strictly
        left of the right hand operand.
        """
        raise NotImplementedError("not yet implemented")

    __lshift__ = strictly_left_of

    def strictly_right_of(self, other):
        """Boolean expression. Returns true if the column is strictly
        right of the right hand operand.
        """
        raise NotImplementedError("not yet implemented")

    __rshift__ = strictly_right_of

    def not_extend_right_of(self, other):
        """Boolean expression. Returns true if the range in the column
        does not extend right of the range in the operand.
        """
        raise NotImplementedError("not yet implemented")

    def not_extend_left_of(self, other):
        """Boolean expression. Returns true if the range in the column
        does not extend left of the range in the operand.
        """
        raise NotImplementedError("not yet implemented")

    def adjacent_to(self, other):
        """Boolean expression. Returns true if the range in the column
        is adjacent to the range in the operand.
        """
        raise NotImplementedError("not yet implemented")

    def __add__(self, other):
        """Range expression. Returns the union of the two ranges.
        Will raise an exception if the resulting range is not
        contiguous.
        """
        raise NotImplementedError("not yet implemented")
```

lelit:
> what are your thoughts on all the other comparison methods?

I'd like to implement those as well: although i seldom used them (on the Python side), I think that sooner or later, someone will surely ask _why_ they are missing :wink:

Please tell me if/how you would prefer go there, ie if it sounds reasonable to have them all, and in the positive case if I should extend this PR or instead if we can install this tranche and opening a new one (I bet the latter option is better).

zzzeek:
lets merge what we have first and we can do that as a new PR etc.    i want beta3 to go out soon

lelit:
Great, thank you!

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4151 has been **merged**. Congratulations! :)

Issue #8765, opened 2022-11-07
Further improvements to new SA 2.0 PG Range


lelit:
### Describe the use case

[This comment](https://github.com/sqlalchemy/sqlalchemy/pull/8707#issuecomment-1302068565) asks for implementation of a new set of methods on the PG Range class to bring it in par with the functionalities of the concrete `XXXRANGE` classes wrt comparison between different ranges.

PR #8707 implemented a few of them, as well as a test infrastructure that asserts their correct behaviour.

### Databases / Backends / Drivers targeted

The new class is PostgreSQL specific, and drivers agnostic.

### Example Use

```python
from datetime import date
from itertools import pairwise

from sqlalchemy import create_engine, select
from sqlalchemy.dialects.postgresql import DATERANGE, Range
from sqlalchemy.orm import DeclarativeBase, Mapped, mapped_column, Session


class Base(DeclarativeBase):
    pass


class Contract(Base):
    __tablename__ = "contracts"

    id: Mapped[int] = mapped_column(primary_key=True)
    validity: Mapped[Range[date]] = mapped_column(DATERANGE)


engine = create_engine("postgresql+psycopg://scott:tiger@pg14/dbname")

Base.metadata.create_all(engine)

with Session(engine) as session:
    old_contract = Contract(validity=Range(date(2021, 1, 1), date(2021, 12, 31)))
    session.add(old_contract)
    cur_contract = Contract(validity=Range(date(2022, 1, 1), date(2022, 12, 31)))
    session.add(cur_contract)
    session.commit()

    validities = session.scalars(select(Contract.validity).order_by(Contract.validity)).all()
    for v1, v2 in pairwise(validities):
        assert v1.adjacent_to(v2), f"There is a gap between {v1} and {v2}"
```

lelit:
The related PR didn't carry a "fixes" marker... should this be closed manually?

zzzeek:
wow that's a weird oversight for me.   yes thanks!

lelit:
Mike, I see a possible "defect" in [current implementation](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/dialects/postgresql/ranges.py#L287) of the `__eq__()` method: it accepts an `other: Any` argument, and that suggests that a code like
```python
if my_range == None:
  pass
```
is valid, even if perhaps not ideal. As written, that raises
```python
>       elif self.empty != other.empty:
[...]

PR #8766, opened 2022-11-07
Issue #8765: implement missing methods on PG Range

lelit:
### Description
This PR implements missing methods on the PG `Range` class, as described by issue #8765.

### Checklist
This pull request is:

- [ ] A short code fix

### DIFF
--- a/doc/build/changelog/unreleased_20/8765.rst
+++ b/doc/build/changelog/unreleased_20/8765.rst
@@ -1,0 +1,10 @@
+.. change::
+    :tags: usecase, postgresql
+    :tickets: 8765
+
+    Complementing :ticket:`8690`, new comparison methods such as
+    ``adjacent_to()``, ``difference()``, ``union()``, etc., were added to the
+    PG-specific range objects, bringing them in par with the standard
+    operators implemented by the underlying
+    :attr:`_postgresql.AbstractRange.comparator_factory`. Pull request
+    courtesy Lele Gaifax.
--- a/lib/sqlalchemy/dialects/postgresql/ranges.py
+++ b/lib/sqlalchemy/dialects/postgresql/ranges.py
@@ -225,72 +225,23 @@ class Range(Generic[_T]):
         if other.empty:
             return False
 
-        olower = other.lower
-        oupper = other.upper
-
-        # A bilateral unbound range contains any other range
-        if olower is oupper is None:
-            return True
-
         slower = self.lower
+        slower_b = self.bounds[0]
+        olower = other.lower
+        olower_b = other.bounds[0]
+
+        if self._compare_edges(slower, slower_b, olower, olower_b) < 0:
+            return False
+
         supper = self.upper
+        supper_b = self.bounds[1]
+        oupper = other.upper
+        oupper_b = other.bounds[1]
 
-        # A lower-bound range cannot contain a lower-unbound range
-        if slower is None and olower is not None:
+        if self._compare_edges(supper, supper_b, oupper, oupper_b) > 0:
             return False
 
-        # Likewise on the right side
-        if supper is None and oupper is not None:
-            return False
-
-        slower_inc = self.bounds[0] == "["
-        supper_inc = self.bounds[1] == "]"
-        olower_inc = other.bounds[0] == "["
-        oupper_inc = other.bounds[1] == "]"
-
-        # Check the lower end
-        step = -1
-        if slower is not None and olower is not None:
-            lside = olower < slower
-            if not lside:
-                if not slower_inc or olower_inc:
-                    lside = olower == slower
-            if not lside:
-                # Cover (1,x] vs [2,x) and (0,x] vs [1,x)
-                if not slower_inc and olower_inc and slower < olower:
-                    step = self._get_discrete_step()
-                    if step is not None:
-                        lside = olower == (slower + step)
-                elif slower_inc and not olower_inc and slower > olower:
-                    step = self._get_discrete_step()
-                    if step is not None:
-                        lside = (olower + step) == slower
-            if not lside:
-                return False
-
-        # Lower end already considered, an upper-unbound range surely contains
-        # this
-        if oupper is None:
-            return True
-
-        # Check the upper end
-        uside = oupper > supper
-        if not uside:
-            if not supper_inc or oupper_inc:
-                uside = oupper == supper
-            if not uside:
-                # Cover (x,2] vs [x,3) and (x,1] vs [x,2)
-                if supper_inc and not oupper_inc and supper < oupper:
-                    if step == -1:
-                        step = self._get_discrete_step()
-                    if step is not None:
-                        uside = oupper == (supper + step)
-                elif not supper_inc and oupper_inc and supper > oupper:
-                    if step == -1:
-                        step = self._get_discrete_step()
-                    if step is not None:
-                        uside = (oupper + step) == supper
-        return uside
+        return True
 
     def contains(self, value: Union[_T, Range]) -> bool:
         "Determine whether this range contains `value`."
--- a/lib/sqlalchemy/dialects/postgresql/ranges.py
+++ b/lib/sqlalchemy/dialects/postgresql/ranges.py
@@ -561,8 +561,7 @@ class Range(Generic[_T]):
             else:
                 return Range(oupper, supper, bounds=rlower_b + supper_b)
 
-        # TODO: figure out if I handled all the cases above
-        assert False
+        assert False, f"Unhandled case computing {self} - {other}"
 
     __sub__ = difference
 
--- a/lib/sqlalchemy/dialects/postgresql/ranges.py
+++ b/lib/sqlalchemy/dialects/postgresql/ranges.py
@@ -525,7 +525,12 @@ class Range(Generic[_T]):
         # middle
         if sl_vs_ol <= 0 and su_vs_ol >= 0 and su_vs_ou <= 0:
             rupper_b = ")" if olower_b == "[" else "]"
-            if self._compare_edges(slower, slower_b, olower, rupper_b) == 0:
+            if (
+                slower_b != "["
+                and rupper_b != "]"
+                and self._compare_edges(slower, slower_b, olower, rupper_b)
+                == 0
+            ):
                 return Range(None, None, empty=True)
             else:
                 return Range(slower, olower, bounds=slower_b + rupper_b)
@@ -533,8 +538,13 @@ class Range(Generic[_T]):
         # If this range starts in the middle of the other and extends to its
         # right
         if sl_vs_ol >= 0 and su_vs_ou >= 0 and sl_vs_ou <= 0:
-            rlower_b = "(" if oupper_b == "]" else "("
-            if self._compare_edges(oupper, rlower_b, supper, supper_b) == 0:
+            rlower_b = "(" if oupper_b == "]" else "["
+            if (
+                rlower_b != "["
+                and supper_b != "]"
+                and self._compare_edges(oupper, rlower_b, supper, supper_b)
+                == 0
+            ):
                 return Range(None, None, empty=True)
             else:
                 return Range(oupper, supper, bounds=rlower_b + supper_b)
--- a/lib/sqlalchemy/dialects/postgresql/ranges.py
+++ b/lib/sqlalchemy/dialects/postgresql/ranges.py
@@ -134,6 +134,86 @@ class Range(Generic[_T]):
         else:
             return None
 
+    def _compare_edges(
+        self,
+        value1: Optional[_T],
+        bound1: str,
+        value2: Optional[_T],
+        bound2: str,
+        only_values: bool = False,
+    ) -> int:
+        """Compare two range bounds.
+
+        Return -1, 0 or 1 respectively when `value1` is less than,
+        equal to or greater than `value2`.
+
+        When `only_value` is ``True``, do not consider the *inclusivity*
+        of the edges, just their values.
+        """
+
+        value1_is_lower_bound = bound1 in {"[", "("}
+        value2_is_lower_bound = bound2 in {"[", "("}
+
+        # Infinite edges are equal when they are on the same side,
+        # otherwise a lower edge is considered less than the upper end
+        if value1 is value2 is None:
+            if value1_is_lower_bound == value2_is_lower_bound:
+                return 0
+            else:
+                return -1 if value1_is_lower_bound else 1
+        elif value1 is None:
+            return -1 if value1_is_lower_bound else 1
+        elif value2 is None:
+            return 1 if value2_is_lower_bound else -1
+
+        value1_inc = bound1 in {"[", "]"}
+        value2_inc = bound2 in {"[", "]"}
+        step = self._get_discrete_step()
+
+        if step is not None:
+            # "Normalize" the two edges as '[)', to simplify successive logic
+            if value1_is_lower_bound:
+                if not value1_inc:
+                    value1 -= step
+                    value1_inc = True
+            else:
+                if value1_inc:
+                    value1 += step
+                    value1_inc = False
+            if value2_is_lower_bound:
+                if not value2_inc:
+                    value2 -= step
+                    value2_inc = True
+            else:
+                if value2_inc:
+                    value2 += step
+                    value2_inc = False
+
+        if value1 < value2:
+            return -1
+        elif value1 > value2:
+            return 1
+        elif only_values:
+            return 0
+        else:
+            # Neither one is infinite but are equal, so we
+            # need to consider the respective inclusive/exclusive
+            # flag
+
+            if value1_inc and value2_inc:
+                return 0
+            elif not value1_inc and not value2_inc:
+                if value1_is_lower_bound == value2_is_lower_bound:
+                    return 0
+                else:
+                    return 1 if value1_is_lower_bound else -1
+            elif not value1_inc:
+                return 1 if value1_is_lower_bound else -1
+            elif not value2_inc:
+                return -1 if value2_is_lower_bound else 1
+            else:
+                return 0
+
     def contained_by(self, other: Range) -> bool:
         "Determine whether this range is a contained by `other`."
 
--- a/lib/sqlalchemy/dialects/postgresql/ranges.py
+++ b/lib/sqlalchemy/dialects/postgresql/ranges.py
@@ -174,7 +174,7 @@ class Range(Generic[_T]):
             # "Normalize" the two edges as '[)', to simplify successive logic
             if value1_is_lower_bound:
                 if not value1_inc:
-                    value1 -= step
+                    value1 += step
                     value1_inc = True
             else:
                 if value1_inc:
@@ -182,7 +182,7 @@ class Range(Generic[_T]):
                     value1_inc = False
             if value2_is_lower_bound:
                 if not value2_inc:
-                    value2 -= step
+                    value2 += step
                     value2_inc = True
             else:
                 if value2_inc:
--- a/lib/sqlalchemy/dialects/postgresql/ranges.py
+++ b/lib/sqlalchemy/dialects/postgresql/ranges.py
@@ -452,6 +452,67 @@ class Range(Generic[_T]):
 
     __add__ = union
 
+    def difference(self, other: Range) -> Range:
+        """Compute the difference between this range and the `other`.
+
+        This raises a ``ValueError`` exception if the two ranges are
+        "disjunct", that is neither adjacent nor overlapping.
+        """
+
+        # Subtracting an empty range is a no-op
+        if self.empty or other.empty:
+            return self
+
+        slower = self.lower
+        slower_b = self.bounds[0]
+        supper = self.upper
+        supper_b = self.bounds[1]
+        olower = other.lower
+        olower_b = other.bounds[0]
+        oupper = other.upper
+        oupper_b = other.bounds[1]
+
+        sl_vs_ol = self._compare_edges(slower, slower_b, olower, olower_b)
+        su_vs_ou = self._compare_edges(supper, supper_b, oupper, oupper_b)
+        if sl_vs_ol < 0 and su_vs_ou > 0:
+            raise ValueError(
+                "Subtracting a strictly inner range is not implemented"
+            )
+
+        sl_vs_ou = self._compare_edges(slower, slower_b, oupper, oupper_b)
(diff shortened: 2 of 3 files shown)


lelit:
Note that I re-used the existing test fixtures, originally designed to exercise the _containment_ methods, also in the new comparison methods tests. This is not ideal, because several corner cases are missing, in particular for the `adjacent_to()` method. I will either expand those fixtures or distill new ones with more appropriate ranges.

lelit:
For completeness, this is the (a bit redundant) test I used to ensure `adjacent_to()` is working on those cases:
```diff
diff --git a/test/dialect/postgresql/test_dialect.py b/test/dialect/postgresql/test_dialect.py
index 27d4a4cf9..aac1a838c 100644
--- a/test/dialect/postgresql/test_dialect.py
+++ b/test/dialect/postgresql/test_dialect.py
@@ -4,6 +4,7 @@
 import logging
 import logging.handlers
 
+from sqlalchemy import literal_column
 from sqlalchemy import BigInteger
 from sqlalchemy import bindparam
 from sqlalchemy import cast
@@ -30,7 +31,9 @@
 from sqlalchemy import TypeDecorator
 from sqlalchemy.dialects.postgresql import base as postgresql
 from sqlalchemy.dialects.postgresql import HSTORE
+from sqlalchemy.dialects.postgresql import INT4RANGE
 from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.dialects.postgresql import NUMRANGE
 from sqlalchemy.dialects.postgresql import psycopg as psycopg_dialect
 from sqlalchemy.dialects.postgresql import psycopg2 as psycopg2_dialect
 from sqlalchemy.dialects.postgresql import Range
@@ -1210,3 +1213,92 @@ def test_get_dialect(self):
     def test_async_version(self):
         e = create_engine("postgresql+psycopg_async://")
         is_true(isinstance(e.dialect, psycopg_dialect.PGDialectAsync_psycopg))
+
+
+class TestRange(fixtures.TestBase):
+    __only_on__ = "postgresql"
+    __backend__ = True
+
+    @testing.combinations(
+        "empty",
+        "(1,2]",
+        "[1,2)",
+        "(,)",
+        "[,1)",
+        "[1,)",
+        "[1,1]",
+        "[1,4)",
+        "[1,4]",
+        "(1,4]",
+        "(1,4)",
+        argnames="r2repr",
+    )
+    @testing.combinations(
+        "empty",
+        "[,)",
+        "[,1)",
+        "[1,)",
+        "[1,4)",
+        "[1,4]",
+        "(1,4]",
+        "(1,4)",
+        "(2,3]",
+        "[2,3]",
+        "(3,4]",
+        "[3,5)",
+        "[0,0]",
+        "[-1,0)",
+        "[-4,1)",
+        "(6,8)",
+        argnames="r1repr",
+    )
+    def test_int4range_adjacent(self, connection, r1repr, r2repr):
+        q = select(
+            literal_column(f"'{r1repr}'::int4range", INT4RANGE).label("r1"),
+            literal_column(f"'{r2repr}'::int4range", INT4RANGE).label("r2"),
+            literal_column(f"'{r1repr}'::int4range -|- '{r2repr}'::int4range"),
+        )
+        r1, r2, adjacent = connection.execute(q).first()
+        eq_(r1.adjacent_to(r2), adjacent)
+
+    @testing.combinations(
+        "empty",
+        "(1.1,2.9]",
+        "[1.1,2.9)",
+        "(,)",
+        "[,1.1)",
+        "[1.1,)",
+        "[1.1,1.1]",
+        "[1.1,4.5)",
+        "[1.1,4.5]",
+        "(1.1,4.5]",
+        "(1.1,4.5)",
+        argnames="r2repr",
+    )
+    @testing.combinations(
+        "empty",
+        "[,)",
+        "[,1.1)",
+        "[1.1,)",
+        "[1.1,4.5)",
+        "[1.1,4.5]",
+        "(1.1,4.5]",
+        "(1.1,4.5)",
+        "(2.9,3.2]",
+        "[2.9,3.2]",
+        "(3.2,4.5]",
+        "[3.2,5.5)",
+        "[0.1,0.1]",
+        "[-1.1,0.1)",
+        "[-4.5,1.1)",
+        "(6.1,8.1)",
+        argnames="r1repr",
+    )
+    def test_numrange_adjacent(self, connection, r1repr, r2repr):
+        q = select(
+            literal_column(f"'{r1repr}'::numrange", NUMRANGE).label("r1"),
+            literal_column(f"'{r2repr}'::numrange", NUMRANGE).label("r2"),
+            literal_column(f"'{r1repr}'::numrange -|- '{r2repr}'::numrange"),
+        )
+        r1, r2, adjacent = connection.execute(q).first()
+        eq_(r1.adjacent_to(r2), adjacent)
```

lelit:
I'd like to know your opinion on the following:

1. Is there a particular reason behind the choice of [`__add__`()](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/dialects/postgresql/ranges.py#L263) method, in contrast to what has been done for example with [strictly_right_of](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/dialects/postgresql/ranges.py#L237) of which `__rshift__` is an alias? For consistency, I'd instead add a (say) `union(self, other)` method, aliased to `__add__`, both on the `Range` class and on the `AbstractRange` one.
2. Was the method [AbstractRange.`__ne__`()](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/dialects/postgresql/ranges.py#L330) omitted on purpose on `Range`? If not, I'd implement it on the latter as well, again for consistency, with the same consideration above about the naming.
3. If there is a `__ne__`, shouldn't be there also `__eq__`?
4. Given that there is `__add__`, shouldn't be there also `__sub__`?

zzzeek:
> I'd like to know your opinion on the following:
> 
>     1. Is there a particular reason behind the choice of [`__add__`()](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/dialects/postgresql/ranges.py#L263) method, in contrast to what has been done for example with [strictly_right_of](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/dialects/postgresql/ranges.py#L237) of which `__rshift__` is an alias? For consistency, I'd instead add a (say) `union(self, other)` method, aliased to `__add__`, both on the `Range` class and on the `AbstractRange` one.

oh you mean why is there no union() method?   by all means we should add one.

> 
>     2. Was the method [AbstractRange.`__ne__`()](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/dialects/postgresql/ranges.py#L330) omitted on purpose on `Range`? If not, I'd implement it on the latter as well, again for consistency, with the same consideration above about the naming.
> 
>     3. If there is a `__ne__`, shouldn't be there also `__eq__`?

`Range` is a dataclass, so ``__eq__()`` and ``__ne__()`` should be implemented already. that was one of the reasons I used dataclasses.   there should be tests for that, if not, we should add them.


> 
>     4. Given that there is `__add__`, shouldn't be there also `__sub__`?

if we have one that works on AbstractRange then yes, if there's a PG operation that this maps towards.

lelit:
Thanks Mike!

On `__add__`, I will then rename existing methods to `union()` method on `Range` and `AbstractRange`, aliasing them to `__add__`.

On `__eq__`/`__ne__`: my question was imprecise, as I knew the dataclass heritage, what I meant was more directed at the different behaviour between `AbstractRange` and `Range`, and whether we care to handle that or not: `Range(1,2,"[]")` is different from `Range(0,3,"()")`, while the respective `Int4Range`s are equal (I mean, `select '[1,2]'::int4range = '(0,3)'::int4range` yields `True`). If we care, I can easily implement the same logic.

And yes, there is a _difference_ operator for ranges in PG (the last one in [this table](https://www.postgresql.org/docs/current/functions-range.html#RANGE-OPERATORS-TABLE)): if get it right, I'll implement that as well, ok?

zzzeek:
> 
> On `__eq__`/`__ne__`: my question was imprecise, as I knew the dataclass heritage, what I meant was more directed at the different behaviour between `AbstractRange` and `Range`, and whether we care to handle that or not: `Range(1,2,"[]")` is different from `Range(0,3,"()")`, while the respective `Int4Range`s are equal (I mean, `select '[1,2]'::int4range = '(0,3)'::int4range` yields `True`). If we care, I can easily implement the same logic.

We probably should!   On my end I'm just trying to build a pipeline, I've not been a user of this particular PostgreSQL feature so I rely upon the users to tell me what non-obvious behaviors it has.  I think making it be as much a mirror of the database type as possible is always best, do you agree?

> 
> And yes, there is a _difference_ operator for ranges in PG (the last one in [this table](https://www.postgresql.org/docs/current/functions-range.html#RANGE-OPERATORS-TABLE)): if get it right, I'll implement that as well, ok?

yup thanks

lelit:
> I think making it be as much a mirror of the database type as possible is always best, do you agree?

Yes, definitely!

CaselIT:
> 1. `union(self, other)` method, aliased to `__add__`, both on the `Range` class and on the `AbstractRange` one.

wouldn't `|` as alias to union be more appropriate? It's what set and dict do. Of we want to follow PG syntax here?

CaselIT:
> PG uses the "+" operator for union(), so we should use `__add__()`

makes sense, IIRC we already have precedence where we opt for the sql syntax vs the python one

lelit:
Also, I think `|` would be slightly misleading: the `+` operator rejects "non-overlapping-non-adjacent" ranges, that is `[1,2] + [3,4]` is ok, while `[1,2] + (3,4]` is not. There does not seem to be an `|` operator that would map to the `range_merge()` function, that would work in that case.

lelit:
I'm going to push the status quo: I implemented the `union()` and the `difference()` methods, but I'm fighting with the test fixtures... I will need to come up with a better scheme :man_shrugging:

lelit:
Uhm, strange: I expected a series of failures... maybe the CI workflow does not exercise the PG specific tests?
Do you have an hint on where to start investigating on the mypy failure? At first, it looks completely unrelated to my changes...

CaselIT:
> Uhm, strange: I expected a series of failures... maybe the CI workflow does not exercise the PG specific tests?

Yes, only sqlite is run on GitHub. Let me bring it on gerrit

> Do you have an hint on where to start investigating on the mypy failure? At first, it looks completely unrelated to my changes...

That's very likely. Mike mentioned that mypy released a new version yesterday that broke things. Don't worry about them for the moment.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision bde39e92290c256ec1baeef9d5975be4e3c7132b of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change bde39e92290c256ec1baeef9d5975be4e3c7132b: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4182

lelit:
As said, I need a more versatile way of generating different ranges sets to feed the various test functions.

I've come up with the following:

```python
    def _modify_range(
            self,
            r: Range,
            lb: str = None,
            lv: str = None,
            uv: str = None,
            ub: str = None,
            empty=False
    ) -> Range:
        """Take a base range `r` and return an arbitrary transformation.

        The two arguments `lb` and `ub` may be either ``=`` to indicate
        "no change" or one of the bound characters, ``(``, ``[``, ``]``
        and ``)``.

        The two arguments `lv` and `uv` may be either ``∞`` to indicate
        an unbound edge, or an arbitrary expression like ``X + 3*Ɛ``,
        where `X` is the original edge value and `Ɛ` is a small value
        specific to the range type. The expression will be evaluated
        using ``eval()``, with a context containing:

          X
            the original edge value
          L
            the lower edge value
          U
            the upper edge value
          Ɛ
            a small amount of the right type

        When `empty` is True, this simply returns an empty range.
        """

        if empty:
            return Range(None, None, empty=True)
        Ɛ = self._epsilon
        if lb == "=":
            lb = r.bounds[0]
        if ub == "=":
            ub = r.bounds[1]
        if lv == "∞":
            lv = None
        else:
            lv = eval(lv, dict(X=r.lower, L=r.lower, U=r.upper, Ɛ=Ɛ))
        if uv == "∞":
            uv = None
        else:
            uv = eval(uv, dict(X=r.upper, L=r.lower, U=r.upper, Ɛ=Ɛ))
        return Range(lv, uv, bounds=lb + ub)

    @testing.combinations(
        dict(empty=True),
        dict(lb="=", lv="X + Ɛ", uv="X + Ɛ*3", ub="="),
        dict(lb="[", lv="L", uv="L", ub="]"),
        argnames="transform",
    )
    def test_foo(self, transform):
        base = self._data_obj()
        other = self._modify_range(base, **transform)
```

that, given the following mixin:

```python
class _Int4RangeTests:

    _col_type = INT4RANGE
    _col_str = "INT4RANGE"

    def _data_str(self):
        return "[1,4)"

    def _data_obj(self):
        return Range(1, 4)

    _epsilon = 1
```

would generate the following `other` ranges:
```
Range(lower=None, upper=None, bounds='[)', empty=True)
Range(lower=2, upper=7, bounds='[)', empty=False)
Range(lower=1, upper=1, bounds='[]', empty=False)
```

while with the mixin:

```
class _NumRangeTests:

    _col_type = NUMRANGE
    _col_str = "NUMRANGE"

    def _data_str(self):
        return "[1.0,9.0)"

    def _data_obj(self):
        return Range(decimal.Decimal("1.0"), decimal.Decimal("9.0"))

    _epsilon = decimal.Decimal(1)
```

would generate

```
Range(lower=None, upper=None, bounds='[)', empty=True)
Range(lower=Decimal('2.0'), upper=Decimal('12.0'), bounds='[)', empty=False)
Range(lower=Decimal('1.0'), upper=Decimal('1.0'), bounds='[]', empty=False)
```

What's your opinion?

CaselIT:
looking at it seems a bit complex.
Just so I understand correctly, you want a list of parameters that's then used using different kind of ranges with different datatypes, so using something like lambdas where you pass the input range and you return a transformed one can be problematic since the types are different?

lelit:
Basically, yes: Mike already implemented [such mechanism](https://github.com/sqlalchemy/sqlalchemy/blob/main/test/dialect/postgresql/test_types.py#L3879), but it's a bit cumbersome to define those sets with different characteristics for the various test functions.
As an example, this is what I wrote to test the `__eq__()` method, where you want to compare, say, `[1,2]` with `(0,2]` and `(0,3)` and...
```python
    @testing.combinations(
        dict(empty=True),
        dict(lb="[", lv="L", uv="L", ub="]"),
        argnames="r1_transform",
    )
    @testing.combinations(
        dict(empty=True),
        dict(lb="[", lv="L", uv="L", ub="]"),
        dict(lb="[", lv="L", uv="L + Ɛ", ub=")"),
        dict(lb="(", lv="L - Ɛ", uv="L", ub="]"),
        dict(lb="(", lv="L - Ɛ", uv="L + Ɛ", ub=")"),
        argnames="r2_transform",
    )
    def test_equal(self, connection, r1_transform, r2_transform):
        base = self._data_obj()
        r1 = self._modify_range(base, **r1_transform)
        r2 = self._modify_range(base, **r2_transform)

        RANGE = self._col_type
        range_typ = self._col_str

        q = select(
            literal_column(f"'{r1}'::{range_typ} = '{r2}'::{range_typ}")
        )
        equal = connection.execute(q).scalar()
        eq_(
            r1 == r2,
            equal,
            f"{r1} == {r2} != {equal}",
        )
```

Anyway, the last word is obviously yours: if you do not like the above, I will try harder to create different variants of Mike's fixtures.

CaselIT:
using lambda can't this become 
```py
@testing.combinations(
    lambda r, e: Range(None, None, empty=True),
    lambda r, e: Range(r.lower, r.lower, bounds="[]"),
    argnames="r1_transform",
)
@testing.combinations(
    lambda r, e: Range(None, None, empty=True),
    lambda r, e: Range(r.lower, r.lower, bounds="[]"),
    lambda r, e: Range(r.lower, r.lower + e, bounds="[)"),
    lambda r, e: Range(r.lower - e, r.lower, bounds="(]"),
    lambda r, e: Range(r.lower - e, r.lower + e, bounds="()"),
    argnames="r2_transform",
)
```

I find it easier to read.

Are some cases where this would not work?

lelit:
Ok, will do that way, thank you.

CaselIT:
note that it's not set in stone, if some cases require something more complex we can add it. I just like starting simple :)

thanks for working on this!

CaselIT:
Not sure if adding to Range the properties `lb` and `ub` that return the two bounds would be useful?
(or called `lbound`/`ubound` or extended as `lower_bound`/`upper_bound` but at that pind doing `bounds[0]`/`bounds[1]` is probably more convenient)

lelit:
I think this is now ready to be reviewed, the test coverage should be complete now.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 1f930058f9b94d714470ee350baa41d7a6f44b08 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 1f930058f9b94d714470ee350baa41d7a6f44b08 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4182

lelit:
I saw the failures with asyncpg, but I won't be able to dig into the issue until later this evening: a quick look make me wonder about why the `DataError` exception isn't properly catched and handled for example [here](https://github.com/lelit/sqlalchemy/blob/pg-range%2B%2B/test/dialect/postgresql/test_types.py#L4361), as it happens with psycopgX...

lelit:
I think I got it: performing an _invalid_ range operation under `psycopg` raises a `DataError`, while with `asyncpg` a more generic `DBAPIError`. Not sure if that's a problem somewhere else though.

Wrt the exception, in [those cases I raised](https://github.com/lelit/sqlalchemy/blob/pg-range%2B%2B/lib/sqlalchemy/dialects/postgresql/ranges.py#L466) a generic `ValueError`: does that sound correct?

zzzeek:
hi -

have been swamped w/ major stability types of issues and im also not around tomorrow, so hopefully next week I can look at what's happening here more assuming i can get things calmed down.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 84c1ba21d84a08421a9d18399570d8e962168bd4 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 84c1ba21d84a08421a9d18399570d8e962168bd4 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4182

lib/sqlalchemy/dialects/postgresql/ranges.py
```diff
@@ -220,52 +284,287 @@ def contains(self, value: Union[_T, Range]) -> bool:
...
+                and supper_b != "]"
+                and self._compare_edges(oupper, rlower_b, supper, supper_b)
+                == 0
+            ):
+                return Range(None, None, empty=True)
+            else:
+                return Range(oupper, supper, bounds=rlower_b + supper_b)
+
+        # TODO: figure out if I handled all the cases above
+        assert False
```
sqla-tester, line 565: **mike bayer** (zzzeek) wrote:

maybe put a message here.  seems like we dont have a good exception class to raise if we got here.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4182

sqla-tester (commented):

this looks great!


remaining changes:

1. changelog file

2. .... I can do it in a subsequent commit, I think it's time to move the range tests into test/dialect/postgresql/test_ranges.py


View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4182

lib/sqlalchemy/dialects/postgresql/ranges.py
lelit: Thank you!

> changelog file

Should I try to draft one? If so, it is going to `doc/build/changelog/unreleased_20/GH_ISSUE_NR.rst`, right?

>  seems like we dont have a good exception class to raise if we got here

Would `NotImplementedError("Unhandled this and that case")` fit?

zzzeek: if i understand the case correctly, it means this part of the code should be unreachable, so there would be nothing to "implement", right?   lets just make it "assert False, 'unhandled code path'"

zzzeek: re: changelog file yes, just add one of those in and I can fix it up if it needs..thanks!

lelit: I did not find a way to check the changelog markup, in particular I'm not sure about the `:attr:` reference that I copied from existing doc.

zzzeek: no worries, restructured text + sphinx directives took me years to learn, and it's not making things easier we have some of our own conventions buried inside of that

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 21c0df86cc0d1502855527e29425fbffc3f45d64 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 21c0df86cc0d1502855527e29425fbffc3f45d64 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4182

doc/build/changelog/unreleased_20/8765.rst
```diff
@@ -0,0 +1,10 @@
+.. change::
+    :tags: usecase, postgresql
+    :tickets: 8765
+
+    Complementing :ticket:`8690`, new comparison methods such as
```
sqla-tester, line 5: **mike bayer** (zzzeek) wrote:

perfect!

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4182

sqla-tester (commented):

this is done.  A+ !!!

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4182

PR #8768, opened 2022-11-07
"Setting Bi-Directional Many-to-many" fix back_populates tablenames

selobu:


selobu:
This change rst "Setting Bi-Directional Many-to-many" located at https://docs.sqlalchemy.org/en/20/orm/basic_relationships.html#setting-bi-directional-many-to-many

where you use class name instead of tablename as the back_populates parameter in relationship

zzzeek:
hi -

back_populates refers to the name of a relationship-bound attribute on the target class.  in this case "parents" refers to Child.parents, and "children" refers to Parent.children.   back_populates does not refer to table names or class names.

closing this as the documentation is correct as written.

PR #8923, opened 2022-12-02
Update rel_1_4 to support changes to GitHub Actions

jvanasco:
GitHub recently began to roll-out an update to GitHub Actions in which the `ubuntu-latest` Operating System label now points to `ubuntu-22.04` instead of `ubuntu-20.04` :

* https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/
* https://github.blog/changelog/2022-12-01-github-actions-larger-runners-using-ubuntu-latest-label-will-now-use-ubuntu-22-04/

This creates a problem for SQLAlchemy 1.4 which is still under support and maintained under the `rel_1_4` branch, as `ubuntu-22.04` does not support Python2.7 or Python3.6, which are still support targets for the legacy SQLAlchemy release series.

### Description
This PR updates the GitHub actions files to continue support by doing the following:

* any actions targeting the Python 2.7/3.6 environments are extended to use both the `ubuntu-latest` and `ubuntu-20.04` operating systems.
  * the Py27/36 environments are excluded from `ubuntu-latest`
  * the remaining environments are excluded from `ubuntu-20.04`
* some tests have been slightly renamed, to use the test runner's job `name` as a prefix.  this greatly simplifies troubleshooting, by making it easier to correlate a failed action to a job name.

### Checklist
This pull request is:

- [x] A CI infrastructure change.  

Please note: This PR includes an erroneous commit that was pushed `rel_1_4` due to a client/developer error (that was me).  The first commit includes a rollback to undo that commit, as it used a totally different approach.

PR #8924, opened 2022-12-02
Returned "GitHub Actions" support for py27 and py36 environments.

jvanasco:
GitHub recently upgraded the `ubuntu-latest` operating system label to point to `ubuntu-22.04` instead of `ubuntu-22.04`.  The `ubuntu-22.04` build does not support Python 2.7 or 3.6.

In order to return support for CI Testing and release builds, the affected jobs now utilize `include` and `exclude` commands in run matrixes to enable and disable jobs under certain environments.

Jobs that requres Py27 and Py36 now run under an explicitly identified `ubuntu-20.04` operating system. The majority of jobs require Py37 or higher, which are all currently supported in the new `ubuntu-latest` operating system ( which points to `ubuntu-22.04`).

Although `ubuntu-20.04` should continue to receive support for several more years, the `ubuntu-22.04` platform is likely to benefit from patch releases on a faster schedule – so it is preferable to continue running all compatible tests on `ubuntu-latest` rather than pinning everything to the earlier os version.

Several jobs were also standardized to use the job "name" as a prefix, a convention that most jobs in the workflows already adapted. This practice greatly simplifies correlating failed tests to specific jobs.

See::

* https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/
* https://github.blog/changelog/2022-12-01-github-actions-larger-runners-using-ubuntu-latest-label-will-now-use-ubuntu-22-04/

This PR replaces #8923

### DIFF
--- a/.github/workflows/create-wheels.yaml
+++ b/.github/workflows/create-wheels.yaml
@@ -15,7 +15,7 @@ jobs:
   # two jobs are defined make-wheel-win-osx and make-wheel-linux.
   # they do the the same steps, but linux wheels need to be build to target manylinux
   make-wheel-win-osx:
-    name: ${{ matrix.python-version }}-${{ matrix.architecture }}-${{ matrix.os }}
+    name: wheel-win-osx-${{ matrix.python-version }}-${{ matrix.architecture }}-${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


jvanasco:
@CaselIT This is the changeset from #8923 consolidated into a single commit against the git commit just-before I accidentally pushed something to sqlalchemy/rel_1_4. 

I'm not sure why tests are running off this PR. 

`run_test.yaml` is fully tested and passing on that PR.
I'm not sure how to test `run_on_pr.yaml` and `create_wheels.yaml`.

You may have to lightly edit those, but this PR should bring GitHub Actions back online for the 1.4 branch!

.github/workflows/create-wheels.yaml
```diff
@@ -281,12 +302,13 @@ jobs:
           twine upload --skip-existing dist/*manylinux*
 
   make-wheel-linux-arm64:
```
CaselIT, line 304: this was probably ok as it was, since it's using a docker images, ubuntu is just used to run docker, but it should work either way

CaselIT (approved):

CaselIT:
You could have used include to add one off things outside of the matrix (so remove 2.7 and 3.6 from the matrix and add them with ubuntu-20.4 with includes.) I don't particularly like the github syntax here. But it's fine as.

thanks, merging

jvanasco:
I had done that in an earlier iteration, but thought it would be more readable and maintainable to explicitly declare all the test platforms, then limit them.

I’m happy to patch this to your preferred syntax.

CaselIT:
it seems to work, so no need to change it :)

PR #8933, opened 2022-12-05
rename some GitHub actions to use test names as prefix

jvanasco:
This just standardizes the  job `name` into the test `name` as a prefix, so it's easier to tell which job failed

Change-Id: I7f3aa96447ade605becbdef01bc19f3b78dbd6d1


CaselIT (approved):
will merge later

Issue #8984, opened 2022-12-14
New PG Range.__eq__() should not raise an exception comparing with a non-Range value


lelit:
### Describe the bug

As written, the method [Range.__eq__()](https://github.com/sqlalchemy/sqlalchemy/blob/rel_2_0_0b4/lib/sqlalchemy/dialects/postgresql/ranges.py#L287) declares its argument as `other: Any`, but it really expects another instance of the same class, otherwise it raises an exception.

### To Reproduce

```python
>>> from sqlalchemy.dialects.postgresql import Range
>>> r = Range(1,10)
>>> r == None
```

### Error

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/lele/wip/sqlalchemy/lib/sqlalchemy/dialects/postgresql/ranges.py", line 294, in __eq__
    elif self.empty != other.empty:
AttributeError: 'NoneType' object has no attribute 'empty'
``` 

### Versions

- OS: any
- Python: any
- SQLAlchemy: 2.0.0b4+
- Database: irrilevant
- DBAPI (eg: psycopg, cx_oracle, mysqlclient): irrilevant

### Additional context

@CaselIT suggested to return `NotImplemented`, that seems the right thing to do accordingly to the [documentation](https://docs.python.org/3/library/constants.html#NotImplemented).

lelit:
Mike, do I do something wrong when creating the PR linked to an issue, that impedes the automatic close of the latter? Or is it a manual process anyway?

zzzeek:
the commit itself has the issue number, which on github will close the issue, provided it's in the main branch which is the case here.   the PR being closed here, as opposed to "merged", doesn't look as nice, but that's per the current PR <-> gerrit design and should not impact the issue getting closed automatically.

lelit:
Ah, ok, so IIUC it gets closed only once you merge it into next beta...

CaselIT:
I think the issue is that `Closes #id` is not understood by github. You need to use `Fixes: #id`

CaselIT:
no sorry, my mistake. This issue is not closed because it's missing the "closes / fixes #issue-id" . 

The commit message is
```
Fix equality comparison between a PG Range and a different class inst…

…ance

This fixes issue https://github.com/sqlalchemy/sqlalchemy/issues/8984, making the method `Range.__eq__` return `NotImplemented` when the argument is an instance of a different class.

[...]

PR #8985, opened 2022-12-14
Fix equality comparison between a PG Range and a different class instance

lelit:
This fixes issue #8984, making the method `Range.__eq__` return `NotImplemented` when the argument is an instance of a different class.

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/ranges.py
+++ b/lib/sqlalchemy/dialects/postgresql/ranges.py
@@ -146,7 +146,7 @@ class Range(Generic[_T]):
         return AbstractRange()
 
     def _contains_value(self, value: _T) -> bool:
-        """return True if this range contains the given value."""
+        """Return True if this range contains the given value."""
 
         if self.empty:
             return False
@@ -681,7 +681,7 @@ class AbstractRange(sqltypes.TypeEngine[Range[_T]]):
         cls: Type[Union[TypeEngine[Any], TypeEngineMixin]],
         **kw: Any,
     ) -> TypeEngine[Any]:
-        """dynamically adapt a range type to an abstract impl.
+        """Dynamically adapt a range type to an abstract impl.
 
         For example ``INT4RANGE().adapt(_Psycopg2NumericRange)`` should
         produce a type that will have ``_Psycopg2NumericRange`` behaviors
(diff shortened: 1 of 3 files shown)


lib/sqlalchemy/dialects/postgresql/ranges.py
```diff
@@ -289,6 +289,9 @@ def __eq__(self, other: Any) -> bool:  # type: ignore[override]  # noqa: E501
         bounds inclusivity, returning ``True`` if they are equal.
         """
 
+        if not isinstance(other, self.__class__):
```
CaselIT, line 292: to better support possible subclasses of range, maybe we could use `Range` directly there instead of `self.__class__`?

lelit: Yes, but I fail to see the difference... can you explain?

CaselIT: If someone were to subclass Range, then compare it with a Range this check would fail. I mean something like

```py
class MyRange(Range):
  pass

print(MyRange() == Range())
```
should return False since it fails the isinstance check (it may still return True, but that would by python that also calls the inverted eq check )

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 5af687912c3b6eea9d3355c5de32883df58572e2 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 5af687912c3b6eea9d3355c5de32883df58572e2: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4304

lib/sqlalchemy/dialects/postgresql/ranges.py
lelit: I see, thanks. I fixed that.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision b8f601f038a2203af02a99ab190ebbc1c489549a of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset b8f601f038a2203af02a99ab190ebbc1c489549a added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4304

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4304 has been **merged**. Congratulations! :)

Issue #9078, opened 2023-01-10
[2.0] Classes which are generics should document their type parameters


sirosen:
### Describe the use case

I'm working on a test branch for sqlalchemy 2.0 in a codebase which uses `mypy --strict`.

This means that custom subclasses of generic types must provide type parameters (`--disallow-any-generics`).
But it isn't clear what the type parameters are or mean in the 2.0 documentation branch.

### Databases / Backends / Drivers targeted

The target is support for `mypy --strict`, but most particularly the `mypy --disallow-any-generics` setting.

Other type checkers may allow for similar checks, though I'm not aware of the various options.

### Example Use

A long time ago, I included the [example of defining a `utcnow` function](https://docs.sqlalchemy.org/en/latest/core/compiler.html#utc-timestamp-function):
```python
from sqlalchemy.ext.compiler import compiles
from sqlalchemy.sql import expression
from sqlalchemy.types import DateTime

class utcnow(expression.FunctionElement):
    inherit_cache = True
    type = DateTime()

@compiles(utcnow, "postgresql")
def _pg_utcnow(element: t.Any, compiler: t.Any, **kw: t.Any) -> str:
    return "TIMEZONE('utc', CURRENT_TIMESTAMP)"
```

`FunctionElement` is a generic type. So this should be written

```python
class utcnow(expression.FunctionElement[SOME_TYPE_HERE]):
```

And the documentation should make it clear what that type parameter means to a type checker.
Similar usage holds for all other generics.

Right now, in my working branch I've taken a _guess_ that the type parameter for FunctionElement is the type produced in python, and I've annotated as
```python
class utcnow(expression.FunctionElement[datetime.datetime]):
```
but I want to emphasize that this is a guess.

### Additional context

I'm happy to make incremental doc contributions towards this, but I'm not sure where or how such documentation should go.

zzzeek:
yes that's the correct type, when you invoke ``utcnow()`` and get a result back from the ``Result``, the data will be a Python ``datetime`` object, so that's the type.

sirosen:
Awesome, thanks! It's great to have my particular case confirmed!

Where should I propose doc for this? As a user trying to figure this out, I looked at `FunctionElement` itself, as well as the relevant example. But perhaps there's another suitable place, like the migration guide.

zzzeek:
[...]

PR #9079, opened 2023-01-10
Document generic type parameters to FunctionElement and GenericFunction

sirosen:
Add a note to FunctionElement which indicates that the type is a typing.Generic class and points at GenericFunction examples for a specific example usage.

A minimal reference is made to type checkers and IDEs as use-cases in order to try to contextualize this as an optional feature which supports particular use cases.

Append to the GenericFunction examples a case which uses `DateTime` but also includes the generic type parameter (`datetime.datetime`).

---

relates to #9078

### DIFF
--- a/lib/sqlalchemy/sql/functions.py
+++ b/lib/sqlalchemy/sql/functions.py
@@ -1088,7 +1088,7 @@ class GenericFunction(Function[_T]):
     The above indicates that the following expression returns a ``datetime``
     object::
 
-        select(func.as_utc())
+        connection.scalar(select(func.as_utc()))
 
     .. versionadded:: 1.3.13  The :class:`.quoted_name` construct is now
        recognized for quoting when used with the "name" attribute of the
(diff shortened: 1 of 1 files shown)


lib/sqlalchemy/sql/functions.py
```diff
@@ -1072,6 +1077,19 @@ class GeoBuffer(GenericFunction):
...
+    and should match the type seen in a :class:`_engine.Result`. For example::
+
+        class as_utc(GenericFunction[datetime.datetime]):
+            type = DateTime()
+            inherit_cache = True
+
+    The above indicates that the following expression returns a ``datetime``
+    object::
+
+        select(func.as_utc())
```
CaselIT, line 1091: we can probably use `conn.scalar(select(func.as_utc()))` since as is that returns a select

CaselIT (commented):
thanks!

left just a suggestion

lib/sqlalchemy/sql/functions.py
sirosen: Ah, my mistake, good catch; thanks!

I looked to see if `conn` was used elsewhere and I saw use of `connection` in another example in this file, so I wrote it as `connection.scalar(...)` to be consistent within the same file.

CaselIT (approved):
@zzzeek can you take a look?

zzzeek:
this looks great and can go right in, thanks very much!

CaselIT:
thanks!

Issue #4751, opened 2019-07-04
need an 88718-style workaround for table names


zzzeek:
from https://github.com/sqlalchemy/alembic/issues/585 , given a model:

```
from sqlalchemy.ext.declarative import declarative_base
from sqlalchemy import MetaData
from sqlalchemy.schema import Column, ForeignKey
from sqlalchemy.types import CHAR, VARCHAR

convention = {
  "ix": 'ix_%(column_0_label)s',
  "uq": "uq_%(table_name)s_%(column_0_name)s",
  "ck": "ck_%(table_name)s_%(constraint_name)s",
  "fk": "fk_%(table_name)s_%(column_0_name)s_%(referred_table_name)s",
  "pk": "pk_%(table_name)s"
}
meta = MetaData(naming_convention=convention)
Base = declarative_base(metadata=meta)

class Table1(Base):
	__tablename__ = "Table1"
	id = Column(CHAR(36), primary_key=True)
	name = Column(VARCHAR(255))

class Table2(Base):
	__tablename__ = "Table2"
	id = Column(CHAR(36), primary_key=True)
	name = Column(VARCHAR(255), unique=True)
	table1_id = Column(CHAR(36), ForeignKey("Table1.id"))
```

user reports that with mysql 8 on osx the FK spec comes back as:
```

For conn_fk_by_sig:
(None, 'Table2', ('table1_id',), None, 'table1', ('id',), None, None, 'not deferrable')

For metadata_fks_by_sig:
(None, 'Table2', ('table1_id',), None, 'Table1', ('id',), None, None, 'not deferrable')
```

since we haven't tested against OSX or a lower case platform we need to either look into that or start turning off the 88718 fix for greater than 8.0.3, or just turn it off entirely and warn for 8.0.1-> 8.0.3

zzzeek:
I have a mysql 8 running on a container with lower case FS and lower_case_table_names=2.

The issue described in https://bugs.mysql.com/bug.php?id=88718 does not exist; the column name is in the correct case.  However, the table name is not:

```
MySQL [test]> CREATE TABLE `Table_One` (
    ->   `Id` INTEGER PRIMARY KEY
    -> );
Query OK, 0 rows affected (0.058 sec)

MySQL [test]> 
MySQL [test]> CREATE TABLE `table_two` (
    ->   `TABLE_ONE_ID` INTEGER,
[...]

PR #9363, opened 2023-02-24
Added upper bound to mysql version needing bugfix correction

MarkWine:
Provided an update to self._needs_correct_for_88718_96365, adding 8.0.13 as the highest mysql version requiring this correction. Follow up to https://github.com/sqlalchemy/sqlalchemy/issues/4751

### Description

https://bugs.mysql.com/bug.php?id=88718 was fixed in MySQL 8.0.14, so updating the check to reflect that.
When running the tests against a db running in a mysql:8.0.13 docker image, the tests pass, but if the upper limit is set to exclude 8.0.13 from using the correction, test_correct_for_mysql_bugs_88718_96365 and test_case_sensitive_column_constraint_reflection in test_reflection.py fail (as expected). When setting the upper limit to the correct 8.0.13, all of the reflection tests pass against mysql:8.0.12, mysql:8.0.13, mysql:8.0.14, and mysql:latest.

I am a bit confused, though, as it looks like https://bugs.mysql.com/bug.php?id=96365 was active as of at least mysql 8.0.17, but I am unable to reproduce in 8.0.14 and would like some clarity on that part.
Here's what I get using 8.0.14.
```sql
CREATE TABLE `Table_One` (
 `Id` INTEGER PRIMARY KEY );
 
 CREATE TABLE `table_two` (`TABLE_ONE_ID` INTEGER, FOREIGN KEY (`TABLE_ONE_ID`) REFERENCES `Table_One` (`Id`) );
 
 SHOW CREATE TABLE `table_two`;
'CREATE TABLE `table_two` (
  `TABLE_ONE_ID` int(11) DEFAULT NULL,
  KEY `TABLE_ONE_ID` (`TABLE_ONE_ID`),
  CONSTRAINT `table_two_ibfk_1` FOREIGN KEY (`TABLE_ONE_ID`) REFERENCES **`Table_One`** (`Id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci'
```
This looks correct, unlike the output in the bug report, which was:
```sql
CREATE TABLE `table_two` (
  `TABLE_ONE_ID` int(11) DEFAULT NULL,
  KEY `TABLE_ONE_ID` (`TABLE_ONE_ID`),
  CONSTRAINT `table_two_ibfk_1` FOREIGN KEY (`TABLE_ONE_ID`) REFERENCES **`table_one`** (`Id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci
```
Please let me know if there are any other steps to test/reproduce this.

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/mysql/base.py
+++ b/lib/sqlalchemy/dialects/mysql/base.py
@@ -2772,7 +2772,7 @@ class MySQLDialect(default.DefaultDialect):
         )
 
         self._needs_correct_for_88718_96365 = (
-            not self.is_mariadb and self.server_version_info >= (8,)
+            not self.is_mariadb and (8, 0, 13) >= self.server_version_info >= (8,)
         )
 
         self.delete_returning = (


zzzeek:
i would check that the lower_case_table_names=2 on the target platform, it requires a case-insensitive filesystem.  I might have actually created an FAT filesystem and mounted it in order to reproduce that issue.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision d79b99ae8a59ac8b383e8e9d919d6892d5ee5571 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change d79b99ae8a59ac8b383e8e9d919d6892d5ee5571: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4465

zzzeek:
it's very likely I will totally lose track of this if we dont have an issue created with a milestone set up can you do that?  all PRs as noted in the box you checked need an issue number that should be noted as "Fixes:".

MarkWine:
Ok, I saw it as an extension of https://github.com/sqlalchemy/sqlalchemy/issues/4751, but can create a new issue tomorrow. Will also try the FAT filesystem. Just mysql installed on Windows should be easy to test. If that fails, might need to make changes or even just close and wait for full fix from MySQL side.

CaselIT:
> Just mysql installed on Windows should be easy to test. If that fails, might need to make changes or even just close and wait for full fix from MySQL side.

Note that iirc ntfs is case sensitive, it's windows explorer that's not. Not sure if it matters here, just mentioning that maybe only trying on windows using the default ntfs may not be enough to reproduce the issue

zzzeek:
anyway, https://bugs.mysql.com/bug.php?id=96365 is still open, so status on this is we can't skip that part of the workaround.   

what is the urgency to remove / change the workaround until both issues have confirmed upstream fixes established?

zzzeek:
OK so, you were able to reproduce the issue then ?   that's fine, will close the gerrit for now.      I certainly would like to take out this workaroud someday though.

sqla-tester:

issue 96365 is still in play so it was decided it's not worth it to remove the workaround right now.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4465

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4465 has been **abandoned**.  That means that at least for the moment I need to close this pull request. Sorry it didn't work out :(

qpkorr:
Just a vote of support for this, when possible - I'm upgrading from debian buster to bookworm, so moving from sqlalchemy 1.2.18 to 1.4.46, but first, we've gone from mysql 5.7 to 8.0.33.  On 5.7, reflecting our databases used to take about half a second - with mysql8, reflecting those same databases is taking more like 15 seconds (causing us to try to cache the reflections with sqlacodegen instead, which is painful).  I think the issue is that where the SHOW CREATE TABLE command takes about 0.001 secs, the subsequent select from information_schema is taking about 0.3 seconds - which adds up quickly.  I'm guessing you're probably aware of this and there's nothing you can do about it - but the sooner mysql fix 96365 and you can add this upper bound, the better! Thank you!

CaselIT:
> Actually - I then realized that all our tables have lower-case names, so we don't need this fix at all. Still, hacking the (package installed) source code to disable it is clearly ugly, so we look forward to the upper bound when it's possible - or else - perhaps your code could detect from 'show tables' if there's any upper-case table names to worry about?

that's a flag on the engine, so you can just:
```py
engine = create_engine('...mysql/mariadb url...')
# ensure dialect is initialized
engine.connect().close()
engine.dialect._needs_correct_for_88718_96365=False
```

qpkorr:
Awesome, thanks very much, that works perfectly!
Yes, noticed that mariadb not affected - we've complex reasons for not going there, though if bugs like this keep popping up we might need to revisit our opinions

Issue #9380, opened 2023-02-27
Regression: MutableDict.pop(..., None) throws an exception


nphilipp:
### Describe the bug

Using `MutableDict.pop(..., None)` worked with versions 1.4.x, but throws an exception if the key is missing with version 2.0.4. I tripped over this when attempting to upgrade a project using `asyncpg` but this also works with the sync API and the `sqlite` driver (see reproducer).

I’ll submit a PR which should fix this.

### SQLAlchemy Version in Use

2.0.4

### DBAPI (i.e. the database driver)

asyncpg, sqlite

### Database Vendor and Major Version

PostgreSQL 14.3, sqlite 3.40

### Python Version

3.11

### Operating system

Linux

### To Reproduce

```python
#!/usr/bin/env python

# sqla20_mutabledict_regression.py

from sqlalchemy import JSON, Column, Integer, MetaData, create_engine, select
from sqlalchemy.ext.mutable import MutableDict
from sqlalchemy.orm import declarative_base, sessionmaker

engine = create_engine('sqlite://', echo=False)
metadata = MetaData()
Base = declarative_base(metadata=metadata)


class Test(Base):

    __tablename__ = "test"

    id = Column(Integer, primary_key=True)
    data = Column(MutableDict.as_mutable(JSON))


metadata.drop_all(bind=engine)
metadata.create_all(bind=engine)

Session = sessionmaker(bind=engine)
session = Session()
session.add(Test(data={"foo": "bar"}))
session.commit()

test = session.execute(select(Test)).scalars().first()
print(test.data.pop("foo", None))  # -> "bar"
print(test.data.pop("foo", None))  # should be `None` but errors
```

### Error

```
Traceback (most recent call last):
  File "/home/nils/src/fedora-infra/duffy/./sqla20_mutabledict_regression.py", line 30, in <module>
    print(test.data.pop("foo", None))  # should be `None` but errors
          ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nils/.virtualenvs/duffy/lib64/python3.11/site-packages/sqlalchemy/ext/mutable.py", line 827, in pop
    result = super().pop(__key)
             ^^^^^^^^^^^^^^^^^^
KeyError: 'foo'
```

sqla-tester:
**Nils Philippsen** has proposed a fix for this issue in the **main** branch:

[...]

PR #9381, opened 2023-02-27
Allow None as default values for MutableDict

nphilipp:
Previously, e.g. `MutableDict.pop(..., None)` wouldn’t return `None` if the key doesn’t exist. This worked in versions 1.4.x.

### Description

`MutableDict._exists()` uses `None` as a sentinel to check if callers specified a default value or not. However, this doesn’t work if `None` is desired as the default value and the specified key doesn’t exist.

Additionally, this PR removes the unused `MutableDict._is_none()` method.

### Checklist

This pull request is:

- [x] A short code fix for issue #9380 
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/ext/mutable.py
+++ b/lib/sqlalchemy/ext/mutable.py
@@ -389,10 +389,13 @@ from ..orm.context import QueryContext
 from ..orm.decl_api import DeclarativeAttributeIntercept
 from ..orm.state import InstanceState
 from ..orm.unitofwork import UOWTransaction
+from ..sql.base import _NoArg
+from ..sql.base import NO_ARG
 from ..sql.base import SchemaEventTarget
 from ..sql.schema import Column
 from ..sql.type_api import TypeEngine
 from ..util import memoized_property
+from ..util.typing import Literal
 from ..util.typing import SupportsIndex
 from ..util.typing import TypeGuard
 
@@ -781,8 +784,10 @@ class MutableDict(Mutable, Dict[_KT, _VT]):
         super().__setitem__(key, value)
         self.changed()
 
-    def _exists(self, value: _T | None) -> TypeGuard[_T]:
-        return value is not None
+    def _exists(
+        self, value: _T | None | Literal[_NoArg.NO_ARG]
+    ) -> TypeGuard[_T]:
+        return value is not NO_ARG
 
     def _is_none(self, value: _T | None) -> TypeGuard[None]:
         return value is None
@@ -795,7 +800,9 @@ class MutableDict(Mutable, Dict[_KT, _VT]):
     def setdefault(self, key: _KT, value: _VT) -> _VT:
         ...
 
-    def setdefault(self, key: _KT, value: _VT | None = None) -> _VT | None:
+    def setdefault(
+        self, key: _KT, value: _VT | None | Literal[_NoArg.NO_ARG] = NO_ARG
+    ) -> _VT | None:
         if self._exists(value):
             result = super().setdefault(key, value)
         else:
@@ -820,7 +827,11 @@ class MutableDict(Mutable, Dict[_KT, _VT]):
     def pop(self, __key: _KT, __default: _VT | _T) -> _VT | _T:
         ...
 
-    def pop(self, __key: _KT, __default: _VT | _T | None = None) -> _VT | _T:
+    def pop(
+        self,
+        __key: _KT,
+        __default: _VT | _T | None | Literal[_NoArg.NO_ARG] = NO_ARG,
+    ) -> _VT | _T:
         if self._exists(__default):
             result = super().pop(__key, __default)
         else:
--- a/lib/sqlalchemy/ext/mutable.py
+++ b/lib/sqlalchemy/ext/mutable.py
@@ -789,9 +789,6 @@ class MutableDict(Mutable, Dict[_KT, _VT]):
     ) -> TypeGuard[_T]:
         return value is not NO_ARG
 
-    def _is_none(self, value: _T | None) -> TypeGuard[None]:
-        return value is None
-
     @overload
     def setdefault(self, key: _KT) -> _VT | None:
         ...
--- a/test/ext/test_mutable.py
+++ b/test/ext/test_mutable.py
@@ -349,6 +349,19 @@ class _MutableDictTestBase(_MutableDictTestFixture):
 
         eq_(f1.data, {"c": "d"})
 
+    def test_pop_default_none(self):
+        sess = fixture_session()
+
+        f1 = Foo(data={"a": "b", "c": "d"})
+        sess.add(f1)
+        sess.commit()
+
+        eq_(f1.data.pop("a", None), "b")
+        eq_(f1.data.pop("a", None), None)
+        sess.commit()
+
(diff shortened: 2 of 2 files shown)


zzzeek:
hey thanks. Can we assume the regressions occurred when we added typing to mutable?  that was added in 2.0.0b4.    if you can get those two github actions to pass we can send this to gerrit.

lib/sqlalchemy/ext/mutable.py
```diff
@@ -752,6 +754,12 @@ def _listen_for_type(mapper: Mapper[_T], class_: type) -> None:
 
 _setup_composite_listener()
 
+class _NoArg(Enum):
+    NO_ARG = 0
+
+
+_NO_ARG = _NoArg.NO_ARG
```
CaselIT, line 761: We already have this somewhere else, not sure if we should reuse it. 

thoughs @zzzeek ?

nphilipp:
> hey thanks. Can we assume the regressions occurred when we added typing to mutable? that was added in 2.0.0b4.

yep

> if you can get those two github actions to pass we can send this to gerrit.

working on it 🙂

lib/sqlalchemy/ext/mutable.py
nphilipp: Yeah that’s where I stole it from 🫣, I wasn’t sure if I should use something kinda buried in the `sql` package in `ext.mutable`.

nphilipp: Hmm, this also doesn’t let me use `Literal[_NO_ARG]` as a type hint…

CaselIT: >  I wasn’t sure if I should use something kinda buried in the sql package

Probably better to create a new one.

Sooner or later we should do a cleanup to optimize a bit the import time, since with all the inline typing it's probably ballooned quite a bit

nphilipp: Ahh: https://github.com/python/typing/issues/689#issuecomment-561425237

CaselIT: working with python typing is a recipe for frustration sadly, especially after working with typescript

nphilipp: I finally used the one from `sql.base` because it’s used elsewhere, too.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 5286df207398b2d0590a830c48f49f9136484cb0 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 5286df207398b2d0590a830c48f49f9136484cb0: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4479

lib/sqlalchemy/ext/mutable.py
```diff
@@ -389,10 +389,11 @@ def __setstate__(self, state):
 from ..orm.decl_api import DeclarativeAttributeIntercept
 from ..orm.state import InstanceState
 from ..orm.unitofwork import UOWTransaction
-from ..sql.base import SchemaEventTarget
+from ..sql.base import _NoArg, NO_ARG, SchemaEventTarget
```
sqla-tester, line 392: **mike bayer** (zzzeek) wrote:

we have a new linter as of 5 minutes ago that will complain about this :).  if you can rebase to latest main you will want to split these into separate lines.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4479

sqla-tester (commented):

has one linting issue and needs a changelog file.  otherwise looks great!

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4479

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision fd903ce1b949d2af26ceb6c2159ad84aab007f3d of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset fd903ce1b949d2af26ceb6c2159ad84aab007f3d added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4479

lib/sqlalchemy/ext/mutable.py
nphilipp: Latest push should solve this.

sqla-tester:

OK so....I'm looking at this and comparing to what the code was before the _exists() thing was added, in ba0e508141206efc55cdab91df21c1 and, that code is much more obvious.   I'm looking here at _exists() and this seems to be changing the flow of an internal method just...for typing, and then clearly breaking something, so I'm going to try just simplifying this and making the code look like it did before.    the pop() / setdefault() in particular should be trying to look as much like [stdlib](https://github.com/python/typeshed/blob/main/stdlib/typing.pyi#L616) as much as possible, I dont think we shoudl be adding NoArg to the the public API like that.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4479

sqla-tester:

OK I see the idea, that pop() takes one or two arguments and we can't tell which one if we aren't doing just *args.    I'm going to use a TYPE_CHECKING block and restore the old way where *args is one element or two, outside of typing.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4479

sqla-tester:

updated.  can we look at gerrit here to make sure this is right.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4479

test/ext/test_mutable.py
```diff
@@ -349,6 +349,19 @@ def test_pop_default(self):
...
+    def test_pop_default_none(self):
+        sess = fixture_session()
+
+        f1 = Foo(data={"a": "b", "c": "d"})
+        sess.add(f1)
+        sess.commit()
+
+        eq_(f1.data.pop("a", None), "b")
+        eq_(f1.data.pop("a", None), None)
+        sess.commit()
```
sqla-tester, line 361: **Federico Caselli** (CaselIT) wrote:

there is no test for set-default (i think)

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4479

sqla-tester (commented):

lying to the type checkers is sometimes the best option

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4479

sqla-tester (commented):

should be fine as is or with the additional test

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4479

* test/ext/test_mutable.py (line 385): does not try setting None, we could add 

  eq_(f1.data.setdefault("w", None), None)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4479 has been **merged**. Congratulations! :)

PR #9543, opened 2023-03-24
Fix creating zero length char with MySQL dialect

### LINKED ISSUE #9544: mysql.CHAR(0) is created as CHAR(1)
### Describe the bug

When creating schema with `mysql.CHAR(0)` it gets rendering to the datatbase as `CHAR(1)`

https://dev.mysql.com/doc/refman/8.0/en/string-type-syntax.html
> MySQL permits you to create a column of type CHAR(0). This is useful primarily when you must be compliant with old applications that depend on the existence of a column but that do not actually use its value. CHAR(0) is also quite nice when you need a column that can take only two values: A column that is defined as CHAR(0) NULL occupies only one bit and can take only the values NULL and  (the empty string).

_No response_

### SQLAlchemy Version in Use

2.0.6

mysqlclient

Mysql 8

### Python Version

3.10

### Operating system

linux

### To Reproduce

```python
`SPEC = CHAR(1).with_variant(mysql.CHAR(0), "mysql", "mariadb")`
```

### Error

N/A

### Additional context

_No response_

bdraco:
### Description
Fix creating char(0) with MySQL dialect

fixes #9544

This is useful for columns that are no longer used but need to remain in the database while data is still being migrated.

https://dev.mysql.com/doc/refman/8.0/en/string-type-syntax.html
> MySQL permits you to create a column of type CHAR(0). This is useful primarily when you must be compliant with old applications that depend on the existence of a column but that do not actually use its value. CHAR(0) is also quite nice when you need a column that can take only two values: A column that is defined as CHAR(0) NULL occupies only one bit and can take only the values NULL and  (the empty string).

TODO: 
- [x] test case

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**


CaselIT:
thanks for opening.

The tests would go in https://github.com/sqlalchemy/sqlalchemy/blob/main/test/dialect/mysql/test_compiler.py

CaselIT:
The ci on GitHub seems fine with the change. Before running on gerrit,. could you add a change note? Look here for examples https://github.com/sqlalchemy/sqlalchemy/tree/main/doc/build/changelog/unreleased_20

Feel free to add "pull request courtesy of <you>" if you like 

Thanks!

CaselIT:
thanks. the bot that creates the gerrit change is down at the moment, will make it run once that's fixed.

thanks for the changes so far!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision dc17fc3e93f0ba90881c4efb06016ddf83c7af8b of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change dc17fc3e93f0ba90881c4efb06016ddf83c7af8b: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4528

zzzeek:
per MySQL's docs, VARCHAR also accepts 0.  this needs to be for all string types.  will update

zzzeek:
BLOB, TEXT, ...

hope nobody has a zero in their current app ....  (should be fine)

CaselIT:
the doc for char made it sound like it was something special to char, but reading more carefully it seems that the length can be 0 in all of them

> MySQL permits you to create a column of type CHAR(0). This is useful primarily when you must be compliant with old applications that depend on the existence of a column but that do not actually use its value. CHAR(0) is also quite nice when you need a column that can take only two values: A column that is defined as CHAR(0) NULL occupies only one bit and can take only the values NULL and '' (the empty string).

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4528 has been **merged**. Congratulations! :)

bdraco:
Thanks

PR #9614, opened 2023-04-07
Fix docs for `Column.default` (in PyCharm)

david-lev:
In PyCharm there is a problem with displaying the docs when a new line starts with `:class:` or any word that is not standard text.

This is how it looked before the fix:

image

And after:
image

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/sql/schema.py
+++ b/lib/sqlalchemy/sql/schema.py
@@ -1613,12 +1613,11 @@ class Column(DialectKWArgs, SchemaItem, ColumnClause[_T]):
             "fast insertmany" feature.  Such features are very new and
             may not yet be well covered in documentation.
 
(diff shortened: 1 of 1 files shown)


CaselIT:
personally I'm -1 on this simply for the fact that:
- there will most likely me other instances of this
- even if fixing all of them it will be difficult to ensure that they stay fixed or new code does not have this behaviour.

A better solution is to report it to pycharm, since this is a bug on their part

zzzeek:
agree, this is valid RST, displays correctly: 

https://docs.sqlalchemy.org/en/20/core/metadata.html#sqlalchemy.schema.Column.params.default

this is a bug in PyCharm, please report it to them!    it is a product you are paying for after all :)

PR #9616, opened 2023-04-07
Fix incorrect literal value in tutorial documentation

cspollar:
In the [Selecting with Textual Column Expressions](https://docs.sqlalchemy.org/en/20/tutorial/data_select.html#setting-the-columns-and-from-clause) I noticed the example and the description text differ.

### Description

The explanation text uses `some label`:
>  such as below where we manufacture a hardcoded string literal 'some label'

But the example uses `some phrase` instead

```python
>>> from sqlalchemy import text
>>> stmt = select(text("'some phrase'"),
```

Please ignore if I'm reading this wrong.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**


CaselIT:
thanks!

PR #9643, opened 2023-04-12
Fix `RowMapping`'s `Mapping` type to reflect that it supports `Column`s or strings

### LINKED ISSUE #9644: `RowMapping.get()` is typed as only supporting string keys, not `Column`s

### Confirm this is not Mypy dataclasses issue 13856

- [X] This is not related to Mypy issue 13856

### Describe the typing issue

`RowMapping.__getitem__` supports `str` and `Column` keys, but `RowMapping.get()` is typed as only supporting `str` keys.

### To Reproduce

```python

from sqlalchemy import Column, Integer, MetaData, Table, select
from sqlalchemy.engine import create_engine

metadata = MetaData()


t_user = Table("user", metadata, Column("user_id", Integer, primary_key=True))


    result = conn.execute(select(Table)).first()
    assert result is not None
    result._mapping.get(t_user.c.user_id)
```

### Error

```
t.py:17: error: No overload variant of "get" of "Mapping" matches argument type "Column[Any]"  [call-overload]
t.py:17: note: Possible overload variants:
t.py:17: note:     def get(self, str, /) -> Optional[Any]
t.py:17: note:     def [_T] get(self, str, /, default: Union[Any, _T]) -> Union[Any, _T]
Found 1 error in 1 file (checked 1 source file)
``` 

### Versions

- OS: macOS 13.3.1
- Python: 3.11.1
- SQLAlchemy: 2.0.9
- Type checker (eg: mypy 0.991, pyright 1.1.290, etc): mypy 1.0.1 (compiled: yes)

### Additional context

_No response_

rouge8:
### Description

I ran into this originally in sqlalchemy2-stubs: https://github.com/sqlalchemy/sqlalchemy2-stubs/pull/251, where `RowMapping` only supported string keys according to the type hints. I ran into a similar issue here upgrading our application where because `RowMapping` subclassed `Mapping[str, Any]`, `Row._mapping.get()` would fail to typecheck when used with `Column` objects.

This patch adds a test to verify that `Row._mapping.get()` continues to work with both strings and `Column`s, though it doesn't look like mypy checks types in the tests.

Fixes #9644.

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/engine/row.py
+++ b/lib/sqlalchemy/engine/row.py
@@ -271,9 +271,11 @@ class ROMappingView(ABC):
     __slots__ = ()
 
     _items: Sequence[Any]
-    _mapping: Mapping[str, Any]
+    _mapping: Mapping["_KeyType", Any]
 
-    def __init__(self, mapping: Mapping[str, Any], items: Sequence[Any]):
+    def __init__(
+        self, mapping: Mapping["_KeyType", Any], items: Sequence[Any]
+    ):
         self._mapping = mapping
         self._items = items
 
@@ -297,16 +299,16 @@ class ROMappingView(ABC):
 
 
 class ROMappingKeysValuesView(
-    ROMappingView, typing.KeysView[str], typing.ValuesView[Any]
+    ROMappingView, typing.KeysView["_KeyType"], typing.ValuesView[Any]
 ):
     __slots__ = ("_items",)
 
 
-class ROMappingItemsView(ROMappingView, typing.ItemsView[str, Any]):
+class ROMappingItemsView(ROMappingView, typing.ItemsView["_KeyType", Any]):
     __slots__ = ("_items",)
 
 
-class RowMapping(BaseRow, typing.Mapping[str, Any]):
+class RowMapping(BaseRow, typing.Mapping["_KeyType", Any]):
     """A ``Mapping`` that maps column names and objects to :class:`.Row`
     values.
 
--- a/test/ext/mypy/plain_files/typed_results.py
+++ b/test/ext/mypy/plain_files/typed_results.py
@@ -8,7 +8,10 @@ from sqlalchemy import column
 from sqlalchemy import create_engine
 from sqlalchemy import insert
 from sqlalchemy import Integer
+from sqlalchemy import MetaData
 from sqlalchemy import select
+from sqlalchemy import String
+from sqlalchemy import Table
 from sqlalchemy import table
 from sqlalchemy.ext.asyncio import AsyncConnection
 from sqlalchemy.ext.asyncio import AsyncSession
@@ -29,6 +32,14 @@ class User(Base):
 
     id: Mapped[int] = mapped_column(primary_key=True)
     name: Mapped[str]
+
+
+t_user = Table(
+    "user",
+    MetaData(),
+    Column("id", Integer, primary_key=True),
+    Column("name", String),
+)
 
 
 e = create_engine("sqlite://")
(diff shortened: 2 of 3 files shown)


zzzeek:
we would super appreciate an issue being opened first, thanks

CaselIT:
> though it doesn't look like mypy checks types in the tests.

type checks are done on likes in this file (that are only type checked, never run with python) https://github.com/sqlalchemy/sqlalchemy/tree/main/test/ext/mypy/plain_files

you can probably add a few lines into https://github.com/sqlalchemy/sqlalchemy/blob/main/test/ext/mypy/plain_files/typed_results.py

rouge8:
> we would super appreciate an issue being opened first, thanks

Done, and updated the commit message / PR description!

CaselIT:
@rouge8 the change looks ok, can you just add a few lines to the typing file mentioned above?

@zzzeek I think _KeyType should allow also for orm columns. At the moment it's just `str | Column` that I think will type error when using `Model.orm_col`.

Also do we support this case?
```py
opcol = tbl.c.col + 5
row = conn.execute(select(opcol)).mappings().first()
row[opcol]
# or even
row[tbl.c.col + 5]
```

zzzeek:
> 
> @zzzeek I think _KeyType should allow also for orm columns. At the moment it's just `str | Column` that I think will type error when using `Model.orm_col`.

do we support using ORM attributes a result map keys?    We've moved very far away from the "use columns as result map keys" use case, we of course use it internally but not via the regular row here, we have a private method that gets us the mappings the ORM needs.

however if the case works then yes we'd have to open this up for one of the open ended "column argument" types, that is , if it's actually unpacking ``__clause_element__()`` and all that which i think it would have to be doing for ORM columns to work here.

> 
> Also do we support this case?
> 
> ```python
> opcol = tbl.c.col + 5
> row = conn.execute(select(opcol)).mappings().first()
> row[opcol]
> # or even
> row[tbl.c.col + 5]
> ```

no, that wouldn't make any sense

CaselIT:
> We've moved very far away from the "use columns as result map keys" use case,

we are talking about the row mapping here, not normal Rows that are index only or using getattr.
Currently the public method is to only use string in getitem, 

The usage with columns is public, so it's not a private thing:
https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.RowMapping
`print("Column b: %s" % row._mapping[table.c.b])`

I guess we can try to see if the current version works with the ORM, and if so we can maybe add a test and update the type.
What would the appropriate type here? `SQLORMExpression` or `QueryableAttribute` seem to general.
What may be problematic is that an orm column_property can map a single table column or an expression, like `col + 5` above. Can we allow the first and not the other? (assuming `col + 5` does not work that is)

rouge8:
> @rouge8 the change looks ok, can you just add a few lines to the typing file mentioned above?

Done! I'm not super familiar with the ORM (especially the 2.0 ORM) but I can either try to implement what you come up with or you can push to my branch.

rouge8:
Hmm, it looks like it might also support `ColumnElement`s, at least on 1.4. I'm seeing this error on 2.0 and I _think_ it's a bug in the type hints, unless it's a behavior change I haven't noticed yet: `error: Invalid index type "ColumnElement[Any]" for "RowMapping"; expected type "Union[str, Column[Any]]"  [index]`

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 6c33fe534cf457d6b5c73f4830a64880830f0f56 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 6c33fe534cf457d6b5c73f4830a64880830f0f56: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4576

PR #9668, opened 2023-04-18
Prebuild the row string to position lookup for Rows

### LINKED ISSUE #9678: Optimize Row index lookup by pre-building the row-index dict
See details at [9668](https://github.com/sqlalchemy/sqlalchemy/pull/9668)

bdraco:
Merge checklist

- [ ] the row and result getstate should accept the current 2.0 serialization even after these changes. It's fine to change the serialization so the new format can't be loaded in older 2.0 versions (best if we don't).

### Description

While I've tried to switch all the performance sensitive places in Home Assistant to use `getitem`, there is some code that can't be easily refactored that deals with large datasets and orm objects. This is a small speed up to the `getattr` implementation.

`python3 test/perf/compiled_extensions.py BaseRow`

Python 3.11.2 Darwin 22.4.0 Darwin Kernel Version 22.4.0: Mon Mar  6 20:59:58 PST 2023; root:xnu-8796.101.5~3/RELEASE_ARM64_T6020 arm64

Before
```
                    | python      | cython      | cy / py     |
getattr             | 1.440994000 | 0.899872834 | 0.624480625 |
getattr_repeats     | 24.36904329 | 14.56666258 | 0.597752748 |
```
After
```
                    | python      | cython      | cy / py     |
getattr             | 1.113157333 | 0.816836375 | 0.733801369 |
getattr_repeats     | 18.37744700 | 13.53884533 | 0.736709802 |
```

~24.58% speed up pure python
~7.05% speed up cython

Python 3.10.5 Linux 4.4.59+ #25556 SMP PREEMPT Sat Aug 28 02:17:25 CST 2021 x86_64 Linux

Before
```
                    | python      | cython      | cy / py     |
getattr             | 5.148779137 | 2.282829123 | 0.443372897 |
getattr_repeats     | 85.49036448 | 34.70639851 | 0.405968541 |
```

After

```
                    | python      | cython      | cy / py     |
getattr             | 3.705751788 | 2.123954093 | 0.573150663 |
getattr_repeats     | 59.90764300 | 31.99583192 | 0.534085975 |
```

~29.92% speed up pure python
~7.80% speed up cython

### Checklist

Fixes #9678

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/cyextension/resultproxy.pyx
+++ b/lib/sqlalchemy/cyextension/resultproxy.pyx
@@ -10,6 +10,7 @@ KEY_OBJECTS_ONLY = _KEY_OBJECTS_ONLY
 
 cdef class BaseRow:
     cdef readonly object _parent
+    cdef readonly dict _name_cache
     cdef readonly tuple _data
     cdef readonly dict _keymap
     cdef readonly int _key_style
@@ -18,6 +19,8 @@ cdef class BaseRow:
         """Row objects are constructed by CursorResult objects."""
 
         self._parent = parent
+
+        self._name_cache = parent._name_cache
 
         if processors:
             self._data = tuple(
@@ -47,10 +50,12 @@ cdef class BaseRow:
         }
 
     def __setstate__(self, dict state):
-        self._parent = state["_parent"]
+        parent = state["_parent"]
+        self._parent = parent
         self._data = state["_data"]
-        self._keymap = self._parent._keymap
+        self._keymap = parent._keymap
         self._key_style = state["_key_style"]
+        self._name_cache = parent._name_cache
 
     def _values_impl(self):
         return list(self)
@@ -68,6 +73,13 @@ cdef class BaseRow:
         return self._data[index]
 
     cpdef _get_by_key_impl_mapping(self, key):
+        return self._real_get_by_key_impl_mapping(key)
+
+    cdef _real_get_by_key_impl_mapping(self, key):
+        cached_index = self._name_cache.get(key)
+        if cached_index is not None:
+            return self._data[cached_index]
+
         try:
             rec = self._keymap[key]
         except KeyError as ke:
@@ -82,13 +94,14 @@ cdef class BaseRow:
         ):
             raise KeyError(key)
 
+        self._name_cache[key] = mdindex
         return self._data[mdindex]
 
     def __getattr__(self, name):
         try:
-            return self._get_by_key_impl_mapping(name)
+            return self._real_get_by_key_impl_mapping(name)
         except KeyError as e:
-           raise AttributeError(e.args[0]) from e
+            raise AttributeError(e.args[0]) from e
 
 
 def rowproxy_reconstructor(cls, state):
--- a/lib/sqlalchemy/cyextension/resultproxy.pyx
+++ b/lib/sqlalchemy/cyextension/resultproxy.pyx
@@ -2,25 +2,19 @@
 # more improvements are likely possible
 import operator
 
-cdef int MD_INDEX = 0  # integer index in cursor.description
-cdef int _KEY_OBJECTS_ONLY = 1
-
-KEY_INTEGER_ONLY = 0
-KEY_OBJECTS_ONLY = _KEY_OBJECTS_ONLY
+cdef object _MISSING_SENTINEL = object()
 
 cdef class BaseRow:
     cdef readonly object _parent
-    cdef readonly dict _name_cache
+    cdef readonly dict _keymap_by_str
     cdef readonly tuple _data
-    cdef readonly dict _keymap
-    cdef readonly int _key_style
 
-    def __init__(self, object parent, object processors, dict keymap, int key_style, object data):
+    def __init__(self, object parent, object processors, object data):
         """Row objects are constructed by CursorResult objects."""
 
         self._parent = parent
 
-        self._name_cache = parent._name_cache
+        self._keymap_by_str = parent._keymap_by_str
 
         if processors:
             self._data = tuple(
@@ -32,10 +26,6 @@ cdef class BaseRow:
         else:
             self._data = tuple(data)
 
-        self._keymap = keymap
-
-        self._key_style = key_style
-
     def __reduce__(self):
         return (
             rowproxy_reconstructor,
@@ -46,16 +36,13 @@ cdef class BaseRow:
         return {
             "_parent": self._parent,
             "_data": self._data,
-            "_key_style": self._key_style,
         }
 
     def __setstate__(self, dict state):
         parent = state["_parent"]
         self._parent = parent
         self._data = state["_data"]
-        self._keymap = parent._keymap
-        self._key_style = state["_key_style"]
-        self._name_cache = parent._name_cache
+        self._keymap_by_str = parent._keymap_by_str
 
     def _values_impl(self):
         return list(self)
@@ -73,35 +60,23 @@ cdef class BaseRow:
         return self._data[index]
 
     cpdef _get_by_key_impl_mapping(self, key):
-        return self._real_get_by_key_impl_mapping(key)
-
-    cdef _real_get_by_key_impl_mapping(self, key):
-        cached_index = self._name_cache.get(key)
-        if cached_index is not None:
-            return self._data[cached_index]
-
-        try:
-            rec = self._keymap[key]
-        except KeyError as ke:
-            rec = self._parent._key_fallback(key, ke)
-
-        mdindex = rec[MD_INDEX]
-        if mdindex is None:
-            self._parent._raise_for_ambiguous_column_name(rec)
-        elif (
-            self._key_style == _KEY_OBJECTS_ONLY
-            and isinstance(key, int)
-        ):
-            raise KeyError(key)
-
-        self._name_cache[key] = mdindex
-        return self._data[mdindex]
+        cached_index = self._keymap_by_str.get(key, _MISSING_SENTINEL)
+        if cached_index is not _MISSING_SENTINEL and cached_index is not None:
+            return self._data[cached_index]        
+        if cached_index is _MISSING_SENTINEL:
+            self._parent._key_fallback(key, KeyError(key))
+        self._parent._raise_for_ambiguous_column_name(self._parent._keymap[key])
 
     def __getattr__(self, name):
-        try:
-            return self._real_get_by_key_impl_mapping(name)
-        except KeyError as e:
-            raise AttributeError(e.args[0]) from e
+        cached_index = self._keymap_by_str.get(name, _MISSING_SENTINEL)
+        if cached_index is not _MISSING_SENTINEL and cached_index is not None:
+            return self._data[cached_index]
+        if cached_index is _MISSING_SENTINEL:
+            try:
+                self._parent._key_fallback(name, KeyError(name))
+            except KeyError as e:
+                raise AttributeError(e.args[0]) from e
+        self._parent._raise_for_ambiguous_column_name(self._parent._keymap[name])
 
 
 def rowproxy_reconstructor(cls, state):
--- a/lib/sqlalchemy/cyextension/resultproxy.pyx
+++ b/lib/sqlalchemy/cyextension/resultproxy.pyx
@@ -2,11 +2,9 @@
 # more improvements are likely possible
 import operator
 
-cdef object _MISSING_SENTINEL = object()
-
 cdef class BaseRow:
     cdef readonly object _parent
-    cdef readonly dict _keymap_by_str
+    cdef readonly dict _key_to_index
     cdef readonly tuple _data
 
     def __init__(self, object parent, object processors, object data):
@@ -14,7 +12,7 @@ cdef class BaseRow:
 
         self._parent = parent
 
-        self._keymap_by_str = parent._keymap_by_str
+        self._key_to_index = parent._key_to_index
 
         if processors:
             self._data = tuple(
@@ -42,7 +40,7 @@ cdef class BaseRow:
         parent = state["_parent"]
         self._parent = parent
         self._data = state["_data"]
-        self._keymap_by_str = parent._keymap_by_str
+        self._key_to_index = parent._key_to_index
 
     def _values_impl(self):
         return list(self)
@@ -59,24 +57,17 @@ cdef class BaseRow:
     def __getitem__(self, index):
         return self._data[index]
 
-    cpdef _get_by_key_impl_mapping(self, key):
-        cached_index = self._keymap_by_str.get(key, _MISSING_SENTINEL)
-        if cached_index is not _MISSING_SENTINEL and cached_index is not None:
-            return self._data[cached_index]        
-        if cached_index is _MISSING_SENTINEL:
-            self._parent._key_fallback(key, KeyError(key))
-        self._parent._raise_for_ambiguous_column_name(self._parent._keymap[key])
+    def _get_by_key_impl_mapping(self, key):
+        return self._get_by_key_impl(key, 0)
+
+    cdef _get_by_key_impl(self, object key, int attr_err):
+        index = self._key_to_index.get(key)
+        if index is not None:
+            return self._data[<int>index]
+        self._parent._key_not_found(key, attr_err != 0)
 
     def __getattr__(self, name):
-        cached_index = self._keymap_by_str.get(name, _MISSING_SENTINEL)
-        if cached_index is not _MISSING_SENTINEL and cached_index is not None:
-            return self._data[cached_index]
-        if cached_index is _MISSING_SENTINEL:
-            try:
-                self._parent._key_fallback(name, KeyError(name))
-            except KeyError as e:
-                raise AttributeError(e.args[0]) from e
-        self._parent._raise_for_ambiguous_column_name(self._parent._keymap[name])
+        return self._get_by_key_impl(name, 1)
 
 
 def rowproxy_reconstructor(cls, state):
--- a/lib/sqlalchemy/engine/_py_row.py
+++ b/lib/sqlalchemy/engine/_py_row.py
@@ -15,6 +15,7 @@ from typing import Union
 
 if typing.TYPE_CHECKING:
     from .result import _KeyMapType
+    from .result import _KeyMapRecType
     from .result import _KeyType
     from .result import _ProcessorsType
     from .result import _RawRowType
@@ -38,9 +39,10 @@ KEY_INTEGER_ONLY, KEY_OBJECTS_ONLY = list(_KeyStyle)
 
 
 class BaseRow:
-    __slots__ = ("_parent", "_data", "_keymap", "_key_style")
+    __slots__ = ("_parent", "_data", "_keymap", "_key_style", "_name_cache")
 
     _parent: ResultMetaData
+    _name_cache: Dict[_KeyType, _KeyMapRecType]
     _data: _RawRowType
     _keymap: _KeyMapType
     _key_style: _KeyStyle
@@ -55,6 +57,8 @@ class BaseRow:
     ):
         """Row objects are constructed by CursorResult objects."""
         object.__setattr__(self, "_parent", parent)
+
+        object.__setattr__(self, "_name_cache", parent._name_cache)
 
         if processors:
             object.__setattr__(
@@ -93,6 +97,7 @@ class BaseRow:
         object.__setattr__(self, "_data", state["_data"])
         object.__setattr__(self, "_keymap", parent._keymap)
         object.__setattr__(self, "_key_style", state["_key_style"])
+        object.__setattr__(self, "_name_cache", parent._name_cache)
 
     def _values_impl(self) -> List[Any]:
         return list(self)
@@ -113,6 +118,10 @@ class BaseRow:
         __getitem__ = _get_by_int_impl
 
     def _get_by_key_impl_mapping(self, key: _KeyType) -> Any:
+        cached_index = self._name_cache.get(key)
+        if cached_index is not None:
+            return self._data[cached_index]
+
         try:
             rec = self._keymap[key]
         except KeyError as ke:
@@ -127,6 +136,7 @@ class BaseRow:
         elif self._key_style == KEY_OBJECTS_ONLY and isinstance(key, int):
             raise KeyError(key)
 
+        self._name_cache[key] = mdindex
         return self._data[mdindex]
 
     def __getattr__(self, name: str) -> Any:
--- a/lib/sqlalchemy/engine/_py_row.py
+++ b/lib/sqlalchemy/engine/_py_row.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import enum
 import operator
 import typing
 from typing import Any
@@ -14,7 +13,6 @@ from typing import Type
 from typing import Union
 
 if typing.TYPE_CHECKING:
-    from .result import _KeyMapType
     from .result import _KeyMapRecType
     from .result import _KeyType
     from .result import _ProcessorsType
@@ -24,41 +22,26 @@ if typing.TYPE_CHECKING:
 
 MD_INDEX = 0  # integer index in cursor.description
 
-
-class _KeyStyle(enum.IntEnum):
-    KEY_INTEGER_ONLY = 0
-    """__getitem__ only allows integer values and slices, raises TypeError
-    otherwise"""
-
-    KEY_OBJECTS_ONLY = 1
-    """__getitem__ only allows string/object values, raises TypeError
-    otherwise"""
-
-
-KEY_INTEGER_ONLY, KEY_OBJECTS_ONLY = list(_KeyStyle)
+_MISSING_SENTINEL = object()
 
 
 class BaseRow:
-    __slots__ = ("_parent", "_data", "_keymap", "_key_style", "_name_cache")
+    __slots__ = ("_parent", "_data", "_keymap_by_str")
 
     _parent: ResultMetaData
-    _name_cache: Dict[_KeyType, _KeyMapRecType]
+    _keymap_by_str: Dict[_KeyType, _KeyMapRecType]
     _data: _RawRowType
-    _keymap: _KeyMapType
-    _key_style: _KeyStyle
 
     def __init__(
         self,
         parent: ResultMetaData,
         processors: Optional[_ProcessorsType],
-        keymap: _KeyMapType,
-        key_style: _KeyStyle,
         data: _RawRowType,
     ):
         """Row objects are constructed by CursorResult objects."""
         object.__setattr__(self, "_parent", parent)
 
-        object.__setattr__(self, "_name_cache", parent._name_cache)
+        object.__setattr__(self, "_keymap_by_str", parent._keymap_by_str)
 
         if processors:
             object.__setattr__(
@@ -74,10 +57,6 @@ class BaseRow:
         else:
             object.__setattr__(self, "_data", tuple(data))
 
-        object.__setattr__(self, "_keymap", keymap)
-
-        object.__setattr__(self, "_key_style", key_style)
-
     def __reduce__(self) -> Tuple[Callable[..., BaseRow], Tuple[Any, ...]]:
         return (
             rowproxy_reconstructor,
@@ -88,16 +67,13 @@ class BaseRow:
         return {
             "_parent": self._parent,
             "_data": self._data,
-            "_key_style": self._key_style,
         }
 
     def __setstate__(self, state: Dict[str, Any]) -> None:
         parent = state["_parent"]
         object.__setattr__(self, "_parent", parent)
         object.__setattr__(self, "_data", state["_data"])
-        object.__setattr__(self, "_keymap", parent._keymap)
-        object.__setattr__(self, "_key_style", state["_key_style"])
-        object.__setattr__(self, "_name_cache", parent._name_cache)
+        object.__setattr__(self, "_keymap_by_str", parent._keymap_by_str)
 
     def _values_impl(self) -> List[Any]:
         return list(self)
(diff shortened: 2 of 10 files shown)


CaselIT:
I wonder if creating this cache upfront would make sense? Have you tried to see if it does make a difference?

test/perf/compiled_extensions.py
```diff
@@ -996,6 +996,82 @@ def getattr(self):
         self.row_long.x
         self.row_long.y
 
+    @test_case
+    def getattr_repeats(self):
+        row = self.row
+        row.a
```
CaselIT, line 1002: I think a plain for here would accomplish the same since Python has no optimization on a for

lib/sqlalchemy/cyextension/resultproxy.pyx
```diff
@@ -68,6 +73,13 @@ cdef class BaseRow:
         return self._data[index]
 
     cpdef _get_by_key_impl_mapping(self, key):
+        return self._real_get_by_key_impl_mapping(key)
+
+    cdef _real_get_by_key_impl_mapping(self, key):
```
CaselIT, line 78: does this make a lot of difference compared just using the cpdef fn?

CaselIT:
I think it may be worth a try. 

> didn't want to make the use cases where the dataset is small and only one or two out of many columns are accessed.

well I think that's not a common case, since it does not make much sense to select 20 columns if only one is needed, especially if a developer care about performance. and if they don't as long as it's not a lot (say ~ 20% ) slower it should be fine

lib/sqlalchemy/cyextension/resultproxy.pyx
bdraco: That shouldn't be the bulk of the improvement.  Getting rid of the `try` helps more I think

`cythonize -X language_level=3 -a -i  lib/sqlalchemy/cyextension/resultproxy.pyx`

Before:
[resultproxy.html.zip](https://github.com/sqlalchemy/sqlalchemy/files/11259648/resultproxy.html.zip)

After:
[resultproxy.html.zip](https://github.com/sqlalchemy/sqlalchemy/files/11259642/resultproxy.html.zip)

bdraco:
> I think it may be worth a try.
> 
> > didn't want to make the use cases where the dataset is small and only one or two out of many columns are accessed.
> 
> well I think that's not a common case, since it does not make much sense to select 20 columns if only one is needed, especially if a developer care about performance. and if they don't as long as it's not a lot slower (say ~ 20% ) it should be fine

Thats true. And they likely don't really care for the tiny cases anyways.

I'll give that a spin tomorrow as its getting late late here.

CaselIT:
Thanks!

@zzzeek since we are at it, do you have other ideas that may be worth trying in this area?

zzzeek:
I looked only briefly and the ratio of code complexity to performance gain under cython does not seem worth it here, my initial reaction, which note is subject to change, is that I really don't like adding more code to resultproxy / row here, caches, all of that is a more risk / cognitive overhead when there are things to fix / maintenance / more difficult to add new features etc.      

I would need to look closely at what the "getattr" and "getattr_repeats" case are testing.      the database row is overall a transitory object that should be thrown away after being read just once.

CaselIT:
The main thing added by this patch is a cache of key -> index in data. This avoids having to do each time key -> record -> index

zzzeek:
OK I see here this is really a view of the existing _keymap and it is carried along in the same place, just that it's built up on a per-access basis rather than all at once.

IMO it would be better to build up such a cache all at once, at the same time that _keymap is built up.  As it is, a multithreaded program would have a lot of threads all putting the same data into _name_cache as they each fail to find the key and the do a set operation in the row.      _keymap already has all these keys.

So I don't like building up the name cache in the row.  if we had such a cache, it should be built up just once after we build keymap in `CursorResultMetaData.__init__`.    then all the other methods here that are digging into _name_cache with get() (which is more expensive than dict indexed access), finding nothing, falling back, looking again, etc. could be simplified to only be looking in _name_cache.

next, _keymap **is** the cache.  why is it not working?  it's because of these checks:

```

        mdindex = rec[MD_INDEX]
        if mdindex is None:
            self._parent._raise_for_ambiguous_column_name(rec)
        # NOTE: keep "== KEY_OBJECTS_ONLY" instead of "is KEY_OBJECTS_ONLY"
        # since deserializing the class from cython will load an int in
        # _key_style, not an instance of _KeyStyle
        elif self._key_style == KEY_OBJECTS_ONLY and isinstance(key, int):
            raise KeyError(key)
```

it's that isinstance() check and the _key_style check that is making it slow.   

So lets get rid of **that** part.  I'd prefer that instead of making _keymap up front, we make two keymaps, _keymap_by_int and _keymap_by_str.     this is basically the same thing as what this would do, if _name_cache were built up front.

So steps:

1. call _name_cache _keymap_by_str
2. built it up front in CursorResultMetaData right along with _keymap
3. change the name-based accessors to look directly in _keymap_by_str and that's it.
4. look to remove KEY_INTEGER_ONLY and KEY_OBJECTS_ONLY entirely.  These are not needed anymore as these came from the legacy model that included KEY_OBJECTS_BUT_WARN and KEY_OBJECTS_NO_WARN; those modes don't exist anymore so the refactoring here should just continue.

That way we remove complexity from the getters while maintaining (and improving) the performance gains.

bdraco:
> look to remove KEY_INTEGER_ONLY and KEY_OBJECTS_ONLY entirely. These are not needed anymore as these came from the legacy model that included KEY_OBJECTS_BUT_WARN and KEY_OBJECTS_NO_WARN; those modes don't exist anymore so the refactoring here should just continue.

Do we want to maintain compat for that? I guess we could load the int values into the dict if it's set?

zzzeek:
> 
> Edit: re-reading I think based on your comment it's safe to remove support for those

correct, those constants are from the 1.4 version of this that was straddling two different APIs

bdraco:
Great. Will do another turn. Timing subject to ✈️ delays

CaselIT:
Thanks!

I agree that it makes sense to simplify that part since it's now not really needed.

bdraco:
That approach is faster.  I have not tried to optimize the cython code yet though. I wrote what I thought would be faster though

`python3 test/perf/compiled_extensions.py BaseRow`

Darwin 22.4.0 Darwin Kernel Version 22.4.0: Mon Mar  6 20:59:58 PST 2023; root:xnu-8796.101.5~3/RELEASE_ARM64_T6020 arm64

Before
```
                    | python      | cython      | cy / py     |
getattr             | 1.440994000 | 0.899872834 | 0.624480625 |
getattr_repeats     | 24.36904329 | 14.56666258 | 0.597752748 |
```
After
```
                    | python      | cython      | cy / py     |
getattr             | 1.113157333 | 0.816836375 | 0.733801369 |
getattr_repeats     | 18.37744700 | 13.53884533 | 0.736709802 |
```

~24.58% speed up pure python
~7.05% speed up cython

bdraco:
I think thats as fast as it can get

Screenshot 2023-04-18 at 11 46 53 AM

bdraco:
Linux 4.4.59+ #25556 SMP PREEMPT Sat Aug 28 02:17:25 CST 2021 x86_64 Linux

Before
```
                    | python      | cython      | cy / py     |
getattr             | 5.148779137 | 2.282829123 | 0.443372897 |
getattr_repeats     | 85.49036448 | 34.70639851 | 0.405968541 |
```

After

```
                    | python      | cython      | cy / py     |
getattr             | 3.705751788 | 2.123954093 | 0.573150663 |
getattr_repeats     | 59.90764300 | 31.99583192 | 0.534085975 |
```

~29.92% speed up pure python
~7.80% speed up cython

bdraco:
I'm going to do some more manual testing now

bdraco:
Performance difference is now enough that its noticeable with profiling on page load times with big queries now

CaselIT:
I think we can further improve the happy case by moving more work the error case

- _keymap_by_str only has non none keys
- we to a `try / except` with return `self._data[self._keymap_by_str[key]]` (not sure if cython likes more the try/except or the explicit get + if)
- in case of keyerror we call a single parent fallback that does check if index is none or not and dispatched
- probably the parent thing should have a flag indicating if we want to raise an attibute error, so `__getattr__` can use the same code

bdraco:
> * _keymap_by_str only has non none keys
> * we to a `try / except` with return `self._data[self._keymap_by_str[key]]` (not sure if cython likes more the try/except or the explicit get + if)
> * in case of keyerror we call a single parent fallback that does check if index is none or not and dispatches to the correct error
> * probably the parent fallback should have a flag indicating if we want to raise an attribute error, so `__getattr__` can use the same code

I expect `try`/`except` pattern will be much more expensive in `cython`
Screenshot 2023-04-18 at 12 10 07 PM

CaselIT:
what version of python are you using to test?
IIRC 3.11 changed things quite a bit regarding the non expection case of try/expect, making it almost zero cost in python. No clue if cython was adapted for 3.11 or not

bdraco:
> what version of python are you using to test? IIRC 3.11 changed things quite a bit regarding the non expection case of try/expect, making it almost zero cost in python. No clue if cython was adapted for 3.11 or not

cython code is the same on 3.11 vs 3.10 AFAICT

Python 3.10.5 x86_64
Python 3.11.2 arm64 mac

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 46039ec963ad7ee669a773478864fac41bfb2bb5 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 46039ec963ad7ee669a773478864fac41bfb2bb5: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4575

zzzeek:
fun stuff I'm running it on gerrit.

I like how a lot of code comes out of row.py and related.   still seems to be a net increase in lines of code but complexity is moved to single spots in cursor.py, moved out of the row logic

bdraco:
Old stats for reference. Going to update the opening text

```
`python3 test/perf/compiled_extensions.py BaseRow`

Darwin 22.4.0 Darwin Kernel Version 22.4.0: Mon Mar  6 20:59:58 PST 2023; root:xnu-8796.101.5~3/RELEASE_ARM64_T6020 arm64

Before
```
                    | python      | cython      | cy / py     |
getattr             | 1.440994000 | 0.899872834 | 0.624480625 |
getattr_repeats     | 24.36904329 | 14.56666258 | 0.597752748 |
```
After
```
                    | python      | cython      | cy / py     |
getattr             | 1.291914583 | 0.840211125 | 0.650361205 |
getattr_repeats     | 21.92209570 | 14.01316120 | 0.639225437 |
```

~10% speed up pure python
~3.8% speed up cython



Linux 4.4.59+ #25556 SMP PREEMPT Sat Aug 28 02:17:25 CST 2021 x86_64 Linux

Before
```
                    | python      | cython      | cy / py     |
getattr             | 5.148779137 | 2.282829123 | 0.443372897 |
getattr_repeats     | 85.49036448 | 34.70639851 | 0.405968541 |
```

After

```
                    | python      | cython      | cy / py     |
getattr             | 4.491845477 | 2.203217393 | 0.490492695 |
getattr_repeats     | 74.81916330 | 33.27543702 | 0.444744843 |
```

~12.48% speed up pure python
~4.12% speed up cython
```

CaselIT:
I'll try some ideas later today

CaselIT:
I've simplified things a bit more.
Performance is a little better compared to the current patch in python, but not by much.

Current version `python ./test/perf/compiled_extensions.py BaseRow`:
```
                    | python      | cython      | cy / py     |
get_by_key          | 0.482758800 | 0.195342800 | 0.404638507 |
getattr             | 1.365923900 | 0.995643600 | 0.728915864 |
get_by_key_recreate | 2.275293100 | 1.432292200 | 0.629497888 |
getattr_recreate    | 4.693909800 | 3.511730600 | 0.748146162 |
```
Previous version:
```
                    | python      | cython      | cy / py     |
get_by_key          | 0.520290200 | 0.212461500 | 0.408351916 |
getattr             | 1.432946000 | 1.001736800 | 0.699075052 |
get_by_key_recreate | 2.292925000 | 1.393086800 | 0.607558817 |
getattr_recreate    | 4.886627000 | 3.522538600 | 0.720852768 |
```
Main:
```
                    | python      | cython      | cy / py     |
get_by_key          | 0.634946500 | 0.253666700 | 0.399508778 |
getattr             | 1.791868800 | 1.143251400 | 0.638021824 |
get_by_key_recreate | 2.510199800 | 1.405295100 | 0.559833962 |
getattr_recreate    | 5.854872000 | 3.743805800 | 0.639434269 |
```

Current / Main:     python 0.834, cython 0.938
Current / Previous: python 0.960, cython 0.996

lib/sqlalchemy/engine/result.py
```diff
@@ -177,6 +179,29 @@ def _row_as_tuple_getter(
...
+            for key, rec in keymap.items()
+            if rec[index] is not None
+        }
+
+    def _key_not_found(self, key: Any, attr_error: bool) -> NoReturn:
+        if key in self._keymap:
+            # the index must be none in this case
+            if attr_error:
+                try:
+                    self._key_fallback(key, KeyError(key))
```
CaselIT, line 196: note to self: remove keyerror from the key_fallback, since it does not make much sense now

bdraco:
Testing now

lib/sqlalchemy/cyextension/resultproxy.pyx
```diff
@@ -67,28 +57,17 @@ cdef class BaseRow:
...
-        ):
-            raise KeyError(key)
+    def _get_by_key_impl_mapping(self, key):
+        return self._get_by_key_impl(key, 0)
 
-        return self._data[mdindex]
+    cdef _get_by_key_impl(self, object key, int attr_err):
+        index = self._key_to_index.get(key)
+        if index is not None:
+            return self._data[<int>index]
```
bdraco, line 66:
```suggestion
            return self._data[index]
```

I think we can drop the `<int>` conversion in the cython

Screenshot 2023-04-19 at 12 10 53 PM

as it means we have to __Pyx_PyInt_As_int to convert it back to the python int since __Pyx_GetItemInt_Tuple expects a python int

Screenshot 2023-04-19 at 12 10 57 PM

bdraco: testing that now to see if its faster or slower because `Pyx_GetItemInt_Tuple` might still be better even with another conversion

bdraco: `<int>` does seem faster.

I'm wondering if we can avoid the `PyInt_as_int` but I guess not

CaselIT: yes, but I wanted the pyx_getitem_tuple instead of the pyx_pyobject_getitem

not sure if there is a way of making it use that without the cast

CaselIT: doing `self._data[int(index)]` make it use `Pyx_GetItemInt_Tuple` but it seems slower

bdraco: The int conversion effective boils it down to `PyTuple_GET_ITEM` with `cpython`

bdraco: `int` seems better 👍

bdraco: I don't see anything else we can squeeze out of this

CaselIT: indeed. I don't know if there is a way of making it do "cast to python int" instead of a c-int.

lib/sqlalchemy/cyextension/resultproxy.pyx
```diff
@@ -2,23 +2,18 @@
 # more improvements are likely possible
```
CaselIT, line 2: I guess we could remove this todo now :)

CaselIT:
> Testing now

thanks!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 86b8ccd1959dbd91b1208f7a648a91f217e1f866 of this pull request into gerrit so we can run tests and reviews and stuff

lib/sqlalchemy/cyextension/resultproxy.pyx
CaselIT: > I don't see anything else we can squeeze out of this

yes, whatever you do it seems to have no effect or make things worse.

Overall I think conceptually there isn't anything left to optimize since we basically do:
- get the index from the dict
- if we find an index return the thing from the tuple
- otherwise raise

not sure if using a linear probe in case of a small row is worth trying

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 86b8ccd1959dbd91b1208f7a648a91f217e1f866 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 86b8ccd1959dbd91b1208f7a648a91f217e1f866 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4575

CaselIT:
uh gerrit took me as the author. Will change that tomorrow.

@zzzeek I think you can have another look to see if there is something more to remove

lib/sqlalchemy/cyextension/resultproxy.pyx
```diff
@@ -2,23 +2,18 @@
...
-    cdef readonly dict _keymap
-    cdef readonly int _key_style
 
-    def __init__(self, object parent, object processors, dict keymap, int key_style, object data):
+    def __init__(self, object parent, object processors, object data):
         """Row objects are constructed by CursorResult objects."""
 
         self._parent = parent
 
+        self._key_to_index = parent._key_to_index
```
CaselIT, line 15: Maybe this dict could be passed from the outside as before to avoid a getattr for each row

lib/sqlalchemy/cyextension/resultproxy.pyx
bdraco: Was looking at the Py_None checks but it looks like they don't matter https://github.com/cython/cython/issues/3797

lib/sqlalchemy/cyextension/resultproxy.pyx
bdraco: I think they would help 
Screenshot 2023-04-19 at 12 59 46 PM

bdraco: We end up with

```diff
--- a/lib/sqlalchemy/cyextension/resultproxy.pyx
+++ b/lib/sqlalchemy/cyextension/resultproxy.pyx
@@ -7,12 +7,12 @@ cdef class BaseRow:
     cdef readonly dict _key_to_index
     cdef readonly tuple _data
 
-    def __init__(self, object parent, object processors, object data):
+    def __init__(self, object parent, object processors, dict key_to_index, object data):
         """Row objects are constructed by CursorResult objects."""
 
         self._parent = parent
 
-        self._key_to_index = parent._key_to_index
+        self._key_to_index = key_to_index
 
         if processors:
             self._data = tuple(
```

We get the `__Pyx_ArgTypeTest` check when passed in instead of `PyDict_CheckExact` but that should be neutral
Screenshot 2023-04-19 at 1 01 45 PM

but not getattr which is a win
Screenshot 2023-04-19 at 1 01 52 PM

CaselIT: did that but it has no measurable impact

CaselIT:
> Just a note so we don't forget once we have a final patch
> 
>     * [x]  the row and result getstate should accept the current 2.0 serialization even after these changes. It's fine to change the serialization so the new format can't be loaded in older 2.0 versions (best if we don't).

To recap, after this patch it will be possible to load old pickles but older sqlalchemy version will *not* be able to load the version coming from this patch.

To retain compact not much would be needed, since it would be enough to hardcode a `_key_style` in `__getstate__`. @zzzeek do you thing it makes sense doing? And if so what value should we set there?

bdraco:
I did some more testing and found that sometimes processors are `[None, None, None, None, None, None]` which means each one is doing the first branch needlessly AFAICT

```
        if processors:
            self._data = tuple(
                [
                    proc(value) if proc else value
                    for proc, value in zip(processors, data)
                ]
            )
        else:
            self._data = tuple(data)
```

bdraco:
```diff
diff --git a/lib/sqlalchemy/engine/cursor.py b/lib/sqlalchemy/engine/cursor.py
index 779f9e9d3..3af6c6d22 100644
--- a/lib/sqlalchemy/engine/cursor.py
+++ b/lib/sqlalchemy/engine/cursor.py
@@ -1454,6 +1454,8 @@ class CursorResult(Result[_T]):
             metadata = self._init_metadata(context, cursor_description)
 
             processors = metadata._processors
+            if processors and all(p is None for p in processors):
+                processors = None
             process_row = Row
             _make_row = functools.partial(process_row, metadata, processors)
             if log_row:
diff --git a/lib/sqlalchemy/engine/result.py b/lib/sqlalchemy/engine/result.py
index 6558efdd4..b0d9874d7 100644
--- a/lib/sqlalchemy/engine/result.py
+++ b/lib/sqlalchemy/engine/result.py
@@ -463,6 +463,8 @@ class ResultInternal(InPlaceGenerative, Generic[_R]):
         metadata = self._metadata
 
         processors = metadata._processors
+        if processors and all(p is None for p in processors):
+            processors = None
         tf = metadata._tuplefilter
 
         if tf and not real_result._source_supports_scalars:
```

Maybe something like this.  But should it get another PR after this one?

CaselIT:
there is code that does this, maybe in a new patch. It's doing `set(processors) == NONE_SET`

sqla-tester:

> Patch Set 4:
> 
> CaselIT@github wrote:
> 
> there is code that does this, maybe in a new patch. It's doing `set(processors) == NONE_SET`

it's NONE_SET.superset(my_list_of_things)

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4575

sqla-tester:

NONE_SET.issuperset(mylist)

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4575

CaselIT:
Right, you catch also the an empty list in that case

CaselIT:
let me try adding it

CaselIT:
Updated the gerrit patch

CaselIT:
> I did some more testing and found that sometimes processors are `[None, None, None, None, None, None]` which means each one is doing the first branch needlessly AFAICT
> 
> ```
>         if processors:
>             self._data = tuple(
>                 [
>                     proc(value) if proc else value
>                     for proc, value in zip(processors, data)
>                 ]
>             )
>         else:
>             self._data = tuple(data)
> ```

turns out we can't do that since the sentinel code that was just merged depends on it to filter the extra columns.

@zzzeek shouldn't this use the tuplefilter feature of the result instead of the processors?

zzzeek:
> > I did some more testing and found that sometimes processors are `[None, None, None, None, None, None]` which means each one is doing the first branch needlessly AFAICT
> > ```
> >         if processors:
> >             self._data = tuple(
> >                 [
> >                     proc(value) if proc else value
> >                     for proc, value in zip(processors, data)
> >                 ]
> >             )
> >         else:
> >             self._data = tuple(data)
> > ```
> 
> turns out we can't do that since the sentinel code that was just merged depends on it to filter the extra columns.

I dont know where that code is.  the sentinel code doesnt do anything with bind or result processors.


> 
> @zzzeek shouldn't this use the tuplefilter feature of the result instead of the processors?

CaselIT:
A lot of tests started failing because the returned rows are `(1,2,<sentinel-value>)` instead of `(1,2)`.

This is because the processors in this case are `(None,None)` so the last value of the original tuple is ignored by the zip. If in this patch we remove these procerrors since they are all none, the zip branch is no longer taken and we get in the case above.

Handling the sentinel this way is not very efficient since a tuple slice is a lot faster in this case.

I think we have a better solution to handle this case in the results, that's called tuplefilters. Do you think the sentinel code could use it?

Otherwise still using the processors for this we can improve this part by passing a slice in the case above, but we need to know the number of sentinel there are in a row. Can the result metadata know that?

CaselIT:
In any case it's probably best to move this change to a follow up patch

zzzeek:
oh the rows themselves are somehow involved.  Well this is that thing we talked about, where we decided to not trim the rows down for sentinels in _deliver_insertmanyvalues_batches.  We can re-add that trimming there

CaselIT:
I think it makes sense doing it here. I'll try some things here (well in another patch really)

zzzeek:
> I think it makes sense doing it here. I'll try some things here (well in another patch really)

ok.  but keep in mind, trimming the rows is never needed in any case, except exactly for the one where we get the rows from deliver_exeutemany_values().   really seems like if it adds expense, it should be localized to the place that it happens. 

unless you want to not rely on the rows being exactly the same length as the description, which you probably also could do

CaselIT:
Ok a recap.

I've moved the processor patch to a new changeset, https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4575 should be ready to review.
There is only one thing left to decide and that's what I asked here https://github.com/sqlalchemy/sqlalchemy/pull/9668#issuecomment-1516869104


The patch of the processor is at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4581 
I've done also some other optimizations, and things improved quite a bit both with and without processors passes.

bdraco:
I did some more testing with this.  The case where `if processors:` is false is a dramatic improvement for some of Home Assistant's logbook cases.  Response time is roughly cut in half.

lib/sqlalchemy/engine/_py_row.py
```diff
@@ -84,15 +64,13 @@ def __getstate__(self) -> Dict[str, Any]:
         return {
             "_parent": self._parent,
```
sqla-tester, line 65: **mike bayer** (zzzeek) wrote:

can confirm (visually) that pickle backwards compat is maintained here.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4575

lib/sqlalchemy/engine/cursor.py
```diff
@@ -919,8 +906,8 @@ def __getstate__(self):
     def __setstate__(self, state):
         self._processors = [None for _ in range(len(state["_keys"]))]
         self._keymap = state["_keymap"]
-
         self._keymap_by_result_column_idx = None
+        self._key_to_index = self._make_key_to_index(self._keymap, MD_INDEX)
```
sqla-tester, line 910: **mike bayer** (zzzeek) wrote:

good this maintains pickle compat

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4575

lib/sqlalchemy/engine/result.py
```diff
@@ -177,6 +182,29 @@ def _row_as_tuple_getter(
         indexes = self._indexes_for_keys(keys)
         return tuplegetter(*indexes)
 
+    def _make_key_to_index(
+        self, keymap: Mapping[_KeyType, Sequence[Any]], index: int
+    ) -> Mapping[_KeyType, int]:
+        return {
+            key: rec[index]
```
sqla-tester, line 189: **mike bayer** (zzzeek) wrote:

great, this is great

seemed like there was something more complicated here previously...

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4575

sqla-tester (commented):

just want to run locally to see what we get

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4575

test/perf/compiled_extensions.py
```diff
@@ -996,6 +985,40 @@ def getattr(self):
...
+            l_row._get_by_key_impl_mapping("o")
+            l_row._get_by_key_impl_mapping("r")
+            l_row._get_by_key_impl_mapping("t")
+            l_row._get_by_key_impl_mapping("y")
+            l_row._get_by_key_impl_mapping("t")
+            l_row._get_by_key_impl_mapping("w")
+            l_row._get_by_key_impl_mapping("o")
+
+    @test_case(number=50_000)
+    def getattr_recreate(self):
```
zzzeek, line 1006: is this the same as "getattr_repeats" mentioned in the change message?  why are they called "recreate" ?

test/perf/compiled_extensions.py
```diff
@@ -996,6 +985,40 @@ def getattr(self):
...
+        self.init_objects()
+        row = self.row
```
sqla-tester, line 1008: **Federico Caselli** (CaselIT) wrote:

> is this the same as "getattr_repeats" mentioned in the change message?  

no, that was an older version that was replaced with this. The commit message should probably be updated.

> why are they called "recreate" 

because I've re-creating the rows in the method, not using a cached one.
(the other methods do not call init_objects())

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4575

sqla-tester (commented):

Ok things to do are:
- update commit message
- either mention in the changelog that pickle format changed or decide a fixed key style

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4575

lib/sqlalchemy/engine/_py_row.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

so pickle compat is as follows:

older v2 pickle loads fine after this patch.
pickle saved in this patch does not load in older v2 code.

The issue is that we removed the key_style.

I think it's fine to keep the current logic (but maybe we mention it in the changelog)
Otherwise we could save a fixed number for the _key_style?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4575

CaselIT:
> I'm not 100% what the status of this PR is. Please let me know if anything is waiting for me.

nope, just busy last week with a relevant fix. It should get merged this week.

thanks for profiling and providing the changes! it turned out there were quite a few of optimizations available

sqla-tester (commented):

commit message and changelog updated. Can you take a final look mike?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4575

* doc/build/changelog/unreleased_20/9678.rst (line 8): Done
* lib/sqlalchemy/engine/cursor.py (line 1462): Ack
* lib/sqlalchemy/engine/result.py (line 501): Done

test/perf/compiled_extensions.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4575

lib/sqlalchemy/engine/result.py
sqla-tester, line 189: **mike bayer** (zzzeek) wrote:

simple

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4575

sqla-tester (commented):

it's great, net reduction in code lines, complexity, great

basically gets us totally off the 1.4 version of things.  this was a goal of 2.0, reduce internal complexity.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4575

* doc/build/changelog/unreleased_20/9678.rst (line 9): +1

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4575 has been **merged**. Congratulations! :)

bdraco:
Really excited about the combination of https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4581 and https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4577 as its going to make a huge different for Home Assistant.  🎆 

Many thanks!

daimor:
Tried to upgrade my dialect to 2.0, finished for 2.0.10, and just switched to the latest 2.0.11 and boom nothing works anymore
```
INTERNALERROR>     all_objects = self._get_all_objects(
INTERNALERROR>   File "/Users/daimor/Projects/sqlalchemy-iris/sqlalchemy_iris/base.py", line 1072, in _get_all_objects
INTERNALERROR>     rows = result.fetchall()
INTERNALERROR>   File "/Users/daimor/Projects/sqlalchemy/lib/sqlalchemy/engine/result.py", line 1331, in fetchall
INTERNALERROR>     return self._allrows()
INTERNALERROR>   File "/Users/daimor/Projects/sqlalchemy/lib/sqlalchemy/engine/result.py", line 560, in _allrows
INTERNALERROR>     made_rows = [make_row(row) for row in rows]
INTERNALERROR>   File "/Users/daimor/Projects/sqlalchemy/lib/sqlalchemy/engine/result.py", line 560, in <listcomp>
INTERNALERROR>     made_rows = [make_row(row) for row in rows]
INTERNALERROR>   File "lib/sqlalchemy/cyextension/resultproxy.pyx", line 17, in sqlalchemy.cyextension.resultproxy.BaseRow.__init__
INTERNALERROR> TypeError: __init__() takes exactly 5 positional arguments (4 given)
```

CaselIT:
Please open a new issue

CaselIT:
oh, that seems an editable install in a sqlalchemy checkout.
@daimor You will need to rebuild the compiled extension locally, since if you just update the git checkout they are not re-built

daimor:
I did only `pip install -e .` after checkout to `rel_2_0_11` and got this, how to rebuild that compiled extension?

CaselIT:
The pip install should do that. Maybe try uninstalling and re-installing sqlalchemy.

Alternatively manually remove the built extensions files in `lib/sqlalchemy/cyextension`

daimor:
thanks, `git clean -fxd` and `pip install -e .` actually helped

PR #9688, opened 2023-04-21
Fixes #9687: Add 'arm64' to greenlet dependency.

### LINKED ISSUE #9687: greenlet dependency not installed on arm64.
### Describe the bug

On M1 Macs, the `greenlet` dependency is not installed because `arm64` is not is in `setup.cfg`.  

When using build tools in CI pipelines that validate generated requirements lock files are equal, the requirement lock files generated on `arm64` (from a local laptop) and `x84_64` (from CI pipelines) differ and tests fail.

When lock files are generated from `arm64` machines, tests fail in CI pipelines. When lock files are generated from `x86_64` machines, tests fail on `arm64`.

_No response_

### SQLAlchemy Version in Use

2.0.10

N/A

N/A

### Python Version

3.10

### Operating system

OSX

### To Reproduce

```python
N/A
```

### Error

From Bazel from `arm64` machine:
```
Repository '@pip_greenlet' is not defined and referenced by '@pip_sqlalchemy//:pkg'
``` 

### Additional context

_No response_

ryan-clancy:
### Description
- Adds arm64 to the greenlet dependency so requirements lock files generated on arm64 machines (e.g., M1 Mac) is equal to requirements lock from x86_64 machines.

Fixes #9687.

### Checklist
This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,7 +37,7 @@ package_dir =
 
 install_requires =
     importlib-metadata;python_version<"3.8"
-    greenlet != 0.4.17;(platform_machine=='aarch64' or (platform_machine=='ppc64le' or (platform_machine=='x86_64' or (platform_machine=='amd64' or (platform_machine=='AMD64' or (platform_machine=='win32' or platform_machine=='WIN32'))))))
+    greenlet != 0.4.17;(platform_machine=='arm64' or (platform_machine=='aarch64' or (platform_machine=='ppc64le' or (platform_machine=='x86_64' or (platform_machine=='amd64' or (platform_machine=='AMD64' or (platform_machine=='win32' or platform_machine=='WIN32')))))))
     typing-extensions >= 4.2.0
 
 [options.extras_require]


zzzeek:
hey there -

there is no arm64 wheel at https://pypi.org/project/greenlet/#files.   We cannot enable any archs for which there isn't a wheel file available for greenlet.  that's the whole reason that byzantine list of requirements is there.

is there any reason you can't use the `sqlalchemy[asyncio]` target as [documented](https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#asyncio-install) ?

the PR submission instructions weren't followed here so closing

zzzeek:
oh you made an issue, never mind

CaselIT:
As mentioned in the issue, I don't think it's advisable to add this by default, since it seems that it's still need to compile greenly on arm mac and that requires the compiler chain to be installed.

Thanks in any case.

We will revisit this decision if greenlet starts creating wheels for arm mac

Issue #8240, opened 2022-07-11
Support NULLS [NOT] DISTINCT in unique indexes and unique constraint


CaselIT:
Currently supported in pg15+ but it's supposedly part of the sql spec: https://news.ycombinator.com/item?id=32054151

https://www.postgresql.org/docs/15/ddl-constraints.html#DDL-CONSTRAINTS-UNIQUE-CONSTRAINTS

Assuming it's sql standard we could have a `null_distinct` equal to `True` to render `NULLS DISTINCT` and `False` to render `NULLS NOT DISTINCT`. The default `None` should not render anything.
If it's implemented for only for postgresql, a flag `postgresql_nulls_not_distinct=True` could render `NULLS NOT DISTINCT`

I would guess that reflection and alembic support would also be needed

zzzeek:
the feature is called F292, see https://modern-sql.com/caniuse/F292

rodrigoestevao:
Will this feature be added on version 1.4 too?

CaselIT:
> Will this feature be added on version 1.4 too?

Unlikely, since v2 is nearing release and this is not currently planned for 2.0.0

I guess if PR are provided before release of 2.0 we could consider backporing to 1.4

sqla-tester:
**pavelserchenia** has proposed a fix for this issue in the **main** branch:

**PG nulls not distinct support** https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4632

network-mapping-dev:
I am trying to use this option with postgres using the async driver but I get a syntax error on the Unique constraint where NULLS NOT DISTINCT is inserted:

E   asyncpg.exceptions.PostgresSyntaxError: syntax error at or near "NULLS"

The above exception was the direct cause of the following exception:
E   sqlalchemy.dialects.postgresql.asyncpg.AsyncAdapt_asyncpg_dbapi.ProgrammingError: <class 'asyncpg.exceptions.PostgresSyntaxError'>: syntax error at or near "NULLS"

The above exception was the direct cause of the following exception:
E   sqlalchemy.exc.ProgrammingError: (sqlalchemy.dialects.postgresql.asyncpg.ProgrammingError) <class 'asyncpg.exceptions.PostgresSyntaxError'>: syntax error at or near "NULLS"
    [SQL: 
    CREATE TABLE aar.rating_scheme (
    	transmission_line_id INTEGER NOT NULL, 
[...]

PR #9834, opened 2023-05-24
pg nulls not distinct support

Bulbashful:
### Description
Adds support for pg 15 nulls not distinct feature https://github.com/sqlalchemy/sqlalchemy/issues/8240. 
There is also changes for alembic https://github.com/sqlalchemy/alembic/pull/1249
https://www.postgresql.org/about/featurematrix/detail/392/

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -2280,12 +2280,11 @@ class PGDDLCompiler(compiler.DDLCompiler):
                 [preparer.quote(c.name) for c in inclusions]
             )
 
-        if self.dialect._supports_nulls_not_distinct:
-            nulls_not_distinct = index.dialect_options["postgresql"]["nulls_not_distinct"]
-            if nulls_not_distinct is True:
-                text += " NULLS NOT DISTINCT"
-            elif nulls_not_distinct is False:
-                text += " NULLS DISTINCT"
+        nulls_not_distinct = index.dialect_options["postgresql"]["nulls_not_distinct"]
+        if nulls_not_distinct is True:
+            text += " NULLS NOT DISTINCT"
+        elif nulls_not_distinct is False:
+            text += " NULLS DISTINCT"
 
         withclause = index.dialect_options["postgresql"]["with"]
         if withclause:
@@ -2320,12 +2319,11 @@ class PGDDLCompiler(compiler.DDLCompiler):
         if len(constraint) == 0:
             return ""
         nulls_not_distinct_option = ""
-        if self.dialect._supports_nulls_not_distinct:
-            nulls_not_distinct = constraint.dialect_options["postgresql"]["nulls_not_distinct"]
-            if nulls_not_distinct is True:
-                nulls_not_distinct_option = "NULLS NOT DISTINCT "
-            elif nulls_not_distinct is False:
-                nulls_not_distinct_option = "NULLS DISTINCT "
+        nulls_not_distinct = constraint.dialect_options["postgresql"]["nulls_not_distinct"]
+        if nulls_not_distinct is True:
+            nulls_not_distinct_option = "NULLS NOT DISTINCT "
+        elif nulls_not_distinct is False:
+            nulls_not_distinct_option = "NULLS DISTINCT "
         text = ""
         if constraint.name is not None:
             formatted_name = self.preparer.format_constraint(constraint)
@@ -3039,7 +3037,6 @@ class PGDialect(default.DefaultDialect):
     _backslash_escapes = True
     _supports_create_index_concurrently = True
     _supports_drop_index_concurrently = True
-    _supports_nulls_not_distinct = False
 
     def __init__(self, json_serializer=None, json_deserializer=None, **kwargs):
         default.DefaultDialect.__init__(self, **kwargs)
@@ -3060,8 +3057,6 @@ class PGDialect(default.DefaultDialect):
             2,
         )
         self.supports_identity_columns = self.server_version_info >= (10,)
-        # https://www.postgresql.org/about/featurematrix/detail/392/
-        self._supports_nulls_not_distinct = self.server_version_info >= (15,)
 
     def get_isolation_level_values(self, dbapi_conn):
         # note the generic dialect doesn't have AUTOCOMMIT, however
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -2280,7 +2280,9 @@ class PGDDLCompiler(compiler.DDLCompiler):
                 [preparer.quote(c.name) for c in inclusions]
             )
 
-        nulls_not_distinct = index.dialect_options["postgresql"]["nulls_not_distinct"]
+        nulls_not_distinct = index.dialect_options["postgresql"][
+            "nulls_not_distinct"
+        ]
         if nulls_not_distinct is True:
             text += " NULLS NOT DISTINCT"
         elif nulls_not_distinct is False:
@@ -2314,12 +2316,13 @@ class PGDDLCompiler(compiler.DDLCompiler):
 
         return text
 
-
     def visit_unique_constraint(self, constraint, **kw):
         if len(constraint) == 0:
             return ""
         nulls_not_distinct_option = ""
-        nulls_not_distinct = constraint.dialect_options["postgresql"]["nulls_not_distinct"]
+        nulls_not_distinct = constraint.dialect_options["postgresql"][
+            "nulls_not_distinct"
+        ]
         if nulls_not_distinct is True:
             nulls_not_distinct_option = "NULLS NOT DISTINCT "
         elif nulls_not_distinct is False:
@@ -2331,7 +2334,7 @@ class PGDDLCompiler(compiler.DDLCompiler):
                 text += "CONSTRAINT %s " % formatted_name
         text += "UNIQUE %s(%s)" % (
             nulls_not_distinct_option,
-            ", ".join(self.preparer.quote(c.name) for c in constraint)
+            ", ".join(self.preparer.quote(c.name) for c in constraint),
         )
         text += self.define_constraint_deferrability(constraint)
         return text
@@ -2998,7 +3001,7 @@ class PGDialect(default.DefaultDialect):
                 "concurrently": False,
                 "with": {},
                 "tablespace": None,
-                "nulls_not_distinct": None
+                "nulls_not_distinct": None,
             },
         ),
         (
@@ -3026,9 +3029,7 @@ class PGDialect(default.DefaultDialect):
         ),
         (
             schema.UniqueConstraint,
-            {
-                "nulls_not_distinct": None
-            },
+            {"nulls_not_distinct": None},
         ),
     ]
 
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -2226,6 +2226,7 @@ class PGDDLCompiler(compiler.DDLCompiler):
         text = "CREATE "
         if index.unique:
             text += "UNIQUE "
+
         text += "INDEX "
 
         if self.dialect._supports_create_index_concurrently:
@@ -2279,6 +2280,13 @@ class PGDDLCompiler(compiler.DDLCompiler):
                 [preparer.quote(c.name) for c in inclusions]
             )
 
+        if self.dialect._supports_nulls_not_distinct:
+            nulls_not_distinct = index.dialect_options["postgresql"]["nulls_not_distinct"]
+            if nulls_not_distinct is True:
+                text += " NULLS NOT DISTINCT"
+            elif nulls_not_distinct is False:
+                text += " NULLS DISTINCT"
+
         withclause = index.dialect_options["postgresql"]["with"]
         if withclause:
             text += " WITH (%s)" % (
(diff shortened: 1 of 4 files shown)


test/dialect/postgresql/test_compiler.py
```diff
@@ -678,6 +678,177 @@ def test_create_index_with_ops(self):
             dialect=postgresql.dialect(),
         )
 
+    def test_nulls_not_distinct(self):
```
zzzeek, line 681: it would be best if these tests broke out into individual cases using `@testing.combinations()`, this is a SQLAlchemy test suite wrapper around [pytest.mark.parametrize](https://docs.pytest.org/en/7.1.x/how-to/parametrize.html):

like this:

```py
@testing.combinations(
   # each case is a 2-tuple of (lambda tbl: DDL, "expected ddl")
    (
         lambda tbl: CreateIndex(Index("test_idx1", tbl.c.data, unique=True,
            postgresql_nulls_not_distinct=True
         )),
         "CREATE UNIQUE INDEX test_idx1 ON test_tbl "
        "(data) NULLS NOT DISTINCT",        
    ),
    (
         lambda tbl: AddConstraint(schema.UniqueConstraint(
            tbl.c.data, name="uq_data1", postgresql_nulls_not_distinct=True,
         )),
           "ALTER TABLE test_tbl ADD CONSTRAINT uq_data1 UNIQUE "
            "NULLS NOT DISTINCT (data)",
    ),

    (
           # next case ...
    ),

    (
           # next case
    ),
   
    # etc ...

)
def test_nulls_not_distinct(self, expr_fn, expected):
    # same setup code

    tbl = Table(.... # same as before )

    # call the lambda using testing.resolve_lambda
    expr = testing.resolve_lambda(expr_fn, tbl=tbl)

   self.assert_compile(expr, expected, dialect=dd)

```

zzzeek (changes requested):
hi thanks for submitting this.

a first round of changes before we move this into gerrit.  thanks!

CaselIT:
Hi, thanks for working on this.

Reflection logic seems to be missing here, and since there is also an alembic change maybe we should add it? I guess it could also be a follow on change.

Another thing to consider is if this option should only be on postgresql or be a global one, since it's in the sql standard. What's your thoughts @zzzeek ?

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2307,6 +2315,29 @@ def visit_create_index(self, create, **kw):
 
         return text
 
+
+    def visit_unique_constraint(self, constraint, **kw):
+        if len(constraint) == 0:
+            return ""
+        nulls_not_distinct_option = ""
+        if self.dialect._supports_nulls_not_distinct:
```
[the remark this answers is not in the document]
Bulbashful: Hi. Yeah It makes sense. Changed

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2994,13 +3026,20 @@ class PGDialect(default.DefaultDialect):
...
+            },
+        ),
     ]
 
     reflection_options = ("postgresql_ignore_search_path",)
 
     _backslash_escapes = True
     _supports_create_index_concurrently = True
     _supports_drop_index_concurrently = True
+    _supports_nulls_not_distinct = False
```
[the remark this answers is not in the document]
Bulbashful: This flag was only for check pg version in methods. I removed this

test/dialect/postgresql/test_compiler.py
Bulbashful: Thank you. I've done this

Bulbashful:
> Hi, thanks for working on this.
> 
> Reflection logic seems to be missing here, and since there is also an alembic change maybe we should add it? I guess it could also be a follow on change.


Hi. Thank you for review.
To be honest I don't know what the changes it should be.

zzzeek:
> Hi, thanks for working on this.
> 
> Reflection logic seems to be missing here, and since there is also an alembic change maybe we should add it? I guess it could also be a follow on change.
> 
> Another thing to consider is if this option should only be on postgresql or be a global one, since it's in the sql standard. What's your thought @zzzeek ?

at the moment it feels this should be PG only even though it's SQL standard.   if *any* other DBs support it though, maybe we should go straight to generic.

CaselIT:
I'll take a look if any other db sqlalchemy internally support (or are fairly known, like db2) supports this

sqla-tester:
I just remembered that SQL Server has the opposite behavior by default but I'm not sure they offer this DDL.  Considering we want to raise if the DDL is not known to be supported, but someone might want a silent skip on SQL server, keeping this per-dialect is probably a better approach, since this is one of those "Flippy / different defaults / not much support" kind of things

CaselIT:
> To be honest I don't know what the changes it should be.

I can take a look while in gerrit

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2279,6 +2280,12 @@ def visit_create_index(self, create, **kw):
                 [preparer.quote(c.name) for c in inclusions]
             )
 
+        nulls_not_distinct = index.dialect_options["postgresql"]["nulls_not_distinct"]
```
CaselIT, line 2283: the default is `None` that means nothing is rendered, correct?

CaselIT:
> I'll take a look if any other db sqlalchemy internally support (or are fairly known, like db2) supports this

seems not, according to https://modern-sql.com/caniuse/F292 and a quick search 

so let's start with PG and then generalize if there is a need to for it

lib/sqlalchemy/dialects/postgresql/base.py
Bulbashful: Yes. Will be rendered only if value is  `True` or `False`.
True - `NULLS NOT DISTINCT`
False - `NULLS DISTINCT`

CaselIT (approved):
Looks fine to me.

The pipeline will probably complain about formatting. Can you run black on the code?

Bulbashful:
> Looks fine to me.
> 
> The pipeline will probably complain about formatting. Can you run black on the code?

Sure. Done

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 825a4ff13a1f428470e184944a167c9d4c57e604 of this pull request into gerrit so we can run tests and reviews and stuff

CaselIT:
I've taken a look at how to reflect the null not distinct, and it's easy for indexes, since there is `indnullsnotdistinct bool` in `pg_index`.
Sadly it's a bit more complex for unique constraints, since it's only in the index, so a bit of work is required since the current reflection logic does not query also indexes.

I'll try taking a look once I've a bit of time

Bulbashful:
> I've taken a look at how to reflect the null not distinct, and it's easy for indexes, since there is `indnullsnotdistinct bool` in `pg_index`. Sadly it's a bit more complex for unique constraints, since it's only in the index, so a bit of work is required since the current reflection logic does not query also indexes.
> 
> I'll try taking a look once I've a bit of time

Thank you. Information about `nulls [not] distinct` can be found in [table_constraints](https://www.postgresql.org/docs/current/infoschema-table-constraints.html) view. Seems that there is no usage of this view for pg dialect. I'll try to manage this case later

CaselIT:
I'm not sure if I would use that view, it's probably better the join with pg_index.
The main difficulty is not the query per se, it's that it requires a bit of work, since at the moment the query to get the unique constraint is shared with primary keys and check constraints, that don't have that type of option.

Bulbashful:
> I'm not sure if I would use that view, it's probably better the join with pg_index. The main difficulty is not the query per se, it's that it requires a bit of work, since at the moment the query to get the unique constraint is shared with primary keys and check constraints, that don't have that type of option.

Yeah. I reckon we can join `pg_index` with `table_constraints` view and check just server version to apply field like here in `_index_query`
`lib/sqlalchemy/dialects/postgresql/base.py:4182`

CaselIT:
I don't think there is a need to join with table_constraints, since we already have all the information in the contraints query.

It may make sense splitting the constraint query though in one for the unique cst and one for the others

CaselIT:
I've updated the gerrit with the reflection logic. Seems to have little to no impact on performance.

sqla-tester (commented):

Should be ok to review, mike

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4632

* doc/build/changelog/unreleased_20/8240.rst (line 9): forgot to credit the constributor. Add

  Pull request courtesy of Pavel Siarchenia.
* lib/sqlalchemy/engine/reflection.py (line 1925): we test this only on pg reflection test, but I was not able to find similar tests for indexes in engine/test_reflection.

Should we add other tests?

Bulbashful:
> I've updated the gerrit with the reflection logic. Seems to have little to no impact on performance.

I've looked the changes. Looks great for me. Thx

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2307,6 +2316,29 @@ def visit_create_index(self, create, **kw):
 
         return text
 
+    def visit_unique_constraint(self, constraint, **kw):
```
sqla-tester, line 2319: **mike bayer** (zzzeek) wrote:

I think I want to make this only def define_unique_constraint_distinct(self, constraint, **kw) and not rewrite all of visit_unique_constraint(), see below

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4632

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2307,6 +2316,29 @@ def visit_create_index(self, create, **kw):
...
+        if len(constraint) == 0:
+            return ""
+        nulls_not_distinct_option = ""
+        nulls_not_distinct = constraint.dialect_options["postgresql"][
```
sqla-tester, line 2323: **mike bayer** (zzzeek) wrote:

does this return None if not present?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4632

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2307,6 +2316,29 @@ def visit_create_index(self, create, **kw):
...
+            "nulls_not_distinct"
+        ]
+        if nulls_not_distinct is True:
+            nulls_not_distinct_option = "NULLS NOT DISTINCT "
+        elif nulls_not_distinct is False:
+            nulls_not_distinct_option = "NULLS DISTINCT "
```
sqla-tester, line 2329: **mike bayer** (zzzeek) wrote:

not that big a deal but I'd find this easier to read if there was a final:

   else:
       nulls_not_distinct_option = ""

instead of having it above

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4632

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2307,6 +2316,29 @@ def visit_create_index(self, create, **kw):
...
+        text = ""
+        if constraint.name is not None:
+            formatted_name = self.preparer.format_constraint(constraint)
+            if formatted_name is not None:
+                text += "CONSTRAINT %s " % formatted_name
+        text += "UNIQUE %s(%s)" % (
+            nulls_not_distinct_option,
```
sqla-tester, line 2336: **mike bayer** (zzzeek) wrote:

OK.    So....this method is copying a *lot* from compiler and NULLS NOT DISTINCT is in the SQL standard...so while I dont yet want to generalize this completely, can we do this instead:


    index d4809ef934..24065d2049 100644
    @@ -6882,7 +6882,8 @@ class DDLCompiler(Compiled):
                formatted_name = self.preparer.format_constraint(constraint)
                if formatted_name is not None:
                    text += "CONSTRAINT %s " % formatted_name
    -        text += "UNIQUE (%s)" % (
    +        text += "UNIQUE %s(%s)" % (
    +            self.define_unique_constraint_distinct(constraint, **kw),
                ", ".join(self.preparer.quote(c.name) for c in constraint)
            )
            text += self.define_constraint_deferrability(constraint)


that is, put it in compiler.py as a new hook, in the same way as define_constraint_deferrability.  since this will someday be a standard hook for other backends also that want to support this.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4632

sqla-tester (commented):

I think I want to enhance the base compiler.py method to have a hook for this rather than restating the whole method.  can we do that?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4632

* doc/build/changelog/unreleased_20/8240.rst (line 9): just add it locally here
* lib/sqlalchemy/engine/reflection.py (line 1925): if the PG dialect is testing this for now that's good enough for the moment, e.g. it's covered
* test/dialect/postgresql/test_reflection.py (line 1510): would be better as inspect(connection).get_indexes("t")
* test/dialect/postgresql/test_reflection.py (line 1536): would be better as inspect(connection).get_unique_constraints("t")

lib/sqlalchemy/dialects/postgresql/base.py
Bulbashful: I thought it's normal because there is similar coping in mssql dialect) 
But I agree with you that this is redundant logic. I'll fix it

CaselIT:
ok I don't think the bot handles the case where gerrit and the change are diverging. I'll unify the two manually tomorrow

lib/sqlalchemy/dialects/postgresql/base.py
Bulbashful: Done

lib/sqlalchemy/dialects/postgresql/base.py
Bulbashful: Yes. And returns `True` or `False` if presented ones

lib/sqlalchemy/dialects/postgresql/base.py
Bulbashful: Done

CaselIT:
> ok I don't think the bot handles the case where gerrit and the change are diverging. I'll unify the two manually tomorrow

updated the gerrit

Bulbashful:
> > ok I don't think the bot handles the case where gerrit and the change are diverging. I'll unify the two manually tomorrow
> 
> updated the gerrit

Thank you. 
There is some troubles with pipeline. I don't know what it is. Tests and linter have passed successfully on my local machine

sqla-tester:

lets see if a rebase fixes final problems

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4632

sqla-tester:

thanks very much for the efforts and tenacity here!!!

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4632

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4632 has been **merged**. Congratulations! :)

Spaceface16518:
Is this going to be backported to 1.4? And is there something I could use in the meantime to work around the lack of this option in the current version?

CaselIT:
most likely no. You can probably subclass the dialect of 1.4 to add support by coping what was done here.

PR #9854, opened 2023-05-29
feat: add `async_creator` argument to `create_async_engine`

### LINKED ISSUE #8215: Support 'async_creator' argument for create_async_engine
### Describe the use case

`sqlalchemy.create_engine` has a `creator` argument that allows a user to pass in a callable which returns a DBAPI connection to be used when creating connection pool engines.

The asyncio engine `create_async_engine` from what I can tell does not have a similar `creator` parameter. It would be nice if the async interface followed a similar pattern and had an `async_creator` parameter for performing the same functionality but with an async function.

Would be used for all asyncio drivers. (`asyncpg`, `aiomysql`, `psycopg2` etc)

### Example Use

The [Cloud SQL Python Connector](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector) is looking to support drivers such as `aiomysql` and `asyncpg` and requires the `creator` argument currently for it's synchronous supported drivers ([Example](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/main/README.md#how-to-use-this-connector)). 

We would like to support the same functionality but with async drivers as follows:
```python
import asyncio
import asyncpg
from google.cloud.sql.connector import Connector

# initialize Cloud SQL Python Connector object
connector = Connector()

# creator function to return the database connection
async def getconn() -> asyncpg.Connection:
    conn: asyncpg.Connection = await connector.connect_async(
        "project:region:instance",  # Cloud SQL connection name
        "asyncpg",  # database driver
        user="root",
        password="shhh",
        db="your-db-name"
    )
    return conn

# create connection pool
pool = create_async_engine(
    "postgresql+asyncpg://",
    async_creator=getconn,
)
```

### Additional context

In progress Cloud SQL Python Connector [PR](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/pull/390) looking to add support for asyncpg which would want this use case. Currently having to use native `asyncpg.create_pool` for connection pooling because SQLAlchemy does not have `async_creator` param for `create_async_engine`.

jackwotherspoon:
### Description
Adding `async_creator` argument to `create_async_engine` to provide similar functionality of synchronous `creator` argument to `create_engine`. Only added to asyncpg currently. If approach looks okay, I will add to other async dialects.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.
         Fixes #8215 

**Have a nice day!**


test/dialect/postgresql/test_async_pg_py3k.py
```diff
@@ -289,3 +283,20 @@ def name_f():
...
             await conn.execute(select(1))
             assert len(cache) > 0
+
+    @async_test
+    async def test_async_creator(self, metadata, async_testing_engine):
+        import asyncpg
+
+        async def async_creator():
+            conn = await asyncpg.connect(
+                "postgresql://scott:tiger@127.0.0.1:5432/test"
```
zzzeek, line 293: our CI doesn't run on localhost like this.  you should transfer the URL from config.db

test/dialect/postgresql/test_async_pg_py3k.py
```diff
@@ -289,3 +283,20 @@ def name_f():
         async with engine.begin() as conn:
             await conn.execute(select(1))
             assert len(cache) > 0
+
+    @async_test
+    async def test_async_creator(self, metadata, async_testing_engine):
```
zzzeek, line 288: this would need `@testing.only_on("postgresql+asyncpg")`

test/ext/asyncio/test_engine_py3k.py
```diff
@@ -66,7 +66,6 @@ def async_trans_ctx_manager_fixture(self, request, metadata):
 
         async def run_test(subject, trans_on_subject, execute_on_subject):
             async with subject.begin() as trans:
-
```
zzzeek, line 69: can we add a test here using mock?

ideally, we'd have a test for each asyncio dialect using each dialect/../provision.py to give us a raw "creator" for each DBAPI

lib/sqlalchemy/dialects/postgresql/asyncpg.py
```diff
@@ -875,6 +873,9 @@ def __init__(self, asyncpg):
 
     def connect(self, *arg, **kw):
         async_fallback = kw.pop("async_fallback", False)
+        creator_fn = kw.pop(
```
zzzeek, line 876: right, this has to be specific to each asyncio DBAPI

so...we would need asyncmy, aiosqlite, aiomysql also...

zzzeek (changes requested):
great start thanks for doing this!

to get this to industrial strength it will need the other backends and some changes to the tests

test/dialect/postgresql/test_async_pg_py3k.py
jackwotherspoon: @zzzeek does it need that? The test class is marked as `__only_on__ = "postgresql+asyncpg"` so i think I might be fine?

test/dialect/postgresql/test_async_pg_py3k.py
jackwotherspoon: Updated to pull from `config.db.url`, thanks!

test/ext/asyncio/test_engine_py3k.py
jackwotherspoon: I'll work on adding a test here now that I've added support for `aiomysql`, `asyncmy` and `aiosqlite` as well

lib/sqlalchemy/dialects/postgresql/asyncpg.py
jackwotherspoon: Added supported for the three additional drivers

test/ext/asyncio/test_engine_py3k.py
jackwotherspoon: @zzzeek do you have an example you could point me at? Not 100% certain how I should be using mock here?

zzzeek: so you would use mock.AsyncMock() , most likely.

here's an example of the existing creator function:

```py
import sqlite3

from sqlalchemy.testing import mock

def creator():
    return sqlite3.connect(":memory:")

mock_creator = mock.Mock(wraps=creator)

e = create_engine("sqlite://", creator=mock_creator)

conn = e.connect()


assert mock_creator.mock_calls == [mock.call()]
```

jackwotherspoon: @zzzeek Thanks for that. I don't see where the code you shared is in the existing code base however?

jackwotherspoon: If you could point me at the part of the code this is being done for the sync `creator` i can go ahead and mimic it for `async_creator` 😄

zzzeek: I just wrote that as an example.  For examples of mocks being used in the test suite, most files in the test suite have some use of mock here and there, such as at https://github.com/sqlalchemy/sqlalchemy/blob/main/test/ext/asyncio/test_engine_py3k.py#L712 .     for actual "creator" tests, there are some in test/engine/test_parseconnect.py but they are  a little more obtuse than they need to be.

lib/sqlalchemy/dialects/mysql/aiomysql.py
```diff
@@ -255,16 +256,17 @@ def _init_dbapi_attributes(self):
 
     def connect(self, *arg, **kw):
         async_fallback = kw.pop("async_fallback", False)
+        creator_fn = kw.pop("creator_fn", partial(self.aiomysql.connect))
```
CaselIT, line 259: There is no need to use partial here or in the other cases

lib/sqlalchemy/ext/asyncio/engine.py
```diff
@@ -82,6 +82,21 @@ def create_async_engine(url: Union[str, URL], **kw: Any) -> AsyncEngine:
             "streaming result set"
         )
     kw["_is_async"] = True
+    async_creator = kw.pop("async_creator", None)
+    if async_creator:
+
+        async def wrap_async_creator() -> Any:
```
CaselIT, line 88: I'm not sure there is a need for this. What advantage does it have compared to just passing `async_creator` to `creator_fn`?

CaselIT (commented):
Personally I preferred the "wrap connect" version more, since it seems a bit more standard than this one, where each and every async dialect needs to do the same check.

Overall seems mostly fine

CaselIT:
It's also missing the implementation for psycopg3 in async mode

jackwotherspoon:
> Personally I preferred the "wrap connect" version more, since it seems a bit more standard than this one, where each and every async dialect needs to do the same check.

Does this mean you think I should change the implementation to the "wrap connect" version?  I went with Mike's version as it was more straightforward for me to implement as I am not very familiar with the SQLAlchemy codebase.

CaselIT:
let me try a couple of things to see how it would go. It may well be that in theory it's nicer, but in practice it's not

lib/sqlalchemy/ext/asyncio/engine.py
```diff
@@ -82,6 +82,21 @@ def create_async_engine(url: Union[str, URL], **kw: Any) -> AsyncEngine:
             "streaming result set"
         )
     kw["_is_async"] = True
+    async_creator = kw.pop("async_creator", None)
+    if async_creator:
+
```
CaselIT, line 87: let's place a `if kw.get("creator", None): raise ArgumentError` mentioning that you cannot pass both async_creator and creator.


Also async_creator should be documented in this function

CaselIT:
> let me try a couple of things to see how it would go. It may well be that in theory it's nicer, but in practice it's not

from a quick try it seems the amount of changes would be pretty much the same, so let's go with the current approach and if needed it can be refined in the future

lib/sqlalchemy/ext/asyncio/engine.py
jackwotherspoon: Good suggestion, will add this

zzzeek:
note that I've just upgraded our gerrit server.  I'm going to try to mirror this PR there now.   It might not work and there might be multiple attempts.  if things get really bad i'll make a real test PR on the testgerrit, but hoping this just works.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 0ae1621aa34ea16520a847f4b382e20e1a42fd20 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Failed to create a gerrit review, git squash against branch 'main' failed

jackwotherspoon:
@zzzeek I made some of the suggestions made by @CaselIT  (great suggestions 😄 ), I still need to get around to add support for psycopg3

lib/sqlalchemy/ext/asyncio/engine.py
jackwotherspoon: good call, removed unnecessary `wrap_async_creator` function and passed `async_creator` directly.

lib/sqlalchemy/dialects/mysql/aiomysql.py
jackwotherspoon: Removed use of `partial` from all dialects

jackwotherspoon:
@CaselIT  @zzzeek  Alright psycopg3 support should be working now. I am going on vacation for about a week starting today so if one of you two has some spare time and can maybe help pull this across the finish line while I'm out that would be great 😄

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 408ab2f51415a11be5e80219eed40553119c527b of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Failed to create a gerrit review, git squash against branch 'main' failed

CaselIT:
@zzzeek this is new. Can you take a look at why the bot is not happy? Os is it just because of a conflict?

zzzeek:
this PR need a rebase and should go in fine after that

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 537073e71e745696f4adb86191b72dd3547b5c95 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 537073e71e745696f4adb86191b72dd3547b5c95: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4722

zzzeek:
this is close to done I do want to add a mock-related test to test_engine_py3k.   I can do that if you want to be done here, let me know.

sqla-tester:

Feel free to add it, mike

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4722

jackwotherspoon:
@zzzeek Feel free to add them! Would be much appreciated 😄

sqla-tester:

ok we're all set thanks all!

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4722

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4722 has been **merged**. Congratulations! :)

jackwotherspoon:
@zzzeek @CaselIT any idea when the next version of SQLAlchemy will be released so that this change will take affect? Thanks in advance 😄

zzzeek:
we've pushed a whole bunch of changes in the past few days so we are due, today or tomorrow

Issue #6759, opened 2021-07-14
Allow generic by implementing __class_getitem__


glyph:
At work we use https://pypi.org/project/sqlalchemy-stubs/ and we prohibit any-generics for code hygiene.  This means we somewhat regularly have stanzas that look like this:

```python3
if TYPE_CHECKING:
     PopulatedColumnGeneric = Column[Any]
     PopulatedBinaryExpression = BinaryExpression[bool, ColumnClause[Any], Any]
 else:
     PopulatedColumnGeneric = Column
     PopulatedBinaryExpression = BinaryExpression
```

so that anything which reads the types using introspection sees actual types, but we've explicitly populated the generic elements of the type.

I realize that sqlalchemy-stubs is a 3rd-party extension here, and you're not responsible for its choices, but I think its generics roughly accurately describe the types within sqlalchemy, and it would be handy and helpful for `[]` to be defined at runtime so we don't need to pull out this escape hatch all the time.

Thank you for your consideration!

CaselIT:
Hi,

We are still supporting python 2 for the 1.4 series, so for now the typing definitions are managed externally since it would otherwise duplicate the work. The plan is to integrate them in the code in version 2.x

Also note that if you are using 1.4 we provide have stubs https://docs.sqlalchemy.org/en/14/orm/extensions/mypy.html that are based on https://github.com/sqlalchemy/sqlalchemy2-stubs

zzzeek:
do we not have an issue already for unifying stubs into main source?

CaselIT:
does not seem so: https://github.com/sqlalchemy/sqlalchemy/labels/pep484

may as well use thig

zzzeek:
hey try this:

```diff
diff --git a/lib/sqlalchemy/sql/visitors.py b/lib/sqlalchemy/sql/visitors.py
index 1cd7b63cde..c16dfb2c62 100644
--- a/lib/sqlalchemy/sql/visitors.py
+++ b/lib/sqlalchemy/sql/visitors.py
@@ -120,6 +120,9 @@ class Traversible(util.with_metaclass(TraversibleType)):
 
     """
 
+    def __class_getitem__(self, key):
+        return self
+
     @util.preload_module("sqlalchemy.sql.traversals")
     def get_children(self, omit_attrs=(), **kw):
         r"""Return immediate child :class:`.visitors.Traversible`
```


now this program runs w/o issue python 3.9 and above:

```py
[...]

PR #9881, opened 2023-06-01
Fix type hint of `key` on `Visitable.__class_getitem__()`

### LINKED ISSUE #9878: Fix type of arg `key` in `Visitable.__class_getitem__()`

### Describe the typing issue

Class `Visitable` defines `__class_getitem__(cls, key: str) -> Any`, to allow generic classes to be used in versions of Python that had not yet implemented support for it (#6759).

However, this causes some type-checkers, including the one built into PyCharm, to become confused: the explicitly defined `__class_getitem__()` is taken to override the implicitly defined version inherited from `typing.Generic`, and this version explicitly hints that `key` must be a string. Therefore indexing the class with another class is a type violation.

The solution is simple: change the type hint to be `str | type`. (Note that a similar definition of `__class_getitem__()` exists in `registry.generate_base()`; this too should be updated the same way.)

### To Reproduce

```python
import sqlalchemy as _sqla
import sqlalchemy.orm as __

class Base(_sqla.orm.DeclarativeBase, _sqla.orm.MappedAsDataclass):
    """Base class for ORM table classes"""

class DBUser(Base):
    """Data model for storing users in PostgreSQL"""

users_query: _sqla.Select[DBUser] = _sqla.select(DBUser)
```

### Error

On line 10, section `[_user.DBUser]`:
```
Expected type 'str', got 'Type[DBUser]' instead
``` 

### Versions

- OS: Win10
- Python: 3.11.2
- SQLAlchemy: 2.0.4
- Type checker (eg: mypy 0.991, pyright 1.1.290, etc): PyCharm 2022.2.5 (Community Edition)

### Additional context

_No response_

macdjord:
### Description
`Visitable.__class_getitem__()` is defined to allow parameterized generics in older versions of Python. (Issue #6759)
Previously, arg `key` of that function was type-hinted as `str`. This caused some type checkers, including PyCharm, to emit a warning if you tried to parameterize a function with the actual class it would return instead of a string reference to it.

Now the type hint has been changed to `Any`, fixing this issue.

The new type of `Any` was selected because:
- `Union[str, type]` would not allow parameterizing with an optional type or a union of types
- `Union[str, type, types.UnionType]` would not work in versions of Python older than 3.10 (when `UnionType` was added)

The definition of `__class_getitem__()` in `registry.generate_base()` has also been updated similarly.

Fixes: #9878

### Checklist

This pull request is:

- [X] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/orm/decl_api.py
+++ b/lib/sqlalchemy/orm/decl_api.py
@@ -1532,7 +1532,7 @@ class registry:
 
         if hasattr(cls, "__class_getitem__"):
 
-            def __class_getitem__(cls: Type[_T], key: str) -> Type[_T]:
+            def __class_getitem__(cls: Type[_T], key: Any) -> Type[_T]:
                 # allow generic classes in py3.9+
                 return cls
 
--- a/lib/sqlalchemy/sql/visitors.py
+++ b/lib/sqlalchemy/sql/visitors.py
@@ -146,7 +146,7 @@ class Visitable:
             cls._original_compiler_dispatch
         ) = _compiler_dispatch
 
-    def __class_getitem__(cls, key: str) -> Any:
+    def __class_getitem__(cls, key: Any) -> Any:
         # allow generic classes in py3.9+
         return cls
 


sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 5aefaaf1065d1f946137a8b2f5be0767856cf9b0 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 5aefaaf1065d1f946137a8b2f5be0767856cf9b0: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4825

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4825 has been **merged**. Congratulations! :)

PR #10108, opened 2023-07-15
Fix annotations

mehdigmira:
### Description
Fixes various type annotations:

1. with_for_update's `of` parameter can be a model (as shown here https://docs.sqlalchemy.org/en/20/orm/queryguide/query.html#sqlalchemy.orm.Query.with_for_update)
2. `session.execute` returns a `CursorResult` and not a `Result` object when ran on an insert, update or delete as described here https://docs.sqlalchemy.org/en/20/tutorial/data_update.html#tutorial-update-delete-rowcount 
3. `_DMLColumnArgument` should use `SQLCoreOperations[Any]` instead of a bare `SQLCoreOperations` otherwise it is partially unknown
4. Add overload to `literal` so that it is inferred that the return type is `BindParameter[NullType]` where `type_` param is None
5. Add overloads to `op` so that the inferred type when `return_type` is not provided is `Callable[[Any], BinaryExpression[Any]]`
6. Add missing overload to `__add__`

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**


CaselIT (changes requested):
great work, thanks. Left a few comments

lib/sqlalchemy/orm/scoping.py
```diff
@@ -639,6 +643,19 @@ def execute(
     ) -> Result[_T]:
         ...
 
+    @overload
+    def execute(
+        self,
+        statement: Delete | Insert | Update,
```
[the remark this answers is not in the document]
CaselIT: it may make sense adding the same overloads to connection / async connection / async session / async scoped session ?

lib/sqlalchemy/sql/elements.py
```diff
@@ -123,11 +124,38 @@
...
     value: Any,
-    type_: Optional[_TypeEngineArgument[_T]] = None,
+    type_: _TypeEngineArgument[_T],
     literal_execute: bool = False,
 ) -> BindParameter[_T]:
+    ...
+
+
+@overload
+def literal(
```
CaselIT, line 137: maybe it should infer the type from the value here? sqlalchemy seems to do this at runtime, since this
```py
foo = sa.literal(123)
print(foo.type)
```
prints
```
INTEGER
```

CaselIT: @zzzeek opinions here?

lib/sqlalchemy/sql/elements.py
```diff
@@ -794,16 +822,50 @@ def reverse_operate(
...
+        @overload
+        def op(
+            self,
+            opstring: str,
+            precedence: int,
+            is_comparison: bool,
+            return_type: _TypeEngineArgument[_OPT],
             python_impl: Optional[Callable[..., Any]] = None,
         ) -> Callable[[Any], BinaryExpression[_OPT]]:
             ...
```
zzzeek, line 846: these overloads dont seem to be correct, or I'm not following

AFAICT there should be two overloads:

1. is_comparison: True  -> return type is `Callable[[Any], BinaryExpression[bool]]`
2. return_type: `_TypeEngineArgument[_OPT]`  -> return type is `Callable[[Any], BinaryExpression[_OPT]]`
3. default return type is `Callable[[Any], BinaryExpression[Any]]`

test/typing/plain_files/orm/typed_queries.py
```diff
@@ -440,6 +440,27 @@ def t_dml_insert() -> None:
     reveal_type(r3)
```
zzzeek, line 442: we should have some tests:

* `op()` change: test/typing/plain_files/sql/operators.py
* `literal` change: test/typing/plain_files/sql/common_sql_expr.py
* `of` change (Core select() version): test/typing/plain_files/sql/common_sql_expr.py , there are some with_for_update tests there already

zzzeek (changes requested):
left some comments

mehdigmira:
@zzzeek  would it be possible to review this so it can be merged ? :)

zzzeek:
OK.  I dont get any kind of email when new commits are added

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 6017526c9cd1282025885cb002de1f984f64205b of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 6017526c9cd1282025885cb002de1f984f64205b: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4794

mehdigmira:
> OK. I dont get any kind of email when new commits are added

Thanks. I requested a new review from both of you a couple of weeks ago. Thought this would notify you.
image

zzzeek:
shrugs i dont know where notifications for that go, if there are any

lib/sqlalchemy/orm/query.py
```diff
@@ -3270,14 +3263,11 @@ def update(
...
         upd._where_criteria = self._where_criteria
-        result: CursorResult[Any] = cast(
-            "CursorResult[Any]",
-            self.session.execute(
-                upd,
-                self._params,
-                execution_options=self._execution_options.union(
-                    {"synchronize_session": synchronize_session}
-                ),
+        result: CursorResult[Any] = self.session.execute(
```
sqla-tester, line 3266: **mike bayer** (zzzeek) wrote:

nice

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4794

lib/sqlalchemy/sql/elements.py
```diff
@@ -794,16 +821,50 @@ def reverse_operate(
...
             self,
             opstring: str,
             precedence: int = 0,
             is_comparison: bool = False,
-            return_type: Optional[_TypeEngineArgument[_OPT]] = None,
+            *,
+            return_type: _TypeEngineArgument[_OPT],
             python_impl: Optional[Callable[..., Any]] = None,
         ) -> Callable[[Any], BinaryExpression[_OPT]]:
             ...
```
sqla-tester, line 835: **mike bayer** (zzzeek) wrote:

OK just this part, what is this getting us?   I dont see any tests added for the op() change here so I dont know what this covers.  can we add something?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4794

sqla-tester (commented):

looks great, it's ready to merge except i have a Q about the op() change

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4794

* test/typing/plain_files/sql/common_sql_element.py (line 25): we have orm.Session in sql/.   welp ok

zzzeek (changes requested):
see sqla-tester's comment re: op() `*` change

lib/sqlalchemy/sql/elements.py
mehdigmira: I think I answered this here here https://github.com/sqlalchemy/sqlalchemy/pull/10108#discussion_r1266297680
With the current implementation, if return_type is None, the overload that matches is the following:

```python
@overload
def op(
            self,
            opstring: str,
            precedence: int = 0,
            is_comparison: bool = False,
            return_type: Optional[_TypeEngineArgument[_OPT]] = None,
            python_impl: Optional[Callable[..., Any]] = None,
        ) -> Callable[[Any], BinaryExpression[_OPT]]:
  ...
```

However the `_OPT` used in the return type `Callable[[Any], BinaryExpression[_OPT]]` cannot be inferred.
mypy will fallback to _OPT=Any, but pyright will report that `_OPT` is unknown.

zzzeek: If I remove the overload that has `*` in it, does a test fail ?  taht's all we need here

lib/sqlalchemy/sql/elements.py
zzzeek: OK as it stands no difference in mypy can be produced, only pyright, but the problem is I dont understand this overload so i have to keep looking at it and figure some way to clarify this

zzzeek: OK I have it now thanks

sqla-tester:

all set thanks very much!

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4794

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4794 has been **merged**. Congratulations! :)

PR #10133, opened 2023-07-22
fix #10131: update type annotations for `load_only`, `defer` and `undefer`

### LINKED ISSUE #10131: `orm.load_only`, `orm.defer` and `orm.undefer` wrongfully typed with `str | QueryableAttribute`

### Describe the typing issue

The functions `orm.load_only`, `orm.defer` and `orm.undefer` are typed as accepting `key: str | QueryableAttribute` as their first argument(s), but will raise an `sqlalchemy.exc.ArgumentError` when strings are passed.

### To Reproduce

```python
from sqlalchemy.orm import load_only, defer, undefer


load_only("foo")
defer("foo")
undefer("foo")
```

### Error

```
  File "/home/***/python3.10/site-packages/sqlalchemy/orm/strategy_options.py", line 2272, in _parse_attr_argument
    insp: InspectionAttr = inspect(attr)  # type: ignore
  File "/home/***/python3.10/site-packages/sqlalchemy/inspection.py", line 152, in inspect
    raise exc.NoInspectionAvailable(
sqlalchemy.exc.NoInspectionAvailable: No inspection system is available for object of type <class 'str'>

The above exception was the direct cause of the following exception:

  File "/home/***/test.py", line 4, in <module>
    load_only("foo")
  File "/home/***/python3.10/site-packages/sqlalchemy/orm/strategy_options.py", line 2325, in load_only
    _, lead_element, _ = _parse_attr_argument(attrs[0])
  File "/home/***/python3.10/site-packages/sqlalchemy/orm/strategy_options.py", line 2274, in _parse_attr_argument
    raise sa_exc.ArgumentError(
sqlalchemy.exc.ArgumentError: expected ORM mapped attribute for loader strategy argument

``` 

### Versions

- OS:
- Python: 3.10.10
- SQLAlchemy: 2.0.19
- Type checker (eg: mypy 0.991, pyright 1.1.290, etc): N/A

### Additional context

_No response_

provinzkraut:
### Description

Change the annotations for the attribute(s) parameters to `QueryableAttribute` and `QueryableAttribute | Literal["*"]` for defer and undefer, to accurately reflect the types they accept.

Fixes #10131.

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/orm/strategy_options.py
+++ b/lib/sqlalchemy/orm/strategy_options.py
@@ -2319,7 +2319,9 @@ def contains_eager(*keys: _AttrType, **kw: Any) -> _AbstractLoad:
 
 
 @loader_unbound_fn
-def load_only(*attrs: _AttrType, raiseload: bool = False) -> _AbstractLoad:
+def load_only(
+    *attrs: QueryableAttribute[Any], raiseload: bool = False
+) -> _AbstractLoad:
     # TODO: attrs against different classes.  we likely have to
     # add some extra state to Load of some kind
     _, lead_element, _ = _parse_attr_argument(attrs[0])
@@ -2376,7 +2378,9 @@ def defaultload(*keys: _AttrType) -> _AbstractLoad:
 
 @loader_unbound_fn
 def defer(
-    key: _AttrType, *addl_attrs: _AttrType, raiseload: bool = False
+    key: Union[QueryableAttribute[Any], Literal["*"]],
+    *addl_attrs: Union[QueryableAttribute[Any], Literal["*"]],
+    raiseload: bool = False,
 ) -> _AbstractLoad:
     if addl_attrs:
         util.warn_deprecated(
@@ -2395,7 +2399,10 @@ def defer(
 
 
 @loader_unbound_fn
-def undefer(key: _AttrType, *addl_attrs: _AttrType) -> _AbstractLoad:
+def undefer(
+    key: Union[QueryableAttribute[Any], Literal["*"]],
+    *addl_attrs: Union[QueryableAttribute[Any], Literal["*"]],
+) -> _AbstractLoad:
     if addl_attrs:
         util.warn_deprecated(
             "The *addl_attrs on orm.undefer is deprecated.  Please use "


lib/sqlalchemy/orm/strategy_options.py
```diff
@@ -2319,7 +2319,9 @@ def contains_eager(*keys: _AttrType, **kw: Any) -> _AbstractLoad:
 
 
 @loader_unbound_fn
-def load_only(*attrs: _AttrType, raiseload: bool = False) -> _AbstractLoad:
+def load_only(
+    *attrs: QueryableAttribute[Any], raiseload: bool = False
```
zzzeek, line 2323: revert all these changes, and simply replace "str" from the `_AttrType` alias with `Literal['*']`

zzzeek (changes requested):
hi, and thanks for this!

this only needs to update the top level `_AttrType` to replace the "str" with a more appropriate `Literal` for '*'.   I would not want `QueryableAttribute` coded directly on the functions themselves as we'd never find this again.

lib/sqlalchemy/orm/strategy_options.py
provinzkraut: I didn't want to touch the other cases where `_AttrType` is being used, so that's why I did this. If it's safe to change this in the alias, that's obviously the better solution. I'll update it accordingly.

zzzeek: well we should have and/or add tests for all elements that use `_AttrType`, then there won't be a problem

zzzeek: typing tests are plain python files in test/typing/plain_files/ .  if there is not already a file that is focused on loader options like `defer()`, `joinedload()`, etc. then one should be added

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 08793ed5bfbffbc8688a2573f021e834fc7de367 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 08793ed5bfbffbc8688a2573f021e834fc7de367: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4826

zzzeek:
any plans to continue this?

provinzkraut:
> any plans to continue this?

Sorry for the delay, I completely lost track of this :grimacing: 

I'll update this shortly!

CaselIT:
@provinzkraut I've updated the gerrit change. It should be fine now

sqla-tester (commented):

hey there -

almost done.  I want to go one step further by replacing that str with a type that comes from path_registry.py.

Also this should have a short changelog file indicating the end-user facing change.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4826

* lib/sqlalchemy/orm/strategy_options.py (line 1671): 'str' is leaking in here.   I would assume this is because we have path tokens that get sent here when we create _TokenStrategyLoad.create.  However I'd prefer we continue typing into those tokens, which would carry this into path_registry.py.  

this is not as hard as it sounds because I think we can start by replacing "str" here with a simple type alias sqlalchemy.orm..path_registry._StrPathToken:

	diff --git a/lib/sqlalchemy/orm/path_registry.py b/lib/sqlalchemy/orm/path_registry.py
	index 2cd8a1412c..8fb148f8df 100644
	--- a/lib/sqlalchemy/orm/path_registry.py
	+++ b/lib/sqlalchemy/orm/path_registry.py
	@@ -58,8 +58,10 @@ else:
	 
	 _SerializedPath = List[Any]
	 
	+_StrPathToken = str
	+
	 _PathElementType = Union[
	-    str, "_InternalEntityType[Any]", "MapperProperty[Any]"
	+    _StrPathToken, "_InternalEntityType[Any]", "MapperProperty[Any]"
	 ]
	 
	 # the representation is in fact
	@@ -433,10 +435,10 @@ class TokenRegistry(PathRegistry):
	 
	     inherit_cache = True
	 
	-    token: str
	+    token: _StrPathToken
	     parent: CreatesToken
	 
	-    def __init__(self, parent: CreatesToken, token: str):
	+    def __init__(self, parent: CreatesToken, token: _StrPathToken):
	         token = PathToken.intern(token)
	 
	         self.token = token



can we integrate this change and carry it outwards to strategy_options here to replace str?  thanks

sqla-tester (commented):

ok, did not add a changelog since trying to use str that are not * on v2 results in an error, so there is really no chance required by users.
but I can add one

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4826

* lib/sqlalchemy/orm/strategy_options.py (line 1671): makes sense

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4826 has been **merged**. Congratulations! :)

PR #10248, opened 2023-08-16
sql/sqltypes/Enum: explicitly document values_callable order

### LINKED ISSUE #9911: Enum `values_callable` assumes the result has the same order as iterating through `__members__`
### Describe the bug

The `values_callable` paramater to `Enum` doesn't mention that the return value should be in the same order that the enum was defined.

This means that if the order of the values returned from values_callable does not match the order of the enum members, then SQLAlchemy will load models with the wrong enum values!

The minimal reproduction below will result in "invalid" values being loaded about 2/3s of the time.

http://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Enum.params.values_callable

### SQLAlchemy Version in Use

1.4.46

asyncpg

PostgreSQL 13.10 (Debian 13.10-0+deb11u1)

### Python Version

3.9.2

### Operating system

Debian 11.4

### To Reproduce

```python
from enum import Enum

import sqlalchemy as sa

Base = declarative_base()


class Visibility(Enum):
    PUBLIC = "public"
    UNLISTED = "unlisted"
    PRIVATE = "private"
    
    @classmethod
    def values(cls) -> set:
        return set(i.value for i in cls)


class FooModel(Base):
    __tablename__ = "foo"
    
    visibility = sa.Column(
        sa.Enum(Visibility, values_callable=lambda m: list(m.values()))
    )
```

### Error

No errors - but lots of confusion

### Additional context

I'd probably be happy with a documentation fix (and maybe an example).
However, I agree with @ethanfurman in https://github.com/sqlalchemy/sqlalchemy/discussions/8934

Another option might be to replace `values_callable` with a `store_values` boolean that iterates through the enum instead of forcing the user to do the conversion themself.

huwcbjones:
Fixes #9911

### Checklist
This pull request is:

- [ ] A short code fix

### DIFF
--- a/lib/sqlalchemy/sql/sqltypes.py
+++ b/lib/sqlalchemy/sql/sqltypes.py
@@ -1395,7 +1395,10 @@ class Enum(String, SchemaType, Emulated, TypeEngine[Union[str, enum.Enum]]):
            compliant enumerated type, which should then return a list of string
            values to be persisted. This allows for alternate usages such as
            using the string value of an enum to be persisted to the database
-           instead of its name.
+           instead of its name. The callable must return the values to be
+           persisted in the same order as iterating through the Enum's
+           ``__member__`` attribute. For example
+           ``lambda x: [i.value for i in x]``.
 
            .. versionadded:: 1.2.3
 


CaselIT (approved):

CaselIT:
thanks. sorry I missed this one

PR #10249, opened 2023-08-17
Generalize pydantic version warning

BayerSe:
### Description

Pydantic 1.x and 2.x both have compatibility issues with sqlalchemy, which we denote by removing the restriction on the 1.x version of pydantic, as discussed [here](https://github.com/sqlalchemy/sqlalchemy/discussions/10243).

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/doc/build/orm/dataclasses.rst
+++ b/doc/build/orm/dataclasses.rst
@@ -530,7 +530,7 @@ Integrating with Alternate Dataclass Providers such as Pydantic
 
 .. warning::
 
-    The dataclass layer of Pydantic version 1.x is **not fully compatible** with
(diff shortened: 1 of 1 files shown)


zzzeek:
yup, thanks for sending.  i almost want to take this whole doc section out :/

CaselIT:
> yup, thanks for sending. i almost want to take this whole doc section out :/

That's what I was thinking too. Doesn't make much sense to mention pydantic if it doesn't work. We can just speak of generic "3rd party libraries that are compatible with dataclasses". Not sure it attrs for example supports it

PR #10274, opened 2023-08-22
Allow using Enum with length=None

### LINKED ISSUE #10269: Allow using Enum with length=None
### Describe the use case

When using string-based enum types, i.e. `Enum(python_enum,  native_enum=False)`, the default behavior is to use `String(N)` with N set to the length of the longest enum member name. This can break if new enum members are added in the future with longer names. If the default behavior cannot be changed, the issue can be solved by explicitly passing `length=None`, which removes the string length limit. This almost works, except that it breaks the check here: https://github.com/sqlalchemy/sqlalchemy/blob/09d247d57ffb6329739deae89019c9e7ad3b9e7f/lib/sqlalchemy/sql/sqltypes.py#L1454
This can be further worked-around by passing an undocumented argument `_disable_warnings=True`. It would be nice if the latter was not required. This requires only a one-line change
```python
if not _disable_warnings and length_arg is not None and length_arg < length:
```

Any database.

### Example Use

**Example:**
```python
import enum
import sqlalchemy as sa

class MyEnum(enum.Enum):
    A = 1

md = sa.MetaData()
table = sa.Table("test", md, sa.Column("enum", sa.Enum(MyEnum, native_enum=False, length=None)))
print(sa.schema.CreateTable(table))
```
**Expected output:**
```
CREATE TABLE test (
        enum VARCHAR
)
```
**Current output:**
```
TypeError: '<' not supported between instances of 'NoneType' and 'int'
```
**Work-around:**
```python
table = sa.Table(
    "test",
    md,
    sa.Column("enum", sa.Enum(MyEnum, native_enum=False, length=None, _disable_warnings=True))
)
```

eltoder:
Fixes: #10269

### Description
Allow passing `length=None` for string-based Enums, which turns them into `VARCHAR` without a length limit.

This pull request is:

- [X] A short code fix

### DIFF
--- a/lib/sqlalchemy/sql/sqltypes.py
+++ b/lib/sqlalchemy/sql/sqltypes.py
@@ -1451,7 +1451,11 @@ class Enum(String, SchemaType, Emulated, TypeEngine[Union[str, enum.Enum]]):
             self._default_length = length = 0
 
         if length_arg is not NO_ARG:
-            if not _disable_warnings and length_arg < length:
+            if (
+                not _disable_warnings
+                and length_arg is not None
+                and length_arg < length
+            ):
                 raise ValueError(
                     "When provided, length must be larger or equal"
                     " than the length of the longest enum value. %s < %s"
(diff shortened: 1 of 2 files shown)


sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 651afaaea76c1ec868402cdd7106ec8b2de76254 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 651afaaea76c1ec868402cdd7106ec8b2de76254: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4830

zzzeek:
looks good

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4830 has been **merged**. Congratulations! :)

PR #10276, opened 2023-08-23
Try pep 646 on Row

mehdigmira:
draft

### DIFF
--- a/lib/sqlalchemy/engine/row.py
+++ b/lib/sqlalchemy/engine/row.py
@@ -28,6 +28,7 @@ from typing import Tuple
 from typing import TYPE_CHECKING
 from typing import TypeVar
 from typing import Union
+from typing_extensions import TypeVarTuple, Unpack
 
 from ..sql import util as sql_util
 from ..util import deprecated
@@ -44,10 +45,20 @@ if TYPE_CHECKING:
     from .result import RMKeyView
 
 _T = TypeVar("_T", bound=Any)
-_TP = TypeVar("_TP", bound=Tuple[Any, ...])
+_TP = TypeVarTuple("_TP")
+
+if TYPE_CHECKING:
+
+    class _Row(BaseRow, tuple[Unpack[_TP]]):
+        ...
+
+else:
+
+    class _Row(BaseRow, Generic[Unpack[_TP]]):
+        ...
 
 
-class Row(BaseRow, Sequence[Any], Generic[_TP]):
+class Row(_Row[Unpack[_TP]]):
     """Represent a single result row.
 
     The :class:`.Row` object represents a row of a database result.  It is
--- a/test/typing/plain_files/orm/session.py
+++ b/test/typing/plain_files/orm/session.py
@@ -97,6 +97,9 @@ with Session(e) as sess:
         User.id
     ).offset(User.id)
 
+    # EXPECTED_TYPE: Optional[User]
+    reveal_type(sess.query(User).get(1))
+
 # more result tests in typed_results.py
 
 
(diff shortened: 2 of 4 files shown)


zzzeek:
wow.    OK just so you know that code is generated dynamically by tools/generate_tuple_map_overloads.py and it also applies to sqlalchemy/sql/selectable.py.

I would be surprised if the pep646 code is actually usable, I was never able to get it to do anything useful despite many attempts, but maybe you've figured something out

lib/sqlalchemy/engine/row.py
```diff
@@ -44,10 +45,20 @@
     from .result import RMKeyView
 
 _T = TypeVar("_T", bound=Any)
-_TP = TypeVar("_TP", bound=Tuple[Any, ...])
+_TP = TypeVarTuple("_TP")
+
+if TYPE_CHECKING:
+
+    class _Row(BaseRow, tuple[Unpack[_TP]]):
```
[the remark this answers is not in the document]
CaselIT: so the main idea is to lie to the type checker making it think we are using a tuple?

CaselIT:
wouldn't we also need to remove this
https://github.com/sqlalchemy/sqlalchemy/blob/4a5831448f1733c80f6b8320f2dbc56627be37ee/lib/sqlalchemy/engine/row.py#L226-L235

test/typing/plain_files/orm/session.py
```diff
@@ -97,6 +97,9 @@ class Address(Base):
         User.id
     ).offset(User.id)
 
+    # EXPECTED_TYPE: Optional[User]
+    reveal_type(sess.query(User).get(1))
```
eltoder, line 101: Can use `assert_type` from typing_extensions.

CaselIT: That would need changes to the mypy test fixtures, but maybe it makes sense?

eltoder: Good point. That would be a separate PR.

CaselIT: Indeed. Also myoy would need to print the type in all cases even if the asset is correct

mehdigmira:
closing this until mypy implements pep646 i'll reopen then

zzzeek:
someone has picked this up again in https://github.com/sqlalchemy/sqlalchemy/pull/10634 .  care to take a look ?

mehdigmira:
I've been involved in testing mypy PEP646 changes as they were being developed (for example [here](https://github.com/python/mypy/pull/16131)).
I already have a [MR](https://github.com/mehdigmira/sqlalchemy/pull/2) ongoing that I used as a base for those tests to detect some edge cases. It should be almost ready, I was planning on picking it up again this weekend to polish it.

zzzeek:
OK then we have two people working on the same thing?  What should we do?  can you coordinate on one PR ?  is there a difference in approach?

CaselIT:
seems that the main thing is still making the type-checker thing that Row is a tuple subclass.

On this thing I like the other PR approach better since we avoid the intermediate class https://github.com/sqlalchemy/sqlalchemy/blob/0352cb931f039cc622618d10d1a7cf2b6ba91aef/lib/sqlalchemy/engine/row.py#L42

CaselIT:
> I already have a [MR](https://github.com/mehdigmira/sqlalchemy/pull/2) ongoing that I used as a base for those tests to detect some edge cases. It should be almost ready, I was planning on picking it up again this weekend to polish it.

btw it's fine to open PR directly here, set them in draft / mention that they are so that they are visible also to other people

Issue #9737, opened 2023-05-03
Missing literal processor for datetime.timedelta in postgresql


web4u2qd:
### Describe the bug

According to the documentation (https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Interval), Interval type is "a type for datetime.timedelta() objects". When I bind a datetime.timedelta parameter and compile a statement I get the following error:
`sqlalchemy.exc.CompileError: No literal value renderer is available for literal value "datetime.timedelta(seconds=60)" with datatype DATETIME`

I would expect it to work since other types (e.g. datetime.datetime) seem to work.

### Optional link from https://docs.sqlalchemy.org which documents the behavior that is expected

https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Interval

### SQLAlchemy Version in Use

2.0.12

### DBAPI (i.e. the database driver)

psycopg2

### Database Vendor and Major Version

PostgreSQL 15

### Python Version

3.10.4

### Operating system

linux

### To Reproduce

```python
import datetime
import sqlalchemy
from sqlalchemy import create_mock_engine


def main():
    def dump(sql, *multiparams, **params):
        stmt = str(sql.compile(dialect=engine.dialect))
        print(stmt)

    engine = create_mock_engine("postgresql+psycopg2://", dump)

    sql = 'SELECT :parameter'
    params = {'parameter': datetime.timedelta(minutes=1)}

    stmt = sqlalchemy.text(sql).bindparams(**params)

    print(str(stmt.compile(dialect=engine.dialect, compile_kwargs={'literal_binds': True})))


if __name__ == '__main__':
    main()
```

### Error

```
Traceback (most recent call last):
  File "/home/project/src/tools/test.py", line 22, in <module>
    main()
  File "/home/project/src/tools/test.py", line 18, in main
    print(str(stmt.compile(dialect=engine.dialect, compile_kwargs={'literal_binds': True})))
  File "/opt/venv/lib/python3.10/site-packages/sqlalchemy/sql/elements.py", line 280, in compile
    return self._compiler(dialect, **kw)
  File "/opt/venv/lib/python3.10/site-packages/sqlalchemy/sql/elements.py", line 288, in _compiler
    return dialect.statement_compiler(dialect, self, **kw)
  File "/opt/venv/lib/python3.10/site-packages/sqlalchemy/sql/compiler.py", line 1424, in __init__
[...]

PR #10383, opened 2023-09-24
fixes: #9737, add literal_processors for interval

indiVar0508:
added literal processor for intervals for postgres and oracle db in their respective types module.
added tests for same in the dialects directory within test_types.py docs refered
postgres: https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-INTERVAL-INPUT oracle: https://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements003.htm#i38598

- Didn't add YEARS to MONTH for oracle as it was not already supported 
- not sure on how to support precision for postgres interval seconds so added a fixme note there for now.

### Checklist

This pull request is:

- [x] A short code fix

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/oracle/types.py
+++ b/lib/sqlalchemy/dialects/oracle/types.py
@@ -4,11 +4,23 @@
 # This module is part of SQLAlchemy and is released under
 # the MIT License: https://www.opensource.org/licenses/mit-license.php
 # mypy: ignore-errors
+from __future__ import annotations
+
+import datetime as dt
+from typing import Callable
+from typing import no_type_check
+from typing import Optional
+from typing import Type
+from typing import TYPE_CHECKING
+
 
 from ... import exc
 from ...sql import sqltypes
 from ...types import NVARCHAR
 from ...types import VARCHAR
+
+if TYPE_CHECKING:
+    from ...engine.interfaces import Dialect
 
 
 class RAW(sqltypes._Binary):
@@ -213,6 +225,21 @@ class INTERVAL(sqltypes.NativeForEmulated, sqltypes._AbstractInterval):
             second_precision=self.second_precision,
             day_precision=self.day_precision,
         )
+
+    @property
+    def python_type(self) -> Type[dt.timedelta]:
+        return dt.timedelta
+
+    @no_type_check
+    def literal_processor(
+        self, dialect: Optional[Dialect]
+    ) -> Callable[[Optional[dt.timedelta]], Optional[str]]:
+        def process(value: Optional[dt.timedelta]) -> Optional[str]:
+            if value is not None:
+                return f"NUMTODSINTERVAL({value.total_seconds()}, 'SECOND')"
+            return None
+
+        return process
 
 
 class TIMESTAMP(sqltypes.TIMESTAMP):
--- a/lib/sqlalchemy/dialects/postgresql/types.py
+++ b/lib/sqlalchemy/dialects/postgresql/types.py
@@ -7,6 +7,8 @@ from __future__ import annotations
 
 import datetime as dt
 from typing import Any
+from typing import Callable
+from typing import no_type_check
 from typing import Optional
 from typing import overload
 from typing import Type
@@ -20,6 +22,7 @@ from ...util.typing import Literal
 if TYPE_CHECKING:
     from ...sql.operators import OperatorType
     from ...sql.type_api import TypeEngine
+    from ...engine.interfaces import Dialect
 
 _DECIMAL_TYPES = (1231, 1700)
 _FLOAT_TYPES = (700, 701, 1021, 1022)
@@ -246,6 +249,17 @@ class INTERVAL(type_api.NativeForEmulated, sqltypes._AbstractInterval):
     @property
     def python_type(self) -> Type[dt.timedelta]:
         return dt.timedelta
+
+    @no_type_check
+    def literal_processor(
+        self, dialect: Optional[Dialect]
+    ) -> Callable[[Optional[dt.timedelta]], Optional[str]]:
+        def process(value: Optional[dt.timedelta]) -> Optional[str]:
+            if value is not None:
+                return f"make_interval(secs=>{value.total_seconds()})"
+            return None
+
+        return process
 
 
 PGInterval = INTERVAL
--- a/lib/sqlalchemy/testing/suite/test_types.py
+++ b/lib/sqlalchemy/testing/suite/test_types.py
@@ -28,6 +28,7 @@ from ... import Date
 from ... import DateTime
 from ... import Float
 from ... import Integer
+from ... import Interval
 from ... import JSON
 from ... import literal
 from ... import literal_column
@@ -450,6 +451,123 @@ class StringTest(_LiteralRoundTripFixture, fixtures.TestBase):
             connection.scalar(select(literal("a") + "b" + "c" + "d" + "e")),
             "abcde",
         )
+
+
+class IntervalTest(_LiteralRoundTripFixture, fixtures.TablesTest):
+    __requires__ = ("datetime_interval",)
+    __backend__ = True
+    compare = None
+    datatype = Interval
+    data = datetime.timedelta(days=1, seconds=4)
+
+    @classmethod
+    def define_tables(cls, metadata):
+        class Decorated(TypeDecorator):
+            impl = cls.datatype
+            cache_ok = True
+
+        Table(
+            "interval_table",
+            metadata,
+            Column(
+                "id", Integer, primary_key=True, test_needs_autoincrement=True
+            ),
+            Column("interval_data", cls.datatype),
+            Column("date_data", DateTime),
+            Column("decorated_interval_data", Decorated),
+        )
+
+    @testing.combinations(
+        (
+            datetime.timedelta(days=1, seconds=4),
+            datetime.timedelta(days=1, seconds=4),
+        ),
+        argnames="input_, compare_",
+    )
+    def test_render_literal_interval(
+        self, literal_round_trip, input_, compare_
+    ):
+        literal_round_trip(Interval(), [input_], [compare_])
+
+    def test_select_direct_literal_interval(self, connection):
+        row = connection.execute(
+            select(literal(self.data, literal_execute=True))
+        ).first()
+        eq_(row, (self.data,))
+
+    # NOTE: There seems to be inconsistence in the behaviour for literal
+    #        for datetime.datetime, if literal_execute=True and
+    #        it works for oracle but doesn't work for postgres because it
+    #        processes query directly, when literal_execute is not provided
+    #        microseconds are not rendered properly for postgres.
+    #        For now, Skipping tests for oracle.
+    @testing.requires.datetime_microseconds
+    def test_arithmetic_operation_literal_interval(self, connection):
+        now = datetime.datetime.now()
+        # Able to subtract
+        row = connection.execute(
+            select(literal(now) - literal(self.data, literal_execute=True))
+        ).first()
+        eq_(row[0], (now - self.data))
+
+        # Able to Add
+        row = connection.execute(
+            select(literal(now) + literal(self.data, literal_execute=True))
+        ).first()
+        eq_(row[0], (now + self.data))
+
+    def test_arithmetic_operation_table_interval_and_literal_interval(
+        self, connection
+    ):
+        interval_table = self.tables.interval_table
+        data = datetime.timedelta(days=2, seconds=5)
+        connection.execute(
+            interval_table.insert(), {"id": 1, "interval_data": data}
+        )
+        # Subtraction Operation
+        row = connection.execute(
+            select(
+                interval_table.c.interval_data
+                - literal(self.data, literal_execute=True)
+            )
+        ).first()
+        eq_(row[0], (datetime.timedelta(days=1, seconds=1)))
+        # Addition Operation
+        row = connection.execute(
+            select(
+                interval_table.c.interval_data
+                + literal(self.data, literal_execute=True)
+            )
+        ).first()
+        eq_(row[0], (datetime.timedelta(days=3, seconds=9)))
+
+    # NOTE: When Fetching from date column from table microseconds are
+    #        not being received so skipping this one for now.
+    @testing.requires.datetime_microseconds
+    def test_arithmetic_operation_table_date_and_literal_interval(
+        self, connection
+    ):
+        interval_table = self.tables.interval_table
+        now = datetime.datetime.now()
+        connection.execute(
+            interval_table.insert(), {"id": 1, "date_data": now}
+        )
+        # Subtraction Operation
+        row = connection.execute(
+            select(
+                interval_table.c.date_data
+                - literal(self.data, literal_execute=True)
+            )
+        ).first()
+        eq_(row[0], (now - self.data))
+        # Addition Operation
+        row = connection.execute(
+            select(
+                interval_table.c.date_data
+                + literal(self.data, literal_execute=True)
+            )
+        ).first()
+        eq_(row[0], (now + self.data))
 
 
 class _DateFixture(_LiteralRoundTripFixture, fixtures.TestBase):
@@ -1940,6 +2058,7 @@ __all__ = (
     "TextTest",
     "NumericTest",
     "IntegerTest",
+    "IntervalTest",
     "CastTypeDecoratorTest",
     "DateTimeHistoricTest",
     "DateTimeCoercedToDateTimeTest",
--- a/test/dialect/oracle/test_types.py
+++ b/test/dialect/oracle/test_types.py
@@ -208,6 +208,33 @@ class DialectTypesTest(fixtures.TestBase, AssertsCompiledSQL):
     def test_float_type_compile(self, type_, sql_text):
         self.assert_compile(type_, sql_text)
 
+    @testing.combinations(
+        (
+            text("select :parameter from dual").bindparams(
+                parameter=datetime.timedelta(days=2)
+            ),
+            "select NUMTODSINTERVAL(172800.0, 'SECOND') from dual",
+        ),
+        (
+            text("SELECT :parameter from dual").bindparams(
+                parameter=datetime.timedelta(days=1, minutes=3, seconds=4)
+            ),
+            "SELECT NUMTODSINTERVAL(86584.0, 'SECOND') from dual",
+        ),
+        (
+            text("select :parameter - :parameter2 from dual").bindparams(
+                parameter=datetime.timedelta(days=1, minutes=3, seconds=4),
+                parameter2=datetime.timedelta(days=0, minutes=1, seconds=4),
+            ),
+            (
+                "select NUMTODSINTERVAL(86584.0, 'SECOND') - "
+                "NUMTODSINTERVAL(64.0, 'SECOND') from dual"
+            ),
+        ),
+    )
+    def test_interval_literal_processor(self, type_, expected):
+        self.assert_compile(type_, expected, literal_binds=True)
+
 
 class TypesTest(fixtures.TestBase):
     __only_on__ = "oracle"
@@ -322,6 +349,31 @@ class TypesTest(fixtures.TestBase):
             row._mapping["day_interval"],
             datetime.timedelta(days=35, seconds=5743),
         )
+
+    def test_interval_literal_processor(self, connection):
+        result = connection.execute(
+            select(
+                literal(
+                    datetime.timedelta(days=1, minutes=3, seconds=4),
+                    literal_execute=True,
+                )
+                - literal(
+                    datetime.timedelta(days=0, minutes=1, seconds=4),
+                    literal_execute=True,
+                )
+            )
+        ).one()
+        eq_(result[0], datetime.timedelta(days=1, seconds=120))
+
+    def test_interval_value(self, connection):
+        """#9737"""
+        stmt = text("select :parameter from dual")
+        result = connection.execute(
+            stmt.bindparams(
+                parameter=datetime.timedelta(days=1, minutes=3, seconds=4),
+            )
+        ).one()
+        eq_(result[0], datetime.timedelta(days=1, minutes=3, seconds=4))
 
     def test_no_decimal_float_precision(self):
         with expect_raises_message(
--- a/test/dialect/postgresql/test_types.py
+++ b/test/dialect/postgresql/test_types.py
@@ -3180,6 +3180,31 @@ class TimestampTest(fixtures.TestBase, AssertsExecutionResults):
         expr = column("bar", postgresql.INTERVAL) == datetime.timedelta(days=1)
         eq_(expr.right.type._type_affinity, types.Interval)
 
+    def test_interval_literal_processor(self, connection):
+        result = connection.execute(
+            select(
+                literal(
+                    datetime.timedelta(days=1, minutes=3, seconds=4),
+                    literal_execute=True,
+                )
+                - literal(
+                    datetime.timedelta(days=0, minutes=1, seconds=4),
+                    literal_execute=True,
+                )
+            )
+        ).one()
+        eq_(result[0], datetime.timedelta(days=1, seconds=120))
+
+    def test_interval(self, connection):
+        """#9737"""
+        stmt = text("select :parameter")
+        result = connection.execute(
+            stmt.bindparams(
+                parameter=datetime.timedelta(days=1, minutes=3, seconds=4),
+            )
+        ).one()
+        eq_(result[0], datetime.timedelta(days=1, minutes=3, seconds=4))
+
 
 class SpecialTypesCompileTest(fixtures.TestBase, AssertsCompiledSQL):
     __dialect__ = "postgresql"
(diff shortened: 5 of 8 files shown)


lib/sqlalchemy/dialects/postgresql/types.py
```diff
@@ -247,6 +247,90 @@ def as_generic(self, allow_nulltype: bool = False) -> sqltypes.Interval:
...
+            def convert_total_seconds_to_interval(seconds):
+                # Define constants for time units
+                SECONDS_IN_MINUTE = 60
+                MINUTES_IN_HOUR = 60
+                HOURS_IN_DAY = 24
+                DAYS_IN_MONTH = 30.44  # Average days in a month
+                DAYS_IN_YEAR = 365.25  # Average days in a year
+
+                # Calculate years
+                years = seconds / (
```
lelit, line 261: Maybe divmod() is clearer/faster/on purpose here and below?

lib/sqlalchemy/dialects/postgresql/types.py
```diff
@@ -247,6 +247,90 @@ def as_generic(self, allow_nulltype: bool = False) -> sqltypes.Interval:
...
         return dt.timedelta
 
+    def literal_processor(self, dialect):
+        def process(value):
+            def convert_total_seconds_to_interval(seconds):
+                # Define constants for time units
+                SECONDS_IN_MINUTE = 60
+                MINUTES_IN_HOUR = 60
+                HOURS_IN_DAY = 24
+                DAYS_IN_MONTH = 30.44  # Average days in a month
```
lelit, line 257: I'd probably go with _precise_ transformation, and avoid going to months/years that necessarily involves using such math with _averages_... AFAICT, PostgreSQL will do the "right" thing when computing, say, `'2024-02-29'::date + '1 year'::interval`, that gives `2025-02-28`, and it could be a surprise that the same expression in Python (where the math is done in fractions of seconds) gives a different result.

lelit: To clarify what I meant, consider starting with `timedelta(seconds=1 * 365.25 * 24 * 60 * 60)`, it would surprise me if that would be rendered as `'1 year'::interval` : 
```
test=# select '2024-02-29'::date + '1 year'::interval;
      ?column?       
---------------------
 2025-02-28 00:00:00

test=# select '2024-02-29'::date + '31557600 seconds'::interval;
      ?column?       
---------------------
 2025-02-28 06:00:00
```

Even Python itself does not involve "imprecise" units:
```python
>>> timedelta(seconds=1 * 365.25 * 24 * 60 * 60)
datetime.timedelta(days=365, seconds=21600)
```

indiVar0508: Hi @lelit ,

Thanks for the feedback and explanation, i updated the approach, ditching the computing years and stuff bit, and directly using what timedelta gives.
saw this in [docs](https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-INTERVAL-INPUT), which should work ig?!
`Quantities of days, hours, minutes, and seconds can be specified without explicit unit markings. For example, '1 12:59:10' is read the same as '1 day 12 hours 59 min 10 sec'.`

indiVar0508: what i was also wondering was for how to control precision of seconds say if i have a interval as `0:00:3.23234234241234` and i want to round it to 3 decimal places for postgres i am not sure how to do it.
in the [docs](https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-DATETIME-INPUT) it says to do `type [ (p) ] 'value'` but where will it fit in expression is something i am not able to figure out, can you point me in the right direction here.
Thanks

lelit: I'm not sure I understand what you are asking: `select interval (3) '0:00:3.1289'` yields `00:00:03.129`, while `select (3) '0:00:3.23234234241234'` gives `00:00:03.232`, but you probably know that. If so, what do you mean with "_where will it fit in expression"_?

indiVar0508: Hi @lelit ,

`but you probably know that` - i am afraid and embarassed, i meant this only :"), i was trying like this `select interval '0:00:3.232323' second (3)` and it also worked for me (but doc said, type [ (p) ] 'value') so wasn't sure and confused, but thanks for helping!, i have updated my commit.

lelit: You're welcome!

lib/sqlalchemy/dialects/postgresql/types.py
```diff
@@ -247,6 +247,14 @@ def as_generic(self, allow_nulltype: bool = False) -> sqltypes.Interval:
     def python_type(self) -> Type[dt.timedelta]:
         return dt.timedelta
 
+    def literal_processor(self, dialect):
+        def process(value):
+            if value is not None:
+                value = f"'{str(value).replace(' days,', '')}'::interval"
```
CaselIT, line 253: seems we can also just do `f"'{value.total_seconds()}'::interval`

CaselIT: `f"make_interval(secs=>{value.total_seconds()})"` works too

lib/sqlalchemy/dialects/oracle/types.py
```diff
@@ -214,6 +214,25 @@ def as_generic(self, allow_nulltype=False):
...
+                        str(value).replace(" days,", "").replace(" day,", "")
+                    )
+                else:
+                    value = "0 " + str(value)
+                value = "INTERVAL '" + value + "' DAY"
+                if self.day_precision is not None:
+                    value += f"({self.day_precision})"
+                value += " TO SECOND"
+                if self.second_precision is not None:
+                    value += f"({self.second_precision})"
```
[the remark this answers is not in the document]
indiVar0508: Hi @CaselIT ,

I saw this method but wasn't sure how to specify precisions in this, but i have updated according to comment because i am realising that for literal query they might not need precision ig or if they require query could be something like `select round_method_for_respective_db_dialect(interval, precision_value) ...`.
have updated the query and tests accordingly

lib/sqlalchemy/dialects/postgresql/types.py
indiVar0508: Updated as per comments, went with `make_interval` method.
i was trying to write tests for execution of queries but was getting this error, have added for compile as of now, not able to figure out how to make tests point to specific db hosted in my local for testing pupose
`test.dialect.oracle.test_types.TypesTest' unsupported on any DB implementation 'sqlite(3.31.1)+pysqlite'`

lib/sqlalchemy/dialects/oracle/types.py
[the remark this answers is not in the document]
CaselIT: If I'm not mistaken precision or length is not taken into consideration when using literal formatting.
For example literal formatter of `String(10)` will not do `value_str[:10]`, so it should be fine to not have precision taken into consideration here.

@zzzeek do you have a general rule here?

lib/sqlalchemy/dialects/postgresql/types.py
CaselIT: by default the query runs using sqlite, that can't execute oracle specific code, so skip are automatically added when tests are marked as "backend".

You can use oracle via docker with the tests, check the readme unit test: https://github.com/sqlalchemy/sqlalchemy/blob/main/README.unittests.rst#docker-configurations
or I can just make it run the pipeline on gerrit ;)

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 9a6f34506780d5953c36eb36a3ace8755b607a3d of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 9a6f34506780d5953c36eb36a3ace8755b607a3d: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4873

lib/sqlalchemy/dialects/postgresql/types.py
CaselIT: mypy seems the only failure here. It logged
```
pep484: commands[0]> mypy ./lib/sqlalchemy
lib/sqlalchemy/dialects/postgresql/types.py:250: error: Function is missing a type annotation  [no-untyped-def]
lib/sqlalchemy/dialects/postgresql/types.py:251: error: Function is missing a type annotation  [no-untyped-def]
Found 2 errors in 1 file (checked 252 source files)
```

indiVar0508: Yeah i have added typing to method(s) but for some reason couldn't figure out how to resolve this, without modifying superclass returntype for now added no_type_check decorator ( saw that oracle had it for entire file, so added this for postgres in method only, and to keep it consistent also added it in oracle )
```py
error: Return type "Callable[[Optional[timedelta]], Optional[str]]" of "literal_processor" incompatible with return type "Optional[_LiteralProcessorType[timedelta]]" in supertype "TypeEngine"  [override]
```
added testcase for db specific backend testing also.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 3019e9fb5a9e7ac94a013a07236719fb6fcee4ee of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 3019e9fb5a9e7ac94a013a07236719fb6fcee4ee added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4873

zzzeek (changes requested):
any chance we can get a suite test ?  here's where the date related tests start: https://github.com/sqlalchemy/sqlalchemy/blob/aea28a9607d3673c680a4ebe7b9fb672db745d0a/lib/sqlalchemy/testing/suite/test_types.py#L455    so this would be a new suite IntervalTest with some basic round trips and literal processor tests.

indiVar0508:
Tried to add a few in [41528f6](https://github.com/sqlalchemy/sqlalchemy/pull/10383/commits/41528f6cb9d8c55e266731031fb8bb7cbc5f972a), you can review once if more are to be added or some changes needed to be done in these

lib/sqlalchemy/testing/suite/test_types.py
```diff
@@ -452,6 +453,100 @@ def test_concatenate_clauselist(self, connection):
         )
 
 
+class IntervalTest(_LiteralRoundTripFixture, fixtures.TestBase):
+    __requires__ = ("datetime",)
```
CaselIT, line 457: we probably need a new require here, called interval or similar

lib/sqlalchemy/testing/suite/test_types.py
```diff
@@ -452,6 +453,100 @@ def test_concatenate_clauselist(self, connection):
...
+    @testing.combinations(
+        [
+            (None, None),
+            (
+                datetime.timedelta(days=1, seconds=4),
+                datetime.timedelta(days=1, seconds=4),
+            ),
+        ]
+    )
+    def test_render_literal_interval(self, literal_round_trip):
```
CaselIT, line 490: combination seems unused, since you always use self.data

test/dialect/oracle/test_types.py
```diff
@@ -323,6 +350,16 @@ def test_interval(self, metadata, connection):
             datetime.timedelta(days=35, seconds=5743),
         )
 
+    def test_interval_literal_processor(self, connection):
```
CaselIT, line 353: this test does not seem to do what it says. no literal processor is used.

maybe you wanted to use `parameter=literal(..., literal_execute=True)`

test/dialect/postgresql/test_types.py
```diff
@@ -3180,6 +3180,16 @@ def test_interval_coercion_literal(self):
         expr = column("bar", postgresql.INTERVAL) == datetime.timedelta(days=1)
         eq_(expr.right.type._type_affinity, types.Interval)
 
+    def test_interval_literal_processor(self, connection):
```
CaselIT, line 3183: same comment as in the oracle file

CaselIT (changes requested):
Thanks for updating!

Can you also add a changelog here https://github.com/sqlalchemy/sqlalchemy/tree/main/doc/build/changelog/unreleased_20 

feel free to add "pull request courtesy of your name"!

lib/sqlalchemy/testing/suite/test_types.py
indiVar0508: Added requires you can review once went with name `datetime_interval`.

While adding only_on, just checked for other DB(s), where i think Interval is supported for arithmetic operations but it is not supported with sqla.

_MSSQL_ : https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/time-date-and-interval-functions?view=sql-server-ver16
_MySQL_ : https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html

CaselIT: not sure if there is a reason why they are supported only on pg and oracle. @zzzeek do you remember why?

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision bf3a73dfd9d329779e12037ae62dc1032e76d0f6 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset bf3a73dfd9d329779e12037ae62dc1032e76d0f6 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4873

indiVar0508:
Hi, 

It seemed some tests were failing, so fixed them some typo by me on comparing tuple vs datetime fixed that,
encountered a different issue in tests for now skipping them for oracle and added a Note on suite test.

It seems to be slight inconsistency in the behaviour of building query for datetime literal for postgres and oracle which causes this issue.
```py
import datetime
import sqlalchemy as sa


def main():
    engine = create_engine("postgresql+psycopg2://user_one:pass@127.0.0.1:5432/sqlalchemy_history_test")
    # engine = create_engine("oracle+cx_oracle://user_one:pass@127.0.0.1")
    conn = engine.connect()
    # This works!
    stmt = sa.select(sa.literal(datetime.datetime.now()) - sa.literal(datetime.timedelta(days=1), literal_execute=True))
    print(str(stmt.compile(dialect=engine.dialect, compile_kwargs={'literal_binds': True})))
    print(conn.execute(stmt).first())
    # This works!
    sql = 'SELECT :parameter - :parameter2 from dual' if engine.dialect.name == 'oracle' else 'SELECT :parameter - :parameter2'
    params = {'parameter': datetime.datetime.now(),'parameter2': datetime.timedelta(days=1)}
    stmt = sa.text(sql).bindparams(**params)
    print(str(stmt.compile(dialect=engine.dialect, compile_kwargs={'literal_binds': True})))
    print(conn.execute(stmt).first())
    # This Fails for postgres but works for oracle!

    # File "/github/sqlalchemy/lib/sqlalchemy/engine/base.py", line 1965, in _exec_single_context
    #   self.dialect.do_execute(
    # File "/github/sqlalchemy/lib/sqlalchemy/engine/default.py", line 921, in do_execute
    #   cursor.execute(statement, parameters)
    # sqlalchemy.exc.DataError: (psycopg2.errors.InvalidDatetimeFormat) invalid input syntax for type interval: "2023-09-29 12:53:23.380940"
    # LINE 1: SELECT '2023-09-29 12:53:23.380940' - make_interval(secs=>86...
    # [SQL: SELECT '2023-09-29 12:53:23.380940' - make_interval(secs=>86400.0) AS anon_1]
    # (Background on this error at: https://sqlalche.me/e/20/9h9h)

    stmt = sa.select(sa.literal(datetime.datetime.now(), literal_execute=True) - sa.literal(datetime.timedelta(days=1), literal_execute=True))
    print(str(stmt.compile(dialect=engine.dialect, compile_kwargs={'literal_binds': True})))
    print(conn.execute(stmt).first())

if __name__ == '__main__':
    main()
```

lib/sqlalchemy/dialects/oracle/types.py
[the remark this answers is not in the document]
indiVar0508: Hi,

Found a corner case on this, if i insert a timedelta of 100 days. then it fails `leading_precision is the number of digits in the leading field. Accepted values are 0 to 9. The default is 2.` so if i have a timedelta of 100 days which has 3 digits it fails.

```py
sqlalchemy.exc.DatabaseError: (cx_Oracle.DatabaseError) ORA-01873: the leading precision of the interval is too small
[SQL: INSERT INTO t (x) VALUES (NUMTODSINTERVAL(8640000.0, 'SECOND'))]
```

CaselIT:
Not too clear what's the issue.

pg wants the cast, IIRC sqlalchemy will add it if it knows the types of the columns.

What's the issue you get in the test where text is not used?

CaselIT:
@zzzeek do you remember if this case was meant to automatically cast in postgresql?

zzzeek:
well PG intervals have to be expressed in interval format, not as timestamps, so I know that much

CaselIT:
the issue does not seem to be on the interval, that uses `make_interval(secs=>86400.0)`, but in the datetime/time/date literal that's used for the subtraction. The test is of the type `select date_or_time_literal - the_interaval`

since this PR is for interval, we can probably just place a cast to the literal, so it renders `select cast(date_or_time_literal, the_type) - the_interval`. @indiVar0508 that should would and I don't think it will be rejected even in other databases.

sqla-tester:

if these are new tests then it's possible that subtracting a date literal and interval literal without casts just wont work.   i would look to see what the non-interval, date related tests are doing, if they have similar tests that are doing arithmetic + literal render + no casts.

I see in the failures a lot of trivial things like getting rows where scalars are expected etc., so I would just be getting as much of the test suite to pass as possible first, go for all green, then add trickier tests like arithmetic with literals.     I would want to review the code + basic test support first.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4873

indiVar0508:
Hi @zzzeek ,

I fixed those tests in 4425fb5f10df567277f5fd4fb8c9a0d553229711, and tested with both DB(s) they should work now.
Was having above mentioned issue for postgres and a microsecond issue for oracle vs postgres so for now skipped the test for oracle and added note [here](https://github.com/sqlalchemy/sqlalchemy/blob/4425fb5f10df567277f5fd4fb8c9a0d553229711/lib/sqlalchemy/testing/suite/test_types.py#L498) and [here](https://github.com/sqlalchemy/sqlalchemy/blob/4425fb5f10df567277f5fd4fb8c9a0d553229711/lib/sqlalchemy/testing/suite/test_types.py#L544) related to microseconds.

Hi @CaselIT , 

`we can probably just place a cast to the literal`
Got it, i will get basic tests verified and then update to cast method if required also did you get a chance to review [this](https://github.com/sqlalchemy/sqlalchemy/pull/10383#discussion_r1341032759), this was another issue which i mentioned in one the comment thread. which seems to have issue for oracle for precision on there default method to make intervals.

sqla-tester:

this PR is a lot of effort so let me run it and see what the current issues are.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4873

indiVar0508:
sure @zzzeek ,

i marked it as draft as #10513 is required for this ( to resolve Day to Second precision issue ) to make round_trip_tests work, 
rest all tests work fine i have updated some test in my local i can push, was waiting for other PR to finish first.

indiVar0508:
Rebased on top of main

lib/sqlalchemy/dialects/postgresql/types.py
```diff
@@ -247,6 +250,17 @@ def as_generic(self, allow_nulltype: bool = False) -> sqltypes.Interval:
...
         return dt.timedelta
 
+    @no_type_check
+    def literal_processor(
+        self, dialect: Optional[Dialect]
+    ) -> Callable[[Optional[dt.timedelta]], Optional[str]]:
+        def process(value: Optional[dt.timedelta]) -> Optional[str]:
+            if value is not None:
+                return f"make_interval(secs=>{value.total_seconds()})"
+            return None
```
sqla-tester, line 260: **mike bayer** (zzzeek) wrote:

this would return the string "NULL".  we dont typically implement or test None for the literal processors since I dont know that's too much of a real use case, ideally we'd address it all at once by having the compiler not actually call literal processors when the value is NULL in any case.  This does get into the issue of JSON "null" etc. but we have flags for that.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4873

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4873

test/dialect/postgresql/test_types.py
```diff
@@ -3180,6 +3180,31 @@ def test_interval_coercion_literal(self):
...
+        result = connection.execute(
+            select(
+                literal(
+                    datetime.timedelta(days=1, minutes=3, seconds=4),
+                    literal_execute=True,
```
sqla-tester, line 3188: **mike bayer** (zzzeek) wrote:

this is the wrong form for parameters also

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4873

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4873

* test/dialect/postgresql/test_types.py (line 6275): this should likely be merged above

sqla-tester:

this is good now thanks!

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4873

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4873 has been **merged**. Congratulations! :)

lib/sqlalchemy/dialects/oracle/types.py
```diff
@@ -214,6 +226,21 @@ def as_generic(self, allow_nulltype=False):
             day_precision=self.day_precision,
         )
 
+    @property
+    def python_type(self) -> Type[dt.timedelta]:
```
sqla-tester, line 230: **Federico Caselli** (CaselIT) wrote:

why this decorator?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4873

lib/sqlalchemy/dialects/postgresql/types.py
```diff
@@ -247,6 +250,17 @@ def as_generic(self, allow_nulltype: bool = False) -> sqltypes.Interval:
     def python_type(self) -> Type[dt.timedelta]:
         return dt.timedelta
 
+    @no_type_check
```
sqla-tester, line 253: **Federico Caselli** (CaselIT) wrote:

here too

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4873

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4873

lib/sqlalchemy/dialects/oracle/types.py
sqla-tester: **mike bayer** (zzzeek) wrote:

somehow i just didnt see that.  will remove

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4873

lib/sqlalchemy/dialects/postgresql/types.py
sqla-tester: **mike bayer** (zzzeek) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4873

lib/sqlalchemy/testing/suite/test_types.py
```diff
@@ -452,6 +453,123 @@ def test_concatenate_clauselist(self, connection):
...
+            Column(
+                "id", Integer, primary_key=True, test_needs_autoincrement=True
+            ),
+            Column("interval_data", cls.datatype),
+            Column("date_data", DateTime),
+            Column("decorated_interval_data", Decorated),
+        )
+
+    @testing.combinations(
+        (
```
indiVar0508, line 481: Hi, 
this is minor but just wanted to bring it up here, 
this testcase was failing with for days > 99days {Precision of more than 2 digits } with oracle DB for round trip test, due #10509, ig we don't have this testcase now with that scenario, in PR for #10513 only compilation tests are added not the round trip as literal processor thing which was being added in this PR.

indiVar0508: Yeah sure, 

So at that time Interval Parameter were ineffective for oracleDB due to issue-10509, 
But my Interval param were `day_precision=9` and `second_precision=9`, 

Now this will work since PR-10513 is merged.

the testcase was this

```py
    @testing.combinations(
        (
            datetime.timedelta(days=103, seconds=4),
            datetime.timedelta(days=103, seconds=4),
        ),
        argnames="input_, compare_",
    )
    def test_render_literal_interval(
        self, literal_round_trip, input_, compare_
    ):
        literal_round_trip(Interval(), [input_], [compare_])
```

i remvoed the 3 digit day test case to make CI work for simple testcases in this PR.

zzzeek: OK that is restored in  https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4939 if you want to review

indiVar0508: Yep looks good!

PR #10450, opened 2023-10-10
Update `table_per_related` example for Declarative API

miketheman:
### Description
When reading more of the Generic Associations, I found that the examples state "Annotated example", and wondered what that meant, since in other parts of the docs that means they are using the new 2.0 style.

I tried to update this example to be more in line with the new style, including a little f-string update.

I completely understand this is unlikely to be merged as-is - but wanted to understand more about the right way to use modern styles to properly code well-hinted, more "exotic" implementations.

Outstanding questions:
- Should examples pass pass `mypy --strict` ? It doesn't right now.
- Are there better ways to apply `Mapped` within the `type(...)` definition, so we could skip importing `Integer`?

### Checklist

This pull request is:

(Do non-running examples count as code fixes/features?)

**Have a nice day!**


zzzeek:
> I tried to update this example to be more in line with the new style, including a little f-string update.
> 
> I completely understand this is unlikely to be merged as-is - but wanted to understand more about the right way to use modern styles to properly code well-hinted, more "exotic" implementations.

it looks fine to me, it uses the new mapping styles and I assume it still works, we'd want to make sure all the generic assocaitions are on 2.0 style ?  (haven't looked)

> 
> Outstanding questions:
> 
>     * Should examples pass pass `mypy --strict` ? It doesn't right now.

it's not critical if we are just updating the examples to use `Mapped[]`, how far off of strict is it?



> 
>     * Are there better ways to apply `Mapped` within the `type(...)` definition, so we could skip importing `Integer`?

I would not seek to apply `Mapped` inside of `type()` since this is imperative code and typing is declarative.  I think the `Integer` can just be omitted completely since the column is a `ForeignKey()` to an existing column.

I know it's months later but if you are still around to work on this we could have all the generic assoc examples updated here and that would be helpful.

CaselIT:
sorry this was forgotten a bit

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision a6a54d69487429c8b3a7758cf22913e6a2156316 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change a6a54d69487429c8b3a7758cf22913e6a2156316: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5995

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5995 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5996 has been **merged**. Congratulations! :)

PR #9324, opened 2023-02-16
Move the metadata for `setuptools` into `PEP 621`-compliant `pyproject.toml`

KOLANICH:
PEP 621 is a standardized backend-neutral format for storing package metadata.

### Description
I had moved the metadata for `setuptools` there. Other configs residing there are not moved yet.

### Checklist

This pull request is:
- [x] A short code fix
- [x] A new feature implementation

PR #10481, opened 2023-10-16
Provide PEP-685-compliant extras

mattoberle:
### Description

`pip==23.3.0` introduces a new "canonical" lookup behavior for package extras:

- https://pip.pypa.io/en/stable/news/#v23-3
- https://github.com/pypa/pip/pull/12002

As a result, a command like this:

```
pip install 'sqlalchemy[postgresql_psycopg2binary]==1.4.47'
```

Results in a warning like this:

```
WARNING: sqlalchemy 1.4.47 does not provide the extra 'postgresql-psycopg2binary'
```

The extra eventually resolves when pip falls back to the original `postgresql_psycopg2binary`.
This PR proposes the addition of the [PEP-685](https://peps.python.org/pep-0685/)-compliant names to `setup.cfg`.

- Maintains backward compatibility for `pip<23.3.0`
- Avoids reliance on the "fallback" behavior of `pip>=23.3.0`

### Checklist

> Note: I'm not sure how to categorize this given the options in the PR template.
> This is a packaging change.

This pull request is:

- [ ] A short code fix
	- please include tests.


zzzeek (changes requested):
new pep not even two years ago already generating warnings...OK

here's what I want to do here:

1. can you confirm `postgresql-psycopg2binary` is the correct format (e.g. "normalized")
2. in the change here, the dashed names should be the primary names, as we will eventually remove the underscore names:


```
mssql-pyodbc = pyodbc
mssql_pyodbc =
    %(mssql-pyodbc)s
```

hoping the dash in the variable name there is accepted

3. put all the "legacy" underscore names in a separate batch way at the bottom so they are out of the way:

```
# legacy pre-pep-685 names
mssql_pymssql =
   %(mssql-pymssql)
mssql_pyodbc =
   %(mssql-pyodbc)
mysql_connector = 
   %(mysql-connector)

# ...

```

use the new names in our tox.ini file in the "extras" section:

```diff
diff --git a/tox.ini b/tox.ini
index ed68dbfcf7..675b80186b 100644
--- a/tox.ini
+++ b/tox.ini
@@ -18,20 +18,20 @@ extras=
      sqlite_file: aiosqlite
      sqlite_file: sqlcipher; python_version < '3.10'
      postgresql: postgresql
-     postgresql: postgresql_asyncpg
-     postgresql: postgresql_pg8000
-     postgresql: postgresql_psycopg
+     postgresql: postgresql-asyncpg
+     postgresql: postgresql-pg8000
+     postgresql: postgresql-psycopg
 
      mysql: mysql
      mysql: pymysql
      mysql: asyncmy
      mysql: aiomysql
-     mysql: mariadb_connector
+     mysql: mariadb-connector
 
      oracle: oracle
      oracle: oracle_oracledb
      py{3,37,38,39,310,311}-mssql: mssql
-     py{3,37,38,39,310,311}-mssql: mssql_pymssql
+     py{3,37,38,39,310,311}-mssql: mssql-pymssql
 
 install_command=
      # TODO: I can find no way to get pip / tox / anyone to have this

```

with all that we can run through gerrit make sure it all works

thanks!

zzzeek:
ok,lets try it

@CaselIT any comments?

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 2431850a1baca38afb82551edcec1cc7cd217da2 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 2431850a1baca38afb82551edcec1cc7cd217da2: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4906

mattoberle:
Thanks for taking a look!

---

>     1. can you confirm `postgresql-psycopg2binary` is the correct format (e.g. "normalized")

The change to pip relies on a function called `canonicalize_name`:
- https://github.com/uranusjr/pip/blob/9ba2bb90fb57e4ee5f624ecd39eade207863c21a/src/pip/_internal/metadata/base.py#L548-L584

Defined here:
- https://github.com/pypa/pip/blob/main/src/pip/_vendor/packaging/utils.py#L27-L35

That follows the PEP which defines the canonical name as:
```
re.sub(r"[-_.]+", "-", name).lower()
```

---

>     2. in the change here, the dashed names should be the primary names, as we will eventually remove the underscore names:
> ...
> hoping the dash in the variable name there is accepted

This was my concern as well, but I made the change and it looks like `%(var-with-dash)s` is acceptable.

---

>     3. put all the "legacy" underscore names in a separate batch way at the bottom so they are out of the way:
> ...
> use the new names in our tox.ini file in the "extras" section:

Done & done.

CaselIT:
> k,lets try it
> 
> @CaselIT any comments?

we must ensure that the fallback also works when using pyproject, an in the PR here https://github.com/sqlalchemy/sqlalchemy/pull/9324

also 2.1+ only or it's good also for 2.0?

sqla-tester:

we have to verify that the fallback works with PEP 621 (https://github.com/sqlalchemy/sqlalchemy/pull/9324). If not we have to find an alternative

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4906

zzzeek:
> > k,lets try it
> > @CaselIT any comments?
> 
> we must ensure that the fallback also works when using pyproject, an in the PR here #9324

well this is the official syntax apparently, so it has to work in pyproject and everywhere else

> 
> also 2.1+ only or it's good also for 2.0?

it looks like 1.4 users are getting this warning already so if this is safe to do for 2.0 at least, that would be good

CaselIT:
> > > k,lets try it
> > > @CaselIT any comments?
> > 
> > 
> > we must ensure that the fallback also works when using pyproject, an in the PR here #9324
> 
> well this is the official syntax apparently, so it has to work in pyproject and everywhere else

the pyformat thing `%(name)s` look a lot like something custom of cfg. As far as I know it's not supported in toml files.

what I'm saying is just to verify that this syntax works also when using toml, if not we can think of a workaround (likely just duplicating the actual requirements) and leave a note in the other PR mentioning that it is something to do before merging that one

zzzeek:
wouldnt the toml use quoted names to handle identifier characters?

CaselIT:
> wouldnt the toml use quoted names to handle identifier characters?

checked out https://github.com/sqlalchemy/sqlalchemy/pull/9324
trying to use 
```toml
postgresql-psycopg = ["psycopg>=3.0.7"]
postgresql_psycopg = %(postgresql-psycopg)s
```
makes pip raise `pip._vendor.tomli.TOMLDecodeError: Invalid value` so that does not work.

trying
```toml
postgresql-psycopg = ["psycopg>=3.0.7"]
postgresql_psycopg = "%(postgresql-psycopg)s"
```
makes pip raise `configuration error: `project.optional-dependencies.postgresql_psycopg` must be array`

trying
```toml
postgresql-psycopg = ["psycopg>=3.0.7"]
postgresql_psycopg = [".[postgresql-psycopg]"]
```
raises instead `configuration error: `project.optional-dependencies.postgresql_psycopg[0]` must be pep508`

suggestions?

zzzeek:
doesnt make much sense if the two peps conflict?  there has to be some simple way to do this

CaselIT:
> doesnt make much sense if the two peps conflict? there has to be some simple way to do this

the problem is the alias. There isn't any problem is we do
```toml
postgresql-psycopg = ["psycopg>=3.0.7"]
postgresql_psycopg = ["psycopg>=3.0.7"]
```

If we find no way of defining an alias that would work too, but should be noted in the other PR

zzzeek:
OK then we just define the `a-b` and `a_b` names direct to the thing they reference

the goal is:

a-b names point to their thing

a_b names are down at the bottom, pointing to whatever

CaselIT:
works for me. will leave a note on the other PR

setup.cfg
```diff
@@ -82,6 +82,30 @@ aiosqlite =
 sqlcipher =
     sqlcipher3_binary
 
+# legacy pre-pep-685 names
+mariadb_connector =
```
zzzeek, line 86: so, towards us being able to move to a pyproject.toml file easily, the request here is to not use the aliases for these legacy names and just point them to their ultimate dependencies directly (yes, duplicating what we have up above).

the goal here is that this batch of names is moved out to the margins where it can be removed at some point (maybe even 2.1, but maybe 2.2, etc) all at once.

zzzeek (changes requested):
@CaselIT just to confirm you want to change these names right now even though we're not on pyproject.toml yet, right?   maybe we can just wait until that time comes to change it further?

CaselIT:
> @CaselIT just to confirm you want to change these names right now even though we're not on pyproject.toml yet, right? maybe we can just wait until that time comes to change it further?

fine either way for me, I've left a comment in the other PR to do that change so we can delay it until then!

setup.cfg
CaselIT: so from my point of view we can delay all the "remove the alias" to the PR that migrates to pyproject. As far as I know the toml version does not support aliases, but maybe there is a way that I'm not aware of

CaselIT:
maybe hope is not lost: https://github.com/pypa/setuptools/issues/3214#issuecomment-1335314441

if I understand correctly we can use (I haven't tried it)
```toml
postgresql-psycopg = ["psycopg>=3.0.7"]
postgresql_psycopg = ["sqlalchemy[psycopg]"]
```

zzzeek:
I'm still on the fence if we should be looking at this for 2.0.  it seems a little soon, we could instead move right to 2.1 with pyproject.toml and just do the whole thing once.        thoughts?

CaselIT:
that works too for me

CaselIT:
This change here will be merged together with #9324, since it seems that when defined in pyproject a the extra `foo-bar` is automatically used when installing `[foo_bar]` and having only an extra `foo_bar` does not seem to work, at least in tox

sqla-tester:

will be merged with https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5028 since otherwise tox foes not work

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4906

CaselIT:
This was never backported to 2.0

I don't remember if we said it should

Issue #10650, opened 2023-11-17
pymysql Connection.ping AttributeError


luoye2018:
### Describe the bug

sqlalchemy with PyMySQL 0.9.3 ping cause exception.

### SQLAlchemy Version in Use

2.0.23 1.4.50

### DBAPI (i.e. the database driver)

pymysql

### Database Vendor and Major Version

MySQL 8

### Python Version

3.9

### Operating system

Linux

### To Reproduce

```python
from sqlalchemy import create_engine
from sqlalchemy import text

SQLALCHEMY_DATABASE_URI='mysql+pymysql://xxxx:xxxx@xxxx:3306/mysql'

engine = create_engine(SQLALCHEMY_DATABASE_URI, pool_pre_ping=True, pool_size=1)

def test():
    with engine.connect() as conn:
        conn.execute(text("select user from mysql.user limit 1"))

for _ in range(2):
    test()
```

### Error

```
Traceback (most recent call last):
  File "/workspace/test-env/../greatrds/.vscode/test.py", line 13, in <module>
    test()
  File "/workspace/test-env/../greatrds/.vscode/test.py", line 9, in test
    with engine.connect() as conn:
  File "/workspace/test-env/venv/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 3268, in connect
    return self._connection_cls(self)
  File "/workspace/test-env/venv/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 145, in __init__
    self._dbapi_connection = engine.raw_connection()
  File "/workspace/test-env/venv/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 3292, in raw_connection
    return self.pool.connect()
  File "/workspace/test-env/venv/lib/python3.9/site-packages/sqlalchemy/pool/base.py", line 452, in connect
    return _ConnectionFairy._checkout(self)
  File "/workspace/test-env/venv/lib/python3.9/site-packages/sqlalchemy/pool/base.py", line 1378, in _checkout
    del fairy
  File "/workspace/test-env/venv/lib/python3.9/site-packages/sqlalchemy/util/langhelpers.py", line 146, in __exit__
    raise exc_value.with_traceback(exc_tb)
  File "/workspace/test-env/venv/lib/python3.9/site-packages/sqlalchemy/pool/base.py", line 1306, in _checkout
    result = pool._dialect._do_ping_w_event(
[...]

PR #10651, opened 2023-11-17
Fixes: #10650 pymysql no attribute ping

luoye2018:
### Description
Fixes: #10650 pymysql no attribute ping

**Have a nice day!**


CaselIT:
Thanks but gonna fix another way.

Issue #10656, opened 2023-11-19
sqlalchemy need to relax certain server capacities assumption and better handle the errors to work with more databases


LLuke:
### Describe the use case

A fix will allow sqlalchemy to work with some database servers implemented enough postgres wire protocol but not all the PostgreSQL features assumed by sqlalchemy.

### Description
sqlalchemyattempted to use queries in 

```
 engine.default.get_server_version_info()
 engine.default.._get_default_schema_name()
 engine.default..get_default_isolation_level()
```
and
```
 dialects.postgresql.PGDialect.self._set_backslash_escapes()
```

to probe and enable features for PostgreSQL and compatible databases. While there are good reasons, they could fail with unhandled errors for database servers that support enough postgres wire protocol and features but otherwise doesn't support these particular functionalities. In particular, sqlalchemy assumed the error would always be NotImplementedError in get_server_version_info() and proceeded to use  server_version_info without further check the None condition arise from NotImplementedError. However, for drivers like pg8000, errors like ProgrammingError would be thrown instead. To make sqlalchemy usable for the many different drivers and servers combinations, it is desirable to make sqlalchemy as lenient as possible without detrimental side effects.

### Databases / Backends / Drivers targeted

A fix is tested with arcadedb 23.10.1 with pg8000. It is possible this will also fix for other servers implemented enough postgres wire protocols

### Example Use

After the fix, the following code works
```
from sqlalchemy import create_engine
import pandas as pd

engine = create_engine(
    'postgresql+pg8000://' + os.environ['DB_USER'] + ':' + os.environ['DB_PASS'] + '@localhost:3333/arcade_testdb', isolation_level="AUTOCOMMIT",
    connect_args={'ssl_context': None}
)
df = pd.read_sql_query("select * from schema:database", con=engine)
engine.dispose()
```

CaselIT:
Hi,

In general sqlalchemy tries to avoid having to support the many flavours and combinations of database that kinda but not really support the postgresql protocol.

The reason is that claiming to do so will lead to a endless stream of issues were people try to use a particular feature that's not implemented or works in a subtly different way.

The usual solution is to create an external dialect for it, that can be based on the internal pg one.

Let's wait for Mike opinion, but personally I'm -1 on this

zzzeek:
hi - 

[...]

PR #10657, opened 2023-11-19
patch for servers not supporting certain assumed postgresql capabilities

LLuke:
This issue is described in [#10656](https://github.com/sqlalchemy/sqlalchemy/issues/10656) with test code there. It is a minimal functionality fix. A caveat is that this fix doesn't deal with ramifications of handling failure of _set_backslash_escapes() which was not in the original sqlalchemy code base either.

Due to otherwise needs to setup an external database, unit tests were not added besideds the test code in the issue.  It is not expected to break any current test cases. It is recommended that sqlacademy to add mocks for all assumptions and handle them as gracefully as possible.

sqlalchemy is widely used and we hope that it is even more widely used. :-)

### Checklist

This pull request is:

- [X ] A short code fix
- [] A new feature implementation
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -3049,15 +3049,18 @@ class PGDialect(default.DefaultDialect):
         super().initialize(connection)
 
         # https://www.postgresql.org/docs/9.3/static/release-9-2.html#AEN116689
-        self.supports_smallserial = self.server_version_info >= (9, 2)
+        self.supports_smallserial = self.server_version_info and self.server_version_info >= (9, 2)
 
-        self._set_backslash_escapes(connection)
+        try: 
+            self._set_backslash_escapes(connection)
+        except:
+            pass
 
-        self._supports_drop_index_concurrently = self.server_version_info >= (
+        self._supports_drop_index_concurrently = self.server_version_info and self.server_version_info >= (
             9,
             2,
         )
-        self.supports_identity_columns = self.server_version_info >= (10,)
+        self.supports_identity_columns = self.server_version_info and self.server_version_info >= (10,)
 
     def get_isolation_level_values(self, dbapi_conn):
         # note the generic dialect doesn't have AUTOCOMMIT, however
--- a/lib/sqlalchemy/engine/default.py
+++ b/lib/sqlalchemy/engine/default.py
@@ -514,20 +514,20 @@ class DefaultDialect(Dialect):
             self.server_version_info = self._get_server_version_info(
                 connection
             )
-        except NotImplementedError:
+        except :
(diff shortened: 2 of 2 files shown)


lib/sqlalchemy/engine/default.py
```diff
@@ -514,20 +514,20 @@ def initialize(self, connection):
             self.server_version_info = self._get_server_version_info(
                 connection
             )
-        except NotImplementedError:
+        except :
```
zzzeek, line 517: we can't suppress all errors here, those would need to be propagated

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -3049,15 +3049,18 @@ def initialize(self, connection):
...
+        except:
+            pass
 
-        self._supports_drop_index_concurrently = self.server_version_info >= (
+        self._supports_drop_index_concurrently = self.server_version_info and self.server_version_info >= (
             9,
             2,
         )
-        self.supports_identity_columns = self.server_version_info >= (10,)
+        self.supports_identity_columns = self.server_version_info and self.server_version_info >= (10,)
```
zzzeek, line 3063: server_version_info must be provided here, it's invalid state for this to not be present

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -3049,15 +3049,18 @@ def initialize(self, connection):
         super().initialize(connection)
 
         # https://www.postgresql.org/docs/9.3/static/release-9-2.html#AEN116689
-        self.supports_smallserial = self.server_version_info >= (9, 2)
+        self.supports_smallserial = self.server_version_info and self.server_version_info >= (9, 2)
 
-        self._set_backslash_escapes(connection)
+        try: 
+            self._set_backslash_escapes(connection)
+        except:
```
zzzeek, line 3056: this would cause silent errors

zzzeek (commented):
overall per discussion in #10656 this is not something we do on our end

lib/sqlalchemy/dialects/postgresql/base.py
LLuke: Agree., also my concern when I hacked it to hear from you. 

A database should better comply but I'm curious whether sqlalchemy has a fallback solution when a database for whatever reason doesn't support this. Is there a legacy solution in history or do we know for sure where needs to be retooled for people to support that dialect without the trials and errors?

CaselIT: the "this" here is custom code that's written to support postgresql.

If a database does need to do something different, the dialect can decide to handle it in a different way.

PR #10719, opened 2023-11-30
Update `TZDateTime` type decorator example to align with python docs

kkirsche:
### Description
This change updates the `TZDateTime` type decorator to use the timezone awareness checks described in the Python documentation located here:

https://docs.python.org/3/library/datetime.html#determining-if-an-object-is-aware-or-naive

The specific lines state:

> A [`datetime`](https://docs.python.org/3/library/datetime.html#datetime.datetime) object `d` is aware if both of the following hold:
>
> `d.tzinfo is not None`
>
> `d.tzinfo.utcoffset(d)` does not return `None`
>
> Otherwise, `d` is naive.

### Checklist

This pull request is:

- [X] A documentation / typographical / small typing error fix
- [ ] A short code fix
	- please include tests.

### DIFF
--- a/doc/build/core/custom_types.rst
+++ b/doc/build/core/custom_types.rst
@@ -156,7 +156,7 @@ denormalize::
 
         def process_bind_param(self, value, dialect):
             if value is not None:
-                if not value.tzinfo:
+                if not value.tzinfo and not value.tzinfo.utcoffset(value):
                     raise TypeError("tzinfo is required")
                 value = value.astimezone(datetime.timezone.utc).replace(tzinfo=None)
             return value


doc/build/core/custom_types.rst
```diff
@@ -156,7 +156,7 @@ denormalize::
 
         def process_bind_param(self, value, dialect):
             if value is not None:
-                if not value.tzinfo:
+                if not value.tzinfo and not value.tzinfo.utcoffset(value):
```
CaselIT, line 159: This requires `is not None` otherwise it will raise for utc dates
`bool(datetime.timedelta(0)) is False`

CaselIT (changes requested):

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision bb30cb3cfe57f326addec21a6cae5f81184c2e74 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change bb30cb3cfe57f326addec21a6cae5f81184c2e74: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4997

doc/build/core/custom_types.rst
CaselIT: also need or

CaselIT: fixed in gerrit

CaselIT:
looking at the docs they could have provided a method / property ...

kkirsche:
> looking at the python docs they could have provided a method / property ...

Fully agreed



> 
> There are places in the code where the test used is only `dt_instance.tzinfo is not None`.
> 
> Do you think it would make sense changing it to the complete test suggested in the python docs at https://docs.python.org/3/library/datetime.html#determining-if-an-object-is-aware-or-naive?
> 
> View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4997

I think it makes sense to change it to match the documentation 1:1 and include a comment or docstring with a link to the Python documentation. I know in the codebases I'm responsible for, I wrapped their method in a function so that if this ever changes, I have both a link to original Python documentation and can change it across the codebase without having to locate each individual instance.

Looking at the repo, I see at least two locations in [lib/sqlalchemy/sql/sqltypes.py](https://github.com/sqlalchemy/sqlalchemy/blob/1669ae65da251bfe7c8afb9b6a983bec81e8ac1b/lib/sqlalchemy/sql/sqltypes.py#L795) that are using this check.

As a result, would it make sense to create a function in https://github.com/sqlalchemy/sqlalchemy/tree/main/lib/sqlalchemy/util to check whether a datetime is aware or naïve?

CaselIT:
Yes it may make sense. I don't think the two conditions will ever disagree, but it's probably better to do as the docs says

@zzzeek what do you think?

sqla-tester:

can we backport?  thanks

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4997

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/4997 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5026 has been **merged**. Congratulations! :)

PR #10831, opened 2024-01-05
Documenting multiprocessing and events

jamesbraza:
### Description

Promoting info from https://github.com/sqlalchemy/sqlalchemy/discussions/10828 to the event docs. Thanks to @zzzeek and @CaselIT for their answers.

### Checklist

This pull request is:

### DIFF
--- a/doc/build/core/event.rst
+++ b/doc/build/core/event.rst
@@ -140,23 +140,21 @@ this value can be supported::
     # it to use the return value
     listen(UserContact.phone, "set", validate_phone, retval=True)
 
-Multiprocessing Support
------------------------
+Events and Multiprocessing
+--------------------------
 
-.. tip::
+Note that events and propagation via hooks occur locally to a Python process.
+In other words, events only propagate objects within the originating Python process.
+Thus, when using Python multiprocessing
+(or other approaches to create separate Python processes),
+there is no means by which a SQLAlchemy event gets transmitted to other processes.
 
-  Note that events and propagation via hooks occur locally to a Python process.
-  In other words, events only propagate objects within the originating Python process.
-  Thus, when using Python multiprocessing
-  (or other approaches to create separate Python processes),
-  there is no means by which a SQLAlchemy event gets transmitted to other processes.
-
-  What happens if one forks the current Python process?
-  If a Python process forks a new child process,
-  existing event listeners will be copied to the child process at that point,
-  along with the rest of the Python interpreter state,
-  but there is no intrinsic mechanism for event listener registration
-  to propagate to other already existing processes.
+What happens if one forks the current Python process?
+If a Python process forks a new child process,
+existing event listeners will be copied to the child process at that point,
+along with the rest of the Python interpreter state,
+but there is no intrinsic mechanism for event listener registration
+to propagate to other already existing processes.
 
 Event Reference
 ---------------
--- a/doc/build/core/event.rst
+++ b/doc/build/core/event.rst
@@ -140,6 +140,17 @@ this value can be supported::
     # it to use the return value
     listen(UserContact.phone, "set", validate_phone, retval=True)
 
+Multiprocessing Support
+-----------------------
+
+The event system supports multiprocessing in a limited manner.
+The parent process can listen to events from within itself or child processes,
+assuming the child process was spawned after the parent has begun listening.
+
+If its desired to share events across two unrelated processes,
+this should be handled through an external mechanism,
+such as a messaging system.
+
 Event Reference
 ---------------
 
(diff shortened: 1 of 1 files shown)


doc/build/core/event.rst
```diff
@@ -140,6 +140,17 @@ this value can be supported::
     # it to use the return value
     listen(UserContact.phone, "set", validate_phone, retval=True)
 
+Multiprocessing Support
+-----------------------
+
+The event system supports multiprocessing in a limited manner.
```
zzzeek, line 146: I disagree with the wording "in a limited manner", from my perspective it does not make any sense for events to "support multiprocessing", since event hooks pass live Python objects around.

This should be a simple ".. tip" section, and just mention, "note that when using Python multiprocessing or other approaches to create separate Python processes, events as they occur are only propagated within the same Python process in which they originated, there is no means by which an event would be transmitted to other running processes."

"In addition, establishment of event hooks themselves occurs locally to a Python process.   If a Python process forks a child process, existing event listeners will be copied to the child process at that point along with the rest of the Python interpreter state, but there is no intrinsic mechanism for event listener registration to propagate to other already existing processes."


will of course propagate from parent to child process, but an event hook established within a child process will not be present in the parent process, or in other sibling processes"

doc/build/core/event.rst
```diff
@@ -140,6 +140,17 @@ this value can be supported::
...
+The parent process can listen to events from within itself or child processes,
+assuming the child process was spawned after the parent has begun listening.
+
+If its desired to share events across two unrelated processes,
```
zzzeek, line 150: IMO prescriptive advice like this is out of scope for these docs.

doc/build/core/event.rst
```diff
@@ -140,6 +140,17 @@ this value can be supported::
     # it to use the return value
     listen(UserContact.phone, "set", validate_phone, retval=True)
 
+Multiprocessing Support
+-----------------------
+
+The event system supports multiprocessing in a limited manner.
+The parent process can listen to events from within itself or child processes,
+assuming the child process was spawned after the parent has begun listening.
```
zzzeek, line 148: that's not accurate either.   if an event happens in the child process, it is only intercepted in that child process and nowhere else.

zzzeek (changes requested):

doc/build/core/event.rst
jamesbraza: Okay thank you, I have removed this info

doc/build/core/event.rst
jamesbraza: Okay sounds good! It's removed now

doc/build/core/event.rst
jamesbraza: Thank you @zzzeek for putting in some real thought into this, appreciated. I have incorporated all of this into the code. I made it a `tip` section.

Let me know if you think the `tip` should be moved to a separate section, or if it can stay in its own `Multiprocessing Support` section

jamesbraza:
Any chance I can bug for a re-review here? Just looking to keep this moving forward

doc/build/core/event.rst
zzzeek: OK I think I misspoke when I used the phrase "tip section" , as this is now an entirely independent section with just a tip inside of it :)  

let's do this:

1. section is named "Events and Multiprocessing", where you have the section right now
2.  it's not a "tip" since this is a section of its own
3. text, will not start with "note that" since this is a first class section.   We also try to stick with more formal language outside of tutorials, so lets go with this

> SQLAlchemy's event hooks are implemented as Python data structures associated with a particular pair of Python functions and objects.   Event propagation itself is implemented as Python function calls.

> From this model, we can describe the behavior of the event system in terms of multiple Python processes, such as a parent child forking a child process, in the same way we would think about any other datastructure / function operation.     Event hooks registered in a parent process will be present in new child processes that are forked from that parent **after** the hooks have been registered, since the child process starts with a copy of all existing Python structures from the parent when it begins.  Child processes that already exist **before** the hooks are registered will **not** receive those new event hooks, as changes made to Python structures in a parent process do not propagate to child processes.

> For the events themselves, these are Python function calls, which do not have any ability to propagate between processes.  SQLAlchemy's event system does not implement any inter-process communication.    It is possible to implement event hooks that use Python inter-process messaging within them however this would need to be implemented by the user.

jamesbraza: Nice, thank you I like your changes and implemented them.  I did tweak the wording slightly as I felt it could be more concise, and have pushed the code. Feel free to take over this PR and directly commit any changes you think are necessary from here

zzzeek (approved):
it's great, thanks!

jamesbraza:
And thank you @zzzeek, appreciate your open-ness! Your open minded-ness doesn't always exist in other open source maintainers when it comes to docs changes, so I appreciate it 🙏

PR #10851, opened 2024-01-09
Parse NOT NULL for MySQL generated columns

### LINKED ISSUE #10850: [MySQL] Nullability of generated columns is not inspected correctly
### Describe the bug

The inspector does not pick up the nullability of generated columns for MySQL.

_No response_

### SQLAlchemy Version in Use

2.0.22

MySQL

MySQL 8

### Python Version

Python 3.11

### Operating system

Linux

### To Reproduce

```python
import os
from sqlalchemy import sql, create_engine, inspect

# Create a table with a generated column in MySQL

engine = create_engine(os.environ["DB_URL"])
conn = engine.connect()


ddl = """CREATE TABLE IF NOT EXISTS mytable (
    my_generated_column int GENERATED ALWAYS AS (1234) VIRTUAL NOT NULL
);"""
conn.execute(sql.text(ddl))

# Try inspecting the generated column
inspector = inspect(conn)
cols = inspector.get_columns("mytable")
cols = [col for col in cols if col["name"] == "my_generated_column"]
assert cols[0]["nullable"] == False
```

### Error

```
  File "reproduction.py", line 19, in <module>
    assert cols[0]["nullable"] == False
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

### Additional context

_No response_

GeorchW:
### Description

As per the MySQL CREATE TABLE specification (see https://dev.mysql.com/doc/refman/8.0/en/create-table.html), the location of the NULL/NOT NULL syntax is slightly different for generated vs. regular columns: in generated columns, the NULL/NOT NULL goes after the VIRTUAL/STORED instead of directly after the data type.

Fixes #10850

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/test/dialect/mysql/test_reflection.py
+++ b/test/dialect/mysql/test_reflection.py
@@ -778,23 +778,26 @@ class ReflectionTest(fixtures.TestBase, AssertsCompiledSQL):
         ).first()
         explicit_defaults_for_timestamp = row[1].lower() in ("on", "1", "true")
 
-        reflected = []
-        for idx, cols in enumerate(
+        test_cases = [
             [
-                [
-                    "x INTEGER NULL",
-                    "y INTEGER NOT NULL",
-                    "z INTEGER",
-                    "q TIMESTAMP NULL",
-                ],
-                ["p TIMESTAMP NULL DEFAULT CURRENT_TIMESTAMP"],
-                ["r TIMESTAMP NOT NULL"],
-                ["s TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP"],
-                ["t TIMESTAMP"],
-                ["u TIMESTAMP DEFAULT CURRENT_TIMESTAMP"],
-                ["v INTEGER GENERATED ALWAYS AS (4711) VIRTUAL NOT NULL"],
-            ]
-        ):
+                "x INTEGER NULL",
+                "y INTEGER NOT NULL",
+                "z INTEGER",
+                "q TIMESTAMP NULL",
+            ],
+            ["p TIMESTAMP NULL DEFAULT CURRENT_TIMESTAMP"],
+            ["r TIMESTAMP NOT NULL"],
+            ["s TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP"],
+            ["t TIMESTAMP"],
+            ["u TIMESTAMP DEFAULT CURRENT_TIMESTAMP"],
+            ["v INTEGER GENERATED ALWAYS AS (4711) VIRTUAL NOT NULL"],
+        ]
+        if connection.dialect.supports_notnull_generated_columns:
+            test_cases.append(
+                ["v INTEGER GENERATED ALWAYS AS (4711) VIRTUAL NOT NULL"])
+
+        reflected = []
+        for idx, cols in enumerate(test_cases):
             Table("nn_t%d" % idx, meta)  # to allow DROP
 
             connection.exec_driver_sql(
--- a/test/dialect/mysql/test_reflection.py
+++ b/test/dialect/mysql/test_reflection.py
@@ -778,94 +778,140 @@ class ReflectionTest(fixtures.TestBase, AssertsCompiledSQL):
         ).first()
         explicit_defaults_for_timestamp = row[1].lower() in ("on", "1", "true")
 
-        test_cases = [
-            [
-                "x INTEGER NULL",
-                "y INTEGER NOT NULL",
-                "z INTEGER",
-                "q TIMESTAMP NULL",
-            ],
-            ["p TIMESTAMP NULL DEFAULT CURRENT_TIMESTAMP"],
-            ["r TIMESTAMP NOT NULL"],
-            ["s TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP"],
-            ["t TIMESTAMP"],
-            ["u TIMESTAMP DEFAULT CURRENT_TIMESTAMP"],
-            ["v INTEGER GENERATED ALWAYS AS (4711) VIRTUAL NOT NULL"],
-        ]
-        if connection.dialect._supports_notnull_generated_columns:
-            test_cases.append(
-                ["v INTEGER GENERATED ALWAYS AS (4711) VIRTUAL NOT NULL"])
-
-        reflected = []
-        for idx, cols in enumerate(test_cases):
-            Table("nn_t%d" % idx, meta)  # to allow DROP
-
-            connection.exec_driver_sql(
-                """
-                    CREATE TABLE nn_t%d (
-                        %s
-                    )
-                """
-                % (idx, ", \n".join(cols))
-            )
-
-            reflected.extend(
-                {
-                    "name": d["name"],
-                    "nullable": d["nullable"],
-                    "default": d["default"],
-                }
-                for d in inspect(connection).get_columns("nn_t%d" % idx)
-            )
-
         if connection.dialect._is_mariadb_102:
             current_timestamp = "current_timestamp()"
         else:
             current_timestamp = "CURRENT_TIMESTAMP"
 
-        eq_(
-            reflected,
-            [
-                {"name": "x", "nullable": True, "default": None},
-                {"name": "y", "nullable": False, "default": None},
-                {"name": "z", "nullable": True, "default": None},
-                {"name": "q", "nullable": True, "default": None},
-                {"name": "p", "nullable": True, "default": current_timestamp},
+        test_cases = [
+            {
+                "ddl_columns": [
+                    "x INTEGER NULL",
+                    "y INTEGER NOT NULL",
+                    "z INTEGER",
+                    "q TIMESTAMP NULL",
+                ],
+                "expected_reflected": [
+                    {"name": "x", "nullable": True, "default": None},
+                    {"name": "y", "nullable": False, "default": None},
+                    {"name": "z", "nullable": True, "default": None},
+                    {"name": "q", "nullable": True, "default": None},
+                ],
+            },
+            {
+                "ddl_columns": ["p TIMESTAMP NULL DEFAULT CURRENT_TIMESTAMP"],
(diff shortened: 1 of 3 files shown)


GeorchW:
btw, the other dialects use the database's metadata views (which would be `information_schema` in MySQL) instead of parsing `CREATE TABLE` statements. Is this a compatibility thing in the MySQL dialect or something? Or is it just "not done yet"? It feels a little spooky to me to parse context-free languages with regular expressions, although I must acknowledge that it's probably "good enough" for the vast majority of cases (i.e. as long as expressions don't contain keywords in strings).

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 5cce2f4208e8b4cd179e67d668720feaa3ff5f04 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 5cce2f4208e8b4cd179e67d668720feaa3ff5f04: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5102

zzzeek:
this PR is great!  thanks for just jumping on it!

GeorchW:
Force-pushed to (hopefully) make CI work, was just a missing trailing comma.

@zzzeek That makes sense. Didn't know about the situation in earlier MySQL.

GeorchW:
@CaselIT 
> Regarding the change, does mariadb behave the same?

[The MariaDB docs](https://mariadb.com/kb/en/generated-columns/) claim that the syntax is compatible, but they don't say anything about nullable generated columns as far as I can see. I just tried both variants:

```sql
CREATE TABLE IF NOT EXISTS mytable ( 
    my_generated_column int  GENERATED ALWAYS AS (1234) VIRTUAL NOT NULL 
);
CREATE TABLE IF NOT EXISTS mytable (
    my_generated_column int NOT NULL GENERATED ALWAYS AS (1234) VIRTUAL
);
```
Neither works. Tested on MariaDB 11.2.2 (via Docker). When leaving out the `NOT NULL` alltogether, the column is considered to be nullable.

So... yeah, that's another bug, I think it's not possible to create non-nullable generated columns with MariaDB+SQLAlchemy right now (not sure though). This PR shouldn't change anything about that, though.

sqla-tester (commented):
Hi, this is **sqla-tester** and I see you've pinged me for review. However, user **GeorchW** is not authorized to initiate CI jobs.  Please wait for a project member to do this!

CaselIT:
> So... yeah, that's another bug, I think it's not possible to create non-nullable generated columns with MariaDB+SQLAlchemy right now (not sure though).

Yes I remember something along there lines for one db.
Maybe mysql added support for it after sqlalchemy implemented computed?

In any case as long as there are no regression with the new regexp (and there should not be) on mariadb then it's fine

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 7cd93171299cc072fcdec581a422c0b9a7ab8d67 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 7cd93171299cc072fcdec581a422c0b9a7ab8d67 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5102

sqla-tester:

recheck

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5102

GeorchW:
https://jenkins.sqlalchemy.org/job/sqlalchemy_21_gerrit/cext=cext,db=mysql,pyv=py311/283/testReport/

Hmm, the tests complain that the specific syntax used in the test is not supported in MariaDB and old MySQL versions...

What's your opinion on how to deal with this? Can/should I somehow differentiate between the backends in the test, only adding the new test case for MySQL versions that actually support non-nullable generated columns?

GeorchW:
I've adjusted the tests to check for the MySQL version, it seems like there are already similar checks in similar places. I think you can recheck if the approach is fine

CaselIT:
> https://jenkins.sqlalchemy.org/job/sqlalchemy_21_gerrit/cext=cext,db=mysql,pyv=py311/283/testReport/
> 
> Hmm, the tests complain that the specific syntax used in the test is not supported in MariaDB and old MySQL versions...
> 
> What's your opinion on how to deal with this? Can/should I somehow differentiate between the backends in the test, only adding the new test case for MySQL versions that actually support non-nullable generated columns?

looking at the update and at the fail I think this should be only a test constraint that we add, since the dialect has no use of that flag.

I'll suggest something later today so that the test automatically skips some test cases depending on the db in use

CaselIT:
refactoring that test does not seem to be warranted. Can you instead of creating an attribute use a property and maybe underscore it since it's a test only thing?

thanks!

GeorchW:
I made it an attribute since the other fields were attributes, too, but sure, I can change that

CaselIT:
well I was looking at the property `_is_mariadb_102` that's a similar one that used only by the test

GeorchW:
@CaselIT Ah right, gotcha. Adjusted it

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 0d58aaa81815264c571ca9ccaca3d256c48d74c7 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 0d58aaa81815264c571ca9ccaca3d256c48d74c7 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5102

GeorchW:
Sorry, my bad, the provided code was broken; I had adjusted the test case to only be included for compatible MySQL versions, but did not alter the expected outcome accordingly.

I've now refactored the test a bit such that SQL DDL input and expected reflected properties are right next to each other, which IMO makes it a bit easier to read (but also a bit longer). Tell me what you think.

The tests take forever to execute on my machine. Even just the initialization of `tox` takes almost a minute. Is there a way to speed this up?

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision fb9a81020c393231ca90a1e88342b11cf64414a1 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset fb9a81020c393231ca90a1e88342b11cf64414a1 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5102

test/dialect/mysql/test_reflection.py
```diff
@@ -778,89 +778,140 @@ def test_nullable_reflection(self, metadata, connection):
...
-        reflected = []
-        for idx, cols in enumerate(
-            [
-                [
+        if connection.dialect._is_mariadb_102:
+            current_timestamp = "current_timestamp()"
+        else:
+            current_timestamp = "CURRENT_TIMESTAMP"
+
+        test_cases = [
```
sqla-tester, line 786: **Michael Bayer** (zzzeek) wrote:

so if we were going to refactor this test, we should do it using `@testing.combinations`.

Here's the options for how we can proceed on this part.


1. we merge it as is, then I will go in after the fact and reorganize this into multiple, independent tests using `@testing.combinations`
2. the submitter Georg reorganizes this using `@testing.combinations` directly.  This is some new API to learn so I dont want to impose this on you unless you were super interested.
3.  I take over the gerrit here (we're not very good at updating github PRs) and I finish this up inline.   since I can work extremely quickly

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5102

sqla-tester (commented):

so this needs a changelog file and I also want to refactor that test to use a more modern form which is `@testing.combinations`, this is a facade on top of pytest parametrize (https://docs.pytest.org/en/7.1.x/example/parametrize.html) that generates individual tests per case.    See inline comment for options on how to proceed.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5102

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5102

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5102

test/dialect/mysql/test_reflection.py
GeorchW: I'm not particularly keen on reorganizing the test case, so not having to do (2) would be nice. Otherwise, I don't really mind whether it's merged before or after

CaselIT: no worries

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5102

test/dialect/mysql/test_reflection.py
sqla-tester: **Michael Bayer** (zzzeek) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5102

sqla-tester:

that's an unrelated failure

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5121

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5121 has been **merged**. Congratulations! :)

GeorchW:
nice, thank you so much! I guess the next release that contains this change is going to happen in the next month or so?

CaselIT:
Yes, it's usually about once a month, so more or less in a couple of weeks

PR #10889, opened 2024-01-15
Update orderinglist annotations, list compatibity

### LINKED ISSUE #10888: OrderingList typing incorrect / incomplete and implementation doesn't accept non-int indices
### Describe the bug

There are a number of issues with the sqlalchemy.ext.orderinglist:OrderingList type hints and implementation:

1. The ordering function is documented as being allowed to return any type of object, and the tests demonstrate this by returning strings. The type annotation requires it to return integers, only, however.
2. The `OrderingList` signature marks the `ordering_attr` as optional, but accepting `None` here would lead to a type error when `getattr(entity, self.ordering_attr)` is called. 
3. List objects accept any `SupportsIndex` type object, one that implements the [`.__index__()` method](https://docs.python.org/3/reference/datamodel.html#object.__index__). The codebase assumes it is always an integer (`count_from_1()` and `count_from_n_factory()` use addition with the value, which `SupportsIndex` objects don't need to support). It's probably best to use `operator.index()` on the value before passing it to the ordering function.
4. The `ordering_list()` ignores the `_T` type variable in the `OrderingFunc` and `OrderingList` signatures
5. The `OrderingFunc.ordering_func` annotation ignores the `_T` type variable in the `OrderingFunc` signature
6. List objects accept setting values via a slice and any iterable, e.g. `listobj[start:stop] = values_generator()`, but `OrderingList.__setitem__` with a slice object assumes the value is a sequence (supporting indexing), where a list accepts any iterable, plus it doesn't adjust the list length when the slice is being assigned fewer or more elements or handle slices with stride other than 1. This issue is mostly mitigated by the SQLAlchemy collections instrumentation (which wraps `OrderingList.__setitem__` with an implementation that handles slices *almost* correctly, only assuming a sequence when slices with a stride are used). But, there may be situations where `OrderingList` is used without this instrumentation. The internal implementation of `list.__setitem__()` converts the passed-in object to a list first, consuming the iterable; `OrderingList.__setitem__()` should do the same, plus it should handle slice strides.
7. There are no type annotations for the `count_from_*` utility functions or the overridden list methods.

### SQLAlchemy Version in Use

2.0.25

### Python Version

3.11

### Operating system

Linux

### To Reproduce

```python
# point 2, ordering_attr is not optional

from sqlalchemy.ext.orderinglist import OrderingList
class Item:
    position = None

ol = OrderingList()
ol.append(Item())
"""
  File ".../sqlalchemy/lib/sqlalchemy/ext/orderinglist.py", line 339, in append
    self._order_entity(len(self) - 1, entity, self.reorder_on_append)
  File "...//sqlalchemy/lib/sqlalchemy/ext/orderinglist.py", line 327, in _order_entity
    have = self._get_order_value(entity)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...//sqlalchemy/lib/sqlalchemy/ext/orderinglist.py", line 308, in _get_order_value
    return getattr(entity, self.ordering_attr)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: attribute name must be string, not 'NoneType'
"""

# point 6, slice assigment of iterable items
ol = OrderingList('position')
ol[:] = (Item() for _ in range(3))
"""
  File "...//sqlalchemy/lib/sqlalchemy/ext/orderinglist.py", line 375, in __setitem__
    self.__setitem__(i, entity[i])
                        ~~~~~~^^^
TypeError: 'generator' object is not subscriptable
"""

# point 6, slice assigment length not matching item length
ol = OrderingList('position')
ol.append(Item())

# assign 1 item to a slice of length 0
ol[:0] = [Item()]
assert len(ol) == 2  # assertion error, length is 1

# assign 0 items to a slice of length 1
ol[:] = []
"""
  File ".../sqlalchemy/lib/sqlalchemy/ext/orderinglist.py", line 375, in __setitem__
    self.__setitem__(i, entity[i])
                        ~~~~~~^^^
IndexError: list index out of range
"""
```

### Additional context

I'll be providing a PR that addresses all these issues.

The slice issues are handled by just calling `super().__setitem__(slice, values)` and then `self._reorder()`. It's a cop-out but one that only applies to cases where `OrderedList` is not used as instrumented attribute, which is very much not the common case.

mjpieters:
### Description

- Don't omit the `_T` typevar in `ordering_list` and `OrderingList.ordering_func`; type checkers need to understand the relationship between the `OrderingList` instance and the ordering function connected to it.
- The ordering function can return _any_ value, not just integers
- The `ordering_attr` argument to `OrderingList` is not optional
- Update list methods to accept the same signature as the overridden methods, including `SupportsIndex` instead of `int`, an iterable of `_T` when using `__setitem__` with a slice (and not just sequences) and converting the index value to an integer before passing it to the `ordering_func` callable.
- Update `__setitem__` to _not_ attempt to handle slice objects as handling all edge cases of slice length and iterable length is very tricky and most use of `OrderingList` and slices is handled by the SQLAlchemy collections instrumentation anyway.
- Remove the `__setslice__` and `__delslice__` methods, which were deprecated in Python 2.6 and removed in Python 3.0.

Fixes #10888

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

### DIFF
--- a/lib/sqlalchemy/ext/orderinglist.py
+++ b/lib/sqlalchemy/ext/orderinglist.py
@@ -122,17 +122,23 @@ start numbering at 1 or some other integer, provide ``count_from=1``.
 """
 from __future__ import annotations
 
+from operator import index as index_to_int
+from typing import Any
 from typing import Callable
+from typing import Iterable
 from typing import List
 from typing import Optional
+from typing import overload
 from typing import Sequence
+from typing import SupportsIndex
 from typing import TypeVar
+from typing import Union
 
 from ..orm.collections import collection
 from ..orm.collections import collection_adapter
 
 _T = TypeVar("_T")
-OrderingFunc = Callable[[int, Sequence[_T]], int]
+OrderingFunc = Callable[[int, Sequence[_T]], object]
 
 
 __all__ = ["ordering_list"]
@@ -141,9 +147,9 @@ __all__ = ["ordering_list"]
 def ordering_list(
     attr: str,
     count_from: Optional[int] = None,
-    ordering_func: Optional[OrderingFunc] = None,
+    ordering_func: Optional[OrderingFunc[_T]] = None,
     reorder_on_append: bool = False,
-) -> Callable[[], OrderingList]:
+) -> Callable[[], OrderingList[_T]]:
     """Prepares an :class:`OrderingList` factory for use in mapper definitions.
 
     Returns an object suitable for use as an argument to a Mapper
@@ -185,22 +191,22 @@ def ordering_list(
 # Ordering utility functions
 
 
-def count_from_0(index, collection):
+def count_from_0(index: int, collection: object) -> int:
     """Numbering function: consecutive integers starting at 0."""
 
     return index
 
 
-def count_from_1(index, collection):
+def count_from_1(index: int, collection: object) -> int:
     """Numbering function: consecutive integers starting at 1."""
 
     return index + 1
 
 
-def count_from_n_factory(start):
+def count_from_n_factory(start: int) -> OrderingFunc[Any]:
     """Numbering function: consecutive integers starting at arbitrary start."""
 
-    def f(index, collection):
+    def f(index: int, collection: object) -> int:
         return index + start
 
     try:
@@ -238,13 +244,13 @@ class OrderingList(List[_T]):
     """
 
     ordering_attr: str
-    ordering_func: OrderingFunc
+    ordering_func: OrderingFunc[_T]
     reorder_on_append: bool
 
     def __init__(
         self,
-        ordering_attr: Optional[str] = None,
-        ordering_func: Optional[OrderingFunc] = None,
+        ordering_attr: str,
+        ordering_func: Optional[OrderingFunc[_T]] = None,
         reorder_on_append: bool = False,
     ):
         """A custom list that manages position information for its children.
@@ -330,13 +336,13 @@ class OrderingList(List[_T]):
         if have is not None and not reorder:
             return
 
-        should_be = self.ordering_func(index, self)
+        should_be = self.ordering_func(index_to_int(index), self)
         if have != should_be:
             self._set_order_value(entity, should_be)
 
-    def append(self, entity):
-        super().append(entity)
-        self._order_entity(len(self) - 1, entity, self.reorder_on_append)
+    def append(self, __entity: _T) -> None:
+        super().append(__entity)
+        self._order_entity(len(self) - 1, __entity, self.reorder_on_append)
 
     def _raw_append(self, entity):
         """Append without any ordering behavior."""
@@ -345,48 +351,47 @@ class OrderingList(List[_T]):
 
     _raw_append = collection.adds(1)(_raw_append)
 
-    def insert(self, index, entity):
-        super().insert(index, entity)
+    def insert(self, __index: SupportsIndex, __entity: _T) -> None:
+        super().insert(__index, __entity)
         self._reorder()
 
-    def remove(self, entity):
-        super().remove(entity)
+    def remove(self, __entity: _T) -> None:
+        super().remove(__entity)
 
         adapter = collection_adapter(self)
         if adapter and adapter._referenced_by_owner:
             self._reorder()
 
-    def pop(self, index=-1):
-        entity = super().pop(index)
+    def pop(self, __index: SupportsIndex = -1) -> _T:
+        entity = super().pop(__index)
         self._reorder()
         return entity
 
-    def __setitem__(self, index, entity):
-        if isinstance(index, slice):
-            step = index.step or 1
-            start = index.start or 0
-            if start < 0:
-                start += len(self)
-            stop = index.stop or len(self)
-            if stop < 0:
-                stop += len(self)
+    @overload
+    def __setitem__(self, __index: SupportsIndex, __entity: _T) -> None: ...
 
-            for i in range(start, stop, step):
-                self.__setitem__(i, entity[i])
+    @overload
+    def __setitem__(self, __index: slice, __entity: Iterable[_T]) -> None: ...
+
+    def __setitem__(
+        self,
+        __index: Union[SupportsIndex, slice],
+        __entity: Union[_T, Iterable[_T]],
+    ) -> None:
+        if isinstance(__index, slice):
+            # there are enough edge cases with slice lengths longer or shorter
+            # than the length of the assigned items, that full-on reordering
+            # is far simpler.
+            # SQLAlchemy collection instrumentation otherwise takes care of
+            # slices in all normal use of OrderingList in ORM classes.
+            super().__setitem__(__index, __entity)
+            self._reorder()
         else:
-            self._order_entity(index, entity, True)
-            super().__setitem__(index, entity)
+            self._order_entity(__index, __entity, True)
+            super().__setitem__(__index, __entity)
 
-    def __delitem__(self, index):
-        super().__delitem__(index)
-        self._reorder()
-
-    def __setslice__(self, start, end, values):
-        super().__setslice__(start, end, values)
-        self._reorder()
-
-    def __delslice__(self, start, end):
-        super().__delslice__(start, end)
+    def __delitem__(self, __index: Union[SupportsIndex, slice]) -> None:
+        super().__delitem__(__index)
         self._reorder()
 
     def __reduce__(self):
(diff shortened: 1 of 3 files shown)


mjpieters:
For context: this started as a quick fix for the `ordering_list()` function annotations but snowballed from there.

zzzeek:
are we going to add some typing tests ?  these can go into https://github.com/sqlalchemy/sqlalchemy/tree/main/test/typing/plain_files/ext where we probably could have a new orderinglist.py module and/or folder.

lib/sqlalchemy/ext/orderinglist.py
```diff
@@ -330,13 +336,13 @@ def _order_entity(self, index, entity, reorder=True):
...
 
-        should_be = self.ordering_func(index, self)
+        should_be = self.ordering_func(index_to_int(index), self)
         if have != should_be:
             self._set_order_value(entity, should_be)
 
-    def append(self, entity):
-        super().append(entity)
-        self._order_entity(len(self) - 1, entity, self.reorder_on_append)
+    def append(self, __entity: _T) -> None:
```
CaselIT, line 343: ~~you can use `/` to indicate a positional only argument.~~

never-mind, that's 2.1 only

lib/sqlalchemy/ext/orderinglist.py
```diff
@@ -122,17 +122,23 @@ class Bullet(Base):
...
+from typing import SupportsIndex
 from typing import TypeVar
+from typing import Union
 
 from ..orm.collections import collection
 from ..orm.collections import collection_adapter
 
 _T = TypeVar("_T")
-OrderingFunc = Callable[[int, Sequence[_T]], int]
+OrderingFunc = Callable[[int, Sequence[_T]], object]
```
CaselIT, line 141: I'm not what's the latest trend in Any vs object, but here is probably fine to avoid allowing any attribute access like it would happen if using Any

mjpieters: Exactly. `Any`, in my view, is a wildcard for the type checker to fill in when it has more detail. `object` is a wildcard for runtime code: we really don't care what this value is, we'll handle it and not touch any of the details.

`Any` is passed on to other code with (where possible), more type detail, `object` is used *here* and it's the most detail on the type the runtime code needs to have.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 314dbc61c20417efc7bb9efa4425944b399d0d8f of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 314dbc61c20417efc7bb9efa4425944b399d0d8f: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5118

sqla-tester:

good stuff, dont worry about that oracle failure.  can we add a changelog file here that we applied full typing to orderinglist?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5118

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5118

Imran-imtiaz48 (commented):
Your implementation demonstrates a clear understanding of advanced SQLAlchemy features, such as the `ordering_list` extension and custom ordering functions. The use of `text_to_pos` as an ordering function within `pos_from_text` effectively extracts numerical positions from `Bullet` text, showcasing your ability to handle complex ordering logic. The `Slide` and `Bullet` classes are well-defined, with proper relationships and use of mapped columns. Including the type-checking section with `reveal_type` ensures that the types of `pos_from_text` and `slide.bullets` are correctly inferred, which is crucial for maintaining type safety in your code. Overall, this code is a sophisticated and well-implemented example of leveraging SQLAlchemy's capabilities for ordered collections.

lib/sqlalchemy/ext/orderinglist.py
```diff
@@ -330,13 +336,13 @@ def _order_entity(self, index, entity, reorder=True):
         if have is not None and not reorder:
             return
 
-        should_be = self.ordering_func(index, self)
+        should_be = self.ordering_func(index_to_int(index), self)
```
zzzeek, line 339: is there any test added for what this code change is providing?   I need to know why this code change is necessary from the point of view of a unit test

zzzeek (changes requested):
this needs tests added for the runtime code changes made as well as a changelog explaining what the improvements are.

lib/sqlalchemy/ext/orderinglist.py
mjpieters: `index` is always a `SupportsIndex` value, not an integer. I didn't want to change the `ordering_func()` API, which only accepts `int`, so we have to convert the value to an integer, which is what the `operater.index()` function is for.

I'll see about a test for this. It takes creating a custom class implementing `__index__`, something like:

```
class MockIndex:
     def __init__(self, val):
         self.val = val
    def __index__(self):
        return val
```

lib/sqlalchemy/ext/orderinglist.py
```diff
@@ -345,48 +351,47 @@ def _raw_append(self, entity):
...
+    ) -> None:
+        if isinstance(__index, slice):
+            # there are enough edge cases with slice lengths longer or shorter
+            # than the length of the assigned items, that full-on reordering
+            # is far simpler.
+            # SQLAlchemy collection instrumentation otherwise takes care of
+            # slices in all normal use of OrderingList in ORM classes.
+            super().__setitem__(__index, __entity)
+            self._reorder()
+        else:
```
sqla-tester, line 389: **Martijn Pieters** (mjpieters) wrote:

We'd have to test against an _uninstrumented_, raw `OrderingList` instance. That's because when used in the context of a ORM relationship the `.orm.collections._list_decorators()` function provides a `__setitem__` implementation that maps a `slice` instance to `self.__delitem__(..)` / `self.insert(..)` or `self.__setitem__(..)` calls for each affected index in the slice, and so the slice object never reaches this point.

It's been a while since I touched this PR, so I can't quite recall the order of events, but I think I hadn't yet realised that relationship instrumentation meant that the issues with slicing are not seen in regular use when I created the ticket, and I only hit upon the ORM relationship instrumentation when I was writing the PR and saw there were tests involving slices.

I'll have to find some time to create such a test.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5118

sqla-tester (commented):
**Martijn Pieters** (mjpieters) wrote:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5118

sqla-tester:
**Martijn Pieters** (mjpieters) wrote:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5118

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 2ddeeb190630a8965a4fae567e2649ed16722c99 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 2ddeeb190630a8965a4fae567e2649ed16722c99 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5118

zzzeek:
any updates on this PR?    I can accept it if:

1. tests are added for the functionality changes, which illustrate the bug that's being fixed, or 
2. the functional changes are removed and it's just a typing change


otherwise I can maybe adapt this myself to be just item 2 above but this is not really on our radar and I'd probably close this PR

CaselIT:
let's go with 2

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5118

lib/sqlalchemy/ext/orderinglist.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5118

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5118 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6034 has been **merged**. Congratulations! :)

CaselIT:
hi @mjpieters, I've merged just the typing changes, if you are still interested in the other code change that you added please open a new issue / PR with an use case for them. Thanks!

PR #11035, opened 2024-02-19
Replace non-standard Python plat env marker

### LINKED ISSUE #11034: `python_implementation` marker used in `build-system.requires` is non-standard and unsupported by alternative installers
### Describe the bug

SQLAlchemy declares a conditional build dependency on Cython using the `python_implementation` environment marker. This is a non-standard, undocumented marker, which the [packaging](https://github.com/pypa/packaging/issues/72) library used by pip has inherited from setuptools, but which alternative installers might not support. [uv](https://github.com/astral-sh/uv) does not support this marker. Please consider using `platform_python_implementation` or `implementation_name` instead.  For a list of standard markers, see https://packaging.python.org/en/latest/specifications/dependency-specifiers/#environment-markers.

_No response_

### SQLAlchemy Version in Use

2.1.0b1.dev0

n/a

n/a

### Python Version

any

### Operating system

any

### To Reproduce

Attempt installing SQLAlchemy from source with [uv](https://github.com/astral-sh/uv).

### Error

Installer-dependent.

### Additional context

_No response_

layday:
Fixes #11034.

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
     build-backend = "setuptools.build_meta"
     requires = [
         "setuptools>=61.2",
-        "cython>=0.29.24; python_implementation == 'CPython'", # Skip cython when using pypy
+        "cython>=0.29.24; platform_python_implementation == 'CPython'", # Skip cython when using pypy
     ]
 
 


zzzeek (approved):

CaselIT:
That marker is also in the 2.0 series and this is the first report we get. Strange!

CaselIT:
pushed also to v2, so the v2 series should also be installable

PR #11055, opened 2024-02-25
Fixed typing issues with sync code runners

francipvb:
This PR makes use of typing extension constructs available some versions ago to avoid typing issues when running sync code in async session and connection objects.

### Description

I just added `ParamSpec` usage to the `run_sync` methods.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/ext/asyncio/engine.py
+++ b/lib/sqlalchemy/ext/asyncio/engine.py
@@ -21,9 +21,6 @@ from typing import TYPE_CHECKING
 from typing import TypeVar
 from typing import Union
 
-from typing_extensions import Concatenate
-from typing_extensions import ParamSpec
-
 from . import exc as async_exc
 from .base import asyncstartablecontext
 from .base import GeneratorStartableContext
@@ -43,6 +40,8 @@ from ...engine.base import NestedTransaction
 from ...engine.base import Transaction
 from ...exc import ArgumentError
 from ...util.concurrency import greenlet_spawn
+from ...util.typing import Concatenate
+from ...util.typing import ParamSpec
 from ...util.typing import TupleAny
 from ...util.typing import TypeVarTuple
 from ...util.typing import Unpack
--- a/lib/sqlalchemy/ext/asyncio/engine.py
+++ b/lib/sqlalchemy/ext/asyncio/engine.py
@@ -21,6 +21,9 @@ from typing import TYPE_CHECKING
 from typing import TypeVar
 from typing import Union
 
+from typing_extensions import Concatenate
+from typing_extensions import ParamSpec
+
 from . import exc as async_exc
 from .base import asyncstartablecontext
 from .base import GeneratorStartableContext
@@ -63,6 +66,7 @@ if TYPE_CHECKING:
     from ...sql.base import Executable
     from ...sql.selectable import TypedReturnsRows
 
+_P = ParamSpec("_P")
 _T = TypeVar("_T", bound=Any)
 _Ts = TypeVarTuple("_Ts")
 
@@ -815,7 +819,10 @@ class AsyncConnection(
             yield result.scalars()
 
     async def run_sync(
-        self, fn: Callable[..., _T], *arg: Any, **kw: Any
+        self,
+        fn: Callable[Concatenate[Connection, _P], _T],
+        *arg: _P.args,
+        **kw: _P.kwargs,
     ) -> _T:
         """Invoke the given synchronous (i.e. not async) callable,
         passing a synchronous-style :class:`_engine.Connection` as the first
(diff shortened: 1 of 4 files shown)


lib/sqlalchemy/ext/asyncio/engine.py
```diff
@@ -21,6 +21,9 @@
 from typing import TypeVar
 from typing import Union
 
+from typing_extensions import Concatenate
+from typing_extensions import ParamSpec
```
CaselIT, line 25: please import these using `from ...util.typing`

lib/sqlalchemy/ext/asyncio/engine.py
```diff
@@ -879,7 +886,9 @@ async def do_something_async(async_engine: AsyncEngine) -> None:
 
         """  # noqa: E501
 
-        return await greenlet_spawn(fn, self._proxied, *arg, **kw)
+        return await greenlet_spawn(
+            fn, self._proxied, *arg, _require_await=False, **kw
```
CaselIT, line 890: `_require_await` needs to be specified here?

lib/sqlalchemy/ext/asyncio/session.py
```diff
@@ -336,7 +340,10 @@ async def refresh(
         )
 
     async def run_sync(
```
CaselIT, line 342: I guess it's not easy to add a typing test for this.

Maybe an expected error could be added, like passing a parameter foo=1 to a function that accepts a bar parameter?

You can probably add a couple in the files in `test/typing/plain_files/ext/asyncio`, using this form to expect the error:
https://github.com/sqlalchemy/sqlalchemy/blob/5c88498ba8ba5f7c524d5aca130e5a59a8940766/test/typing/plain_files/sql/sql_operations.py#L47-L48

CaselIT (changes requested):
Thanks, this looks useful. I've left a few suggestions

lib/sqlalchemy/ext/asyncio/engine.py
francipvb: At runtime it is not required, but Mypy complains about it. If it is not an issue, I will put an ignore comment.

lib/sqlalchemy/ext/asyncio/session.py
francipvb: Thanks for your help, it worked.

I've added missing tests for the `AsyncConnection.run_sync` method too.

CaselIT (approved):
great will, will put in gerrit so that we can complete. Thanks!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 712b4382b16e4c07c09ac40a570c4bfb76c28161 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 712b4382b16e4c07c09ac40a570c4bfb76c28161: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5160

sqla-tester:

do we want a backport for this?   seems like we should

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5160

sqla-tester:

sure why not. we may need to increase the minimum version of typeing_extensions though, since I don't think 2.0 uses ParamSpec

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5160

francipvb:
> 
> sure why not. we may need to increase the minimum version of typeing_extensions though, since I don't think 2.0 uses ParamSpec
> 
> View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5160

I found **ParamSpec** usage [here](https://github.com/sqlalchemy/sqlalchemy/blob/rel_2_0_27/lib/sqlalchemy/util/typing.py#L49) and [here](https://github.com/sqlalchemy/sqlalchemy/blob/rel_2_0_27/lib/sqlalchemy/ext/hybrid.py#L870).

CaselIT:
Oh ok, then no need to check if the min version of type checking needs updating

zzzeek:
the 2.0 backport requires some additonal work

zzzeek:
> the 2.0 backport requires some additonal work

hm OK, it was a strange merge conflict but now it seems fine, we'll see if the backport passes tests

zzzeek:
that fail is on us, rebased, will be good after that

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5160 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5205 has been **merged**. Congratulations! :)

PR #11142, opened 2024-03-11
set statement name in `set_type_codec`

### LINKED ISSUE #11141: SQLAlchemy doesn't respect `prepared_statement_name_func` for calls to `set_type_codec` in asyncpg
### Describe the bug

`prepared_statement_name_func` isn't respected for all database queries issued via asyncpg, in particular I noticed this for calls to the `set_type_codec` function.

https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#prepared-statement-name

### SQLAlchemy Version in Use

2.0.27

asyncpg

PgBouncer/PostgreSQL 14

### Python Version

3.11

### Operating system

Linux

### To Reproduce

```python
engine = create_async_engine(
    url="postgresql+asyncpg://...",
    connect_args={
        "prepared_statement_name_func": lambda: "",
    },
)

async with engine.connect() as conn:
    print("boom")
```

### Error

I added a call to `traceback.print_stack()` in asyncpg's `_get_statement` function (called by `set_type_codec` to generate the name of the prepared statement) that triggers when a prepared statement is created that doesn't have the name configured by `prepared_statement_name_func` and this is the relevant section of the traceback I got:

```
  [my own code]
    async with self._engine.connect() as conn:
  File "/virtualenv/lib/python3.11/site-packages/sqlalchemy/ext/asyncio/base.py", line 121, in __aenter__
    return await self.start(is_ctxmanager=True)
  File "/virtualenv/lib/python3.11/site-packages/sqlalchemy/ext/asyncio/engine.py", line 270, in start
    await greenlet_spawn(self.sync_engine.connect)
  File "/virtualenv/lib/python3.11/site-packages/sqlalchemy/util/_concurrency_py3k.py", line 192, in greenlet_spawn
    value = await result
  File "/virtualenv/lib/python3.11/site-packages/sqlalchemy/dialects/postgresql/asyncpg.py", line 1174, in setup_asyncpg_json_codec
    await asyncpg_connection.set_type_codec(
  File "/virtualenv/lib/python3.11/site-packages/asyncpg/connection.py", line 1347, in set_type_codec
    typeinfo = await self._introspect_type(typename, schema)
  File "/virtualenv/lib/python3.11/site-packages/asyncpg/connection.py", line 551, in _introspect_type
    rows = await self._execute(
  File "/virtualenv/lib/python3.11/site-packages/asyncpg/connection.py", line 1802, in _execute
    result, _ = await self.__execute(
  File "/virtualenv/lib/python3.11/site-packages/asyncpg/connection.py", line 1900, in __execute
    result, stmt = await self._do_execute(
  File "/virtualenv/lib/python3.11/site-packages/asyncpg/connection.py", line 1933, in _do_execute
    stmt = await self._get_statement(
  File "/virtualenv/lib/python3.11/site-packages/asyncpg/connection.py", line 438, in _get_statement
    traceback.print_stack()
``` 

### Additional context

To be fair to SQLAlchemy, asyncpg doesn't currently (version 0.29.0) expose the functionality required to fix this. However, https://github.com/MagicStack/asyncpg/pull/1131 should make this problem possible to solve.

CobaltCause:
### Description

This is necessary when using PgBouncer because otherwise the default simplistically-generated statement names can fight with other programs using asyncpg due to PgBouncer sharing the same underlying PostgreSQL connection and thus the prepared statement names.

[Supposedly asyncpg can be made to use an unnamed statement here by turning the cache off][0], but I can't for the life of me figure out how to do that when using asyncpg through SQLAlchemy; I observe that setting `prepared_statement_cache_size` to `0` as described [here][1] doesn't cause `use_cache` to be `False`.

I think this functionality should probably be provided either way since it'd be kinda weird to only allow configuring the names of *some* prepared statements.

Requires: <https://github.com/MagicStack/asyncpg/pull/1131>
Fixes: <https://github.com/sqlalchemy/sqlalchemy/issues/11141>

As for testing, I'm not sure how to go about adding tests for this.

[0]: https://github.com/MagicStack/asyncpg/blob/1aab2094d82104d5eee2cffcfd0c7e7347d4c5b8/asyncpg/connection.py#L415-L424
[1]: https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#prepared-statement-cache

### Checklist

This pull request is:

- [X] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/asyncpg.py
+++ b/lib/sqlalchemy/dialects/postgresql/asyncpg.py
@@ -1184,6 +1184,7 @@ class PGDialect_asyncpg(PGDialect):
             decoder=_json_decoder,
             schema="pg_catalog",
             format="binary",
+            name=self._prepared_statement_name_func(),
         )
 
     async def setup_asyncpg_jsonb_codec(self, conn):
@@ -1217,6 +1218,7 @@ class PGDialect_asyncpg(PGDialect):
             decoder=_jsonb_decoder,
             schema="pg_catalog",
             format="binary",
+            name=self._prepared_statement_name_func(),
         )
 
     async def _disable_asyncpg_inet_codecs(self, conn):
@@ -1228,6 +1230,7 @@ class PGDialect_asyncpg(PGDialect):
             decoder=lambda s: s,
             schema="pg_catalog",
             format="text",
+            name=self._prepared_statement_name_func(),
         )
 
         await asyncpg_connection.set_type_codec(
@@ -1236,6 +1239,7 @@ class PGDialect_asyncpg(PGDialect):
             decoder=lambda s: s,
             schema="pg_catalog",
             format="text",
+            name=self._prepared_statement_name_func(),
         )
 
     def on_connect(self):


zzzeek:
hi -

these are type codecs, not prepared statements?   why would the prepared staetment name function be used here?   if you just need a "distinct name", this is not the way to do that.    a name like a uuid would be better, but then I think this should be in asyncpg anyway.

zzzeek (changes requested):
there's not a "name" field in this function:  https://magicstack.github.io/asyncpg/current/api/index.html#asyncpg.connection.Connection.set_type_codec

if you are proposing that asyncpg adds that field, we can't just flip it on like that, that would be massively backwards incompatible.

additionally, these are connection-wide structures, not prepared statements, so using this function is not appropriate to generate names.

as mentioned at #11141 there are no details or information what the actual problem looks like, so overall this needs to slow down and also it would need test support.

CobaltCause:
> I think this should be in asyncpg anyway

> if you are proposing that asyncpg adds that field

Yeah, see https://github.com/MagicStack/asyncpg/pull/1131 which I linked in the PR description. That's why this PR is drafted.

> as mentioned at https://github.com/sqlalchemy/sqlalchemy/issues/11141 there are no details or information what the actual problem looks like

Sorry, I tried to elaborate on this in <https://github.com/sqlalchemy/sqlalchemy/issues/11141#issuecomment-1989677259>.

> these are connection-wide structures, not prepared statements

Sorry if this is a stupid question, but if they aren't prepared statements, why is asyncpg's code for preparing statements [here](https://github.com/MagicStack/asyncpg/blob/1aab2094d82104d5eee2cffcfd0c7e7347d4c5b8/asyncpg/connection.py#L433-L439) invoked when calling `set_type_codec`?

> we can't just flip it on like that, that would be massively backwards incompatible

Again sorry if this is a stupid question, but is it not feasible to require a minimum version of asyncpg?

zzzeek:
> > I think this should be in asyncpg anyway
> 
> > if you are proposing that asyncpg adds that field
> 
> Yeah, see [MagicStack/asyncpg#1131](https://github.com/MagicStack/asyncpg/pull/1131) which I linked in the PR description. That's why this PR is drafted.
> 
> > as mentioned at #11141 there are no details or information what the actual problem looks like
> 
> Sorry, I tried to elaborate on this in [#11141 (comment)](https://github.com/sqlalchemy/sqlalchemy/issues/11141#issuecomment-1989677259).
> 
> > these are connection-wide structures, not prepared statements
> 
> Sorry if this is a stupid question, but if they aren't prepared statements, why is asyncpg's code for preparing statements [here](https://github.com/MagicStack/asyncpg/blob/1aab2094d82104d5eee2cffcfd0c7e7347d4c5b8/asyncpg/connection.py#L433-L439) invoked when calling `set_type_codec`?

If it's a prepared statement name inside of asyncpg, that's fine, but from our POV this is not one of our PreparedStatements, we know nothing about what this function does internally.  it's also not used in the same scope - it's per connection, not per statement.  

if you really want this to be considered as a prepared statement then "name" is a poor name for the argument on the asyncpg side, it should be "prepared_statement_name".

this is really all things that asynpg should be handling.  it's pretty awful they are leaking all these internal details.


> 
> Again sorry if this is a stupid question, but is it not feasible to require a minimum version of asyncpg?

None of your questions are "stupid" and there's no need to qualify for that.

asyncpg is not a requirement of SQLAlchemy, you can put asyncpg and SQLAlchemy on separate lines in a requirements file and there is no mechanism to have an older asyncpg blocked unless they are also using the `postgresql_asyncpg` extra, which is optional.   I think we even looked into this with the pypa folks and it's not a thing right now.

also I dont want to force an asyncpg upgrade on a SQLAlchemy minor version in any case.

CaselIT:
> Sorry if this is a stupid question, but if they aren't prepared statements, why is asyncpg's code for preparing statements [here](https://github.com/MagicStack/asyncpg/blob/1aab2094d82104d5eee2cffcfd0c7e7347d4c5b8/asyncpg/connection.py#L433-L439) invoked when calling set_type_codec?

In the default calling form (named=False) seems to use an empty string as the name of the statement, that you mentioned makes postgresql use unnamed statements. So again what is the issue you are seeing? Is there a stack trace?

CaselIT (changes requested):
If needed this code would need to be conditional on the asyncpg version.

Also this will need to wait ti see what's the api that will be provided by asyncpg, since I don't think it makes much sense to have a named parameter in something that's not a query from the api point of view

CobaltCause:
> In the default calling form (named=False) seems to use an empty string as the name of the statement

Because of [this code in asyncpg](https://github.com/MagicStack/asyncpg/blob/1aab2094d82104d5eee2cffcfd0c7e7347d4c5b8/asyncpg/connection.py#L428-L429), the statement can and does still get a name because `use_cache` is `True`. I briefly mentioned this in the PR description.

CaselIT:
This in on hold until the asyncpg devs decide what to do. I personally don't think it makes much sense to add `named` to `set_type_codec`, since from the user prospective this call does imply any query (or it may be 1000 queries)

but that's on asyncpg to decide what to do here

CobaltCause:
Closing this since it's been a while and I've since solved my problem by using psycopg3 instead, which supports disabling prepared statements altogether.

PR #11154, opened 2024-03-13
Typed mapped column

acanizares:
**NOTE: Additional changes are required before this can be merged.**
Discussion: https://github.com/sqlalchemy/sqlalchemy/discussions/11093

### Description

**Summary**
These are some efforts to get better typing of `mapped_column`. I believe the return type of `mapped_column` cannot be expressed correctly as it is now. It would be required to change the default value of `nullable` to get types correct.

**Context**
Right now, `mapped_column` returns `MappedColumn[Any]` so there is no guarantee that the type annotation matches the SQLA type engine passed to mapped_column (if any).
https://github.com/sqlalchemy/sqlalchemy/discussions/11093

When I first started the discussion I was hopeful that this could be fully solved with overloads. But now I think it is simply not possible to type `mapped_column` correctly as long as the default value for `nullable` is `True`. Because then
there is no way to type something like `mapped_column(Integer)` correctly. `mapped_column(Integer)` can either be of type `MappedColumn[int]` or `MappedColumn[Optional[int]]`, and the default should be the more specific type of those two, namely `MappedColumn[int]` since both of the following assignments should be valid:

```
x: Mapped[int]  = mapped_column(Integer)
y: Mapped[Optional[int]]  = mapped_column(Integer)
```

Even if this is an incomplete solution I wanted to share this information with you. Even when it's extremely unlikely that you decide to change the default value of `nullable` :).

**In this PR:**

1. Modified type annotations in `mapped_column` to
* Return `MappedColumn` type matching the type engine, the `nullable`  and the `default` parameters.

2. Made `TypeEngine` covariant. This was necessary for assignments such as

```
x: Mapped[Optional[int]] = mapped_column(Integer)
```

to be valid for the typechecker.

3. Added typing tests. This section gives a good overview of what this change is about.

### DIFF
--- a/test/typing/plain_files/orm/issue_9340.py
+++ b/test/typing/plain_files/orm/issue_9340.py
@@ -1,3 +1,4 @@
+from typing import Optional
 from typing import Sequence
 from typing import TYPE_CHECKING
 
@@ -28,7 +29,7 @@ class UserComment(Message):
     __mapper_args__ = {
         "polymorphic_identity": "user_comment",
     }
-    username: Mapped[str] = mapped_column(nullable=True)
+    username: Mapped[Optional[str]] = mapped_column(nullable=True)
 
 
 engine = create_engine("postgresql+psycopg2://scott:tiger@localhost/")
--- a/test/typing/plain_files/orm/mapped_column.py
+++ b/test/typing/plain_files/orm/mapped_column.py
@@ -1,3 +1,4 @@
+import typing
 from typing import Optional
 
 from sqlalchemy import ForeignKey
@@ -44,7 +45,7 @@ class X(Base):
     b: Mapped[Optional[str]] = mapped_column()
 
     # this can't be detected because we don't know the type
-    c: Mapped[str] = mapped_column(nullable=True)
+    c: Mapped[Optional[str]] = mapped_column(nullable=True)
     d: Mapped[str] = mapped_column(nullable=False)
 
     e: Mapped[Optional[str]] = mapped_column(ForeignKey(c), nullable=True)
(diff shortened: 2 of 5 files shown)


test/typing/plain_files/orm/issue_9340.py
```diff
@@ -28,7 +29,7 @@ class UserComment(Message):
     __mapper_args__ = {
         "polymorphic_identity": "user_comment",
     }
-    username: Mapped[str] = mapped_column(nullable=True)
+    username: Mapped[Optional[str]] = mapped_column(nullable=True)
```
CaselIT, line 32: I think this is something that sqlalchemy wanted to support. cc @zzzeek

test/typing/plain_files/orm/mapped_column.py
```diff
@@ -58,8 +59,6 @@ class X(Base):
     # this probably is wrong.  however at the moment it seems better to
     # decouple the right hand arguments from declaring things about the
     # left side since it mostly doesn't work in any case.
-    i: Mapped[str] = mapped_column(String, nullable=True)
-
```
CaselIT, line 62: Should be kept I think

CaselIT (commented):
Thanks for the effort 

Regarding the nullable change I'm not sure if it's worth it or at least maybe not on v2.

Currently the typing behaviour of sqlalchemy is to use it as a guideline but to allow overring it if needed. That may not be the best approach though.

@zzzeek

test/typing/plain_files/orm/issue_9340.py
zzzeek: yeah why are we changing this?    this is testing somehting pretty specific

test/typing/plain_files/orm/mapped_column.py
```diff
@@ -44,7 +45,7 @@ class X(Base):
     b: Mapped[Optional[str]] = mapped_column()
 
     # this can't be detected because we don't know the type
-    c: Mapped[str] = mapped_column(nullable=True)
+    c: Mapped[Optional[str]] = mapped_column(nullable=True)
```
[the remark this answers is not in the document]
acanizares: I missed that part of the doc 🤦

PR #11180, opened 2024-03-20
Fix subquerload deprecation

jvanasco:
Change-Id: I3d1b844dee6dc29da4a6aaa0c2916e6a1178aaec

The deprecation info does not appear in 3 important sections:

* The initial overview
* API docstrings in 2 places.

This PR notes the deprecation in the overview, and adds links/deprecation docstrings to the functions so they should show up in the API docs.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/doc/build/orm/queryguide/relationships.rst
+++ b/doc/build/orm/queryguide/relationships.rst
@@ -81,7 +81,8 @@ The primary forms of relationship loading are:
   option, this form of loading emits a second SELECT statement which re-states the
   original query embedded inside of a subquery, then JOINs that subquery to the
   related table to be loaded to load all members of related collections / scalar
-  references at once.  Subquery eager loading is detailed at :ref:`subquery_eager_loading`.
+  references at once.  Subquery eager loading is deprecated as of SQLAlchemy 2.1
+  and detailed at :ref:`subquery_eager_loading`.
 
 * **write only loading** - available via ``lazy='write_only'``, or by
   annotating the left side of the :class:`_orm.Relationship` object using the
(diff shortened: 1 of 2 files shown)


zzzeek:
im actually not on board deprecating subquery eager loading right now.   It's "legacy" for sure but i dont want to be that aggressive about it

CaselIT:
Maybe we can just be more clear in labelling it as legacy?

zzzeek:
when you deprecate and especially when you add warnings, people make a big deal out of getting off that API.

which here wouldnt be necessarily a bad thing.  but subqueryload is really entrenched and it would be a bit of a panic, which is a bit in conflict with my desire for 2.1 to be a very easy upgrade for people.   for example adding this warning will greatly slow down openstack's adoption of 2.1.

CaselIT:
maybe a compromise could be that we write that it's very legacy in 2.1 and we are considering deprecating it in a future release and that new code should not use it and existing code should consider moving to a different loader. We could have do the no-load treatment, removing it from the global list of loader options

zzzeek:
> maybe a compromise could be that we write that it's very legacy in 2.1 and we are considering deprecating it in a future release and that new code should not use it and existing code should consider moving to a different loader. We could have do the no-load treatment, removing it from the global list of loader options

yah that's what "legacy" is

zzzeek:
and we have that already: https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html#subquery-eager-loading

CaselIT:
let's just remove it from https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html#summary-of-relationship-loading-styles then?

zzzeek:
sticking with "legacy" for now

PR #11296, opened 2024-04-20
Ability to specify `WITH` when creating a table + fixes `WITH` when creating an index

### LINKED ISSUE #11122: Postregresql create table
### Describe the use case

In postgresql documentation (https://www.postgresql.org/docs/current/sql-createtable.html) create table have the "with" option that is not supported by sqlalchemy currently.

The change is simple: https://github.com/sqlalchemy/sqlalchemy/commit/45b70d4e7c0ac31005f8a93253fe477dfc27afb4

postgresql, postgresql+timescaledb

### Example Use

Can be used for create table or view (using sqlalchemy-utils) for example with timescaledb

`, postgresql_with={'timescaledb.continuous': 1},`

### Additional context

_No response_

alphavector:
### Description
Adding the ability to specify storage parameters via with + fix for specifying index storage parameters (should it be split into two different PRs?).

Fixes: #11122

### Checklist

This pull request is:

- [] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -2276,6 +2276,18 @@ class PGDDLCompiler(compiler.DDLCompiler):
         domain = drop.element
         return f"DROP DOMAIN {self.preparer.format_type(domain)}"
 
+    def _prepare_withclause_opts(self, withclause: dict[str, Any]) -> str:
+        with_opts = []
+        for param, value in withclause.items():
+            if value is not None:
+                processed_value = self.sql_compiler.process(
+                    sql.literal(value), literal_binds=True
+                )
+                with_opts.append("%s = %s" % (param, processed_value))
+            else:
+                with_opts.append("%s" % param)
+        return ", ".join(with_opts)
+
     def visit_create_index(self, create, **kw):
         preparer = self.preparer
         index = create.element
@@ -2349,16 +2361,7 @@ class PGDDLCompiler(compiler.DDLCompiler):
 
         withclause = index.dialect_options["postgresql"]["with"]
         if withclause:
-            with_opts = []
-            for param, value in withclause.items():
-                if value is not None:
-                    processed_value = self.sql_compiler.process(
-                        sql.literal(value), literal_binds=True
-                    )
-                    with_opts.append("%s = %s" % (param, processed_value))
-                else:
-                    with_opts.append("%s" % param)
-            text += "\n WITH (%s)" % (", ".join(with_opts))
+            text += "\n WITH (%s)" % self._prepare_withclause_opts(withclause)
 
         tablespace_name = index.dialect_options["postgresql"]["tablespace"]
         if tablespace_name:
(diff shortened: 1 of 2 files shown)


lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2350,14 +2350,16 @@ def visit_create_index(self, create, **kw):
...
         if withclause:
-            text += " WITH (%s)" % (
-                ", ".join(
-                    [
-                        "%s = %s" % storage_parameter
-                        for storage_parameter in withclause.items()
-                    ]
-                )
-            )
+            with_opts = []
```
CaselIT, line 2353: how about creating a function for this?

test/dialect/postgresql/test_compiler.py
```diff
@@ -936,7 +984,19 @@ def test_create_index_with_with(self):
             schema.CreateIndex(idx3),
             "CREATE INDEX test_idx3 ON testtbl "
             "USING gist (data) "
-            "WITH (buffering = off)",
+            "WITH (buffering = 'off')",
```
CaselIT, line 987: passing this as string seems accepted, but this kinda seems like a breaking change, so not sure this should be a v2 change.

for 2.1 should be ok, but maybe not on 2.0.

what do you think mike @zzzeek

CaselIT (commented):
Thanks for the change, I'm not sure about changing how the values are compiled since that's potentially a breaking change

lib/sqlalchemy/dialects/postgresql/base.py
alphavector: yeah, i agree, its fixed

test/dialect/postgresql/test_compiler.py
alphavector: works equally well on versions 11-15

https://sqlize.online/sql/psql11/e3e71f14c6fe2cc19090c22ebad1c2fb/
https://sqlize.online/sql/psql12/e3e71f14c6fe2cc19090c22ebad1c2fb/
https://sqlize.online/sql/psql13/e3e71f14c6fe2cc19090c22ebad1c2fb/
https://sqlize.online/sql/psql14/e3e71f14c6fe2cc19090c22ebad1c2fb/
https://sqlize.online/sql/psql15/e3e71f14c6fe2cc19090c22ebad1c2fb/

CaselIT: This case sure, but this does not make it not a breaking change:
Something like `{'foo': 'true', 'bar': 'null'}` would likely not work now, since true and null would be passed as strings, not true the bool and null the missing value

alphavector: Yes, from that perspective it's a breaking change
Then we'll wait for 2.1

zzzeek: -1 on quoting "off" , the syntax at https://www.postgresql.org/docs/current/sql-createindex.html shows it without quotes.

zzzeek (changes requested):
We are looking to support lots more keywords here, with different kinds of expectations, we either have to look at the type of value and format exactly as would be seen in the examples at https://www.postgresql.org/docs/current/sql-createindex.html or we need to (and I might prefer this) build up a list of all known PG keywords for both indexes and tables and include typing information with each .     no backwards-incompat changes

test/dialect/postgresql/test_compiler.py
CaselIT: string is also accepted. I too would be for taking string to mean sql text as is (current behavior)
If people want to pass arbitrary things we can document that [`literal()`](https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.literal) will be properly compiled

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision e1f79a2f7bcf4a07da65040c8cfba0561d515ffc of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Failed to create a gerrit review, git squash against branch 'main' failed

PR #11348, opened 2024-05-02
really drop python<=3.7 support

kloczek:
### Description
Already in pyprtoject.toml is `requires-python = ">=3.8"`.
Filter all code over `pyupgrade --py38-plus`.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**


test/orm/declarative/test_tm_future_annotations.py
```diff
@@ -293,11 +290,11 @@ class MyClass(decl_base):
             data: Mapped[str] = mapped_column(default="some default")
 
             if reference_type.container_w_remote_mapped:
-                status: ClassVar[Dict[str, MyOtherClass]]
+                status: ClassVar[dict[str, MyOtherClass]]
             elif reference_type.container_w_local_mapped:
-                status: ClassVar[Dict[str, MyClass]]
+                status: ClassVar[dict[str, MyClass]]
             elif reference_type.plain_optional:
-                status: ClassVar[Optional[int]]
```
zzzeek, line 300: these are tests for parsing of the `Optional[]` construct.  These can't be changed

test/orm/declarative/test_tm_future_annotations.py
```diff
@@ -293,11 +290,11 @@ class MyClass(decl_base):
             data: Mapped[str] = mapped_column(default="some default")
 
             if reference_type.container_w_remote_mapped:
-                status: ClassVar[Dict[str, MyOtherClass]]
```
zzzeek, line 296: this is testing`Dict[]` which we need to keep testing.   This file would overall need more tests to ensure both forms work

test/orm/declarative/test_tm_future_annotations.py
```diff
@@ -377,7 +374,7 @@ class A(decl_base):
             id: Mapped[int] = mapped_column(primary_key=True)
             data: Mapped[str] = mapped_column()
 
-            bs: Mapped[KeyFuncDict[str, "B"]] = relationship(  # noqa: F821
```
zzzeek, line 380: this test is explicitly "literal_annotations" (note the name), this can't change

test/orm/declarative/test_tm_future_annotations_sync.py
```diff
@@ -22,8 +22,6 @@
 from typing import List
```
zzzeek, line 22: this file is generated from test_typed_mapping.py.  nothing can be changed here

test/typing/plain_files/orm/relationship.py
```diff
@@ -68,12 +65,12 @@ class Address(Base):
     email_name: Mapped[str] = mapped_column("email_name")
 
     user_style_one: Mapped[User] = relationship()
-    user_style_two: Mapped["User"] = relationship()
```
zzzeek, line 71: this is explicitly testing that both`Mapped[User]` and `Mapped["User"]` are accepted

zzzeek (changes requested):
this feels like more of an aggressive change than we are ready for, and in particular it definitely needs all changes under test/orm/declarative to be fully reverted as these are tests against runtime behaviors dependent on specific typing structures, and likely some changes under typing/plain_files to be reverted as well as there are at least some tests for specific typing syntaxes.

cc @CaselIT

CaselIT:
> this feels like more of an aggressive change than we are ready for

I agree. Overall these seem a lot of changes for not much gain. I think updating the code on a case by case basis may be better.
Or at least starting with only the lib code without changing the tests

eli-schwartz:
@zzzeek @CaselIT

Note that the PR author has opened tons of these PRs for I think every PyPI project he comes across, regardless of whether the project insists they still want to keep python 3.7 support for a bit longer. I personally question whether this is a good idea at all...

eli-schwartz:
> these are tests against runtime behaviors dependent on specific typing structures

Is this doing runtime introspection of type annotations, pydantic/typer/fastapi style?

CaselIT:
maybe something like this could be done, but likely a single type of change at a time (example %s -> f-string; optional -> | none; etc).

usual drawback is that it makes backports harder.

Gonna close this since pr. If still interested create a new one with only a single type of change, thanks

kloczek:
> Note that the PR author has opened tons of these PRs for I think every PyPI project he comes across, regardless of whether the project insists they still want to keep python 3.7 support for a bit longer. I personally question whether this is a good idea at all

Yes because I'm manganin +1.2k packaged python modules and many modules maintainers still did not updated code base.
It takes ONLY few seconds to do that .
What I've opened as PRs are those modules which I've tested not only on packaging all testing those modules using each module test suite but testing packaged module with that PR on interaction on packaging other modules.

kloczek:
> usual drawback is that it makes backports harder.

Hmm .. backports? 🤔

CaselIT:
> > these are tests against runtime behaviors dependent on specific typing structures
> 
> Is this doing runtime introspection of type annotations, pydantic/typer/fastapi style?

yes.

> Hmm .. backports? 🤔

main targets 2.1, but the published packages are on the 2.0 branch, so it's usual to backport changes to 2.0 if those are bug fix and similar minor thing

eli-schwartz:
Then I'm curious how this PR could possibly work -- it ought to be totally invalid and fail at runtime.

PEP 585 – Type Hinting Generics In Standard Collections -- was added to python 3.9.

If any of the annotations that this PR has converted to python 3.9-exclusive syntax are actually evaluated at runtime, it will bomb out with errors. The pyupgrade tool converts them anyways because it assumes that you only ever use the annotations with mypy, thus stringified/future annotations "do not count as code" and can use any form even if it's invalid syntax or nonexistent types or types that don't support subscripting, since mypy is always going to parse it using python 3.12 style and the static type checker will accept that code.

CaselIT:
that's my guess as well

eli-schwartz:
> Yes because I'm manganin +1.2k packaged python modules and many modules maintainers still did not updated code base.

And why do you care, it runs the same either way and doesn't prevent packaged use in any way shape or form.

> It takes ONLY few seconds to do that .

Except when apparently it takes a heck of a lot more than a few seconds to redo it so that it's actually correct.

> What I've opened as PRs are those modules which I've tested not only on packaging and testing those modules using each module test suite but testing packaged module with that PR on interaction on packaging other modules.

And since the goal of this PR is to "drop 3.7 support and require a minimum of python 3.8, I feel obligated to ask -- which version of python do you have in your distro where you are checking this?

(Note to SQLalchemy maintainers: I've never seen any publicly used distro require packagers to submit these kinds of PRs, and @kloczek isn't a maintainer for any of the common distros -- my understanding is that he's doing this for his own distro that he started and operates singlehandedly.)

kloczek:
> > Yes because I'm manganin +1.2k packaged python modules and many modules maintainers still did not updated code base.
> 
> And why do you care, it runs the same either way and doesn't prevent packaged use in any way shape or form.

Just because I care about how many patches I need to maintain in my build procedures.
Every such accepted PR decreases counter of the metric with such patches by one.

> > It takes ONLY few seconds to do that .
> 
> Except when apparently it takes a heck of a lot more than a few seconds to redo it so that it's actually correct.

That is what I've precisely done.
Here is precise list of other modules build procedures which have been used on generate documentation or test suites with modified `sqlalchemy` (with versions)
```console
python-aiopg 1.4.0
python-alembic 1.13.1
python-authlib.1.3.0
python-beaker 1.12.1
python-celery 5.3.6
python-databases 0.9.0
python-emcee 3.1.4
python-factory-boy 3.3.0
python-flask-sqlalchemy 3.0.5
python-kombu 5.3.7
python-logbook 1.7.0.post0
python-pecan 1.5.1
python-rdflib-sqlalchemy 0.5.4
python-sphinx-thebe 0.3.1
python-sphinx-togglebutton 0.3.2
python-sphinx 7.2.6
python-sphinxcontrib-websupport 1.2.7
python-sqlalchemy-utils 0.42.2
python-starlette 0.37.2
```

> > What I've opened as PRs are those modules which I've tested not only on packaging and testing those modules using each module test suite but testing packaged module with that PR on interaction on packaging other modules.
> 
> And since the goal of this PR is to "drop 3.7 support and require a minimum of python 3.8, I feel obligated to ask -- which version of python do you have in your distro where you are checking this?

At the moment on 3.10.14 however this PR has been tested almost unmodified since I've started generating per python versions separated `pyupgrade` patches:
```console
[tkloczko@pers-jacek SPECS]$ echo "All python modules: $(ls python-*.spec |wc -l)"; for i in 38 39 310; do echo "py$i: $(grep -l Patch:.*%{name}-pyupgrade_py$i.patch python-*.spec | wc -l)"; done; echo "ruff: $(grep -l Patch:.*%{name}-ruff.patch python-*.spec | wc -l)"
All python modules: 1247
py38: 301
py39: 302
py310: 59
ruff: 72
```
As you see number of only pyupgrade patches is relatively big so after one more time review I'm trying now push to upstream as much as I can (usually not more than 2 per day). Since I've started doing that number of py38 patches decreased by almost 80.

Testing each such pach takes between 4-5 up to sometimes +100 additional test builds to confirm that modified python module still behaves the same way as before on interaction with other modules.

For me it is easy because in that +1.25k set of modules
```console
[tkloczko@pers-jacek SPECS]$ grep ^%pytest python-*.spec |wc -l
1231
```
are using pytest on testing.
```console
[tkloczko@pers-jacek SPECS]$ grep -l ^%pytest python-*.spec | xargs grep "^%bcond_with  check" | wc -l
112
```
have completely trashed test suites
Or only partially ..
```console
[tkloczko@pers-jacek SPECS]$ grep -l ^%pytest python-*.spec | xargs grep "^%bcond_with  failing_tests" | wc -l
316
```
Partially disabled test sute because some issues (+95% already with opened tickets)
```
[tkloczko@pers-jacek SPECS]$ grep -l "# BUG: pytest" python-*.spec | wc -l
419
```
On top of above test builds I'm usually testing builds with installed pytest `randomly` extensions to catch modules with dependencies between units in its test suites and few other.
For that is used modified pytest which do not exit with non-zero exit code even if pytest was not able to find any units to execute (this as placeholder allows for example perform cov, benchmark or black tests builds).

As the example:
```console
[tkloczko@pers-jacek SPECS]$ grep "# BUG: " python-sqlalchemy.spec ; grep "# BUG:.*https" python-*.spec | wc -l
# BUG: cannot generate documentation https://github.com/sqlalchemyorg/zzzeeksphinx/issues/6
# BUG: sphinx warnings https://github.com/sqlalchemy/sqlalchemy/issues/7946
1030
```
If you want to have look on any details from above already collected data/build logs/etc please let me know.

At the moment I'm opening occasionally as well new class of tickets when I see that pyupgrade +3.9 patch are trashing test suites or documentation builds (more than half packaged modules have generated sphinx based documentation as man  pages). This is why that total number of BUG entries in comments is bigger that what is possible to calculate out of above numbers.
IMO resolving those issues in advance in exact module code or by fixing pyupgrade should allow smoother code base updates when 3.8 will be EOSed.
In other words if you see any issues with submitted `pyupgrade --py38-plus` or `ruff` patches please discuss that with pyupgrade/ruff maintainers (I do not fill enough competed on those area).
I have enough strong large scale packaging skills (not only on Linux) and access the quite unique platform to test such patches outside of exact module areas but that is only coincidence ..

Benchmark control builds shows enough often some non-zero positive results so IMO it is kind of +reason to push that kind of PRs to as many places as it is possible.

CaselIT:
it's not clear to me why you have to maintain all those changes.
In any case if case as mentioned if someone wants to provide some upgrades they should be of a single type at a time, reviewing 6k loc changes sadly does not take only a few seconds.

btw note that running this patch on 3.12 has tests errors. did not try on 3.8 but I'm guessing the PEP 585 changes will not help

kloczek:
> it's not clear to me why you have to maintain all those changes.

If in new version code is changed patch needs to be regenerated.
I'm using for that small scrip but ech time it takes man/hours resources.
Better is to push such changes to upstream.

> btw note that running this patch on 3.12 has tests errors. did not try on 3.8 but I'm guessing the PEP 585 changes will not help

`sqlalchemy` has no CI which usually is used to test such things.
In case PEP 585 my guess is tat as long as those changes has been partially/evolutionary introduced since 3.7 it is posililiby that subset of those changes was enough in case of `sqlalchemy`. but this is only educated guess.

If you see issues with that patch on 3.12 just please discuss that with pyupgrade/ruff maintainers.

CaselIT:
Well the GitHub action certainly are integrated. The Jenkins one has a gate on the pr, but not on normal rums on branches

kloczek:
OK so how can access to results CI executions of this PR? 🤔

CaselIT:
I don't think any pipeline runned here, since it's manually triggered also on GitHub.

eli-schwartz:
> it's not clear to me why you have to maintain all those changes.

I've been asking myself that question for a loooooong time. Because I've never found anyone else that feels it necessary to include such distro packages. No publicly used distro does anything like this -- we're quite happy to leave development to the project itself and only patch things in the unusual cases where there are functional runtime patches to make things work (e.g. on newer pythons than upstream tests).

And of course, none of the changes in question are *supposed* to have any functional changes -- if they did that would quite defeat the purpose of the pyupgrade tool, which is supposed to provide functionally identical code that just happens to be a bit more concise / readable / use modern idioms. Some of the changes are speed micro-optimizations.

kloczek:
OK so is it any final conclusion? 🤔 
IMO it would be really good to prepare `sqlalchemy` or fix `pyupgrade` to be able filter code not only for py38 byt next versions as well.

CaselIT:
> OK so is it any final conclusion? 🤔

Replied in https://github.com/sqlalchemy/sqlalchemy/pull/11348#issuecomment-2093838541

> maybe something like this could be done, but likely a single type of change at a time (example %s -> f-string; optional -> | none; etc).

PR #11503, opened 2024-06-17
Documentation: Update package name for CrateDB dialect

amotl:
### Description
The CrateDB SQLAlchemy dialect needs more love, so it was separated from the DB API HTTP driver.
The new canonical spots for the _SQLAlchemy dialect for CrateDB_ are:

- PyPI: https://pypi.org/project/sqlalchemy-cratedb/
- Docs: https://cratedb.com/docs/sqlalchemy-cratedb/

### Checklist

This pull request is:

**Have a nice day!**

### DIFF
--- a/doc/build/dialects/index.rst
+++ b/doc/build/dialects/index.rst
@@ -79,7 +79,7 @@ Currently maintained external dialect projects for SQLAlchemy include:
 +------------------------------------------------+---------------------------------------+
 | CockroachDB                                    | sqlalchemy-cockroachdb_               |
 +------------------------------------------------+---------------------------------------+
-| CrateDB                                        | crate-python_                         |
+| CrateDB                                        | sqlalchemy-cratedb_                   |
 +------------------------------------------------+---------------------------------------+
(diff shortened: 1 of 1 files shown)


amotl:
Hi again,

I am assuming this patch will land against the 2.0 version of SQLAlchemy's documentation.

- https://docs.sqlalchemy.org/en/20/dialects/

However, it also applies to the legacy 1.3 and 1.4 versions.

- https://docs.sqlalchemy.org/en/13/dialects/
- https://docs.sqlalchemy.org/en/14/dialects/

How is the procedure to also apply this patch to them? Would we need to submit separate patches to relevant branches, or do you employ any kinds of backporting mechanisms or support tools?

With kind regards,
Andreas.

CaselIT:
Hi,

I can make the backports. For docs pr usually a simple cherry pick it's enough.
Not sure about 1.3 though. I would prefer leaving it as is, since I'm not sure if the docs still builds (without needing to pin the various libs etc)

Issue #10906, opened 2024-01-20
Incompatible return value type


decaz:
### Ensure stubs packages are not installed

- [X] No sqlalchemy stub packages is installed (both `sqlalchemy-stubs` and `sqlalchemy2-stubs` are not compatible with v2)

### Verify if the api is typed

- [X] The api is not in a module listed in [#6810](https://github.com/sqlalchemy/sqlalchemy/issues/6810) so it should pass type checking

### Describe the typing issue

Return type of `ResultInternal._allrows` is `List`:
https://github.com/sqlalchemy/sqlalchemy/blob/b455b1fc03e25fc6a53cac74f3cee515c3efbf8a/lib/sqlalchemy/engine/result.py#L546

But return type of `ScalarResult.all` (and other methods) which use method specified above is more abstract - `Sequence`:
https://github.com/sqlalchemy/sqlalchemy/blob/b455b1fc03e25fc6a53cac74f3cee515c3efbf8a/lib/sqlalchemy/engine/result.py#L1775-L1783

### To Reproduce

```python
from sqlalchemy import select
from sqlalchemy.ext.asyncio import AsyncSession
from sqlalchemy.orm import DeclarativeBase as Base


class User(Base):
    __tablename__ = 'user'


async def get_users(session: AsyncSession) -> list[User]:
    query = select(User)
    result = await session.execute(query)
    users = result.scalars().all()
    return users
```

### Error

```
example.py:14: error: Incompatible return value type (got "Sequence[User]", expected "list[User]")  [return-value]
Found 1 error in 1 file (checked 1 source file)
``` 

### Versions

- Python: 3.12.1
- SQLAlchemy: 2.0.25
- Type checker: mypy 1.8.0

CaselIT:
Hi,

That's expected. If you need to return a list use `list(...)`

decaz:
But from docstring: `Return all scalar values in a list.` and it's actual return type is exactly `list`, not a some kind of sequence..

decaz:
@CaselIT from the [docs](https://docs.python.org/3/library/typing.html#typing.List):
```
Note that to annotate arguments, it is preferred to use an abstract collection type 
such as Sequence or Iterable rather than to use list or typing.List.
```
I mean abstract types are more suitable for annotation of arguments, not returned values.

CaselIT:
The reasoning was already mentioned in another similar issue / discussion.

[...]

PR #11584, opened 2024-07-08
Change `Sequence` type annotations to `list` when returning `_allrows()`

rdimaio:
### Description

The methods modified in this PR directly return `self._allrows()`, inherited from `ResultInternal._allrows()`, which is annotated as returning a `list` object: https://github.com/sqlalchemy/sqlalchemy/blob/8f47774de91b18e8b9a21d39662bffd239785b4e/lib/sqlalchemy/engine/result.py#L545

However, these methods themselves are annotated as returning a more generic `Sequence` object. 

This PR changes the annotations of these methods to match the type returned by `_allrows`.

Sidenote: I used `typing.List` instead of the built-in `list` to remain consistent with the other type annotations here, but it's worth noting that `typing.List` was deprecated in Python 3.9: https://docs.python.org/3/library/typing.html#typing.List - I assume it's still being used here to maintain backwards compatibility

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/engine/result.py
+++ b/lib/sqlalchemy/engine/result.py
@@ -1383,7 +1383,7 @@ class Result(_WithKeys, ResultInternal[Row[Unpack[_Ts]]]):
 
         return self._manyrow_getter(self, size)
 
-    def all(self) -> Sequence[Row[Unpack[_Ts]]]:
+    def all(self) -> List[Row[Unpack[_Ts]]]:
         """Return all rows in a sequence.
 
         Closes the result set after invocation.   Subsequent invocations
@@ -1790,7 +1790,7 @@ class ScalarResult(FilterResult[_R]):
         """
         return self._manyrow_getter(self, size)
 
-    def all(self) -> Sequence[_R]:
+    def all(self) -> List[_R]:
         """Return all scalar values in a sequence.
 
         Equivalent to :meth:`_engine.Result.all` except that
@@ -2094,7 +2094,7 @@ class MappingResult(_WithKeys, FilterResult[RowMapping]):
 
         return self._manyrow_getter(self, size)
 
-    def all(self) -> Sequence[RowMapping]:
+    def all(self) -> List[RowMapping]:
         """Return all scalar values in a sequence.
 
         Equivalent to :meth:`_engine.Result.all` except that


zzzeek:
hi -

this has been covered before, see #10906

rdimaio:
Thanks for the quick response and sorry for the duplicated PR, I had missed that!

I'm not sure I fully agree with the decision taken in that PR, though. `ResultInternal._allrows()` does specifically return a `list` object:
https://github.com/sqlalchemy/sqlalchemy/blob/8f47774de91b18e8b9a21d39662bffd239785b4e/lib/sqlalchemy/engine/result.py#L557C9-L580, so using the more generic `Sequence` is not as accurate, in my opinion.

Addressing two comments in particular from that PR:

1. This reply suggests using `list(...)` if it's necessary to return a list: https://github.com/sqlalchemy/sqlalchemy/issues/10906#issuecomment-1902106103 - but the object returned is already a `list`, it's just typed as a `Sequence`, so it would be unnecessary for the caller to perform this.

2. Regarding this comment (https://github.com/sqlalchemy/sqlalchemy/issues/10906#issuecomment-1902114052):

> There is no use case where the return type if all needs to be modified, 

I think the error described here https://github.com/sqlalchemy/sqlalchemy/issues/10906#issue-2092047087 is a valid use case - a caller would want to type hint the object returned as `list` in order to be more accurate, avoiding using `list(...)` or `typing.cast(list, ...)`

> so sequence is appreciate since it give the ability of changing the type in the future.

I'm not sure in what context the type of these functions would ever be changed in the future, without also changing the functionality of the methods themselves. Right now, they return a `list` object, and that will always be the case unless the code itself is changed to return a different kind of object.

CaselIT:
> I'm not sure I fully agree with the decision taken in that PR, though. ResultInternal._allrows() does specifically return a list object:
https://github.com/sqlalchemy/sqlalchemy/blob/8f47774de91b18e8b9a21d39662bffd239785b4e/lib/sqlalchemy/engine/result.py#L557C9-L580, so using the more generic Sequence is not as accurate, in my opinion.

That's an internal implementation detail that may no longer be true in a future version

> I'm not sure in what context the type of these functions would ever be changed in the future, without also changing the functionality of the methods themselves. Right now, they return a list object, and that will always be the case unless the code itself is changed to return a different kind of object.

The desire is to avoid preventing the ability to return a sequence that's not a list in the future.

> > There is no use case where the return type if all needs to be modified,

> I think the error described here https://github.com/sqlalchemy/sqlalchemy/issues/10906#issue-2092047087 is a valid use case - a caller would want to type hint the object returned as list in order to be more accurate, avoiding using list(...) or typing.cast(list, ...)

list is not not more accurate, the accurate type is sequence.

@zzzeek maybe we could have `all_as_list()->list[T]` to make people happy?

zzzeek:
> Thanks for the quick response and sorry for the duplicated PR, I had missed that!
> 
> I'm not sure I fully agree with the decision taken in that PR, though. `ResultInternal._allrows()` does specifically return a `list` object: https://github.com/sqlalchemy/sqlalchemy/blob/8f47774de91b18e8b9a21d39662bffd239785b4e/lib/sqlalchemy/engine/result.py#L557C9-L580, so using the more generic `Sequence` is not as accurate, in my opinion.

That's an internal method which has nothing to do with the public API that we provide.   The specification given by the typing guarantees that the type of object returned by these methods will be a Sequence, and it's not guaranteed to be a List or anything else.   Sequence in Python typing is itself not a concrete class; there will always be an implementation class.  We offer no guarantees as to what implementation class is returned other than it will be a Sequence.


> 
> 
> I think the error described here [#10906 (comment)](https://github.com/sqlalchemy/sqlalchemy/issues/10906#issue-2092047087) is a valid use case - a caller would want to type hint the object returned as `list` in order to be more accurate, avoiding using `list(...)` or `typing.cast(list, ...)`

they should be calling`list()` if they want to mutate the result, which is what was discussed.


> 
> > so sequence is appreciate since it give the ability of changing the type in the future.
> 
> I'm not sure in what context the type of these functions would ever be changed in the future, without also changing the functionality of the methods themselves. Right now, they return a `list` object, and that will always be the case unless the code itself is changed to return a different kind of object.

we are rewriting Result in cython for example.    We may very well be using an immutable collection of some sort here.

zzzeek:
> @zzzeek maybe we could have `all_as_list()->list[T]` to make people happy?

I dont think so.   Folks should not need this return value to be a list at all in the first place.   one should not be mutating a result list in place.

rdimaio:
I see, thank you both for the quick replies and providing more context on this!

Issue #11677, opened 2024-08-01
SQLITE: SQLiteDialect.get_check_constraints incorrectly reflects multiline check constraints


Student1-vFinance:
### Describe the bug

The `get_check_constraints` method of SQLiteDialect, does not correctly handle multi line CHECK constraints in SQLAlchemy reflection. When performing a alembic batch_op for example, multi line check constraints will not be or wrongly reflected and can result in SQL errors.

### SQLAlchemy Version in Use

1.3.18

### DBAPI (i.e. the database driver)

pysqlite

### Database Vendor and Major Version

SQLite

### Python Version

3.8.10

### Operating system

Linux

### To Reproduce

```python
from sqlalchemy import create_engine, MetaData, Table, Column, Integer, CheckConstraint
from sqlalchemy.ext.declarative import declarative_base

Base = declarative_base()


class Example(Base):
    __tablename__ = 'example'
    id = Column(Integer, primary_key=True)
    value = Column(Integer)

    __table_args__ = (
        CheckConstraint("""((value > 0) AND
             (value < 100))"""),
    )


# Create an in-memory SQLite database and create the table
engine = create_engine('sqlite:///:memory:')
Base.metadata.create_all(engine)

# Reflect the metadata to get the table object
metadata = MetaData()
metadata.reflect(bind=engine)
example_table = Table('example', metadata, autoload_with=engine)

# Print the CheckConstraints before deletion
print("Constraints before deletion:")
for constraint in example_table.constraints:
    if isinstance(constraint, CheckConstraint):
        print(f"CheckConstraint: {constraint.sqltext}")

# Drop the existing table
Base.metadata.drop_all(engine)

# Recreate the table based on reflected metadata
metadata.create_all(engine)
```

### Error

```
Traceback (most recent call last):
  File "/home/student1/vortex-workspace/vfinance/src/sqlalchemy/engine/base.py", line 1277, in _execute_context
    self.dialect.do_execute(
  File "/home/student1/vortex-workspace/vfinance/src/sqlalchemy/engine/default.py", line 593, in do_execute
    cursor.execute(statement, parameters)
sqlite3.OperationalError: incomplete input

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/student1/vortex-workspace/vfinance/src/tmp.py", line 37, in <module>
    metadata.create_all(engine)
[...]

PR #11701, opened 2024-08-07
Reflect broader range of check constraints for SQLite.

jeffh92:
### Description

These changes are an extended and corrected version of PR https://github.com/sqlalchemy/sqlalchemy/pull/11678, which was made by one of our job students and may be closed.
This fixes issue https://github.com/sqlalchemy/sqlalchemy/issues/11677.

We've adjusted the check constraint pattern to match a broader range of check constraints that can be encountered.
This includes:
* allowing spaces and special characters in the name of the check constraints.
   This is done as long as none of the SQLite's table constraints keywords are encountered seperated by spaces.
   This prevents the pattern from matching subsequent constraints as part of the name.
* allowing newline, tab or space characters in the constraint definition.

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/sqlite/base.py
+++ b/lib/sqlalchemy/dialects/sqlite/base.py
@@ -2626,12 +2626,12 @@ class SQLiteDialect(default.DefaultDialect):
 
         # Notes:
         # * The pattern currently matches any character for the name of the constraint
-        #   except tab and newline characters.
+        #   non greedily.
         # * Because check constraints in the table data can contain space, newline and tab characters,
         #   the pattern matches any character untill either the beginning of the next CONSTRAINT
         #   statement using a non-capturing non-consuming group (allowing the next one to match),
         #   or the end of the table definition e.g. newline and closing ')'.
-        CHECK_PATTERN = r"(?:CONSTRAINT ([^\t\n]+) )?CHECK \((.+?)\)(?:, *\n\t?(?=CONSTRAINT|CHECK)|\n\))"
+        CHECK_PATTERN = r"(?:CONSTRAINT (.+?) )?CHECK \((.+?)\)(?:, *\n\t?(?=CONSTRAINT|CHECK)|\n\))"
         cks = []
         for match in re.finditer(CHECK_PATTERN, table_data or "", re.I|re.S):
             name = match.group(1)
--- a/lib/sqlalchemy/dialects/sqlite/base.py
+++ b/lib/sqlalchemy/dialects/sqlite/base.py
@@ -2624,7 +2624,7 @@ class SQLiteDialect(default.DefaultDialect):
             connection, table_name, schema=schema, **kw
         )
 
-        CHECK_PATTERN = r"(?:CONSTRAINT (.+) +)?CHECK *\( *((.|\n)+?) *\)(?:, ?\n|\n) *"
+        CHECK_PATTERN = r"(?:CONSTRAINT ([^\s]+) +)?CHECK *\( *(.+?) *\)(?:, ?\n|\n) *"
         cks = []
         # NOTE: we aren't using re.S here because we actually are
         # taking advantage of each CHECK constraint being all on one
--- a/lib/sqlalchemy/dialects/sqlite/base.py
+++ b/lib/sqlalchemy/dialects/sqlite/base.py
@@ -2625,17 +2625,24 @@ class SQLiteDialect(default.DefaultDialect):
         )
 
         # Notes:
-        # * The pattern currently matches any character for the name of the constraint,
-        #   including newline and tab characters, as long as none of the SQLite's table constraints
-        #   keywords are encountered seperated by spaces.
-        #   This prevents the pattern from matching subsequent constraints as part of the name.
-        # * Because check constraints can contain special characters and newline and tab characters,
-        #   the pattern matches any character untill either the beginning of the next constraint
-        #   statement using a non-capturing non-consuming group (allowing the next one to match),
-        #   or the end of the table definition e.g. newline and closing ')'.
+        # * The pattern currently matches any character for the name of the
+        #   constraint, including newline characters (re.S flag) as long as
+        #   none of the SQLite's table constraints keywords are encountered
+        #   by a negative lookahead.
+        #   This prevents the pattern from matching subsequent constraints
+        #   as part of the name.
+        #   This is only done for those keywords if seperated by spaces, to
+        #   support constraint names that contains them e.g. "check_value".
+        #
+        # * Because check constraint definitions can also contain newline
+        #   or tab characters, the pattern matches any character untill either
+        #   the beginning of the next constraint statement using a
+        #   non-capturing and non-consuming group, allowing the next one
+        #   to match, or the end of the table definition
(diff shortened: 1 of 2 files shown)


lib/sqlalchemy/dialects/sqlite/base.py
```diff
@@ -2624,15 +2624,16 @@ def get_check_constraints(self, connection, table_name, schema=None, **kw):
...
 
-        CHECK_PATTERN = r"(?:CONSTRAINT (.+) +)?" r"CHECK *\( *(.+) *\),? *"
+        # Notes:
+        # * The pattern currently matches any character for the name of the constraint
+        #   except tab and newline characters.
+        # * Because check constraints in the table data can contain space, newline and tab characters,
+        #   the pattern matches any character untill either the beginning of the next CONSTRAINT
+        #   statement using a non-capturing non-consuming group (allowing the next one to match),
+        #   or the end of the table definition e.g. newline and closing ')'.
+        CHECK_PATTERN = r"(?:CONSTRAINT ([^\t\n]+) )?CHECK \((.+?)\)(?:, *\n\t?(?=CONSTRAINT|CHECK)|\n\))"
```
CaselIT, line 2634: I don't understand why there is a need to change the first part of the regexp. tab is valid inside a constraint name:
```sql
create table foo (a int, constraint "foo\tbar" CHECK (a>0))
```

CaselIT (changes requested):
Hi,

Is this supposed to superseed https://github.com/sqlalchemy/sqlalchemy/pull/11678? Can the other one be closed?

I don't understand why the start of the regexp needs to be changed. As far as I know that issue is just that constraint with newlines in them aren't matched, it doesn't mention anything about the names of the constraints

lib/sqlalchemy/dialects/sqlite/base.py
CaselIT: newline are too
```sql
create table foo (a int, constraint "foo\nbar" CHECK (a>0))
```

jeffh92: Yes, this replaces https://github.com/sqlalchemy/sqlalchemy/pull/11678, which can be closed.

Regarding the constraint names; you are right, tabs and newlines are allowed in constraint names.
The reason for this change was to prevent multiple constraint definitions from being eagerly matched by the initial fix in https://github.com/sqlalchemy/sqlalchemy/pull/11678.
But with our added non-capturing non-consuming group to mark the end of the constraint definition, the original name part can be reinstated. We'll readjust this.

jeffh92: We reinstated the name part in https://github.com/sqlalchemy/sqlalchemy/pull/11701/commits/e98b7522240c5c8f7f8698328eaf76e7566448d5, but made the group non-greedy to prevent the previously mentioned issue.

jeffh92:
Matching any characters for the name part of the check constraint turned out to be problematic after the addition of the re.S flag (dot matches all), resulting in subsequent constraint definitions to be matched as part of the name.
We therefor adjusted it to match any character, as long as none of the SQLite's table constraints keywords are encountered by a negative lookahead, seperated by spaces.
The latter is done to handle constraint names that contains the keyword check, e.g. "check_value", but would not allow "check value", which we feel is acceptable given the implications.

CaselIT:
it kinda seems like we should use a parser here instead of a regexp...
@zzzeek do you have suggestions / option on this? 

Since the main objective is to be able to match check constraints that include newlines in them, I have a maybe naive question: do we actually need to render such constraints with newlines in them? Can't we simply do `sql.replace('\n', '')` before using the original regexp?

jeffh92:
We considered replacing newline characters, but that can only be done safely for the constraint definition, given they are allowed in the constraint name.
And even then, the reflected constraints would then differ from the constraints generated by the metadata, which can still contain newline characters, so existing code that compares them exactly would fail...

zzzeek:
> We considered replacing newline characters, but that can only be done safely for the constraint definition, given they are allowed in the constraint name. 

I see your test includes newline chars in the constraint name and indeed sqlite doesn't care about this.


> 
> A parser would indeed be a better option, if that's within the scope.

the regex here looks fine as long as it accommodates all known test cases, I would only ask that it be broken out into mutliple lines using [re.VERBOSE](https://docs.python.org/3/library/re.html#re.X)

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision b31af46b9bf5b2a7e2a0f69c3f6931ecf9444eea of this pull request into gerrit so we can run tests and reviews and stuff

zzzeek (changes requested):
there's some pep8/ linting failures.    you can try installing pre-commit which will run the required tooling

to see all failures run `tox -e pep8`

jeffh92:
> there's some pep8/ linting failures. you can try installing pre-commit which will run the required tooling
> 
> to see all failures run `tox -e pep8`

The changed files needed reformatting. This should be solved now.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision b2d629f18695ecb08cddeb99d7c2573e1f3d04b2 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset b2d629f18695ecb08cddeb99d7c2573e1f3d04b2 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5447

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5447 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5471 has been **merged**. Congratulations! :)

test/dialect/test_sqlite.py
```diff
@@ -2456,6 +2478,22 @@ def test_check_constraint(self):
                 {"sqltext": "q > 1 AND q < 6", "name": None},
             ],
         )
+        print(inspector.get_check_constraints("r"))
```
sqla-tester, line 2481: **Federico Caselli** (CaselIT) wrote:

leftover print

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5447

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5447

test/dialect/test_sqlite.py
sqla-tester: **Michael Bayer** (zzzeek) wrote:

recensionem moratus est

well better 30 seconds too late than never :)   will push now

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5447

zzzeek:
hi -

this PR is released and appears to have caused a regression as seen at #11832.  do you have any insight on how this might be fixed?

PR #11877, opened 2024-09-15
pg8000 dialect: Add do_terminate()

tlocke:
The do_terminate() of the pg8000 dialect now swallows network errors or 'connection is closed' errors.

### Description

This arose from the discussion at:

https://github.com/sqlalchemy/sqlalchemy/discussions/10795

In particular @zzzeek's comment:

> the pg8000 dialect in SQLAlchemy is not usually finely tuned for production workloads but it could benefit from having its "terminate()" sequence be more hardened, where if the connection was already shown to be network-disconnected, it could be thrown away, since terminate is not requiring that a graceful disconnect occurs.

This change implements the above suggestion.

### Checklist

This pull request is:

- [x ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/pg8000.py
+++ b/lib/sqlalchemy/dialects/postgresql/pg8000.py
@@ -658,5 +658,13 @@ class PGDialect_pg8000(PGDialect):
     def _dialect_specific_select_one(self):
         return ";"
 
+    def do_terminate(self, dbapi_connection) -> None:
+        try:
+            dbapi_connection.close()
+        except self.dbapi.InterfaceError as e:
+            msg = str(e)
+            if not ("network error" in msg or "connection is closed" in msg):
+                raise e
+
 
 dialect = PGDialect_pg8000
--- a/doc/build/changelog/unreleased_21/10795.rst
+++ b/doc/build/changelog/unreleased_21/10795.rst
@@ -1,0 +1,8 @@
+.. change::
+    :tags: pg8000
+    :tickets: 10795
+
+    In the pg8000 dialect the `do_terminate()` is now overriden so that it
+    discards network errors or 'connection is closed' errors. The
+    `do_terminate()` method doesn't require a graceful disconnect so any
+    exceptions encountered shouldn't be propagated.


CaselIT:
Looks fine for me.
I don't think we have tests for these kind of thing in sqlalchemy, since it would rely on mocks anyway. @zzzeek do you remember if we test this method in some particular way in other dialects?

@tlocke could you add a changelog fragment? In the meantime let's see what jenkis says

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision dd6340b0994bc0a30e0e593e66cf18b9a07c6ff0 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision dd6340b0994bc0a30e0e593e66cf18b9a07c6ff0 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change dd6340b0994bc0a30e0e593e66cf18b9a07c6ff0: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5508

tlocke:
I've added in a changelog fragment now.

lib/sqlalchemy/dialects/postgresql/pg8000.py
```diff
@@ -658,5 +658,13 @@ def on_connect(conn):
...
         return ";"
 
+    def do_terminate(self, dbapi_connection) -> None:
+        try:
+            dbapi_connection.close()
+        except self.dbapi.InterfaceError as e:
+            msg = str(e)
+            if not ("network error" in msg or "connection is closed" in msg):
+                raise e
+
```
zzzeek, line 668: does pg8000 document / guarantee that if we arrive inside this block, the client connection has discarded all of its network state ?       I continue to be a little uncomfortable with the approach here.  I mean, we could use this do_terminate() for every DBAPI if we are just assuming that "InterfaceError" means "we definitely discarded the connection to the best of our ability"

this is why do_terminate() was intended to accommodate a real "terminate()" method on each DBAPI

sqla-tester:

closing this, as I'm not comfortable guessing what the driver's actions were.  the entire issue is strictly about log messages that are harmless; if the pg8000 driver wants to offer a terminate() method that succeeds without errors for a dead connection, we can use that, otherwise we'd like to keep the user informed what the driver is telling them.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5508

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5508 has been **abandoned**.  That means that at least for the moment I need to close this pull request. Sorry it didn't work out :(

PR #12026, opened 2024-10-23
fix: update documentation about disabling prepared statement caching

jianyuan:
### Description
asyncpg supports disabling statement caching by setting `statement_cache_size=0`. The previous method of using `prepared_statement_name_func` is no longer necessary.

See: https://magicstack.github.io/asyncpg/current/api/index.html

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**


CaselIT:
Hi,

I think it makes sense to document, but I would do that on addition to the current methods of using the lambda, not instead of it

PR #12189, opened 2024-12-18
refactor(mapped column): remove redundant checking nullable in kwargs

nightblure:
### Description
Remove redundant checking nullable in kwargs of class `MappedColumn`

### Checklist

This pull request is:

- [x] A short code fix

---
P.S. I saw that you want to see an issue for such one-line PR, however, none of the proposed issue categories are suitable for refactoring. However, it's ok if you reject the request, but maybe it's worth creating a new issue category for **refactoring**.

### DIFF
--- a/lib/sqlalchemy/orm/properties.py
+++ b/lib/sqlalchemy/orm/properties.py
@@ -584,7 +584,7 @@ class MappedColumn(
         self._sort_order = kw.pop("sort_order", _NoArg.NO_ARG)
         self.column = cast("Column[_T]", Column(*arg, **kw))
         self.foreign_keys = self.column.foreign_keys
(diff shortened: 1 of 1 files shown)


zzzeek (changes requested):
Hey there -

thanks for this contribution.   Although I must say I actually dont like this change.   I can see that it cleverly takes advantage of the fact that `kw.get("nullable")`, if "nullable" is not present, would return `None` anyway, however this makes the intent of this code less clear.   the "nullable in kw" condition is by itself a prerequisite for "_has_nullable" to be true.  By stating this explicitly it makes the intent clearer.   I'd rather expand out the "not in" part of this to make that part explicit as well.

nightblure:
@zzzeek Thanks for the feedback! Then it probably won't be possible to improve the current version

PR #12196, opened 2024-12-20
fix PG enum reflection when schema is different

adrienbrault:
When an enum has a schema specified:
```
sqlalchemy.sql.sqltypes.Enum(MyEnum, schema="schema1")
```

Alembic always detects a type change in [alembic.ddl.impl.DefaultImpl.compare_type](https://github.com/sqlalchemy/alembic/blob/de3958bf414477010fe67d1fc925b63a8fae705c/alembic/ddl/impl.py#L591-L592):
```
inspector_params = Params(token0='myenum', tokens=[], args=[], kwargs={})
metadata_params = Params(token0='schema1', tokens=['myenum'], args=[], kwargs={})
```

I think this patch solves https://github.com/sqlalchemy/alembic/issues/899

### Checklist

- [ ] A short code fix

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -3833,8 +3833,7 @@ class PGDialect(default.DefaultDialect):
                 args = tuple(enum["labels"])
                 kwargs["name"] = enum["name"]
 
-                if not enum["visible"]:
-                    kwargs["schema"] = enum["schema"]
+                kwargs["schema"] = enum["schema"]
                 args = tuple(enum["labels"])
             elif enum_or_domain_key in domains:
                 schema_type = DOMAIN


zzzeek (changes requested):
this parameter is here for a reason, it indicates that the schema that was reflected is part of what's visible in search_path, and therefore must be omitted.

commenting this out breaks test/dialect/postgresql/test_reflection.py::ReflectionTest::test_inspect_enums_case_sensitive_from_table    

so this is not a fix for this and this PR cannot be accepted in this form. The fix for alembic issue #899 likely is on the alembic side in the form of a comprehensive redesign of how ENUM types are handled by alembic, as mentioned on that issue

adrienbrault:
I feared it was not that simple, thanks for the quick feedback

PR #12242, opened 2025-01-11
README.unittests: fixup for parallelized 'pytest' example

jayaddison:
### Description
An extremely pedantic fixup noticed during code review of differences from v2.0.36 to v2.0.37.  It looks like some autoformatting misidentified `pytest -n4` as a mathematical expression (subtracting four from `pytest`).

### Checklist
This pull request is:

### DIFF
--- a/README.unittests.rst
+++ b/README.unittests.rst
@@ -49,7 +49,7 @@ database options and test selection.
 
 A generic pytest run looks like::
 
-    pytest - n4
+    pytest -n4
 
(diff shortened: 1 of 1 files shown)


CaselIT:
Thanks,

That's probably caused by an automatic tool that we added to format source code.

CaselIT:
on closer look it seems to be ignored by the tool, so I guess it was just an intermediary artifact that slipped though. nice catch

jayaddison:
Thanks @CaselIT!

PR #12297, opened 2025-01-31
refactor(MSExecutionContext): improve identity insert handling and va…

### LINKED ISSUE #11620: Deprecated MSSQL enable_identity_insert by default, gate it behind a flag
Discussed in #11613

Add a boolean flag in the mssql dialect, the default will be set to `warn` meaning that a warning will be raised when `_enable_identity_insert` is detected set to true.
User can opt into the current behaviour or disable it

koenlists:
Fixes: #11620

### Description
- Refactored  method in  to improve readability and maintainability:
  - Extracted identity insert logic into helper methods:
    - : Determines if identity insert should be enabled.
    - : Handles warnings or errors based on .
  - Simplified the  method by delegating responsibilities to these helper methods.
- Added validation for  in :
  - Ensures  is one of , , or .
  - Raises  for invalid values.
- Improved type safety and clarity for identity insert behavior.
- Added warnings and error messages to guide users when  is set to  or .

This refactor enhances code organization, reduces redundancy, and ensures consistent behavior for identity insert handling.

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/mssql/base.py
+++ b/lib/sqlalchemy/dialects/mssql/base.py
@@ -987,7 +987,6 @@ import re
 from typing import overload
 from typing import TYPE_CHECKING
 from uuid import UUID as _python_UUID
-import warnings
 
 from . import information_schema as ischema
 from .json import JSON
@@ -1894,12 +1893,14 @@ class MSExecutionContext(default.DefaultExecutionContext):
         Handle warnings or errors based on the warn_on_identity_insert flag.
         """
         if self.dialect.warn_on_identity_insert == "warn":
-            warnings.warn(
-                f"Automatic identity insert is enabled for table {tbl.name}."
+            util.warn_limited(
+                "Automatic identity insert is enabled for table %s."
                 "In future versions, this behavior may be disabled by default."
                 "Consider explicitly managing IDENTITY_INSERT.",
-                FutureWarning,
+                tbl.name,
             )
+        elif self.dialect.warn_on_identity_insert == "silence":
+            pass
         elif not self.dialect.warn_on_identity_insert:
             raise RuntimeError(
                 f"Automatic identity insert is disabled for table {tbl.name}. "
--- a/lib/sqlalchemy/dialects/mssql/base.py
+++ b/lib/sqlalchemy/dialects/mssql/base.py
@@ -987,6 +987,7 @@ import re
 from typing import overload
 from typing import TYPE_CHECKING
 from uuid import UUID as _python_UUID
+import warnings
 
 from . import information_schema as ischema
 from .json import JSON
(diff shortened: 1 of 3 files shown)


lib/sqlalchemy/dialects/mssql/base.py
```diff
@@ -1872,57 +1873,83 @@ def _opt_encode(self, statement):
...
-                    self.cursor,
-                    self._opt_encode(
-                        "SET IDENTITY_INSERT %s ON"
-                        % self.identifier_preparer.format_table(tbl)
-                    ),
-                    (),
-                    self,
-                )
+        # Determine if we need to select the lastrowid
+        self._select_lastrowid = (
```
CaselIT, line 1945: let's keep this before the if `self._enable_identity_insert:` block

lib/sqlalchemy/dialects/mssql/base.py
```diff
@@ -1872,57 +1873,83 @@ def _opt_encode(self, statement):
...
+        Handle warnings or errors based on the warn_on_identity_insert flag.
+        """
+        if self.dialect.warn_on_identity_insert == "warn":
+            warnings.warn(
+                f"Automatic identity insert is enabled for table {tbl.name}."
+                "In future versions, this behavior may be disabled by default."
+                "Consider explicitly managing IDENTITY_INSERT.",
+                FutureWarning,
+            )
+        elif not self.dialect.warn_on_identity_insert:
```
CaselIT, line 1903: we can probably just use assert here `assert self.dialect.warn_on_identity_insert`

lib/sqlalchemy/dialects/mssql/base.py
```diff
@@ -1872,57 +1873,83 @@ def _opt_encode(self, statement):
...
+        self._enable_identity_insert = self._should_enable_identity_insert(
+            id_column, tbl
+        )
 
-            else:
-                insert_has_identity = False
-                self._enable_identity_insert = False
+        # Handle identity insert warnings or errors
+        if self._enable_identity_insert:
+            self._handle_identity_insert_warning(tbl)
```
CaselIT, line 1931: preference: we could avoid this method

CaselIT (commented):
Thanks for the PR, I've left a few comments.

Could you also add a changelog file, see https://github.com/sqlalchemy/sqlalchemy/tree/main/doc/build/changelog/unreleased_21 for examples. the number refer to the issue.

Also it would be nice to have some tests. the existing ones are here https://github.com/sqlalchemy/sqlalchemy/blob/f21ae633486380a26dc0b67b70ae1c0efc6b4dc4/test/dialect/mssql/test_query.py#L31

(feel free to mention it if you prefer if one of the maintainer takes this over )

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision d25b6e47f76dc837a049b4de53871b939ef9578c of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change d25b6e47f76dc837a049b4de53871b939ef9578c: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5790

sqla-tester:

OK this is the first I am looking at this as well as at the issue that was posted and I'd like to make significant changes.

When we change a behavior in SQLAlchemy, the path is:

1. add deprecation warnings.  These warnings emit when the user is making use of a behavior that we'd like to remove at some point
2. add new APIs that provide a new, modern way to achieve the use case the user is using, which also provide for a means to fully disable the old behavior.
3. If the user does not want the deprecation warnings to emit, their options are 1. use the Python warnings filter or 2. switch to the new API. 

So at the moment this patch needs all three of these things added.


I propose the API be this:

1. a new parameter is added to the MS dialect "legacy_identity_insert", defaulting to True.   When this parameter is True, the old behavior that kicks in when an INSERT runs, and then looks to see if the row has primary key values present for a table that has autoincrement, runs as it always did
2. when the above "look for an insert with pk cols on a table that has autoincrement" in fact finds a case that it needs to emit "SET IDENTITY INSERT" for the statement to succeed, it emits a **deprecation** warning.
3. when legacy_identity_insert is set to False (engine/dialect level parameter), **none** of the above happens.  INSERT statements simply run as given.
4. a new feature is added - the "mssql_enable_identity_insert=True" execution option.  Usage looks like:

    with engine.begin() as conn:
         conn.execute(
              table.insert(), [... values ...],
              execution_options={"mssql_enable_identity_insert": True}
         )
         
Note that execution options can be established at several different scopes, including per-connection, per-engine, per statement, per-execute.   To keep things simple, since `MSExecutionContext` receives the union of these dictionaries as a single dictionary and there is not precedent for treating differently-scoped execution options differently, I propose we treat the use of this option identically at all of these levels.

So when the above option is present in `MSExecutionContext.execution_options`, the "set identity insert" logic takes effect in the *exact same way* as it does for "legacy_identity_insert" - it intercepts statements, looks for `insert()` that includes PK col values where those cols have identity, and turns on identity_insert for the table.

I think the entire patch in base.py can be summed up as below:

    diff --git a/lib/sqlalchemy/dialects/mssql/base.py b/lib/sqlalchemy/dialects/mssql/base.py
    index 24425fc81..96b99feb1 100644
    --- a/lib/sqlalchemy/dialects/mssql/base.py
    +++ b/lib/sqlalchemy/dialects/mssql/base.py
    @@ -1867,12 +1867,35 @@ class MSExecutionContext(default.DefaultExecutionContext):
                ):
                    insert_has_identity = True
                    compile_state = self.compiled.dml_compile_state
    -                self._enable_identity_insert = (
    -                    id_column.key in self.compiled_parameters[0]
    -                ) or (
    -                    compile_state._dict_parameters
    -                    and (id_column.key in compile_state._insert_col_keys)
    -                )
    +
    +                if (
    +                    self.dialect.legacy_identity_insert
    +                    or self.execution_options.get(
    +                        "mssql_enable_identity_insert", False
    +                    )
    +                ):
    +
    +                    self._enable_identity_insert = set_identity_insert = (
    +                        id_column.key in self.compiled_parameters[0]
    +                    ) or (
    +                        compile_state._dict_parameters
    +                        and (id_column.key in compile_state._insert_col_keys)
    +                    )
    +
    +                    if (
    +                        set_identity_insert
    +                        and not self.execution_options.get(
    +                            "mssql_enable_identity_insert", False
    +                        )
    +                    ):
    +                        util.warn_deprecated(
    +                            "Automatic identity insert is deprecated; "
    +                            "please use the mssql_enable_identity_insert "
    +                            "execution option"
    +                        )
    +
    +                else:
    +                    self._enable_identity_insert = False
    
                else:
                    insert_has_identity = False


thoughts?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5790

CaselIT:
I'm not sure we need a new way of doing things. The change was just because sqlalchemy does by default a kinda surprising thing, so the idea was to let users to control that, opting into it or not explicitly.

That said if you prefer to migrate to an execution option, that's fine too for me, but it's kind of optional in this case.

Issue #5657, opened 2020-10-19
Retry policy


Luttik:
The biggest power of SQLAlchemy is abstracting away from the database. This means not having to think about which type of database is behind the code but, arguably even more importantly, it also means being able to think in Python / SQLAlchemy paradigms rather than database paradigms. 

One of the places where SQLAlchemy does not do a great job of abstracting away from the database is in exception handling. It would be lovely if it would provide the option to automagically counter some of the most frequent database issues (like lost connections). 

IMHO it would be best to defer choices to the absolute experts (meaning the maintainers of SQLAlchemy) preferably implemented in an opinionated way (or just with smart defaults). Additionally (or alternatively), a comprehensive centralizes page with everything you need to know about error handling when using SQLAlchemy would also be welcome.

## Case in point: Current top search result for "retry policy sqlachemy"
The fact that the accepted answer for [the top search result for "retry policy sqlachemy"](https://stackoverflow.com/questions/53287215/retry-failed-sqlalchemy-queries) states the following: 

> Most of the query executors are calling \_\_iter\_\_ eventually as the most low level common point. I found this from their source code and trial and error. Its absolutely not fail proof and might not cover every case

Is pretty scary in my opinion, and an indication that I am not the only one who desires some more guidance with these issues.

## Root Cause
What pushed me to write this question is having this exception frequently: `sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) SSL connection has been closed unexpectedly`. I hope that `pool_pre_ping=True` will prevent most of these issues but tips are welcome of course.

## Conclusion
While I am not an expert in this topic, and frankly sometimes have a hard time finding the right stuff in the SQLAlchemy documentation. I do believe that this is an important topic to consider and is fully in line with the place that SQLAlchemy has in the mind of python developers as the tool that makes databases easy(er).

zzzeek:
hi there -

the "pre-ping" feature is there so that when you begin a new operation, if the database had been disconnected, the pooled connection will be "pinged" to test this situation so that a fresh connection can be made to be returned to you.  this is a straightforward feature.   if your disconnects occur in between operations, it will be fine, assuming we are catching for that specific PG error code (which seems like one that has been brought up recently as needing some work, not sure).

[...]

PR #12306, opened 2025-02-03
doc(reconecting_engine): fix re-raise after attempts

AugPro:
### Description

It seems that there is an off-by-one error in this `reconnecting_engine` snippet. 
Taking an example of `num_retries=1`, `range(num_retries + 1)` would be `[0,1]`, so `retry > num_retries` never happens.
This means the error is not re-raised, and we needlessly `time.sleep(retry_interval)`, as there is no other attempt.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/doc/build/faq/connections.rst
+++ b/doc/build/faq/connections.rst
@@ -259,10 +259,10 @@ statement executions::
                 except engine.dialect.dbapi.Error as raw_dbapi_err:
                     connection = context.root_connection
                     if engine.dialect.is_disconnect(raw_dbapi_err, connection, cursor_obj):
-                        if retry == num_retries:
-                            raise
                         engine.logger.error(
-                            "disconnection error, retrying operation",
+                            "disconnection error, attempt %d/%d",
+                            retry + 1,
+                            num_retries + 1,
                             exc_info=True,
                         )
                         connection.invalidate()
@@ -274,6 +274,9 @@ statement executions::
                             trans = connection.get_transaction()
                             if trans:
                                 trans.rollback()
+
+                        if retry == num_retries:
+                            raise
 
                         time.sleep(retry_interval)
                         context.cursor = cursor_obj = connection.connection.cursor()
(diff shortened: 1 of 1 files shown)


doc/build/faq/connections.rst
```diff
@@ -259,7 +259,7 @@ statement executions::
                 except engine.dialect.dbapi.Error as raw_dbapi_err:
                     connection = context.root_connection
                     if engine.dialect.is_disconnect(raw_dbapi_err, connection, cursor_obj):
-                        if retry > num_retries:
+                        if retry == num_retries:
                             raise
```
AugPro, line 263: If we do reraise the error after enough attempt, shouldn't we `connection.invalidate()` and `connection.rollback()` before re-raising  ?

zzzeek:
sure, looks like I wrote this in a hurry one day over at #5657, I initially thought this was from some production code somehwere but it's not.  so feel free to refactor / rewrite if you have a production ready routine

AugPro:
> sure, looks like I wrote this in a hurry one day over at #5657, I initially thought this was from some production code somehwere but it's not. so feel free to refactor / rewrite if you have a production ready routine

We've used this code in prod for a few days only. I'll submit another issue/PR if we find other suggestions 🙂.

Do you plan on including this feature in the lib some day @zzzeek ? Dropped connections are a pretty recurrent issue on our end, and this retry mechanism is really useful for read-only operations.

doc/build/faq/connections.rst
AugPro: Do you think we should invalidate/rollback the connection in all error cases (before this if-statement) @zzzeek ? I'm not knowledgable enough to assess if this makes sense 😕

CaselIT: it likely makes sense to move this if after the rollback and maybe update the log message to be `"disconnection error, retrying operation %s", retry == num_retries`

AugPro: updated in 61f3e5e, WDYT ?

CaselIT (approved):
💯

PR #12316, opened 2025-02-05
Add Support for Oracle VECTOR Datatype in SQLAlchemy

suraj-ora-2020:
### Description

This PR adds support for the VECTOR datatype in the Oracle dialect of SQLAlchemy, which corresponds to the vector data type introduced in Oracle Database 23ai. This new VECTOR type allows for efficient storage and retrieval of high-dimensional vector data, enabling SQLAlchemy users to leverage vector-based queries. By treating VECTOR as a "first-class" type, this integration makes it easier to work with Oracle’s vector-based capabilities directly in Python.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**


CaselIT:
Hi,

Could you also open an issue so it can be better tracked and referenced?
From a very quick look, it seems like a good start but I have a few initial comments:

There seem to be two changes here:
- vector type
- fetch type

let's split those with their own issues and PR please.

Also does oracle support additional operator with the vector type, similar to the postgresql ones (I mean things like @> <@ etc)?

CaselIT:
also for the fetch change, I think this is something that could be a parameter in the `fetch()`, similar to with_ties and percent already present https://docs.sqlalchemy.org/en/20/core/selectable.html#sqlalchemy.sql.expression.Select.fetch

for reference spec seems to be https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/SELECT.html#GUID-CFA006CA-6FF1-4972-821E-6996142A51C6__BABBADDD

Issue #6810, opened 2021-07-24
Inline pep484 typing in the SQLAlchemy source


CaselIT:
The plan would be to phase out the stubs package at https://github.com/sqlalchemy/sqlalchemy2-stubs and move all the type annotation inline with the source.

Targeting for 2.0 but this may realistically be pushed to 2.x

CaselIT:
We should try https://github.com/ambv/retype/

sqla-tester:
**Federico Caselli** referenced this issue:

**Properly type _generative** https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3423

sqla-tester:
**Mike Bayer** has proposed a fix for this issue in the **main** branch:

**initial reorganize for static typing** https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3447

zzzeek:
related issues:

#7535 
#7519

zzzeek:
also there's the issue of replacing existing sqlalchemy-stubs packages in type checkers and it's suggested we use a "positive" masking approach, as a "negative" uninstall approach is not supported by type checkers that bundle the stubs: https://github.com/microsoft/pylance-release/issues/840#issuecomment-1008329878

sqla-tester:
**Mike Bayer** has proposed a fix for this issue in the **main** branch:

**WIP for ORM typing** https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3495

zzzeek:
current goals are:

1. get https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3495, based on https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/3447, working.   POC here is that I would like to include updating the mypy plugin and removing the dependency on sqlalchemy2-stubs, w/ mypy tests passing.  that will be the proof of concept for 3495, which will also validate 3447.
2. then merge 3447 and 3495
3. after those two are in, that's the "hard part" of typing, that is, the column expression / operator / ORM mapping stuff.  then we can do a package-by-package or module-by-module approach in separate gerrits 
[...]

PR #12384, opened 2025-02-28
Add type annotations to `postgresql.array`

dlax:
Improved static typing for `postgresql.array()` by making the type parameter (the type of array's elements) inferred from the `clauses` and `type_` arguments while also ensuring they are consistent.

Also completed type annotations of `postgresql.ARRAY` following commit 0bf7e02afbec557eb3a5607db407f27deb7aac77 and added type annotations for functions `postgresql.Any()` and `postgresql.All()`.
 
Finally, fixed shadowing `typing.Any` by the `Any()` function through aliasing as `typing_Any` and removed unused `_type_tuple` attribute of `postgresql.array`.

Related to #6810

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/array.py
+++ b/lib/sqlalchemy/dialects/postgresql/array.py
@@ -4,7 +4,6 @@
 #
 # This module is part of SQLAlchemy and is released under
 # the MIT License: https://www.opensource.org/licenses/mit-license.php
-# mypy: ignore-errors
 
 
 from __future__ import annotations
@@ -29,6 +28,7 @@ from ...sql import operators
 
 if TYPE_CHECKING:
     from ...engine.interfaces import Dialect
+    from ...sql._typing import _ColumnExpressionArgument
     from ...sql._typing import _TypeEngineArgument
     from ...sql.elements import ColumnElement
     from ...sql.elements import Grouping
@@ -45,22 +45,30 @@ if TYPE_CHECKING:
 _T = TypeVar("_T", bound=typing_Any)
 
 
-def Any(other, arrexpr, operator=operators.eq):
+def Any(
+    other: typing_Any,
+    arrexpr: _ColumnExpressionArgument[_T],
+    operator: OperatorType = operators.eq,
+) -> ColumnElement[bool]:
     """A synonym for the ARRAY-level :meth:`.ARRAY.Comparator.any` method.
     See that method for details.
 
     """
 
-    return arrexpr.any(other, operator)
+    return arrexpr.any(other, operator)  # type: ignore[no-any-return, union-attr]  # noqa: E501
 
 
-def All(other, arrexpr, operator=operators.eq):
+def All(
+    other: typing_Any,
+    arrexpr: _ColumnExpressionArgument[_T],
+    operator: OperatorType = operators.eq,
+) -> ColumnElement[bool]:
     """A synonym for the ARRAY-level :meth:`.ARRAY.Comparator.all` method.
     See that method for details.
 
     """
 
-    return arrexpr.all(other, operator)
+    return arrexpr.all(other, operator)  # type: ignore[no-any-return, union-attr]  # noqa: E501
 
 
 class array(expression.ExpressionClauseList[_T]):
--- a/lib/sqlalchemy/dialects/postgresql/array.py
+++ b/lib/sqlalchemy/dialects/postgresql/array.py
@@ -10,7 +10,7 @@
 from __future__ import annotations
 
 import re
-from typing import Any
+from typing import Any as typing_Any
 from typing import Optional
 from typing import TypeVar
 
@@ -24,7 +24,7 @@ from ...sql import operators
 from ...sql._typing import _TypeEngineArgument
 
 
-_T = TypeVar("_T", bound=Any)
+_T = TypeVar("_T", bound=typing_Any)
 
 
 def Any(other, arrexpr, operator=operators.eq):
@@ -237,7 +237,7 @@ class ARRAY(sqltypes.ARRAY):
 
     def __init__(
         self,
-        item_type: _TypeEngineArgument[Any],
+        item_type: _TypeEngineArgument[typing_Any],
         as_tuple: bool = False,
         dimensions: Optional[int] = None,
         zero_indexes: bool = False,
--- a/lib/sqlalchemy/dialects/postgresql/array.py
+++ b/lib/sqlalchemy/dialects/postgresql/array.py
@@ -11,8 +11,11 @@ from __future__ import annotations
 
 import re
 from typing import Any as typing_Any
+from typing import Iterable
 from typing import Optional
+from typing import TYPE_CHECKING
 from typing import TypeVar
+from typing import Union
 
 from .operators import CONTAINED_BY
 from .operators import CONTAINS
@@ -21,7 +24,15 @@ from ... import types as sqltypes
 from ... import util
 from ...sql import expression
 from ...sql import operators
-from ...sql._typing import _TypeEngineArgument
+
+if TYPE_CHECKING:
+    from ...sql._typing import _TypeEngineArgument
+    from ...sql.elements import Grouping
+    from ...sql.expression import BindParameter
+    from ...sql.operators import OperatorType
+    from ...sql.selectable import _SelectIterable
+    from ...sql.type_api import TypeEngine
+    from ...util.typing import Self
 
 
 _T = TypeVar("_T", bound=typing_Any)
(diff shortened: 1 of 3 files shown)


lib/sqlalchemy/dialects/postgresql/array.py
```diff
@@ -4,30 +4,44 @@
...
 from .operators import CONTAINED_BY
 from .operators import CONTAINS
 from .operators import OVERLAP
 from ... import types as sqltypes
 from ... import util
+from ...engine.interfaces import Dialect
 from ...sql import expression
 from ...sql import operators
 from ...sql._typing import _TypeEngineArgument
+from ...sql.elements import ColumnElement
```
Polandia94, line 30: I think is better to use a 
`if TYPE_CHECKING: 
`import block, to avoid unnecesary imports

dlax: Indeed; done.

lib/sqlalchemy/dialects/postgresql/array.py
```diff
@@ -4,15 +4,20 @@
...
 # This module is part of SQLAlchemy and is released under
 # the MIT License: https://www.opensource.org/licenses/mit-license.php
-# mypy: ignore-errors
 
 
 from __future__ import annotations
 
 import re
-from typing import Any
+from typing import Any as Any_
```
CaselIT, line 12: use typing_Any (it's used in other places)

lib/sqlalchemy/dialects/postgresql/array.py
```diff
@@ -4,15 +4,20 @@
...
+from typing import Iterable
 from typing import Optional
+from typing import Self
```
CaselIT, line 15: iirc we need to import this from util.typing since it's not present in all python version we support

lib/sqlalchemy/dialects/postgresql/array.py
```diff
@@ -21,13 +26,25 @@
...
+    from ...sql.type_api import _ResultProcessorType
+    from ...sql.type_api import TypeEngine
 
-_T = TypeVar("_T", bound=Any)
 
+_T = TypeVar("_T", bound=Any_)
 
-def Any(other, arrexpr, operator=operators.eq):
+
+def Any(other, arrexpr, operator=operators.eq):  # type: ignore[no-untyped-def]
```
CaselIT, line 47: may as well type this instead of ignoring it. same below

lib/sqlalchemy/dialects/postgresql/array.py
```diff
@@ -127,15 +149,21 @@ def __init__(self, clauses, **kw):
                     if main_type.dimensions is not None
                     else 2
                 ),
-            )
+            )  # type: ignore[assignment]
```
CaselIT, line 152: strange that we have this type error. self.type is a typeengine

lib/sqlalchemy/dialects/postgresql/array.py
```diff
@@ -127,15 +149,21 @@ def __init__(self, clauses, **kw):
...
         return (self,)
 
-    def _bind_param(self, operator, obj, _assume_scalar=False, type_=None):
+    def _bind_param(
+        self,
+        operator: OperatorType,
+        obj: Any_,
+        type_: Optional[TypeEngine[_T]] = None,
+        _assume_scalar: bool = False,
+    ) -> BindParameter[_T]:
```
CaselIT, line 166: I guess we need to change the return type of this function in the base type.

@zzzeek would the correct type be `BindParameter | ExpressionClauseList`?

lib/sqlalchemy/dialects/postgresql/array.py
```diff
@@ -322,81 +354,87 @@ def overlap(self, other):
...
 
     @property
-    def python_type(self):
+    def python_type(self) -> Type[Any_]:
         return list
 
-    def compare_values(self, x, y):
-        return x == y
+    def compare_values(self, x: Any_, y: Any_) -> bool:
+        return x == y  # type: ignore[no-any-return]
```
CaselIT, line 365: we could probably just remove hashable, python_type and compare_values since they are the same as the base type (just a consideration)

CaselIT (commented):
Thanks, left a few suggestion.

I think to do a better job the base array type should also be typed.

let us know if you want to do this or if you prefer we take this over

dlax:
> I think to do a better job the base array type should also be typed.

Indeed; thanks for the suggestion.

> let us know if you want to do this or if you prefer we take this over

No; that's fine, I can work on this too.

I'll then open another dedicated PR for this and we can get back to this one later on. (I'll probably still address (some of) your other comments here in the meantime, while things are fresh in my mind.)

lib/sqlalchemy/dialects/postgresql/array.py
dlax: Not sure, but it might be due to `postgresql.array` being generic (`class array(expression.ExpressionClauseList[_T]):`) whereas the base `ARRAY` is not (`class ARRAY(SchemaEventTarget, Indexable, Concatenable, TypeEngine[Sequence[Any]]):`. So when assigning an `ARRAY()` value to `self.type`(which is a `TypeEngine[_T]`, the generic is lost, isn't it?

dlax:
> I think to do a better job the base array type should also be typed.

See https://github.com/sqlalchemy/sqlalchemy/pull/12386; I adjusted the code here to use annotations from the other branch but did not rebase, hence test failures).

lib/sqlalchemy/dialects/postgresql/array.py
dlax: I would say:
~~~python
def Any(other: Any, arrexpr, operator: OperatorType = operators.eq) -> ColumnElement[bool]: ...
~~~
but it's not clear to me what `arrexpr` should be?

dlax: I went on with `arrexpr: ColumnElement[Any]` .

CaselIT: sorry, missed this. You can look at these: https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/sql/_elements_constructors.py

it seems that it uses `_ColumnExpressionArgument[_T]`

dlax: Thank you for the hint; fixed.

dlax:
Rebased to benefit from 0bf7e02afbec557eb3a5607db407f27deb7aac77 (https://github.com/sqlalchemy/sqlalchemy/pull/12386).

dlax:
The improvements to `postgresql.array()` might be worth a changelog entry? (Other changes would probably not be visible from end users).

lib/sqlalchemy/dialects/postgresql/array.py
CaselIT: @zzzeek sorry to ping you again here

CaselIT:
there don't seem to be behavior changes introduced, so I'm not sure how useful it would be a change log just mentioning typing.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 78eea29f1de850afda036502974521969629de7e of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 78eea29f1de850afda036502974521969629de7e: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5785

CaselIT:
I've manually updated the gerrit with the last commit

sqla-tester:

recheck


(testing something)

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5785

sqla-tester:

recheck

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5785

sqla-tester:

recheck

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5787

sqla-tester:

do you want to merge w/o a changelog?  I guess it's fine

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5785

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5785 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5787 has been **merged**. Congratulations! :)

Issue #6274, opened 2021-04-15
'CompoundSelect' object has no attribute 'mapper'


andrewcooke:
**Describe the bug**
Calling aliased(union(Query, Query)) gives the error 'CompoundSelect' object has no attribute 'mapper'

**Expected behavior**
I expect it to return something that represents the SQL ALIAS construct.

**To Reproduce**

```py
class SQLAlchemyTest(TestCase):

    def test_sqlalchemy(self):

        engine = create_engine('sqlite:///:memory:')
        Base = declarative_base()
        Session = sessionmaker(engine)

        class Table(Base):
            __tablename__ = 'table'
            id = Column(Integer, primary_key=True)

        Base.metadata.create_all(engine)

        with Session() as s:
            q1 = s.query(Table).filter(Table.id == 1)
            q2 = s.query(Table).filter(Table.id == 2)
            q3 = aliased(union(q1, q2))
```

**Error**

```
Ran 1 test in 0.010s

FAILED (errors=1)

Error
Traceback (most recent call last):
  File "/home/andrew/project/choochoo/py/tests/test_search.py", line 52, in test_sqlalchemy
    q3 = aliased(union(q1, q2))
  File "/home/andrew/project/choochoo/py/env/lib/python3.9/site-packages/sqlalchemy/orm/util.py", line 1201, in aliased
    return AliasedClass(
  File "/home/andrew/project/choochoo/py/env/lib/python3.9/site-packages/sqlalchemy/orm/util.py", line 495, in __init__
    mapper = insp.mapper
AttributeError: 'CompoundSelect' object has no attribute 'mapper'
```

**Versions.**
 - OS: OpenSuse 15.2
 - Python: 3.9.0
 - SQLAlchemy: 1.4.7
 - Database: sqlite3
 - DBAPI: sqlite

**Additional context**
https://github.com/sqlalchemy/sqlalchemy/discussions/6271

**Have a nice day!**

andrewcooke:
Actually, select() has the same problem:

```py
        with Session() as s:
            q1 = select(Table).filter(Table.id == 1)
            q2 = select(Table).filter(Table.id == 2)
            q3 = aliased(union(q1, q2))
```

gives the same error

```
Traceback (most recent call last):
  File "/home/andrew/project/choochoo/py/tests/test_search.py", line 52, in test_sqlalchemy
    q3 = aliased(union(q1, q2))
[...]

Issue #10236, opened 2023-08-15
remove select().c  / .columns, completely.   no trace


zzzeek:
I thought we already removed this in 2.0 but we didn't.   Erase it completely for 2.1 please

sqla-tester:
**Mike Bayer** has proposed a fix for this issue in the **main** branch:

**remove _implicit_subquery and all derived functions** https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5570

PR #12433, opened 2025-03-14
Coerce aliased Select and CompoundSelect to subquery

rgroothuijsen:
Fixes issue #6274

### Description

When `aliased()` is used on a select statement or compound select statement, it currently throws an error. The proposed solution in #6274 was to instead return a subquery and emit a warning to the user that aliases are not supported for these types of statements.

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/orm/util.py
+++ b/lib/sqlalchemy/orm/util.py
@@ -85,6 +85,7 @@ from ..sql.cache_key import MemoizedHasCacheKey
 from ..sql.elements import ColumnElement
 from ..sql.elements import KeyedColumnElement
 from ..sql.selectable import FromClause
+from ..sql.selectable import GenerativeSelect
 from ..util.langhelpers import MemoizedSlots
 from ..util.typing import de_stringify_annotation as _de_stringify_annotation
 from ..util.typing import eval_name_only as _eval_name_only
@@ -1015,6 +1016,19 @@ class AliasedInsp(
         flat: bool = False,
         adapt_on_names: bool = False,
     ) -> Union[AliasedClass[_O], FromClause]:
+        if isinstance(element, GenerativeSelect):
+            util.warn(
+                f"{type(element).__name__} can't be aliased, "
+                f"coercing to subquery instead."
+            )
+            if name:
+                return coercions.expect(
+                    roles.FromClauseRole, element.subquery(name=name)
+                )
+            else:
+                return coercions.expect(
+                    roles.AnonymizedFromClauseRole, element.subquery()
+                )
         if isinstance(element, FromClause):
             if adapt_on_names:
                 raise sa_exc.ArgumentError(
--- a/lib/sqlalchemy/orm/util.py
+++ b/lib/sqlalchemy/orm/util.py
@@ -1021,15 +1021,13 @@ class AliasedInsp(
                 f"{type(element).__name__} can't be aliased, "
                 f"coercing to subquery instead."
             )
-            if name:
-                return coercions.expect(
-                    roles.FromClauseRole, element.subquery(name=name)
-                )
-            else:
-                return coercions.expect(
-                    roles.AnonymizedFromClauseRole, element.subquery()
-                )
-        if isinstance(element, FromClause):
+            return coercions.expect(
+                roles.FromClauseRole,
+                element.subquery(name=name),
+                flat=flat,
+                allow_select=True
+            )
+        elif isinstance(element, FromClause):
             if adapt_on_names:
                 raise sa_exc.ArgumentError(
                     "adapt_on_names only applies to ORM elements"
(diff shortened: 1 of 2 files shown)


zzzeek (changes requested):
great job. I think we can simplify the fix and also add some more test support, see inline comments

lib/sqlalchemy/orm/util.py
```diff
@@ -1015,6 +1016,19 @@ def _alias_factory(
         flat: bool = False,
         adapt_on_names: bool = False,
     ) -> Union[AliasedClass[_O], FromClause]:
+        if isinstance(element, GenerativeSelect):
+            util.warn(
```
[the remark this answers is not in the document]
rgroothuijsen: I couldn't get the exact code above to produce the expected result. It appears a GenerativeSelect can't be automatically coerced to FromClause without first turning it into a subquery.

lib/sqlalchemy/orm/util.py
```diff
@@ -1015,7 +1016,18 @@ def _alias_factory(
         flat: bool = False,
         adapt_on_names: bool = False,
     ) -> Union[AliasedClass[_O], FromClause]:
-        if isinstance(element, FromClause):
+        if isinstance(element, GenerativeSelect):
+            util.warn(
+                f"{type(element).__name__} can't be aliased, "
+                f"coercing to subquery instead."
+            )
+            return coercions.expect(
```
CaselIT, line 1024: how about we instead just raise an informative unsupported error here? This is what other cases like this behave in 2.0 so I don't think it makes much sense to add support for a previously unsupported feature (from by understanding this raised an exception in all cases) and also deprecate it right away

CaselIT (changes requested):
I would prefer just go to the "raise informative error"  step without passing to the deprecation step, since we had an uninformative error already

lib/sqlalchemy/orm/util.py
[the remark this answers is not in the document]
zzzeek: the coercion itself does the subquery() call.  I tested the above patch with the original case at #6274, and if it is not working for an additional case here then I'd want to add that support to coercions, I can do that if you show me the failure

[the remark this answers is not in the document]
zzzeek: somethings weird here because this worked in my main branch and isnt working in this PR, checking

zzzeek (changes requested):
OK we changed from coercion + warning to raise in #10236.  So this PR needs to have a separate version for main vs. 2.0.  in main, these test should expect a raise.  in 2.0, we need to use expect_deprecated.  the tests themselves should be in test/orm/test_deprecated.py.

@CaselIT this is getting complicated do we want to just move to gerrit and straighten it out

CaselIT:
> @CaselIT this is getting complicated do we want to just move to gerrit and straighten it out

it's likely the best option

lib/sqlalchemy/orm/util.py
zzzeek: agree, this would be an exception raise.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 416dde8509ac05b209400603d9e00bc82aa47c79 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 416dde8509ac05b209400603d9e00bc82aa47c79: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6829

lib/sqlalchemy/orm/util.py
zzzeek: i believe perhaps we wanted to add a 2.0 (2.1 at this point) deprecation for:

    union(...).alias()

being a synonym of subquery()

likely a separate issue though (since that one woujld be 2.1 only)

zzzeek: OK what is happening is the coercion that I wanted to use here does in fact do a deprecation in the 2.0 branch.  so we'd just use that.   this is all fine

sqla-tester:

i just remembered that in changelogs that apply to 2.0 and 2.1 im supposed to be using the "backports" feature so they show up appropriately in both.  ill address those en masse later

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6829

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6829 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6830 has been **merged**. Congratulations! :)

PR #12459, opened 2025-03-24
Increase minimum required greenlet version

cjw296:
SQLAlchemy currently has an odd requirement around greenlet: `greenlet!=0.4.17`.
I couldn't spot why this was there from some lightweight annotation digging, but it's a really old version.

Unfortunately, this lets dependency resolvers pick *very* old version of greenlet that don't work with Python 3.
I discovered this while moving [pytest-sqlalchemy](https://github.com/toirl/pytest-sqlalchemy) to be [`uv`](https://github.com/astral-sh/uv)-based and using GitHub Actions for CI, where I've been using `uv sync --all-extras --dev --resolution lowest` to check that the minimum required versions of dependencies actually work.

Here's the failing CI run: https://github.com/toirl/pytest-sqlalchemy/actions/runs/14029663597/job/39274427927?pr=11

I've picked `greenlet 3.0.0` as the minimum version, since it's the first version that appears to have dropped Python 2 support.

This isn't a massive deal, but I think the change does help SQLAlchemy present a more correct view of its dependencies.
It should also be backported to the 1.4 branch, but I don't know what the process is for that.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,7 @@ Changelog = "https://docs.sqlalchemy.org/latest/changelog/index.html"
 Discussions = "https://github.com/sqlalchemy/sqlalchemy/discussions"
 
 [project.optional-dependencies]
-asyncio = ["greenlet!=0.4.17"]
+asyncio = ["greenlet>=3"]
 mypy = [
     "mypy >= 1.7",
     "types-greenlet >= 2"
@@ -59,7 +59,7 @@ oracle-oracledb = ["oracledb>=1.0.1"]
 postgresql = ["psycopg2>=2.7"]
 postgresql-pg8000 = ["pg8000>=1.29.3"]
 postgresql-asyncpg = [
-    "greenlet!=0.4.17",  # same as ".[asyncio]" if this syntax were supported
+    "greenlet>=3",  # same as ".[asyncio]" if this syntax were supported
     "asyncpg",
 ]
 postgresql-psycopg2binary = ["psycopg2-binary"]
@@ -68,19 +68,19 @@ postgresql-psycopg = ["psycopg>=3.0.7,!=3.1.15"]
 postgresql-psycopgbinary = ["psycopg[binary]>=3.0.7,!=3.1.15"]
 pymysql = ["pymysql"]
 aiomysql = [
-    "greenlet!=0.4.17",  # same as ".[asyncio]" if this syntax were supported
+    "greenlet>=3",  # same as ".[asyncio]" if this syntax were supported
     "aiomysql",
 ]
 aioodbc = [
-    "greenlet!=0.4.17",  # same as ".[asyncio]" if this syntax were supported
+    "greenlet>=3",  # same as ".[asyncio]" if this syntax were supported
     "aioodbc",
 ]
 asyncmy = [
-    "greenlet!=0.4.17",  # same as ".[asyncio]" if this syntax were supported
+    "greenlet>=3",  # same as ".[asyncio]" if this syntax were supported
     "asyncmy>=0.2.3,!=0.2.4,!=0.2.6",
 ]
 aiosqlite = [
-    "greenlet!=0.4.17",  # same as ".[asyncio]" if this syntax were supported
+    "greenlet>=3",  # same as ".[asyncio]" if this syntax were supported
     "aiosqlite",
 ]
 sqlcipher = ["sqlcipher3_binary"]


cjw296:
Full output of the failing situation:

```bash
Run uv sync --all-extras --dev --resolution lowest
warning: The direct dependency `sqlalchemy-utils` is unpinned. Consider setting a lower bound when using `--resolution lowest` to avoid using outdated versions.
  × Failed to build `greenlet==0.1`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `setuptools.build_meta:__legacy__.build_wheel` failed (exit
      status: 1)

      [stderr]
        File "<string>", line 14, in <module>
        File
      "/home/runner/work/_temp/setup-uv-cache/builds-v0/.tmpsr7Fjr/lib/python3.11/site-packages/setuptools/build_meta.py",
      line 334, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=[])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File
      "/home/runner/work/_temp/setup-uv-cache/builds-v0/.tmpsr7Fjr/lib/python3.11/site-packages/setuptools/build_meta.py",
      line 304, in _get_build_requires
          self.run_setup()
        File
      "/home/runner/work/_temp/setup-uv-cache/builds-v0/.tmpsr7Fjr/lib/python3.11/site-packages/setuptools/build_meta.py",
      line 522, in run_setup
          super().run_setup(setup_script=setup_script)
        File
      "/home/runner/work/_temp/setup-uv-cache/builds-v0/.tmpsr7Fjr/lib/python3.11/site-packages/setuptools/build_meta.py",
      line 320, in run_setup
          exec(code, locals())
        File "<string>", line 3, in <module>
        File
      "/home/runner/work/_temp/setup-uv-cache/sdists-v[9](https://github.com/toirl/pytest-sqlalchemy/actions/runs/14029663597/job/39274427927?pr=11#step:4:10)/pypi/greenlet/0.1/ueULOOmf3fJ6ZtCM-Vmit/src/ez_setup/__init__.py",
      line 176
          print "Setuptools version",version,"or greater has been installed."
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      SyntaxError: Missing parentheses in call to 'print'. Did you mean
      print(...)?

      hint: This usually indicates a problem with the package or the build
      environment.
  help: `greenlet` (v0.1) was included because `pytest-sqlalchemy`
        (v0.2.0) depends on `sqlalchemy>=1.4` (v1.4.0) which depends on
        `greenlet<0.4.[17](https://github.com/toirl/pytest-sqlalchemy/actions/runs/14029663597/job/39274427927?pr=11#step:4:18) | >0.4.17`
```

CaselIT:
Hi,

> Unfortunately, this lets dependency resolvers pick very old version of greenlet that don't work with Python 3.

This seems like a bug in the resolver, not of sqlalchemy. Have you considered reporting this to uv? I doesn't make any sense for it to select a version that doesn't advertise python3 support.

> Given that greenlet<3.1 [appears to not work on Python 3.13](https://github.com/toirl/pytest-sqlalchemy/actions/runs/14030537434/job/39276989709), it may be pragmatic to just go for 3.1 as the minimum:

This again seems uv that's broken. Why does it select an incompatible version?? the resolver has one job and it seems to fail here.

> It should also be backported to the 1.4 branch, but I don't know what the process is for that.

I'm not sure if this is possible, as 1.4 still supports python 2 (Don't remember if we had a constraint to install greenlet only on python 3). Also it's unlikely that 1.4 will have further releases

> This isn't a massive deal, but I think the change does help SQLAlchemy present a more correct view of its dependencies.

I don't agree that sqlalchemy currently has an incorrect view of the dependencies.

To me the issue appears to be on uv side. Overall uv still seems to still have issues, since more than an user reported broken installs of sqlalchemy when using uv.

That said I'm not attached to that constraint, but I believe you should first report this to uv since these seems uv bugs

CaselIT (changes requested):
Let's first try to solve the uv bug

cjw296:
I'm afraid you may be missing the crux of the issue here...

uv is resolving dependencies correctly, and `uv sync --resolution lowest` is a useful tool for discovering bugs like this.

So, to try and explain the problem differently, SQLAlchemy puts no lower bound on its constraint for `greenlet`, which means that `greenlet 0.1`, which [does not declare that it is incompatible with Python 3](https://pypi.org/project/greenlet/0.1/), is a valid choice for a dependency resolver to make.

The "bug" here could be argued to be with`greenlet`, for not declaring that 0.1 doesn't work with Python 3, but that's a tenuous one given that Python 3 may not have been released when greenlet 0.1 was released (not actually true, since Python 3 was released by 2006, but the state of Python packaging metadata significantly poorer back then.)

There's a potential "feature" here, for `uv` to exclude releases "so old they can't be expected to work" when using the `--resolution lowest" mode, but that seems like quite a big and heuristic ask.

My view is that the "bug" here is SQLAlchemy not declaring the minimum version of `greenlet` with which it will actually work correctly. The reality is that this won't hit many people, since most dependency resolution is "eager" and picks the newest version that will work, but there may well be cases where other requirements that also have `greenlet` as a dependency resulting in the final resolved version being one that doesn't work with SQLAlchemy.

So, I see this as a small change that improves the correctness of SQLAlchemy's metadata. What are the potential downsides of this change?

zzzeek:
hi Chris !

> So, I see this as a small change that improves the correctness of SQLAlchemy's metadata. What are the potential downsides of this change?

In the general sense, what we're doing is trying very hard not to put ourselves on the hook for supporting things that should be fixed by others.   This is both about keeping proper boundaries for the project itself as well as doing a net good for the community in getting other software improved.  I'll point to a big giant example that just happened, where [python 3.14 decided that it would silently skip over conditionals when evaluating type annotations](https://github.com/python/cpython/issues/130881#issuecomment-2704089137).  And indeed, as I suspected, [this really odd behavior was in fact part of the spec they made!](https://peps.python.org/pep-0649/#backwards-compatibility-with-stock-semantics).  However after much discussion, the cpython folks [decided they can in fact fix it](https://github.com/python/cpython/pull/130935).   Even though I have a workaround all ready to go for that issue, it's a lot better that they're just going to fix that problem after being pushed.

That's a dramatic example, but the general idea that we try really hard to not resort to workarounds without even suggesting improvements for the upstream systems we depend on.       

For the category of non-standard tooling, things like trio/anyio, Poetry, fastapi, when these tools illustrate problems that dont exist at all in the python standard tooling we are understandably even more skeptical overall since to us these appear as tools that duplicate functionality already found elsewhere, with the promise of being "better", but end up introducing weird problems that place a burden on everyone else.   We've gotten things fixed in Poetry at least.

as though I weren't droning on long enough, another reason we dig into "small" changes is that we want to see what the implications of the change are and if we are doing this correctly, because if there is a bigger problem to be solved, we'd like to hit it just once and not fix the same problem over and over again for fragments of the software.   In this case, the change to **only** greenlet looks very arbitrary?    This command `uv --lowest-resolution` seems to be introducing the concept that Python requirements have no intrinsic means of lower bounding software versions automatically under any circumstances, which would appear to suggest that *any* requirement in Python that does not have a lower bound is essentially broken.   So why not all the other requirements in pyproject.toml like `pymssql`, `mysql-connector-python`, `pymysql` ?   These projects all have old versions that are incompatible with newer pythons as well.    did your `uv --lowest-resolution` test omit these because you aren't working with those databases?

Bigger picture why is it even valid to have a name in a requirements file that has no lower bound?   Shouldn't that simply be an error practically by definition?      I think if this PR came in with lower bounds for **all** requirements that would at least make more sense to me?

zzzeek:
here is the same feature being requested for pip, coming from the openstack world (that I work in): https://github.com/pypa/pip/issues/8085

zzzeek:
reading the use case there, this tool makes sense for an *application* that has the final requirements file set up.   Note that openstack  names [all of its dependencies explicitly](https://github.com/openstack/requirements/blob/4a006431f60be70eea411a5c5f2e9119b8cdb567/upper-constraints.txt#L558), even for libraries that are not directly referenced by the application itself.

Chris I think this tool is intended for that approach, where the ultimate downstream application fully names all dependencies and dependencies-of-dependencies.     I dont think pointing it at libraries with open ended constraints is going to be a good path.

pfmoore:
> Chris I think this tool is intended for that approach, where the ultimate downstream application fully names all dependencies and dependencies-of-dependencies. I dont think pointing it at libraries with open ended constraints is going to be a good path.

I found this thread from the linked pip feature request. I agree with @zzzeek here - applications can reasonably use something like `--resolution lowest` to validate that lower bounds specified in the application constraints file are correct, but "testing" dependency specifiers in individual libraries using that feature is neither helpful nor scalable.

For the specific case of `pytest-sqlalchemy`, you should probably be testing with `--resolution lowest-direct`, as that will validate *your* lower bounds, while not putting you in a position where you can't test code under your control without getting other projects to make changes.

CaselIT:
so to summarise:
- supporting `uv sync --all-extras --dev --resolution lowest` is not something we want to do, since it would require manually setting a min version for each dependency for each different python version. This isn't feasible.
- that said adding a min version to greenlet may make sense. Since sqlalchemy is used by many other project I think 3.0 is a bit too recent, especially for the 2.0 series. Something like `>=1.0` may be more appropriate there, since we don't want do cause unneeded issues to users for no particular reason.
- on 2.1 it's likely fine to use >=3.0, but then again I question why should we do that since `>=1.0` works fine in 2.1 too, so it may just be simpler to keep it

thoughs?

pfmoore:
@CaselIT Who is "we" in your comment? I suggested `--resolution lowest-direct` for `pytest-sqlalchemy`, would that not work for you (and if not, why not?) Then all you have to do is fix any constraints on *your* dependencies, which are under your control.

zzzeek:
I would vote for >= 1.0 and leave it at that, that's apparently just one non-beta release past the 0.4.17 version.

CaselIT:
We refers to sqlalchemy devs, that don't maintain `pytest-sqlalchemy`.

I don't have an opinion regarding the alternative resolution, seems like it would work for pytest-sqlalchemy.
Regarding why not supporting lower from what I gather we would have to list different constraints for each python version, (example 3.13 needs >=3.1 for greenlet), and that not something that I think would add value

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 4bd856b9c164df984f05c094c977686470ed4244 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 4bd856b9c164df984f05c094c977686470ed4244: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5823

sqla-tester:

Mike do you want a change note?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5823

zzzeek:
> 
> Mike do you want a change note?
> 
> View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5823

this can just go in, 1.0 is just one bump past the 0.4.17 version

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5823 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5824 has been **merged**. Congratulations! :)

cjw296:
Apologies for the delay in replying...

Just a note that greenlet 1.0 [doesn't work with Python 3.10 and above](https://github.com/toirl/pytest-sqlalchemy/actions/runs/14102537430/job/39501688650?pr=12), so may not be a good choice.

I agree with @pfmoore 's [comment](https://github.com/sqlalchemy/sqlalchemy/pull/12459#issuecomment-2748269676) above and [said similar](https://github.com/pypa/pip/issues/8085#issuecomment-2757196495) on the [pip issue](https://github.com/pypa/pip/issues/8085): `lowest-direct` is probably the pragmatic choice to achieve my aims.

I don't agree with @zzzeek that this should only be done for apps; for apps, I think a fully locked down set of versions in a uv.lock file is the only sane way to go, maybe Mike meant frameworks? (like Django, etc, where they're not the final app, but do have a much more constrained set of dependencies to sanely target?) I _do_ think libraries, such as `pytest-sqlalchemy` and indeed `SQLAlchemy` should correctly specify the minimum versions they require of their dependencies. In the day and age of supply chain attacks and [ever shorter support windows](https://devguide.python.org/versions/), my personal preference is to have these be more aggressive, but that's definitely "my thing" rather than what I'd suggest others necessarily do.

To [Mike's point](https://github.com/sqlalchemy/sqlalchemy/pull/12459#issuecomment-2747917560) about "why only greenlet?": because it's the one that caused the CI to blow up ;-) I know the pain of bigger-than-needed changes and really appreciate all the free work OSS maintainers do, so only want to introduce needed changes rather things-that-could-be-changed, unless I want to support that going forwards.

I'm a little confused as to why `greenlet` is being installed at all, I wonder if `uv sync --all-extras` is doing all extras *even for dependencies*? That would certainly feel like a `uv` bug, let me check...

CaselIT:
> Just a note that greenlet 1.0 [doesn't work with Python 3.10 and above](https://github.com/toirl/pytest-sqlalchemy/actions/runs/14102537430/job/39501688650?pr=12), so may not be a good choice.

sure, but that's only a problem if you use that resolver option of uv, that as mentioned is not something we want to support.

> SQLAlchemy should correctly specify the minimum versions they require of their dependencies. 

we do. sqlalchemy works with greenlet 1.0 (or before that 0.4.17). It's not feasible to list the min version for each python version, since that's by default incomplete, since we cannot list the constraints for newer python version.

> I'm a little confused as to why greenlet is being installed at all

it's required in 2.0. will move to an extra in 2.0

cjw296:
So, it looks like SQLAlchemy 1.4 [*always* requires greenlet](https://github.com/sqlalchemy/sqlalchemy/blob/rel_1_4_54/setup.cfg#L46), is that intentional? 

I think that's the root cause of the CI failures I'm seeing.

CaselIT:
2.0 is the same too, so yes, it wasn't left there for 4+ years by mistake

cjw296:
Looks like [lowest-direct is working](https://github.com/toirl/pytest-sqlalchemy/actions/runs/14104691930/job/39508445890).

CaselIT:
that's great. lower should work also on 2.1 since we removed greenlet as required lib.

notatallshaw:
@cjw296 It occurs to me that using `lowest` with `uv sync`, which does universal resolution, for libraries like greenlet which does not support being built against newer Python versions with older releases, you likely need to mark greenlet as `no-build-package` in the uv configuration: https://docs.astral.sh/uv/reference/settings/#no-build-package

If this doesn't make a lock file that avoids building greenlet for Python versions it doesn't support you should discuss with the uv team.

PR #12463, opened 2025-03-25
Type postgresql.aggregate_order_by()

dlax:
Overloading of `__init__()` is needed, probably for the same reason as it is in `ReturnTypeFromArgs`.

Related to #6810.

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/ext.py
+++ b/lib/sqlalchemy/dialects/postgresql/ext.py
@@ -8,6 +8,10 @@
 from __future__ import annotations
 
 from typing import Any
+from typing import Iterable
+from typing import List
+from typing import Optional
+from typing import overload
 from typing import Sequence
 from typing import TYPE_CHECKING
 from typing import TypeVar
@@ -28,12 +32,17 @@ from ...sql.visitors import InternalTraversal
 
 if TYPE_CHECKING:
     from ...sql._typing import _ColumnExpressionArgument
+    from ...sql.elements import ClauseElement
+    from ...sql.elements import ColumnElement
+    from ...sql.operators import OperatorType
+    from ...sql.selectable import FromClause
+    from ...sql.visitors import _CloneCallableType
     from ...sql.visitors import _TraverseInternalsType
 
 _T = TypeVar("_T", bound=Any)
 
 
-class aggregate_order_by(expression.ColumnElement):
+class aggregate_order_by(expression.ColumnElement[_T]):
     """Represent a PostgreSQL aggregate order by expression.
 
     E.g.::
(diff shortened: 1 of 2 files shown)


CaselIT:
let's also what in what happens in https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5799 since I proposed a slight change on the current overloads. Sadly I could not remove them, but I think I reduced at least the scope of the issue at https://github.com/sqlalchemy/sqlalchemy/blob/aae34df0b5aa7dfe02bdc19744b1b6bc8533ee91/lib/sqlalchemy/sql/_typing.py#L88-L91

or at least that is part of the problem

dlax:
> let's also what in what happens in https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5799 since I proposed a slight change on the current overloads.

I have to admit that this problem is not clear to me at the moment... Maybe we could track this as a dedicated issue, possibly with a reference in code? And maybe gather help from mypy people if we manage to produce a minimal example?

CaselIT:
I guess in this case we don't allow a literal here, so the changes done in 084eac6916bdb8cfa85fd39cfbd5a0401245454e do not apply

dlax:
> but maybe only 2 overload could be enough?

Can you clarify what you mean? There are only 2 overloads in my proposal.

In fact, I only implemented what's tested (i.e. Column and mapper's attribute). If there are more use cases (now or later), we can add more overloads and tests.

CaselIT:
indeed sorry, I counted 3 inits, but the 3rd on is the implementation.

I wounder if the second overload could be `_ColumnExpressionArgument` instead

dlax:
> I wounder if the second overload could be `_ColumnExpressionArgument` instead

It seems to work. Changed, thanks.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 701d979e20c6ca3e32b79145c20441407007122f of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 701d979e20c6ca3e32b79145c20441407007122f: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5839

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5839 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5840 has been **merged**. Congratulations! :)

PR #12527, opened 2025-04-15
chore: add type hint for reconstructor

SaidBySolo:
### Description

Add type hinting to the reconstructor decorator.

Related https://github.com/sqlalchemy/sqlalchemy/issues/6810

Function types are applied flexibly.

If necessary, type hints can be applied more strictly.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/orm/mapper.py
+++ b/lib/sqlalchemy/orm/mapper.py
@@ -4246,7 +4246,7 @@ def _dispose_registries(registries: Set[_RegistryType], cascade: bool) -> None:
         reg._new_mappers = False
 
 
-def reconstructor(fn):
+def reconstructor(fn: Callable[..., Any]) -> Callable[..., Any]:
     """Decorate a method as the 'reconstructor' hook.
 
     Designates a single method as the "reconstructor", an ``__init__``-like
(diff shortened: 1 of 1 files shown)


CaselIT:
Hi,

It probably would make more sense to use a generic `_FN=TypeVar('_FN', bound=Callable)` so that we could do `reconstructor(fn: _FN)->_FN`

SaidBySolo:
> Hi,
> 
> It probably would make more sense to use a generic `_FN=TypeVar('_FN', bound=Callable)` so that we could do `reconstructor(fn: _FN)->_FN`


https://github.com/sqlalchemy/sqlalchemy/blob/f2a9ecde29bb9d5daadd0626054ff8b54865c781/lib/sqlalchemy/orm/mapper.py#L124

Is it okay to use _Fn declared on this line?

CaselIT:
Either one should be fine. Type ignore is likely better so we don't change the code, but setattr is equivalent

SaidBySolo:
Since this is a PR that resolves type annotations, it seems better not to change the code, so I ignored it with ``# type: ignore[attr-defined]``.

No issues were found when testing locally with ``tox -e pep484``

CaselIT:
thanks!

PR #12537, opened 2025-04-18
refactor (orm): remove unused variables and simplify key lookups

ashm-dev:
Redundant variables and unnecessary conditions were removed across several modules. Improved readability and reduced code complexity without changing functionality.

### DIFF
--- a/lib/sqlalchemy/orm/clsregistry.py
+++ b/lib/sqlalchemy/orm/clsregistry.py
@@ -72,7 +72,7 @@ def _add_class(
         # class already exists.
         existing = decl_class_registry[classname]
         if not isinstance(existing, _MultipleClassMarker):
-            existing = decl_class_registry[classname] = _MultipleClassMarker(
+            decl_class_registry[classname] = _MultipleClassMarker(
                 [cls, cast("Type[Any]", existing)]
             )
     else:
--- a/lib/sqlalchemy/orm/clsregistry.py
+++ b/lib/sqlalchemy/orm/clsregistry.py
@@ -69,11 +69,11 @@ def _add_class(
 
     """
     if classname in decl_class_registry:
-        if not isinstance(
-            decl_class_registry[classname], _MultipleClassMarker
-        ):
-            decl_class_registry[classname] = _MultipleClassMarker(
-                [cls, cast("Type[Any]", decl_class_registry[classname])]
+        # class already exists.
+        existing = decl_class_registry[classname]
+        if not isinstance(existing, _MultipleClassMarker):
+            existing = decl_class_registry[classname] = _MultipleClassMarker(
+                [cls, cast("Type[Any]", existing)]
             )
     else:
         decl_class_registry[classname] = cls
--- a/lib/sqlalchemy/orm/clsregistry.py
+++ b/lib/sqlalchemy/orm/clsregistry.py
@@ -69,11 +69,11 @@ def _add_class(
 
     """
     if classname in decl_class_registry:
-        # class already exists.
-        existing = decl_class_registry[classname]
-        if not isinstance(existing, _MultipleClassMarker):
-            existing = decl_class_registry[classname] = _MultipleClassMarker(
-                [cls, cast("Type[Any]", existing)]
+        if not isinstance(
+            decl_class_registry[classname], _MultipleClassMarker
+        ):
+            decl_class_registry[classname] = _MultipleClassMarker(
+                [cls, cast("Type[Any]", decl_class_registry[classname])]
             )
     else:
         decl_class_registry[classname] = cls
@@ -317,7 +317,7 @@ class _ModuleMarker(_ClsRegistryToken):
                 else:
                     raise
         else:
-            existing = self.contents[name] = _MultipleClassMarker(
+            self.contents[name] = _MultipleClassMarker(
                 [cls], on_remove=lambda: self._remove_item(name)
             )
 
--- a/lib/sqlalchemy/orm/decl_base.py
+++ b/lib/sqlalchemy/orm/decl_base.py
@@ -1277,8 +1277,6 @@ class _ClassScanMapperConfig(_MapperConfig):
                     or isinstance(attr_value, _MappedAttribute)
                 )
             )
-        else:
-            is_dataclass_field = False
 
         is_dataclass_field = False
         extracted = _extract_mapped_subtype(
--- a/lib/sqlalchemy/orm/identity.py
+++ b/lib/sqlalchemy/orm/identity.py
@@ -167,6 +167,7 @@ class _WeakInstanceDict(IdentityMap):
         else:
             if existing is state:
                 return None
+            assert existing is not None
             self._manage_removed_state(existing)
 
         self._dict[state.key] = state
--- a/lib/sqlalchemy/orm/identity.py
+++ b/lib/sqlalchemy/orm/identity.py
@@ -101,7 +101,7 @@ class IdentityMap:
         if state.modified:
             self._modified.add(state)
 
-    def _manage_removed_state(self, state: InstanceState[Any]) -> None:
+    def _manage_removed_state(self, state: InstanceState[Any] | None) -> None:
         del state._instance_dict
         if state.modified:
             self._modified.discard(state)
--- a/lib/sqlalchemy/orm/identity.py
+++ b/lib/sqlalchemy/orm/identity.py
@@ -101,7 +101,7 @@ class IdentityMap:
         if state.modified:
             self._modified.add(state)
 
-    def _manage_removed_state(self, state: InstanceState[Any] | None) -> None:
+    def _manage_removed_state(self, state: InstanceState[Any]) -> None:
         del state._instance_dict
         if state.modified:
             self._modified.discard(state)
--- a/lib/sqlalchemy/orm/identity.py
+++ b/lib/sqlalchemy/orm/identity.py
@@ -160,19 +160,14 @@ class _WeakInstanceDict(IdentityMap):
         self, state: InstanceState[Any]
     ) -> Optional[InstanceState[Any]]:
         assert state.key is not None
-        if state.key in self._dict:
-            try:
-                existing = existing_non_none = self._dict[state.key]
-            except KeyError:
-                # catch gc removed the key after we just checked for it
-                existing = None
-            else:
-                if existing_non_none is not state:
-                    self._manage_removed_state(existing_non_none)
-                else:
-                    return None
-        else:
+        try:
+            existing = self._dict[state.key]
+        except KeyError:
             existing = None
+        else:
+            if existing is state:
+                return None
+            self._manage_removed_state(existing)
 
         self._dict[state.key] = state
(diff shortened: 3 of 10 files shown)


lib/sqlalchemy/orm/clsregistry.py
```diff
@@ -69,11 +69,11 @@ def _add_class(
 
     """
     if classname in decl_class_registry:
-        # class already exists.
-        existing = decl_class_registry[classname]
-        if not isinstance(existing, _MultipleClassMarker):
-            existing = decl_class_registry[classname] = _MultipleClassMarker(
-                [cls, cast("Type[Any]", existing)]
+        if not isinstance(
```
CaselIT, line 72: the previous code was better in my opinion, since it avoided re-accessing the dict multiple times

lib/sqlalchemy/orm/decl_base.py
```diff
@@ -1277,8 +1277,6 @@ def _collect_annotation(
                     or isinstance(attr_value, _MappedAttribute)
                 )
             )
-        else:
-            is_dataclass_field = False
```
CaselIT, line 1281: this code looked wired.

CaselIT (commented):
looks ok, just prefer the previous code in one spot

lib/sqlalchemy/orm/clsregistry.py
ashm-dev: Hmm, and you're right, I didn't pay attention to this)

lib/sqlalchemy/orm/decl_base.py
ashm-dev: Return it?

CaselIT: no, it's fine as is, since that variable is set outside the if one line below https://github.com/sqlalchemy/sqlalchemy/pull/12537/files#diff-fb37e8bc18ebc4c14f6780e17648d92847ccbfd5fe96567b76d686f2d7ae210dL1283

lib/sqlalchemy/orm/clsregistry.py
```diff
@@ -69,11 +69,11 @@ def _add_class(
...
     if classname in decl_class_registry:
-        if not isinstance(
-            decl_class_registry[classname], _MultipleClassMarker
-        ):
-            decl_class_registry[classname] = _MultipleClassMarker(
-                [cls, cast("Type[Any]", decl_class_registry[classname])]
+        # class already exists.
+        existing = decl_class_registry[classname]
+        if not isinstance(existing, _MultipleClassMarker):
+            existing = decl_class_registry[classname] = _MultipleClassMarker(
```
CaselIT, line 75: the assignment to existing can be removed since it's not accessed anymore. maybe this was the spot indicated by the tool

ashm-dev: fixed it)

CaselIT (approved):

lib/sqlalchemy/orm/identity.py
```diff
@@ -160,19 +160,14 @@ def replace(
...
-                    return None
-        else:
+        try:
+            existing = self._dict[state.key]
+        except KeyError:
             existing = None
+        else:
+            if existing is state:
+                return None
+            self._manage_removed_state(existing)
```
CaselIT, line 170: ok I guess the old code used that `existing_non_none` variable to avoid having to change the signature of that function.

let's add an `assert existing is not None` before the call instead

ashm-dev: Added it!

lib/sqlalchemy/orm/identity.py
```diff
@@ -101,7 +101,7 @@ def _manage_incoming_state(self, state: InstanceState[Any]) -> None:
         if state.modified:
             self._modified.add(state)
 
-    def _manage_removed_state(self, state: InstanceState[Any]) -> None:
+    def _manage_removed_state(self, state: InstanceState[Any] | None) -> None:
```
CaselIT, line 104:
```suggestion
    def _manage_removed_state(self, state: InstanceState[Any]) -> None:
```

optional should no longer be needed here

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 4b851862598f5791827f9c745de9f5c90fca012c of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 4b851862598f5791827f9c745de9f5c90fca012c: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5870

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision ab53f8c3487e8cfb4d4a0235c27d8a5b8557d193 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset ab53f8c3487e8cfb4d4a0235c27d8a5b8557d193 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5870

lib/sqlalchemy/orm/identity.py
```diff
@@ -160,19 +160,15 @@ def replace(
...
-                # catch gc removed the key after we just checked for it
-                existing = None
-            else:
-                if existing_non_none is not state:
-                    self._manage_removed_state(existing_non_none)
-                else:
-                    return None
-        else:
+        try:
+            existing = self._dict[state.key]
```
sqla-tester, line 164: **Michael Bayer** (zzzeek) wrote:

this key is usually not going to be present.   so catching keyerror here even in modern python is 5x slower than using "key in dict"

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5870

sqla-tester (commented):

I know a lot of these complex key gets look "dumb" but these date back to the Python 2.3, 2.4, etc days when interpreters were much slower.

I did a scan on a modern Python 3.12 and while @caselit is correct that "key in dict"  / "dict.get" etc. are all essentially the same speed, the case where KeyError is thrown and caught is still around 5x slower:


```
def key_in_map(map, key):
    if key in map:
        return map[key]
    else:
        return None

def key_via_get(map, key):
    value = map.get(key)
    return value

def key_w_keyerror(map, key):
    try:
        value = map[key]
    except KeyError:
        return None
    else:
        return value


import timeit


print(f"key not present in map: {timeit.timeit("key_in_map({}, 'foo')", "from __main__ import key_in_map")}")
print(f"key not present via get: {timeit.timeit("key_via_get({}, 'foo')", "from __main__ import key_via_get")}")
print(f"key not present w keyerror: {timeit.timeit("key_w_keyerror({}, 'foo')", "from __main__ import key_w_keyerror")}")

print(f"key present in map: {timeit.timeit("key_in_map({'foo': 'bar'}, 'foo')", "from __main__ import key_in_map")}")
print(f"key present via get: {timeit.timeit("key_via_get({'foo': 'bar'}, 'foo')", "from __main__ import key_via_get")}")
print(f"key present w keyerror: {timeit.timeit("key_w_keyerror({'foo': 'bar'}, 'foo')", "from __main__ import key_w_keyerror")}")
```

output:

```
key not present in map: 0.042835595028009266
key not present via get: 0.049074911046773195
key not present w keyerror: 0.18940914497943595
key present in map: 0.07440703903557733
key present via get: 0.0777056619990617
key present w keyerror: 0.0765952980145812

```


so for the one in identity map replace I dont want to catch KeyError in the normal "key missing" case

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5870

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5870

lib/sqlalchemy/orm/identity.py
sqla-tester: **Michael Bayer** (zzzeek) wrote:

so it likely would need to be a "get()" here.   because the case where our "key in dict" would return True, then be False on the next go definitely did happen, that's a real race we need to be very careful about.   this code is pretty delicate and overall it's likely better not to change it at all.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5870

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5870

lib/sqlalchemy/orm/identity.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

probably using a single get it the best option since that's atomic. We also avoid the assert check

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5870

sqla-tester: **Federico Caselli** (CaselIT) wrote:

on second look this module uses a similar pattern in many places, so it make sense to simplify all of them with get, but it's likely better to do this in a separate change. will restore the old code

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5870

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5870

ashm-dev (commented):
get is much slower (about 10 times) than getting the key value in the try-except block.

At least use it .get is both simpler and clearer, but it is slow

zzzeek:
> get is much slower (about 10 times) than getting the key value in the try-except block.
> 
> At least use it .get is both simpler and clearer, but it is slow

This is important: whether or not you expect the key to be present in the typical case.  The typical case here is that we do not.   So catching KeyError in the usual case is much slower.

This is why we don't trust automated code tools to make decisions like this

lib/sqlalchemy/orm/identity.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

the simplification of identity is done in https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5874

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5870

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5870

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5870 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5872 has been **merged**. Congratulations! :)

PR #12558, opened 2025-04-27
typing: pg: type NamedType create/drops (fixes #12557)

### LINKED ISSUE #12557: dialects: postgresql: bind on `ENUM` types is partially unknown

- [x] No sqlalchemy stub packages is installed (both `sqlalchemy-stubs` and `sqlalchemy2-stubs` are not compatible with v2)

- [x] The api is not in a module listed in [#6810](https://github.com/sqlalchemy/sqlalchemy/issues/6810) so it should pass type checking

### Describe the typing issue

The bind is partially unknown. This is typed in `sqlalchemy.sql.schema` but other places (such as postgres) can use it too.

### To Reproduce

```python
import sqlalchemy as sa


enum = sa.create_engine("sqlite:///", echo=True)
with e.connect() as c:
    enum = postgresql.ENUM('EGGS', 'HAM', name='spam', create_type=False)
    enum.create(c) # <-- Error here
```

### Error

```
Type of "create" is partially unknown (reportUnknownMemberType)
 Type of "create" is "(bind: Unknown | None = None, checkfirst: bool = True) -> None"

Type of "drop" is partially unknown (reportUnknownMemberType)
 Type of "drop" is "(bind: Unknown | None = None, checkfirst: bool = True) -> None"
``` 

### Versions

- OS: Windows 10
- Python: 3.13.0
- SQLAlchemy: First noticed @ 2.0.38, but appears to exist at HEAD and has existed for a while before
- Type checker (eg: mypy 0.991, pyright 1.1.290, etc): pyright 1.1.398

### Additional context

Initially observed while seeing that Alembic migrations for Postgres enums were failing type checking

jkrejcha:
### Description

Type the `create` and `drop` functions for `NamedType`s

Also partially type the SchemaType create/drop functions more generally

One change to this is that the default parameter of `None` is removed. It doesn't work and will fail with a `AttributeError` at runtime since it immediately tries to access a property of `None` which doesn't exist.

fixes #12557

### Checklist
This pull request is:

- [X] A documentation / typographical / small typing error fix
- [X] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/named_types.py
+++ b/lib/sqlalchemy/dialects/postgresql/named_types.py
@@ -7,6 +7,7 @@
 # mypy: ignore-errors
 from __future__ import annotations
 
+from types import ModuleType
 from typing import Any
 from typing import Optional
 from typing import Type
@@ -25,6 +26,7 @@ from ...sql.ddl import InvokeCreateDDLBase
 from ...sql.ddl import InvokeDropDDLBase
 
 if TYPE_CHECKING:
+    from ...sql._typing import _CreateDropBind
     from ...sql._typing import _TypeEngineArgument
 
 
@@ -36,7 +38,7 @@ class NamedType(sqltypes.TypeEngine):
     DDLDropper: Type[NamedTypeDropper]
     create_type: bool
 
-    def create(self, bind, checkfirst=True, **kw):
+    def create(self, bind: _CreateDropBind, checkfirst: bool = True, **kw: Any) -> None:
         """Emit ``CREATE`` DDL for this type.
 
         :param bind: a connectable :class:`_engine.Engine`,
@@ -50,7 +52,7 @@ class NamedType(sqltypes.TypeEngine):
         """
         bind._run_ddl_visitor(self.DDLGenerator, self, checkfirst=checkfirst)
 
-    def drop(self, bind, checkfirst=True, **kw):
+    def drop(self, bind: _CreateDropBind, checkfirst: bool = True, **kw: Any) -> None:
         """Emit ``DROP`` DDL for this type.
 
         :param bind: a connectable :class:`_engine.Engine`,
@@ -63,7 +65,7 @@ class NamedType(sqltypes.TypeEngine):
         """
         bind._run_ddl_visitor(self.DDLDropper, self, checkfirst=checkfirst)
 
-    def _check_for_name_in_memos(self, checkfirst, kw):
+    def _check_for_name_in_memos(self, checkfirst: bool, **kw: Any) -> None:
         """Look in the 'ddl runner' for 'memos', then
         note our name in that collection.
 
@@ -87,7 +89,7 @@ class NamedType(sqltypes.TypeEngine):
         else:
             return False
 
-    def _on_table_create(self, target, bind, checkfirst=False, **kw):
+    def _on_table_create(self, target, bind: _CreateDropBind, checkfirst: bool = False, **kw: Any) -> None:
         if (
             checkfirst
             or (
@@ -97,7 +99,7 @@ class NamedType(sqltypes.TypeEngine):
         ) and not self._check_for_name_in_memos(checkfirst, kw):
             self.create(bind=bind, checkfirst=checkfirst)
 
-    def _on_table_drop(self, target, bind, checkfirst=False, **kw):
+    def _on_table_drop(self, target, bind: _CreateDropBind, checkfirst: bool = False, **kw: Any) -> None:
         if (
             not self.metadata
             and not kw.get("_is_metadata_operation", False)
@@ -105,11 +107,11 @@ class NamedType(sqltypes.TypeEngine):
         ):
             self.drop(bind=bind, checkfirst=checkfirst)
 
-    def _on_metadata_create(self, target, bind, checkfirst=False, **kw):
+    def _on_metadata_create(self, target, bind: _CreateDropBind, checkfirst: bool = False, **kw: Any) -> None:
         if not self._check_for_name_in_memos(checkfirst, kw):
             self.create(bind=bind, checkfirst=checkfirst)
 
-    def _on_metadata_drop(self, target, bind, checkfirst=False, **kw):
+    def _on_metadata_drop(self, target, bind: _CreateDropBind, checkfirst: bool = False, **kw: Any) -> None:
         if not self._check_for_name_in_memos(checkfirst, kw):
             self.drop(bind=bind, checkfirst=checkfirst)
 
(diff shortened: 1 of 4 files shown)


lib/sqlalchemy/dialects/postgresql/named_types.py
```diff
@@ -36,7 +38,7 @@ class NamedType(sqltypes.TypeEngine):
     DDLDropper: Type[NamedTypeDropper]
     create_type: bool
 
-    def create(self, bind, checkfirst=True, **kw):
+    def create(self, bind: _CreateDropBind, checkfirst: bool = True, **kw):
```
CaselIT, line 41: please complete the typing of the function:
- kw should be typed as any
- return is None

this applies also to the other functions

CaselIT (commented):
thanks for starting this, could you complete the typing so that of those functions so that we can consider them "done"?
The target argument is likely fine to type as any, since there are isinstance checks on it in any case.

Thanks!

lib/sqlalchemy/dialects/postgresql/named_types.py
jkrejcha: Alright, should be done :)

CaselIT: thanks!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 75c8d81bfb68f45299a9448d45dda446532205d3 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 75c8d81bfb68f45299a9448d45dda446532205d3: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5881

lib/sqlalchemy/sql/_typing.py
```diff
@@ -72,7 +72,10 @@
     from .sqltypes import TableValueType
     from .sqltypes import TupleType
     from .type_api import TypeEngine
+    from ..engine import Connection
     from ..engine import Dialect
+    from ..engine import Engine
+    from ..engine.mock import MockConnection
     from ..util.typing import TypeGuard
```
sqla-tester, line 80: **Federico Caselli** (CaselIT) wrote:

As an alternative we may create a protocol so that we can avoid this import (that's for typing only, but still not great)

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5881

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5881

lib/sqlalchemy/sql/_typing.py
```diff
@@ -72,7 +72,10 @@
     from .sqltypes import TableValueType
     from .sqltypes import TupleType
     from .type_api import TypeEngine
+    from ..engine import Connection
     from ..engine import Dialect
+    from ..engine import Engine
+    from ..engine.mock import MockConnection
```
sqla-tester, line 78: **Michael Bayer** (zzzeek) wrote:

these can't be here, we cannot have anything from sqlalchemy.dialects.* in sql/.   This implies that third party dialects have no path to being typed.   So all such problems must be solved as though sqlalchemy.dialects are not even locally present.   so remove these imports

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5881

sqla-tester (commented):

we don't need _DDLVisitor or _DDLItem, we should use the existing InvokeDropDDLBase and SchemaEventTarget base classes.   Use these and substitute SchemaEventTarget into ddl.py where it currently refers to SchemaItem.   We can clean up and clarify the names as a separate step.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5881

* lib/sqlalchemy/sql/_typing.py (line 318): I dont think there is any reason to add `_DDLVisitor`, as a union that explicitly names its members, when we can use the existing `InvokeDropDDLBase` class that is already a common base to all four of these classes, and does not need to name its subclasses since OOP handles this for us.   So I would remove this
* lib/sqlalchemy/sql/_typing.py (line 324): same here; `NamedType` is actually a mixin, the actual `NamedType` classes are `ENUM` and `DOMAIN` which are both subclasses of `SchemaEventTarget`, which is the common base of `SchemaItem` as well:


```

SchemaEventTarget 
      ----> SchemaItem

      ----> SchemaType
            ----> Enum
                  ----> postgresql.ENUM(NamedType, Enum)
                        
            ----> postgresql.DOMAIN(NamedType, SchemaType)

```



there is already a hierarchy to use here based on `SchemaEventTarget`; so remove both `_DDLItem` and `_DDLVisitor`.

We can make nicer names for `InvokeDropDDLBase` and `SchemaEventTarget` as a separate step.

lib/sqlalchemy/sql/_typing.py
sqla-tester: **Michael Bayer** (zzzeek) wrote:

protocols are great but we dont need python tricks here, we can use normal OOP concepts because these are classes

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5881

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5881

* lib/sqlalchemy/sql/_typing.py (line 318): main issue is the init that's different, but I can fix that
* lib/sqlalchemy/sql/_typing.py (line 324): yes, but named type is not a SchemaEventTarget, so the call with self will raise. it's expected since it's a mixin, but that's also something not 100% clean

lib/sqlalchemy/sql/_typing.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5881

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5881

* lib/sqlalchemy/sql/_typing.py (line 318): Done
* lib/sqlalchemy/sql/_typing.py (line 324): in the end I just typed it as visitable, since that's all it's needed

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5881 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/5882 has been **merged**. Congratulations! :)

PR #12612, opened 2025-05-20
Add missing requires in the tests for older postgresql version

dlax:
Follow up commit 39bb17442ce6ac9a3dde5e2b72376b77ffce5e28.

Issue #12600, opened 2025-05-16
Reflection regression in postgresql


CaselIT:
Apologies if this is a different issue, but I'm also seeing a regression with postgres after updating from 2.0.40 to 2.0.41. Something relating to how tables are reflected seems to have broken.

I'm using psycopg as the driver. Here's the relevant error and traceback:

```
sqlalchemy.exc.InternalError: (psycopg.errors.InternalError_) failed to find conversion function from unknown to text
[SQL: SELECT attr.conrelid, array_agg(CAST(attr.attname AS TEXT) ORDER BY attr.ord) AS cols, attr.conname, min(attr.description) AS description, min(attr.indnkeyatts) AS indnkeyatts, bool_and(attr.indnullsnotdistinct) AS indnullsnotdistinct 
FROM (SELECT con.conrelid AS conrelid, con.conname AS conname, con.description AS description, con.ord AS ord, con.indnkeyatts AS indnkeyatts, con.indnullsnotdistinct AS indnullsnotdistinct, pg_catalog.pg_attribute.attname AS attname 
FROM pg_catalog.pg_attribute JOIN (SELECT pg_catalog.pg_constraint.conrelid AS conrelid, pg_catalog.pg_constraint.conname AS conname, unnest(pg_catalog.pg_index.indkey) AS attnum, generate_subscripts(pg_catalog.pg_index.indkey, %(generate_subscripts_1)s::INTEGER) AS ord, NULL AS indnkeyatts, false AS indnullsnotdistinct, pg_catalog.pg_description.description AS description 
FROM pg_catalog.pg_constraint JOIN pg_catalog.pg_index ON pg_catalog.pg_constraint.conindid = pg_catalog.pg_index.indexrelid LEFT OUTER JOIN pg_catalog.pg_description ON pg_catalog.pg_description.objoid = pg_catalog.pg_constraint.oid 
WHERE pg_catalog.pg_constraint.contype = %(contype)s::VARCHAR AND pg_catalog.pg_constraint.conrelid IN (%(oids_1)s)) AS con ON pg_catalog.pg_attribute.attnum = con.attnum AND pg_catalog.pg_attribute.attrelid = con.conrelid 
WHERE con.conrelid IN (%(oids_1)s)) AS attr GROUP BY attr.conrelid, attr.conname ORDER BY attr.conrelid, attr.conname]
[parameters: {'generate_subscripts_1': 1, 'contype': 'p', 'oids_1': 1067720}]
[...]

PR #12613, opened 2025-05-20
Skip test using MACADDR8 for PostgreSQL < 10

dlax:


dlax:
(I'm running the test suite with an old Postgres following the work on #12600 and thus submitting fixes for those missing "skips".)

CaselIT:
can we consolidate these in #12612?

dlax:
> can we consolidate these in #12612?

Done

PR #12642, opened 2025-06-03
Add create_type parameter to sqltypes.SchemaType

### LINKED ISSUE #10604: add "create_type" to sqltypes.Enum ?

ovangle:
Adds support for `create_type` to `sqlalchemy.sql.sqltypes.SchemaType`.

Fixes #10604 

### Description

I don't expect this PR to be accepted -- the changes I've made are far more extensive than would be strictly required to actually implement the issue as described, but  I believe them to be mostly positive and enhance readability and maintainabilty of the code surrounding DDL generation. 

Firstly, I refactored the types so that the `SchemaType` mixin is now a fully fledged `SchemaItem`. Secondly, I rewrote the DDL  generation so that `SchemaType`s DDL is generated by `SchemaGenerator` and `SchemaDropper` in a manner consistent with other schema items, rather than relying on table and metadata DDL events. 

The implementation is still a little bit of a WIP, I expect to be working on this for another week yet. There are a lot of tests to write and documentation to add, but I wanted to advertise the progress I've already made and that I am actively working on this issue and to hopefully gain in-principle approval for the changes I'm proposing. The implementation _should_ at least be the entire *existing* test suite (excluding mypy and lint), which, considering what has been reworked was an important milestone.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -1594,6 +1594,7 @@ from ... import select
 from ... import sql
 from ... import util
 from ...engine import characteristics
+from ...engine import Connection
 from ...engine import default
 from ...engine import interfaces
 from ...engine import ObjectKind
@@ -1604,6 +1605,7 @@ from ...engine.reflection import ReflectionDefaults
 from ...sql import bindparam
 from ...sql import coercions
 from ...sql import compiler
+from ...sql import ddl
 from ...sql import elements
 from ...sql import expression
 from ...sql import roles
@@ -2366,7 +2368,7 @@ class PGDDLCompiler(compiler.DDLCompiler):
             constraint.ondelete, self._fk_ondelete_pattern
         )
 
-    def visit_create_enum_type(self, create, **kw):
+    def visit_create_enum(self, create, **kw):
         type_ = create.element
 
         return "CREATE TYPE %s AS ENUM (%s)" % (
@@ -2377,7 +2379,7 @@ class PGDDLCompiler(compiler.DDLCompiler):
             ),
         )
 
-    def visit_drop_enum_type(self, drop, **kw):
+    def visit_drop_enum(self, drop, **kw):
         type_ = drop.element
 
         return "DROP TYPE %s" % (self.preparer.format_type(type_))
@@ -2842,6 +2844,34 @@ class PGTypeCompiler(compiler.GenericTypeCompiler):
         return "JSONPATH"
 
 
+class PGDDLGenerator(ddl.SchemaGenerator):
+    def __init__(
+        self, connection, item: schema.SchemaItem, checkfirst=False, **kwargs: Any
+    ) -> None:
+        super().__init__(connection, item, **kwargs)
+        self.checkfirst = checkfirst
+
+    def visit_DOMAIN(self, domain: DOMAIN, create_ok: bool = False, **kw):
+        if not create_ok and not self._can_create_schema_type(domain, **kw):
+            return
+        with self.with_ddl_events(domain):
+            CreateDomainType(domain)._invoke_with(self.connection)
+
+
+class PGDDLDropper(ddl.SchemaDropper):
+    def __init__(
+        self, connection, item: schema.SchemaItem, checkfirst=False, **kwargs: Any
+    ) -> None:
+        super().__init__(connection, item, **kwargs)
+        self.checkfirst = checkfirst
+
+    def visit_DOMAIN(self, domain, drop_ok: bool = False, **kw):
+        if not drop_ok and not self._can_drop_schema_type(domain, **kw):
+            return
+        with self.with_ddl_events(domain):
+            DropDomainType(domain)._invoke_with(self.connection)
+
+
 class PGIdentifierPreparer(compiler.IdentifierPreparer):
     reserved_words = RESERVED_WORDS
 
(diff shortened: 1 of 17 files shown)


zzzeek:
OK, so I sort of see what's happening here, but this is the important part

what is the problem that hopes to be solved here?      

if readability alone, it's not clear this is helping, this looks a lot more complicated.  Especially that it removes postgresql.NamedType.   The "named type" concept is entirely a PostgreSQL thing right now, there's no other database I'm familiar with (PG variants excluded) that has this concept of CREATE TYPE.

The other reason SchemaType is around is to support types that also carry along a CHECK constraint with them, that is, the Boolean and base Enum datatypes.

That latter concept has proven to be mostly a disaster, particularly on the Alembic side, which is why the "create_constraint" parameter of Boolean/Enum now defaults to False.   Once we defaulted these to False, the user reported issues we we had with these features went to zero, and I'm pretty sure nobody is turning on these parameters.

this is all to say that SchemaType's movement is just as likely (if not moreso) to be **out** of SQLAlchemy entirely, rather than more *in* to SQLAlchemy.       If I'm looking to make a big change it could be to remove SchemaType entirely.  It's been the source of confusion and bugs for years.   

So I'm pretty skeptical about a big refactor making SchemaType more of a first class citizen without a really good reason.    IMO if we are looking for readability, removing SchemaType entirely for like 2.2 and just having PG stick to its NamedType thing would improve readability the most.

ovangle:
@zzzeek Fair enough. I'm just looking to do something with my pathetic, hopeless, jobless existence (and that volunteering with an open source project I liked was better than applying for yet another cynical code chop-shop).

My reading of the code as it existed when I came across this issue was that the "ability to contribute DDL statements for a type" was the primary purpose of `SchemaType` and that as such it should have been a `SchemaItem`. If this notion existed in core despite being largely postgres specific (which it did), then the generation of said DDL statements would be better handled consistently with other `SchemaItem`s, by contributing those statements via the `SchemaGenerator`. 

In addition, I felt that  `util.portable_instancemethod` was a hack which violates the principle of least surprise and should be nuked out of existence (although I wanted to keep my changes limited, so I refrained from removing it completely and it is still used in the `CheckConstraint` stuff). 

I could conceivably move most of this code into `NamedType` if you prefer.

zzzeek:
@ovangle OK, I'm not very good at big code reviews, so I will say this PR does a great job at understanding the schema DDL system, does a great job at making SchemaType more of a first class citizen, for sure.   It's done very well.     And yes portable_instancemethod is awful, that's some really old stuff and it's stuck there due to some obscure pickling use case we haven't dropped.

So let me clarify with my comment what I was getting at before I spewed too much negativity - we are looking here for a rationale for the SchemaType idea being better integrated, which would drive just how we'd want to refactor it for best results, if we are doing a big refactor.    And also, if we were doing that, besides getting rid of portable_instancemethod, there's a whole lot of other [pain](https://github.com/sqlalchemy/sqlalchemy/blob/742a3a6f9b320514e51e8f882143665a8edf31b9/lib/sqlalchemy/sql/type_api.py#L1466) that would be great to simplify though the "emulated" idea might be a different track.

When I think about this problem and how confusing the whole "we make a boolean by doing int + CHECK constraint" idea of SchemaType is (that was its original purpose), I think more about something like a separate DDL concept, like "TYPE" or "TYPEDDL" or something, that is then somehow integrated as a thing that can accompany *any* type, that is, instead of "SchemaType inherits from TypeEngine", there would be "TypeDDLThing is used by TypeEngine", that is, the inheritance part of this is a PITA and causes problems like [this](https://github.com/sqlalchemy/sqlalchemy/discussions/12618).

Put another way, this is a big PR but for what it seeks to do I think it would need to be *bigger* and have some upfront thinking for getting the most bang for the buck.    I like a lot that you are willing to tackle big refactors, so I don't want to be discouraging.

CaselIT:
@ovangle not sure if it works for you time wise, given Australia from the profile, but we usually have dev meetings at https://app.gitter.im/#/room/#sqlalchemy_devel:gitter.im on Thursdays at 15 utc time, if you want to participate.

ovangle:
@zzzeek 
I wasn't bothered by the negativity, I was leaning towards this not being accepted in the first place. It seemed like an uneccessarily large change considering the issue I'd originally looked at, but the implication of that issue was that "some of the machinery for actually creating the enum" should be migrated as well, which led me to "schema type seems the natural place for that, but it's not behaving consistently with other classes that contribute DDL statements".

But I wasn't interested in trying to argue justifications or authorial intent or for better integration. If you believed that there was "no problem to be solved here", then I'm not going to argue. I'm just looking for something to do atm and I've given up on trying to continue my career (since I know better than taking a job with any of the VC startups who actually invite me in for an interview). I spent most of today moving the code into `NamedType` (except for configuring the DDL generator/dropper on the dialect and removing the event listeners from `SchemaType`), which seems sensible, since "types as part of the DDL" is indeed postgres specific. MS SQL is the only dialect with a similar DDL construct, and it requires that you add C# assemblies to SQL Server. 

Regarding inheritance vs composition, yeah I think a marker class for "can contribute DDL" is useful and should be what schematype is, but the hook for what happens in a particular dialect should happen in the dialect, I can see how assuming emulation and then overriding that emulation would cause problems. 

@CaseIT
I'm willing to take this more seriously, but that is 1am here. I've joined the room, but I can't guarantee that I'll be an active participant.

CaselIT:
> I'm willing to take this more seriously, but that is 1am here. I've joined the room, but I can't guarantee that I'll be an active participant.

if you are in the chat feel free to write also before 15, we usually are there also before that time

PR #12649, opened 2025-06-07
Fix missing data type in Article.writer_id mapping example

krave1986:
The code example in the "Overlapping Foreign Keys" section contains a bug that causes a runtime error when using modern SQLAlchemy versions.

```
writer_id = mapped_column()  # Missing data type
```


CaselIT:
thanks!

PR #12510, opened 2025-04-08
Add support for reflection of collation in types on PostgreSQL

dlax:
The 'collation' returned is `None` if it matches the default collation for the type. On the other hand, if the column collation matches the one of the database but is explicitly set at column creation, the value is reflected.

Related to #6511.

PR #12697, opened 2025-06-25
Support schema-qualified collation in types for PostgreSQL

### LINKED ISSUE #9693: Text type columns using schema qualified collations compile a quoted DDL statement on Postgres
### Describe the bug

If you use a `Text` type that includes a fully schema qualified collation (ie. `Text(collation="my_schema.my_collation")`), the DDL statement created when running `metadata.create_all()` is incorrect for Postgres.

The statement created presently is:
`TEXT COLLATE "my_schema.my_collation"`
Which is a reference to a collation called "my_schema.my_collation" in the default/public namespace.  

What we actually want is to reference the collation "my_collation" that resides in "my_schema" schema.  The correct statement to use the schema qualified collation instead needs to be unquoted:
`TEXT COLLATE my_schema.my_collation`

It seems, however, that the implementation in `sqlalchemy.sql.compiler.GenericTypeCompiler::_render_string_type` always builds a quoted collation name in the DDL statement, which isn't want we want in the case of schema qualified collations.

_No response_

### SQLAlchemy Version in Use

2.0.10

psycopg2

PostgreSQL 14.1

### Python Version

3.10

### Operating system

OSX

### To Reproduce

```python
from sqlalchemy import Column, DDL, event, Integer, Text, create_engine, MetaData

schema_name = "my_schema"
metadata = MetaData(schema=schema_name)
Base = declarative_base(metadata=metadata)


class MyClass(Base):
    __tablename__ = "my_table"
    column_with_schema_qualified_collation = Column(Text(collation="my_schema.my_collation"))


# This builds our schema-qualified collation when we run metadata.create_all
event.listen(
    metadata.tables["my_schema.my_table"],
    "before_create",
    DDL(
        """
        CREATE COLLATION my_schema.my_collation (
            provider = icu,
            locale = 'en-US-u-kn-true'
        )
    """
    ),
)

engine = create_engine("postgresql+psycopg2://localhost/<my_db>")
metadata.create_all(engine)
```

### Error

```
  File "/Users/user/repos/sqlalchemy_bug_repro/main.py", line 30, in <module>
    metadata.create_all(engine)
  File "/Users/user/repos/sqlalchemy_bug_repro/.venv/lib/python3.10/site-packages/sqlalchemy/sql/schema.py", line 5796, in create_all
    bind._run_ddl_visitor(
  File "/Users/user/repos/sqlalchemy_bug_repro/.venv/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 3239, in _run_ddl_visitor
    conn._run_ddl_visitor(visitorcallable, element, **kwargs)
  File "/Users/user/repos/sqlalchemy_bug_repro/.venv/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 2443, in _run_ddl_visitor
    visitorcallable(self.dialect, self, **kwargs).traverse_single(element)
  File "/Users/user/repos/sqlalchemy_bug_repro/.venv/lib/python3.10/site-packages/sqlalchemy/sql/visitors.py", line 670, in traverse_single
    return meth(obj, **kw)
  File "/Users/user/repos/sqlalchemy_bug_repro/.venv/lib/python3.10/site-packages/sqlalchemy/sql/ddl.py", line 922, in visit_metadata
    self.traverse_single(
  File "/Users/user/repos/sqlalchemy_bug_repro/.venv/lib/python3.10/site-packages/sqlalchemy/sql/visitors.py", line 670, in traverse_single
    return meth(obj, **kw)
  File "/Users/user/repos/sqlalchemy_bug_repro/.venv/lib/python3.10/site-packages/sqlalchemy/sql/ddl.py", line 961, in visit_table
    )._invoke_with(self.connection)
  File "/Users/user/repos/sqlalchemy_bug_repro/.venv/lib/python3.10/site-packages/sqlalchemy/sql/ddl.py", line 315, in _invoke_with
    return bind.execute(self)
  File "/Users/user/repos/sqlalchemy_bug_repro/.venv/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1413, in execute
    return meth(
  File "/Users/user/repos/sqlalchemy_bug_repro/.venv/lib/python3.10/site-packages/sqlalchemy/sql/ddl.py", line 181, in _execute_on_connection
    return connection._execute_ddl(
  File "/Users/user/repos/sqlalchemy_bug_repro/.venv/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1525, in _execute_ddl
    ret = self._execute_context(
  File "/Users/user/repos/sqlalchemy_bug_repro/.venv/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1841, in _execute_context
    return self._exec_single_context(
  File "/Users/user/repos/sqlalchemy_bug_repro/.venv/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1982, in _exec_single_context
    self._handle_dbapi_exception(
  File "/Users/user/repos/sqlalchemy_bug_repro/.venv/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 2339, in _handle_dbapi_exception
    raise sqlalchemy_exception.with_traceback(exc_info[2]) from e
  File "/Users/user/repos/sqlalchemy_bug_repro/.venv/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1963, in _exec_single_context
    self.dialect.do_execute(
  File "/Users/user/repos/sqlalchemy_bug_repro/.venv/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 918, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedObject) collation "my_schema.my_collation" for encoding "UTF8" does not exist
LINE 4:  column_with_schema_qualified_collation TEXT COLLATE "my_sch...
                                                     ^

[SQL: 
CREATE TABLE my_schema.my_table (
        id SERIAL NOT NULL, 
        column_with_schema_qualified_collation TEXT COLLATE "my_schema.my_collation", 
        PRIMARY KEY (id)
)

]
``` 

### Additional context

I was able to get around this issue by adding a new Text type that has a custom compiler when using postgresql.  The custom compiler is a re-implementation of `_render_string_type`, where we build the unquoted DDL when the collation is schema qualified.

The example below is the same as the reproduction code, just with the new custom type added.  No error when running the below:

```
from sqlalchemy.ext.compiler import compiles
from sqlalchemy.types import Text, TypeDecorator
from sqlalchemy import Column, DDL, event, Integer, Text, create_engine, MetaData


class SchemaQualifiedCollationText(TypeDecorator):
    """
    This gets around a limitation/bug in SQLAlchemy where schema-qualified collations applied to the Text type produce
    the wrong CREATE/ALTER statements in Postgres

    eg:
    Text(collation='my_custom_schema.my_custom_collation') produces statement:
        'COLLATE "my_custom_schema.my_custom_collation"'
    But the proper statement to use a schema-qualified collation in Postgres is actually:
        'COLLATE my_custom_schema.my_custom_collation'
    """

    impl = Text


@compiles(SchemaQualifiedCollationText, "postgresql")
def postgresql_text(type_, compiler, **kw):
    """
    This is a reimplementation of sqlalchemy.sql.compiler.GenericTypeCompiler::_render_string_type function
    When we're using postgresql, it'll build the correct unquoted collation name when supplying a schema-qualified
    collation (ie. my_custom_schema.my_custom_collation).
    """

    text = "TEXT"

    if type_.length:
        text += "(%d)" % type_.length
    if type_.collation:
        if len(type_.collation.split(".")) > 1:
            collation_str = type_.collation
        else:
            collation_str = '"%s"' % type_.collation
        text += " COLLATE %s" % collation_str
    return text

schema_name = "my_schema"
metadata = MetaData(schema=schema_name)
Base = declarative_base(metadata=metadata)


class MyClass(Base):
    __tablename__ = "my_table"
    column_with_schema_qualified_collation = Column(SchemaQualifiedCollationText(collation="my_schema.my_collation"))


# This builds our schema-qualified collation when we run metadata.create_all
event.listen(
    metadata.tables["my_schema.my_table"],
    "before_create",
    DDL(
        """
        CREATE COLLATION my_schema.my_collation (
            provider = icu,
            locale = 'en-US-u-kn-true'
        )
    """
    ),
)

engine = create_engine("postgresql+psycopg2://localhost/<my_db>")
metadata.create_all(engine)
```

dlax:
We add a `collation_schema` in types supporting `collation`, along with
`postgresql.DOMAIN`, allowing to create respective elements with a
schema-qualified collation name.

Reflection is also handled for both types and domains.

Fix #9693.

### DIFF
--- a/doc/build/changelog/unreleased_20/9693.rst
+++ b/doc/build/changelog/unreleased_20/9693.rst
@@ -1,0 +1,5 @@
+.. change::
+    :tags: usecase, postgresql
+    :tickets: 9693
+
+    Added support for schema-qualified collation names in types for PostgreSQL.
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -2566,7 +2566,14 @@ class PGDDLCompiler(compiler.DDLCompiler):
 
         options = []
         if domain.collation is not None:
-            options.append(f"COLLATE {self.preparer.quote(domain.collation)}")
+            collation = self.preparer.quote(domain.collation)
+            if domain.collation_schema is not None:
+                collation = (
+                    self.preparer.quote(domain.collation_schema)
+                    + "."
+                    + collation
+                )
+            options.append(f"COLLATE {collation}")
         if domain.default is not None:
             default = self.render_default_string(domain.default)
             options.append(f"DEFAULT {default}")
@@ -3118,6 +3125,8 @@ class ReflectedDomain(ReflectedNamedType):
     """
     collation: Optional[str]
     """The collation for the domain."""
+    collation_schema: Optional[str]
+    """The name of the schema in which the collation is defined."""
 
 
 class ReflectedEnum(ReflectedNamedType):
@@ -4119,7 +4128,7 @@ class PGDialect(default.DefaultDialect):
             else_=sql.null(),
         ).label("default")
 
-        # get the name of the collate when it's different from the default one
+        # get collation (name and schema), when different from the default one
         collate = sql.case(
             (
                 sql.and_(
@@ -4133,7 +4142,19 @@ class PGDialect(default.DefaultDialect):
                     .scalar_subquery()
                     != pg_catalog.pg_attribute.c.attcollation,
                 ),
-                select(pg_catalog.pg_collation.c.collname)
+                select(
+                    sql.func.json_build_object(
+                        "name",
+                        pg_catalog.pg_collation.c.collname,
+                        "schema",
+                        pg_catalog.pg_namespace.c.nspname,
+                    ),
+                )
+                .outerjoin(
+                    pg_catalog.pg_namespace,
+                    pg_catalog.pg_namespace.c.oid
+                    == pg_catalog.pg_collation.c.collnamespace,
+                )
                 .where(
                     pg_catalog.pg_collation.c.oid
                     == pg_catalog.pg_attribute.c.attcollation
@@ -4216,6 +4237,7 @@ class PGDialect(default.DefaultDialect):
         named_type_loader: _NamedTypeLoader,
         type_description: str,
         collation: Optional[str],
+        collation_schema: Optional[str],
     ) -> sqltypes.TypeEngine[Any]:
         """
         Attempts to reconstruct a column type defined in ischema_names based
@@ -4322,10 +4344,12 @@ class PGDialect(default.DefaultDialect):
                     named_type_loader,
                     type_description="DOMAIN '%s'" % domain["name"],
                     collation=domain["collation"],
+                    collation_schema=domain["collation_schema"],
                 )
                 args = (domain["name"], data_type)
 
                 kwargs["collation"] = domain["collation"]
+                kwargs["collation_schema"] = domain["collation_schema"]
                 kwargs["default"] = domain["default"]
                 kwargs["not_null"] = not domain["nullable"]
                 kwargs["create_type"] = False
@@ -4356,6 +4380,7 @@ class PGDialect(default.DefaultDialect):
 
         if collation is not None:
             kwargs["collation"] = collation
+            kwargs["collation_schema"] = collation_schema
 
         data_type = schema_type(*args, **kwargs)
         if array_dim >= 1:
@@ -4381,7 +4406,10 @@ class PGDialect(default.DefaultDialect):
                 row_dict["format_type"],
                 named_type_loader,
                 type_description="column '%s'" % row_dict["name"],
-                collation=collation,
+                collation=collation["name"] if collation is not None else None,
+                collation_schema=(
+                    collation["schema"] if collation is not None else None
+                ),
             )
 
             default = row_dict["default"]
@@ -5482,6 +5510,10 @@ class PGDialect(default.DefaultDialect):
 
     @lru_cache()
     def _domain_query(self, schema):
+        collation_namespace = pg_catalog.pg_namespace.alias(
+            "collation_namespace"
+        )
+
         con_sq = (
             select(
                 pg_catalog.pg_constraint.c.contypid,
@@ -5517,7 +5549,8 @@ class PGDialect(default.DefaultDialect):
                 pg_catalog.pg_namespace.c.nspname.label("schema"),
                 con_sq.c.condefs,
                 con_sq.c.connames,
-                pg_catalog.pg_collation.c.collname,
+                pg_catalog.pg_collation.c.collname.label("collation"),
+                collation_namespace.c.nspname.label("collation_schema"),
             )
             .join(
                 pg_catalog.pg_namespace,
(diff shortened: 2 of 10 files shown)


dlax:
notes:
- this is based on  #12510 (first 4 commits)
- ~~5th commit (related to reflection caching to avoid `JOIN`s) might be worth considering apart from this PR~~

dlax:
Rebased (and changed according to the rework of https://github.com/sqlalchemy/sqlalchemy/pull/12510).

CaselIT:
thanks, sorry for the delay, I'll come to it

zzzeek (changes requested):
hi -

I'm rethinking this change as this is way too much complexity dug into compiler.py for a feature that's postgresql specific.  

to keep things simple for this odd use case why dont we do just this patch:

```diff
diff --git a/lib/sqlalchemy/sql/compiler.py b/lib/sqlalchemy/sql/compiler.py
index 0cd866b96..dcaf960d0 100644
--- a/lib/sqlalchemy/sql/compiler.py
+++ b/lib/sqlalchemy/sql/compiler.py
@@ -7325,7 +7325,9 @@ class GenericTypeCompiler(TypeCompiler):
         if length:
             text += f"({length})"
         if collation:
-            text += f' COLLATE "{collation}"'
+            text += (
+                f" COLLATE {self.dialect.identifier_preparer.quote(collation)}"
+            )
         return text
 
     def visit_CHAR(self, type_: sqltypes.CHAR, **kw: Any) -> str:

```

then a collation can be schema qualified like this:

```
class MyClass(Base):
    __tablename__ = "my_table"
    column_with_schema_qualified_collation = Column(
        Text(collation=quoted_name("test_schema.my_collation", quote=False))
    )

```

we can document this and even have an alias for it like `sqlalchemy.dialects.postgresql.collation_identifier()` that creates that construct.


I understand this PR is 6 months old so if I dont hear from anyone in a few days I'll submit something for this

CaselIT:
works for me.

zzzeek:
> works for me.

there's a slight risk in that we are no longer unconditionally quoting the collation name in this case, it's based on identifier rules.   should be fine unless we want to say it's a 2.1 thing (id be happier if this were just 2.1)

CaselIT:
the issue is 2.5 years old, and it doesn't seem too common, so I think it's fine to target 2.1

dlax:
> I'm rethinking this change as this is way too much complexity dug into compiler.py for a feature that's postgresql specific.
> 
> to keep things simple for this odd use case why dont we do just this patch:

A significant part of complexity of this patch is to support reflection. But maybe it's not worth that complexity and we can go with the simple approach you suggest; fine be me.

zzzeek:
we'll go with #13418.  thanks

zzzeek:
change of plans.    however I need to understand what's going on in #12698 and why that is separate.

zzzeek:
we will adapt some of whats here for a new gerrit in #13447

PR #12698, opened 2025-06-25
Handle schema-qualified collation value in collate()

dlax:
based on #12697 (and #12510).

### DIFF
--- a/doc/build/changelog/unreleased_20/12698.rst
+++ b/doc/build/changelog/unreleased_20/12698.rst
@@ -1,0 +1,5 @@
+.. change::
+    :tags: usecase, postgresql
+    :tickets: 9693
+
+    Handle schema-qualified collation value passed to :func:`_sql.collate`.
--- a/lib/sqlalchemy/sql/sqltypes.py
+++ b/lib/sqlalchemy/sql/sqltypes.py
@@ -192,6 +192,7 @@ class String(Concatenable, TypeEngine[str]):
         self,
         length: Optional[int] = None,
         collation: Optional[str] = None,
+        collation_schema: Optional[str] = None,
     ):
         """
         Create a string-holding type.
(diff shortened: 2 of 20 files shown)


zzzeek:
OK but we went with #13418 for the string collations, adding all new "collation_schema" feels too heavy handed.    seeing if i can take this through with the new approach

lib/sqlalchemy/sql/sqltypes.py
```diff
@@ -215,21 +216,30 @@ def __init__(
...
+          datatypes should be used for a :class:`_schema.Column` that expects
+          to store non-ascii data. These datatypes will ensure that the
+          correct types are used on the database.
 
         """
 
         self.length = length
         self.collation = collation
+        if collation_schema is not None and collation is None:
+            raise TypeError(
```
zzzeek, line 234: sqlalchemy.ArgumenetError

lib/sqlalchemy/sql/sqltypes.py
```diff
@@ -215,21 +216,30 @@ def __init__(
             >>> print(select(cast("some string", String(collation="utf8"))))
             {printsql}SELECT CAST(:param_1 AS VARCHAR COLLATE utf8) AS anon_1
 
-          .. note::
+        :param collation_schema: Optional, the name of the schema in which the
+          collation is defined.
```
zzzeek, line 220: versionadded

doc/build/changelog/unreleased_20/12698.rst
```diff
@@ -0,0 +1,5 @@
+.. change::
```
zzzeek, line 1: this should be merged into one changelog file for expediency (can have multiple ..change:: sections in one file) , this is also 2.1 milestone for all of these.   12698 is not an issue number either

doc/build/changelog/unreleased_20/12698.rst
```diff
@@ -0,0 +1,5 @@
...
+    :tags: usecase, postgresql
+    :tickets: 9693
```
zzzeek, line 3: is this the right issue?   func.collate wasn't addressed in #9693 ?

lib/sqlalchemy/sql/sqltypes.py
```diff
@@ -192,6 +192,7 @@ def __init__(
         self,
         length: Optional[int] = None,
         collation: Optional[str] = None,
+        collation_schema: Optional[str] = None,
```
zzzeek, line 195: see im very confused about scope here, this is a type.   PR #12697 is "Support schema-qualified collation in types for PostgreSQL".  this is "Handle schema-qualified collation value in collate()".  e.g. just func.collate().   im very confused

zzzeek (commented):
im working up a new plan that will supersede this but we are looking at this general idea

zzzeek:
we will adapt some of whats here for a new gerrit in #13447

PR #12712, opened 2025-07-02
Fixes: #12711 Added sparse vector support in Oracle

### LINKED ISSUE #12711: Add support for sparse vector in oracle
### Describe the use case

A Sparse vector is a vector which has zero value for most of its dimensions.
`

        CREATE TABLE t_s (
	    id INTEGER NOT NULL, 
	    name VARCHAR2(20 CHAR), 
	    embedding VECTOR(10,FLOAT32,SPARSE), 
	    PRIMARY KEY (id)
        )

`

Oracle

### Example Use

`VECTOR(
                dim=3,
                storage_format=VectorStorageFormat.FLOAT32,
                vector_format=VectorFormat.SPARSE
                ),`

### Additional context

_No response_

suraj-ora-2020:
### Description

A Sparse vector is a vector which has zero value for most of its dimensions.
Sparse vectors are supported in Oracle Database 23.7 and later.

This PR extends vector support by adding functionality for sparse vectors in Oracle

Fixes #12711

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/oracle/base.py
+++ b/lib/sqlalchemy/dialects/oracle/base.py
@@ -737,8 +737,22 @@ VECTOR Datatype
 
 Oracle Database 23ai introduced a new VECTOR datatype for artificial intelligence
 and machine learning search operations. The VECTOR datatype is a homogeneous array
-of 8-bit signed integers, 8-bit unsigned integers (binary), 32-bit floating-point numbers,
-or 64-bit floating-point numbers.
+of 8-bit signed integers, 8-bit unsigned integers (binary), 32-bit floating-point
+numbers, or 64-bit floating-point numbers.
+
+A vector's storage type can be either DENSE or SPARSE. A dense vector contains
+meaningful values in most or all of its dimensions. In contrast, a sparse vector
+has non-zero values in only a few dimensions, with the majority being zero.
+
+Sparse vectors are represented by the total number of vector dimensions, an array
+of indices, and an array of values where each value’s location in the vector is
+indicated by the corresponding indices array position. All other vector values are
+treated as zero.
+
+The storage formats that can be used with sparse vectors are float32, float64, and
+int8. Note that the binary storage format cannot be used with sparse vectors.
+
+Sparse vectors are supported when you are using Oracle Database 23.7 or later.
 
 .. seealso::
 
@@ -746,17 +760,24 @@ or 64-bit floating-point numbers.
     <https://python-oracledb.readthedocs.io/en/latest/user_guide/vector_data_type.html>`_ - in the documentation
     for the :ref:`oracledb` driver.
 
-.. versionadded:: 2.0.41
+.. versionadded:: 2.0.41 - Added VECTOR datatype
 
 CREATE TABLE support for VECTOR
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-With the :class:`.VECTOR` datatype, you can specify the dimension for the data
-and the storage format. Valid values for storage format are enum values from
-:class:`.VectorStorageFormat`. To create a table that includes a
-:class:`.VECTOR` column::
+With the :class:`.VECTOR` datatype, you can specify the number of dimensions,
+the storage format, and the storage type for the data. Valid values for the
+storage format are enum members of :class:`.VectorStorageFormat`. Valid values
+for the storage type are enum members of :class:`.VectorStorageType`. If
+storage type is not specified, a DENSE vector is created by default.
 
-    from sqlalchemy.dialects.oracle import VECTOR, VectorStorageFormat
+To create a table that includes a :class:`.VECTOR` column::
+
+    from sqlalchemy.dialects.oracle import (
+        VECTOR,
+        VectorStorageFormat,
+        VectorStorageType,
+    )
 
     t = Table(
         "t1",
@@ -764,7 +785,11 @@ and the storage format. Valid values for storage format are enum values from
         Column("id", Integer, primary_key=True),
         Column(
             "embedding",
-            VECTOR(dim=3, storage_format=VectorStorageFormat.FLOAT32),
+            VECTOR(
+                dim=3,
+                storage_format=VectorStorageFormat.FLOAT32,
+                storage_type=VectorStorageType.SPARSE,
+            ),
         ),
         Column(...),
         ...,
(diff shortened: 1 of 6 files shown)


CaselIT:
Hi, can you fix the link error? after that I'll add it to gerrit to

thanks.

You can install pre-commit to run the linters

suraj-ora-2020:
I have fixed them. Thanks.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision d0525220f837a5a559033ebfb43b0205c5f5ea08 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change d0525220f837a5a559033ebfb43b0205c5f5ea08: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6023

lib/sqlalchemy/dialects/oracle/base.py
```diff
@@ -806,13 +838,28 @@
             {"id": 1, "embedding": [1, 2, 3]},
         )
 
+Data can be inserted into a sparse vector using a tuple consisting of the number of
+dimensions, an array of indices, and a corresponding array of values::
+
+    from sqlalchemy import insert, select
+
+    sparse_val = (10, [1, 2], array.array("d", [23.45, 221.22]))
```
sqla-tester: **Federico Caselli** (CaselIT) wrote:

we are adding python side objects to ease usage of more advantage types, like Range or BitString in postgresql. links https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#sqlalchemy.dialects.postgresql.Range and https://docs.sqlalchemy.org/en/21/dialects/postgresql.html#sqlalchemy.dialects.postgresql.BitString

Maybe we could provide a class for this? Or a named tuple? Something like

```py
class SparseValue(NamedTuple):
  num_dimensions: int
  indexes: Sequence[int]
  values: array.array[Any] | np.ndarray  # assuming this is supported by the driver
```

This could be used also as the return type of sparse array columns? 

Threat this as a suggestion, it's likely ok as is

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6023

sqla-tester (commented):

left a suggestion regarding an improvement on the typing of the sparse array.

Also a question considering typing etc? Wouldn't this make more sense as another type? So VECTOR could be typed as `VECTOR(TypeEngine[list[Any]])` and an ipotetical different class SparseVector could be typed as `SparseVector(TypeEngine[SparseValue])`?


again just a suggestion

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6023

lib/sqlalchemy/dialects/oracle/base.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

we could even just reuse the value defined in the driver? https://github.com/oracle/python-oracledb/blob/2325059859ca9b4a517b8bfae7fa3c4f5b536ab2/src/oracledb/sparse_vector.py#L41

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6023

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6023

suraj-ora-2020:
> 
> left a suggestion regarding an improvement on the typing of the sparse array.
> 
> Also a question considering typing etc? Wouldn't this make more sense as another type? So VECTOR could be typed as `VECTOR(TypeEngine[list[Any]])` and an ipotetical different class SparseVector could be typed as `SparseVector(TypeEngine[SparseValue])`?
> 
> again just a suggestion
> 
> View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6023

I feel it might be more consistent to keep **Vector** and **SparseVector** together, since **SparseVector** is essentially a **format** or **variant** of **Vector**, rather than a separate type. Treating it separately could potentially lead to redundancy or fragmentation. Would love to hear your thoughts on this approach.

lib/sqlalchemy/dialects/oracle/base.py
suraj-ora-2020: @CaselIT Including the SparseVector object in sqlalchemy.dialects.oracle seems reasonable to me.

Also importing the value directly from oracledb also feels like a sensible approach—I’d go ahead with that change. The only consideration is that, with this setup, users will need to explicitly `import oracledb` and then create the object using `oracledb.SparseVector()`.

CaselIT: We can probably alias that in the oracle dialect module.

@zzzeek would you be ok doing a try/expect at module lever to try and import that oracledb class and otherwise use a function that would error when invoked?
I know that's not how we usually do this, but it may make sense in this case?

CaselIT:
It's ok for me, it was just an idea if useful from a typing prospective. We can always add such "useful for typing" subclasses later on if we get requests for it

lib/sqlalchemy/dialects/oracle/base.py
zzzeek: MMMmmmm I'm a pretty strong -1 on having SQLAlchemy users need to import symbols from the oracledb DBAPI module directly.  as you noted, this is exactly the thing we've been trying to move away from in the few cases where it was happening.

I would much prefer if we made our own SparseVector object in sqlalchemy.dialects.oracle , which would likely look very similar to oracledb's.   But I think having our own object presented to the user is best.

CaselIT: the idea was to try and avoid the marshaling of sa-sparevectors into oracledb-sparevectors and vice versa.

but it's probably the easiest solution. As mentione the alternative could be something similer to the following somewhere in the oracledb module

```py
try:
  from oracledb import SparseVector
except ImportError:
  def SparseVector(*args, **kwargs):
    raise Exception('OracleDB is needed for this')
```

zzzeek: is this a value object returned per-row?  even then I'd be +1 marshall

CaselIT: yes, it's a value object returned per row.

suraj-ora-2020: @CaselIT Based on the points raised, I see a couple of possible approaches we could take. Before proceeding, I wanted to check with you on which approach you’d prefer I follow.

CaselIT: So the easiest thing it likely to just make a simple class in sqlalchemy and marshal to-from that in the type, like mike suggested.

suraj-ora-2020: I've uploaded the latest changes based on your suggestion. Could you please review and let me know if anything needs to be updated?

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 0be516b522b3c5085c35695ed6176d360599ab34 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 0be516b522b3c5085c35695ed6176d360599ab34 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6023

suraj-ora-2020:
Hi @CaselIT,  just following up on the PR.  Let me know if you need anything from my side to help with the review.

CaselIT:
sorry, just lack of time. Will try to look at this PR this week, sorry for the delay

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 5a4199de1e89785129ee6fce4c7e65570419a1c7 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 5a4199de1e89785129ee6fce4c7e65570419a1c7 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6023

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6023 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6069 has been **merged**. Congratulations! :)

PR #12726, opened 2025-07-10
Make __aiter__ in AsyncTupleReulst a sync function.

golovasteek:
### Description
According to PEP 525 __aiter__ should be a sync function.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**


CaselIT:
thanks, will merge later

zzzeek:
OK let's figure out what happened then?

~~1. can we look where this `__aiter__` came from in the first place?~~ doenst matter it's just TYPE_CHECKING
~~2. was it wrong from the start?~~ yes but it's only TYPE_CHECKING, the real `__aiter__`s are correct
~~3. has it never been useful?  can we add a test to see that it works as intended?~~

zzzeek:
this is just in TYPE_CHECKING, OK.      should be backported at least?

CaselIT:
> this is just in TYPE_CHECKING, OK. should be backported at least?

has been merged to 2.0

PR #12731, opened 2025-07-13
fix(typing): allow DeclarativeBase subclasses in with_for_update(of=...)

vedant713:
This commit updates the `_ForUpdateOfArgument` type alias used in the `with_for_update(of=...)` method to explicitly allow `Type[DeclarativeBase]` subclasses.

In SQLAlchemy 2.x, `DeclarativeBase` is the recommended way to define ORM models. While runtime behavior already accepts these classes as valid arguments for the `of` parameter, static type checkers like mypy and pyright currently report type errors when `DeclarativeBase` subclasses are passed.

This patch resolves issue #12730 by expanding `_ForUpdateOfArgument` to include `Type[DeclarativeBase]` (and sequences containing them), aligning type hints with actual valid runtime behavior.

Fixes #12730.

### DIFF
--- a/lib/sqlalchemy/sql/selectable.py
+++ b/lib/sqlalchemy/sql/selectable.py
@@ -113,6 +113,7 @@ and_ = BooleanClauseList.and_
 
 
 if TYPE_CHECKING:
+    from sqlalchemy.orm import DeclarativeBase
     from ._typing import _ColumnExpressionArgument
     from ._typing import _ColumnExpressionOrStrLabelArgument
     from ._typing import _FromClauseArgument
@@ -175,9 +176,10 @@ _ForUpdateOfArgument = Union[
     Union[
         "_ColumnExpressionArgument[Any]",
         "_FromClauseArgument",
+        Type[DeclarativeBase],
     ],
     # or sequence of single column elements
-    Sequence["_ColumnExpressionArgument[Any]"],
+     Sequence[Union["_ColumnExpressionArgument[Any]", Type[DeclarativeBase]]],
 ]
 
 


lib/sqlalchemy/sql/selectable.py
```diff
@@ -113,6 +113,7 @@
 
 
 if TYPE_CHECKING:
+    from sqlalchemy.orm import DeclarativeBase
```
zzzeek, line 116: this isn't needed (also we can't have ORM imports in core in any case, DeclarativeBase is too narrow as well)

lib/sqlalchemy/sql/selectable.py
```diff
@@ -175,9 +176,10 @@ def entity_namespace(self) -> _EntityNamespace: ...
     Union[
         "_ColumnExpressionArgument[Any]",
         "_FromClauseArgument",
+        Type[DeclarativeBase],
```
zzzeek, line 179: this is not needed, `_FromClauseArgument` accommodates mapped classes

lib/sqlalchemy/sql/selectable.py
```diff
@@ -175,9 +176,10 @@ def entity_namespace(self) -> _EntityNamespace: ...
...
     ],
     # or sequence of single column elements
-    Sequence["_ColumnExpressionArgument[Any]"],
+     Sequence[Union["_ColumnExpressionArgument[Any]", Type[DeclarativeBase]]],
```
zzzeek, line 182: this list needs `_FromClauseArgument` if the "of" can point to lists of tables, which will accommodate the ORM mapped classes as well

zzzeek (changes requested):
hey there-

thanks for this!   I've tested locally and this just needs `_FromClauseArgument`:

```diff
diff --git a/lib/sqlalchemy/sql/selectable.py b/lib/sqlalchemy/sql/selectable.py
index 349f18930..f1959db43 100644
--- a/lib/sqlalchemy/sql/selectable.py
+++ b/lib/sqlalchemy/sql/selectable.py
@@ -177,7 +177,7 @@ _ForUpdateOfArgument = Union[
         "_FromClauseArgument",
     ],
     # or sequence of single column elements
-    Sequence["_ColumnExpressionArgument[Any]"],
+    Sequence[Union["_ColumnExpressionArgument[Any]","_FromClauseArgument"]],
 ]

```

mypy passes your original test in that case.

if you can also add a test case right in test/typing/plain_files/sql/common_sql_element.py you'll see plenty of other cases for "of" pointing to ORM entities

zzzeek:
closing due to no activity

PR #12756, opened 2025-07-27
Add ability to specify numeric start num when compile stmt

uriyyo:
### Description

Discussion #12755

Here is brief summary.

We are working on a feature where we need to compile part of a raw SQL statement. Basically, we want to create a WHERE clause based on input data dynamically.

There is a small issue that SQLAlchemy always compiles query and starts counting the number of parameters from 1, and there is no way to adjust this value. You can post-process the query with a regex expression where we replace all numeric params, but it's rather a workaround.

I propose to add the ability to configure numeric_start_num.

I imagine code to look smth like this:

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/sql/compiler.py
+++ b/lib/sqlalchemy/sql/compiler.py
@@ -1386,7 +1386,7 @@ class SQLCompiler(Compiled):
          these values may be disabled which means SQL expressions may
          be rendered inline, RETURNING may not be rendered, etc.
 
-         :param numeric_start_num: the starting number for numeric
+        :param numeric_start_num: the starting number for numeric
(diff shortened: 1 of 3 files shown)


CaselIT:
Hi,

I don't think it's a good feature add, since it's used exclusively by the asyncpg driver. It's also really unclear what the use case would be.

Let's first continue in the discussions to see what the actual use case is for this.

uriyyo:
@CaselIT Yup, makes sense to me, sorry for being to fast with PR

CaselIT:
no problem.

PR #12757, opened 2025-07-28
There have been universal2 wheels for greenlet for a while now.

glyph:
### Description

Remove a stray mention of "Apple M1" from the documentation which is no longer accurate.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/doc/build/faq/installation.rst
+++ b/doc/build/faq/installation.rst
@@ -13,7 +13,7 @@ I'm getting an error about greenlet not being installed when I try to use asynci
 
 The ``greenlet`` dependency does not install by default for CPU architectures
 for which ``greenlet`` does not supply a `pre-built binary wheel <https://pypi.org/project/greenlet/#files>`_.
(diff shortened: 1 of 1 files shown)


CaselIT:
Hi, thanks for bringing this to our attention.

It's likely that a different change is required, since in 2.1 greenlet will be optional, and in 2.0 I believe it's still not installed by default on arm macs.

glyph:
> Hi, thanks for bringing this to our attention.
> 
> It's likely that a different change is required, since in 2.1 greenlet will be optional, and in 2.0 I believe it's still not installed by default on arm macs.

The removed text in the documentation is just incorrect here, though; there *is* a prebuilt greenlet binary wheel for "Apple M1". The stuff about needing to install `sqlalchemy[asyncio]` might still be needed but that's a separate issue, and why my change is extremely targeted :)

zzzeek:
i think this is fine to merge

PR #12765, opened 2025-07-29
Add list subtype to ARRAY

matthaigh27:
### Description

When using [sqlacodegen](https://github.com/agronholm/sqlacodegen), it is unable to correctly generate the type for postgres columns that use `ARRAY`, it just gets left as `list`. I'm hoping to have it correctly generate a type like `list[str]`

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/sql/sqltypes.py
+++ b/lib/sqlalchemy/sql/sqltypes.py
@@ -3199,7 +3199,7 @@ class ARRAY(
 
     @property
     def python_type(self) -> Type[Any]:
-        return list
+        return list[self.item_type.python_type]
 
     def compare_values(self, x: Any, y: Any) -> bool:
         return x == y  # type: ignore[no-any-return]


CaselIT:
Hi,

I don't think this works as proposed, but let's see what the pipeline says

lib/sqlalchemy/sql/sqltypes.py
```diff
@@ -3199,7 +3199,7 @@ def hashable(self) -> bool:  # type: ignore[override]
 
     @property
     def python_type(self) -> Type[Any]:
-        return list
+        return list[self.item_type.python_type]
```
zzzeek, line 3202: this was not the intent of python_type

zzzeek:
I'm a -1 on this.  this is not what `python_type` was intended for, it was intended for runtime types not pep-484 types, and bigger picture as we can see it's not really worth relying on for more complex tasks.   sqlacodegen should fix this issue on their end (which is in the short term add a specific rule for ARRAY, longer term not rely on `python_type` and use a more comprehensive lookup system suitable to their needs)

lib/sqlalchemy/sql/sqltypes.py
CaselIT: I don't remember if we do something similar in other places, but I may miss-remember

matthaigh27:
No worries, I’ll close this one out and head over to sqlacodegen. Thanks for looking through it anyways!

PR #12814, opened 2025-08-18
Fix typing for `hybrid_property.__set__` to properly validate setter values

macro1:
While iterating on some typing improvements, my colleague @seamuswn pointed out mypy wasn't catching when values with invalid types were set using a `hybrid_property` setter. I believe this is all that's needed to fix the typing.

### Description
Adjust `hybrid_property.__set__` to expect a value of the type that matches the generic's type variable.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/ext/hybrid.py
+++ b/lib/sqlalchemy/ext/hybrid.py
@@ -1454,7 +1454,7 @@ class hybrid_property(interfaces.InspectionAttrInfo, ORMDescriptor[_T]):
         else:
             return self.fget(instance)
 
-    def __set__(self, instance: object, value: Any) -> None:
+    def __set__(self, instance: object, value: _T) -> None:
         if self.fset is None:
             raise AttributeError("can't set attribute")
         self.fset(instance, value)
--- a/lib/sqlalchemy/ext/hybrid.py
+++ b/lib/sqlalchemy/ext/hybrid.py
@@ -1454,10 +1454,12 @@ class hybrid_property(interfaces.InspectionAttrInfo, ORMDescriptor[_T]):
         else:
             return self.fget(instance)
 
-    def __set__(self, instance: object, value: _T) -> None:
+    def __set__(
+        self, instance: object, value: Union[SQLCoreOperations[_T], _T]
+    ) -> None:
         if self.fset is None:
             raise AttributeError("can't set attribute")
-        self.fset(instance, value)
+        self.fset(instance, value)  # type: ignore[arg-type]
 
     def __delete__(self, instance: object) -> None:
         if self.fdel is None:


CaselIT:
Hi,

I think it would need to be also include the ability to set sql constructs, similar to `Mapped`?
https://github.com/sqlalchemy/sqlalchemy/blob/9bd8f205edae4d8fab3b7d188c2c809176edbaae/lib/sqlalchemy/orm/base.py#L822-L824

what do you thin @zzzeek ?

zzzeek:
> Hi,
> 
> I think it would need to be also include the ability to set sql constructs, similar to `Mapped`?
> 
> https://github.com/sqlalchemy/sqlalchemy/blob/9bd8f205edae4d8fab3b7d188c2c809176edbaae/lib/sqlalchemy/orm/base.py#L822-L824
> 
> what do you thin @zzzeek ?

not sure because `__get__` is only returning `_T`.   I'd need to see some examples.   The Any here might be intentional.

macro1:
>  ... The Any here might be intentional.
Could it be due to needing covariance?

At lease in my use cases, `fset` is defined to accept `_T` (when compared to `fget`), and `__set__()` will only handle input using `fset`, so it feels fairly safe to use `_T`. I'm not sure what the `SQLCoreOperations` situation might be.. We wouldn't see that while working with an instance where `fget` is returning values of type `_T` would we?

Example of what it might look like if instead of `Any` we only accept `_T` in `__set__()`:
```python
class MyModel(BaseModel):
    __tablename__ = "my_model"

    int_col: Mapped[int | None] = mapped_column()

    @hybrid_property
    def str_col(self) -> dict[str, int | None]:
        return {"val": self.int_col}

    @str_col.inplace.setter
    def _str_col_setter(self, value: dict[str, int | None]) -> None:
        self.int_col = value.get("val")


my_instance = MyModel()
reveal_type(my_instance.str_col)  # Revealed type is "builtins.str | None"

none_val: dict[str, None] = {}
my_instance.str_col = none_val  # Incompatible types in assignment

matching_val: dict[str, int | None] = {}
my_instance.str_col = matching_val

complex_val: dict[str, int | str] = {"val": 123, "foo": "bar"}
my_instance.str_col = complex_val  # Incompatible types in assignment
reveal_type(my_instance.str_col)  # Revealed type is "builtins.str | None"
```
In this case the code will work, but we have type errors all over. I'd say it's an issue with the typing of the hybrid property rather than an indication we should use `Any` though.. And the implementer could use to type as `Any` if they felt `Mapping` was inappropriate.

CaselIT:
> I'm not sure what the `SQLCoreOperations` situation might be..

well hybrid is very open ended, so you can do basically whatever in the set.

An example of `SQLCoreOperations` that's valid is something like this:

```py
from sqlalchemy.ext.hybrid import hybrid_property
from sqlalchemy.orm import DeclarativeBase, Mapped, mapped_column, Session
import sqlalchemy as sa

class Base(DeclarativeBase):
    pass

class MyModel(Base):
    __tablename__ = "my_model"

    int_col: Mapped[int | None] = mapped_column()

    @hybrid_property
    def some_col(self) -> int:
        return (self.int_col or 0) + 1

    @some_col.inplace.setter
    def _str_col_setter(self, value: int) -> None:
        self.int_col = value - 1

m = MyModel(id=42, int_col=1)
e = sa.create_engine("sqlite://", echo=True)

Base.metadata.create_all(e)
with Session(e) as s:
    s.add(m)
    s.flush()
    m.some_col = sa.select(sa.func.max(MyModel.id)).scalar_subquery()
    s.flush()
    print(m.some_col, m.int_col)
```

But of course you could find cases where any is appropriate, for example say we do a setter like this:
```
    @some_col.inplace.setter
    def _str_col_setter(self, value: Any) -> None:
        self.int_col = id(value)
```

So to recap: if we want to narrow from Any, I think the same type accepted by mapped are likely fine as type, it doesn't cover every case, but it should mostly be ok. What's your opinition mike? You prefer leaving Any?

zzzeek:
I guess we can put the _T there but this all feels weird becasue the callables being decorated have the types right there.  What does like `@property` do?   probably some magic hardcoded-to-property thing

CaselIT:
it seems that the set and get are not connected in property.

I believe we could do something similar, using two generics on hybrid, but without using default value for generics that are 3.12+ only, it would look quite ugly, so I think a simpler generic similar to the one used in mapped is best for now

macro1:
updated to match `Mapped`

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 8a17b26aea264acf70c52de324b8ccb92b469f2d of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 8a17b26aea264acf70c52de324b8ccb92b469f2d: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6106

macro1:
> I believe we could do something similar, using two generics on hybrid, but without using default value for generics that are 3.12+ only, it would look quite ugly, so I think a simpler generic similar to the one used in mapped is best for now

my 2c, i think this will make sense longer term. currently mypy allows narrowing of the bound type arg tied to the generic, so the type revealed on the getter is impacted by values set on the setter. maybe there are situations where that's useful, but due to how they're implemented with two functions and often underlying state of an entirely different type underneath.. probably not helpful in the general sense

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6106 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6107 has been **merged**. Congratulations! :)

PR #12816, opened 2025-08-18
Add arm64 wheel builds and tests

finnagin:
This would add a wheel build and run tests for arm64 windows in the wheel build workflow and tests workflow.

### Description

Resolves: #12815

This adds windows arm64 builds to `create-wheels.yml` using the `windows-11-arm` runner image and adds windows arm64 tests to `run-test.yml` skipping python versions not supported for arm64 windows in the setup-python action.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/.github/workflows/create-wheels.yaml
+++ b/.github/workflows/create-wheels.yaml
@@ -27,6 +27,7 @@ jobs:
           - compiled
         os:
           - "windows-2022"
+          - "windows-11-arm"
           # TODO: macos-14 uses arm macs (only python 3.10+) - make arm wheel on it
           - "macos-13"
           - "ubuntu-22.04"
@@ -44,6 +45,8 @@ jobs:
 
         exclude:
           - os: "windows-2022"
+            linux_archs: "aarch64"
+          - os: "windows-11-arm"
             linux_archs: "aarch64"
           - os: "macos-13"
             linux_archs: "aarch64"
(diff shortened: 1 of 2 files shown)


CaselIT:
we may need to update also the cibuildwheel config here https://github.com/sqlalchemy/sqlalchemy/blob/67f62aac5b49b6d048ca39019e5bd123d3c9cfb2/pyproject.toml#L204-L222

have you tried running it to confirm a successful log? you should be able to run everything in your fork (commenting the push to pypi to avoid it failing)

finnagin:
> have you tried running it to confirm a successful log?

[I did](https://github.com/finnagin/sqlalchemy/actions/runs/17035572167), though when I ran it I didn't think to skip the pypi upload step so it failed there, but clicking through the tests that were run by cibuildwheel on each of the windows arm64 builds they look like they each passed the tests it ran. Also, the test workflow did [all pass as well.](https://github.com/finnagin/sqlalchemy/actions/runs/17048358848)

>we may need to update also the cibuildwheel config here 

I think since the arm64 build is being run on an arm64 machine arm64 gets included for the default arch value (`"auto"`) and we don't need to add arm64 explicitly.

CaselIT (approved):
looks ok to me. I think the safest option would be to target 2.1 only

PR #12822, opened 2025-08-22
fix: Handle Enum naming conflict with built-in types #12761

### LINKED ISSUE #12761: Postgresql named that have the same name as an internal type are not rendered correctly in `create table`
### Describe the bug

It seems that postgresql allows named types (enum, domains) with the same name as internal types. To use them in a create type they need to be fully qualified with the schema name.

For example, in the following sql only `c` is of type enum
```sql
create type "text" as enum ('foo');

create table t (
  a text,
  b "text",
  c public."text"
)
```

Also check reflection, create type, drop type, etc

_No response_

### SQLAlchemy Version in Use

2.x

any pg

any

### Python Version

any

### Operating system

any

### To Reproduce

```python
see above
```

### Error

the type in create table is not correct, since it requires to be fully qualified

### Additional context

Regarding implementation, it's likely best to only render the full qualified name when there is a name collision.

kapildagur:
### Title

`fix: Handle ENUM and DOMAIN naming conflicts with PostgreSQL built-in types #12761`

### Description

This PR fixes a bug in PostgreSQL where user-defined `ENUM` and `DOMAIN` types with names matching PostgreSQL built-in types (e.g., `text`, `char`) were incorrectly resolved to the built-in type instead of the intended user-defined type.

### Root Cause

PostgreSQL prioritizes `pg_catalog` when resolving unqualified type names. This can shadow user-defined types in other schemas when their names collide with built-in types.

### Solution

* Centralized schema resolution for `ENUM` and `DOMAIN` in `PGIdentifierPreparer.format_type()`.
* The schema resolution logic:

  * Respects explicit schema on the type.
  * Applies schema translation maps when available.
  * Falls back to the dialect’s `default_schema_name` (if configured).
  * Raises `CompileError` if the type name collides with a PostgreSQL built-in type and no schema can be determined.

This avoids collisions with built-in types while maintaining expected behavior for explicitly namespaced and schema-inherited types.

### Test Plan

* New regression tests were added to cover both `ENUM` and `DOMAIN` under different schema configurations, including:

  * Compile-time failure when no schema or default schema is configured.
  * Explicit schema provided.
  * Schema inherited from table definition.
  * Schema inherited from metadata.
  * Schema translation mapping.
* Tests are placed in `test_types` and `test_compiler` and pass across PostgreSQL drivers (`asyncpg`, `pg8000`, `psycopg2`, `psycopg2cffi`).

### Logic Explanation

`PGIdentifierPreparer.format_type()` now performs schema resolution internally instead of requiring explicit `schema` arguments. This ensures consistent handling of:

* Explicit schema and schema translation maps.
* Fallback to dialect’s `default_schema_name`.
* Compile-time errors for unresolved collisions with built-in types.

This centralization avoids duplicate logic across compilers, respects DRY & SRP principles, and ensures correctness in all compilation paths.

Fixes: #12761

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -2394,7 +2394,7 @@ class PGDDLCompiler(compiler.DDLCompiler):
     def visit_drop_enum_type(self, drop, **kw):
         type_ = drop.element
 
-        return "DROP TYPE " f"{self.preparer.format_type(type_)}"
+        return "DROP TYPE %s" % (self.preparer.format_type(type_))
 
     def visit_create_domain_type(self, create, **kw):
         domain: DOMAIN = create.element
@@ -2419,15 +2419,14 @@ class PGDDLCompiler(compiler.DDLCompiler):
             options.append(f"CHECK ({check})")
 
         return (
-            "CREATE DOMAIN "
-            f"{self.preparer.format_type(domain)} AS "
+            f"CREATE DOMAIN {self.preparer.format_type(domain)} AS "
             f"{self.type_compiler.process(domain.data_type)} "
             f"{' '.join(options)}"
         )
 
     def visit_drop_domain_type(self, drop, **kw):
         domain = drop.element
-        return "DROP DOMAIN" f" {self.preparer.format_type(domain)}"
+        return f"DROP DOMAIN {self.preparer.format_type(domain)}"
 
     def visit_create_index(self, create, **kw):
         preparer = self.preparer
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -2873,7 +2873,8 @@ class PGIdentifierPreparer(compiler.IdentifierPreparer):
         schema (excluding "public" to preserve legacy behavior), and
         enforces explicit schema for builtin names when unresolved.
         """
-        schema = getattr(type_, "schema", None)
+
+        schema = self.schema_for_object(type_)
 
         # Apply schema translation first (if available)
         if getattr(self, "schema_translate_map", None):
@@ -2907,9 +2908,7 @@ class PGIdentifierPreparer(compiler.IdentifierPreparer):
         name = self.quote(type_.name)
 
         # prefer schema from object, fall back to override
-        effective_schema = self.schema_for_object(
-            type_
-        ) or self._get_schema_for_type(type_)
+        effective_schema = self._get_schema_for_type(type_)
 
         if (
             not self.omit_schema
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -2394,7 +2394,7 @@ class PGDDLCompiler(compiler.DDLCompiler):
     def visit_drop_enum_type(self, drop, **kw):
         type_ = drop.element
 
-        return "DROP TYPE %s" % (self.preparer.format_type(type_))
+        return "DROP TYPE " f"{self.preparer.format_type(type_)}"
 
     def visit_create_domain_type(self, create, **kw):
         domain: DOMAIN = create.element
@@ -2419,14 +2419,15 @@ class PGDDLCompiler(compiler.DDLCompiler):
             options.append(f"CHECK ({check})")
 
         return (
-            f"CREATE DOMAIN {self.preparer.format_type(domain)} AS "
+            "CREATE DOMAIN "
+            f"{self.preparer.format_type(domain)} AS "
             f"{self.type_compiler.process(domain.data_type)} "
             f"{' '.join(options)}"
         )
 
     def visit_drop_domain_type(self, drop, **kw):
         domain = drop.element
-        return f"DROP DOMAIN {self.preparer.format_type(domain)}"
+        return "DROP DOMAIN" f" {self.preparer.format_type(domain)}"
 
     def visit_create_index(self, create, **kw):
         preparer = self.preparer
@@ -2866,6 +2867,38 @@ class PGIdentifierPreparer(compiler.IdentifierPreparer):
             )
         return value
 
+    def _get_schema_for_type(self, type_):
+        """Resolve effective schema for ENUM/DOMAIN.
+
+        Applies schema translation maps, falls back to dialect default
+        schema (excluding "public" to preserve legacy behavior), and
+        enforces explicit schema for builtin names when unresolved.
+        """
+        schema = getattr(type_, "schema", None)
+
+        # Apply schema translation first (if available)
+        if getattr(self, "schema_translate_map", None):
+            schema = self.schema_translate_map.get(schema, None)
+
+        # Fall back to dialect default schema (omit "public")
+        if schema is None:
+            default_schema = getattr(self.dialect, "default_schema_name", None)
+            if default_schema != "public":
+                schema = default_schema
+
+        # Raise error for builtin names if schema still unresolved
+        if (
+            schema is None
+            and getattr(type_, "name", None) in self.dialect.ischema_names
+        ):
+            raise exc.CompileError(
+                f"{type_.__class__.__name__.upper()} with name "
+                f"'{type_.name}' requires an explicit schema when "
+                "no default_schema_name is configured"
+            )
+
+        return schema
+
     def format_type(self, type_, use_schema=True):
         if not type_.name:
             raise exc.CompileError(
@@ -2873,7 +2906,11 @@ class PGIdentifierPreparer(compiler.IdentifierPreparer):
             )
 
         name = self.quote(type_.name)
-        effective_schema = self.schema_for_object(type_)
+
+        # prefer schema from object, fall back to override
+        effective_schema = self.schema_for_object(
+            type_
+        ) or self._get_schema_for_type(type_)
 
         if (
             not self.omit_schema
--- a/test/dialect/postgresql/test_types.py
+++ b/test/dialect/postgresql/test_types.py
@@ -84,6 +84,7 @@ from sqlalchemy.orm import Session
 from sqlalchemy.sql import bindparam
 from sqlalchemy.sql import operators
 from sqlalchemy.sql import sqltypes
+from sqlalchemy.testing import config
 from sqlalchemy.testing import expect_raises
 from sqlalchemy.testing import expect_raises_message
 from sqlalchemy.testing import fixtures
@@ -1240,6 +1241,111 @@ class NamedTypeTest(
         assert "my_enum" not in [
             e["name"] for e in inspect(connection).get_enums()
         ]
+
+    def _setup_type(self, type_kind, name, **kw):
+        """Create ENUM or DOMAIN with given name and args."""
+        if type_kind.enum:
+            return ENUM("a", "b", "c", name=name, **kw)
+        else:
+            return DOMAIN(name=name, data_type=Integer, **kw)
+
+    def _method_for(self, type_kind):
+        """Return inspector method name for given type."""
+        return "get_enums" if type_kind.enum else "get_domains"
+
+    def _type_names_for_schema(self, connection, method_name, schema):
+        """Fetch type names from inspector for a schema."""
+        insp = inspect(connection)
+        method = getattr(insp, method_name)
+        return [t["name"] for t in method(schema=schema)]
+
+    @testing.variation("type_kind", ["enum", "domain"])
+    def test_builtin_name_conflict_compiletime_failure(
+        self, connection, type_kind
+    ):
+        """Case 1: Fails if builtin name used without schema/default."""
+        dialect = postgresql.dialect()
+        dialect.default_schema_name = None
+        builtin_name = "text"
+
+        t = self._setup_type(type_kind, builtin_name)
+
+        assert_raises_message(
+            exc.CompileError,
+            f"{t.__class__.__name__.upper()} with "
+            f"name '{builtin_name}' requires an explicit schema "
+            "when no default_schema_name is configured",
+            t.compile,
+            dialect=dialect,
+        )
+
+    @testing.variation("type_kind", ["enum", "domain"])
+    def test_builtin_name_conflict_explicit_schema(
+        self, connection, type_kind
+    ):
+        """Case 2: Works with explicit schema."""
+        explicit_schema = config.test_schema
+        builtin_name = "text"
+        t = self._setup_type(type_kind, builtin_name, schema=explicit_schema)
+
+        metadata = MetaData()
+        Table("t", metadata, Column("c", t))
+        metadata.create_all(connection)
+
+        eq_(
+            self._type_names_for_schema(
+                connection, self._method_for(type_kind), explicit_schema
+            ),
+            [builtin_name],
+        )
+        metadata.drop_all(connection)
+
+    @testing.variation("type_kind", ["enum", "domain"])
+    def test_builtin_name_conflict_inherit_schema_from_table(
+        self, connection, type_kind
+    ):
+        """Case 3: Inherits schema from table."""
+        explicit_schema = config.test_schema
+        inherited_name = (
+            "status_enum_inherit" if type_kind.enum else "age_domain_inherit"
+        )
+
+        t = self._setup_type(type_kind, inherited_name, inherit_schema=True)
+
+        metadata = MetaData()
+        Table("t", metadata, Column("c", t), schema=explicit_schema)
+        metadata.create_all(connection)
+
+        eq_(
+            self._type_names_for_schema(
+                connection, self._method_for(type_kind), explicit_schema
+            ),
+            [inherited_name],
+        )
+        metadata.drop_all(connection)
+
+    @testing.variation("type_kind", ["enum", "domain"])
+    def test_builtin_name_conflict_inherit_schema_from_metadata(
+        self, connection, type_kind
+    ):
+        """Case 4: Inherits schema from metadata."""
+        explicit_schema = config.test_schema
+        meta_inherit_name = (
+            "priority_enum_meta" if type_kind.enum else "score_domain_meta"
+        )
+
+        metadata = MetaData(schema=explicit_schema)
+        t = self._setup_type(type_kind, meta_inherit_name)
+        Table("t", metadata, Column("c", t))
+        metadata.create_all(connection)
+
+        eq_(
+            self._type_names_for_schema(
+                connection, self._method_for(type_kind), explicit_schema
+            ),
+            [meta_inherit_name],
+        )
+        metadata.drop_all(connection)
 
 
 class DomainTest(
--- a/test/dialect/postgresql/test_compiler.py
+++ b/test/dialect/postgresql/test_compiler.py
@@ -68,6 +68,7 @@ from sqlalchemy.sql import operators
 from sqlalchemy.sql import table
 from sqlalchemy.sql import util as sql_util
 from sqlalchemy.sql.functions import GenericFunction
+from sqlalchemy.testing import config
 from sqlalchemy.testing import expect_raises
 from sqlalchemy.testing import expect_raises_message
 from sqlalchemy.testing import fixtures
@@ -356,6 +357,60 @@ class CompileTest(fixtures.TestBase, AssertsCompiledSQL):
             render_schema_translate=True,
         )
 
+    def test_enum_with_builtin_name_schema_is_qualified(self):
+        for sch in (config.test_schema, config.test_schema_2):
+            e = Enum("active", "inactive", name="text", schema=sch)
+            self.assert_compile(
+                postgresql.CreateEnumType(e),
+                f"CREATE TYPE {sch}.text AS ENUM ('active', 'inactive')",
+            )
+
+            t = Table("user_status", MetaData(), Column("status", e))
+            self.assert_compile(
+                schema.CreateTable(t),
+                f"CREATE TABLE user_status (status {sch}.text)",
+            )
+
+    def test_drop_enum_with_builtin_name_is_qualified(self):
+        for sch in (config.test_schema, config.test_schema_2):
+            e = Enum("a", "b", name="text", schema=sch)
+            self.assert_compile(
+                postgresql.DropEnumType(e),
+                f"DROP TYPE {sch}.text",
+            )
+
+    def test_cast_enum_with_builtin_name(self):
+        for sch in (config.test_schema, config.test_schema_2):
+            e = Enum("opt1", "opt2", name="text", schema=sch)
+            stmt = select(cast(column("foo"), e))
+            self.assert_compile(
+                stmt,
+                f"SELECT CAST(foo AS {sch}.text) AS foo",
+            )
+
+    def test_compileerror_enum_without_schema_builtin_name(self):
+        e = Enum("yes", "no", name="text")
+        dialect = postgresql.dialect()
+        dialect.default_schema_name = None
+
+        with expect_raises_message(
+            exc.CompileError,
+            "ENUM with name 'text' requires an explicit schema "
+            "when no default_schema_name is configured",
+        ):
+            e.compile(dialect=dialect)
+
+    def test_schema_translate_for_builtin_name_enum(self):
+        for sch in (config.test_schema, config.test_schema_2):
+            e = Enum("hot", "cold", name="text", schema=sch)
+            schema_translate_map = {sch: "translated_schema"}
+            self.assert_compile(
+                postgresql.CreateEnumType(e),
+                "CREATE TYPE translated_schema.text AS ENUM ('hot', 'cold')",
+                schema_translate_map=schema_translate_map,
+                render_schema_translate=True,
+            )
+
     def test_domain(self):
         self.assert_compile(
             postgresql.CreateDomainType(
@@ -428,6 +483,60 @@ class CompileTest(fixtures.TestBase, AssertsCompiledSQL):
             schema_translate_map=schema_translate_map,
             render_schema_translate=True,
         )
+
+    def test_domain_with_builtin_name_schema_is_qualified(self):
+        for sch in (config.test_schema, config.test_schema_2):
+            d = DOMAIN("text", Integer, schema=sch)
+            self.assert_compile(
+                postgresql.CreateDomainType(d),
+                f"CREATE DOMAIN {sch}.text AS INTEGER ",
+            )
+
+            t = Table("doc", MetaData(), Column("body", d))
+            self.assert_compile(
+                schema.CreateTable(t),
+                f"CREATE TABLE doc (body {sch}.text)",
+            )
+
+    def test_drop_domain_with_builtin_name_is_qualified(self):
+        for sch in (config.test_schema, config.test_schema_2):
+            d = DOMAIN("text", Integer, schema=sch)
+            self.assert_compile(
+                postgresql.DropDomainType(d),
+                f"DROP DOMAIN {sch}.text",
+            )
+
+    def test_cast_domain_with_builtin_name(self):
+        for sch in (config.test_schema, config.test_schema_2):
+            d = DOMAIN("text", Integer, schema=sch)
+            stmt = select(cast(column("bar"), d))
+            self.assert_compile(
+                stmt,
+                f"SELECT CAST(bar AS {sch}.text) AS bar",
+            )
+
+    def test_compileerror_domain_without_schema_builtin_name(self):
+        d = DOMAIN("text", Integer)
+        dialect = postgresql.dialect()
+        dialect.default_schema_name = None
+
+        with expect_raises_message(
+            exc.CompileError,
+            "DOMAIN with name 'text' requires an explicit schema "
+            "when no default_schema_name is configured",
+        ):
+            d.compile(dialect=dialect)
+
+    def test_schema_translate_for_builtin_name_domain(self):
+        for sch in (config.test_schema, config.test_schema_2):
+            d = DOMAIN("text", Integer, schema=sch)
+            schema_translate_map = {sch: "translated_schema"}
+            self.assert_compile(
+                postgresql.CreateDomainType(d),
+                "CREATE DOMAIN translated_schema.text AS INTEGER ",
+                schema_translate_map=schema_translate_map,
+                render_schema_translate=True,
+            )
 
     def test_create_table_with_schema_type_schema_translate(self):
         e1 = Enum("x", "y", "z", name="somename")


kapildagur:
Hii @CaselIT!

I just raised a PR with a test and fix for issue #12761.
Please have a look when you get a chance. Thanks!

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -1627,6 +1628,8 @@ def update():
 
 IDX_USING = re.compile(r"^(?:btree|hash|gist|gin|[\w_]+)$", re.I)
 
+PG_DEFAULT_SCHEMA: Final[str] = "public"
```
zzzeek: This should not be hardcoded.   A dialect gets this value from _get_default_schema_name() up front and it's available as `dialect.default_schema_name`.

if we want to support the use case of:

1. code is using a PGDialect() without it being associated with a connection (therefore it can't look up the default schema)
2. code is using Enum types with PG reserved words as names and does not indicate a "schema" argument
3. we want to generate string SQL without any connection

for now, if this is in tests, they should generate a PGDialect() object and apply `dialect.default_schema_name = "some_schema"` directly.    When we do this in tests, we should *not* use the word "public", because I want to make sure we are not hardcoding the word "public".

If the compilation of an ENUM / DOMAIN is called where we need to render the schema, there is no default_schema_name set up , there's no schema on the type, and there's no way to get a schema, it should raise `sqlalchemy.exc.CompileError` with a descriptive message that the schema on the type must be set.  We need tests for that also.

The PG search path is described In https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PATH ; so this not only indicates that "public" is not necessarily the default place to create objects, it also indicates "public" can be dropped entirely.   We dont want to get in the habit of hardcoding to something that is true 99% of the time, but when someone with the 1% case comes along, everything falls apart due to the hardcoding assumption, so we just dont hardcode it.

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2383,6 +2386,13 @@ def define_constraint_ondelete_cascade(self, constraint):
     def visit_create_enum_type(self, create, **kw):
         type_ = create.element
 
+        if (
+            hasattr(type_, "schema")
+            and not getattr(type_, "schema")
+            and type_.name in ischema_names
+        ):
+            setattr(type_, "schema", PG_DEFAULT_SCHEMA)
```
zzzeek: we cannot mutate the type_ in place here.   it should not be needed, the "schema" would be embedded into the string we return and we would need to send the "schema" argument along to self.preparer.format_type(), for those cases where the schema is not actually set on the type.

The actual default schema name comes from self.dialect.default_schema_name.  if it's None, raise CompileError.

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2775,6 +2785,13 @@ def visit_enum(self, type_, **kw):
     def visit_ENUM(self, type_, identifier_preparer=None, **kw):
         if identifier_preparer is None:
             identifier_preparer = self.dialect.identifier_preparer
+        if (
+            hasattr(type_, "schema")
+            and not getattr(type_, "schema")
+            and type_.name in ischema_names
+        ):
+            setattr(type_, "schema", PG_DEFAULT_SCHEMA)
```
zzzeek: same instructions as my previous comment for line 2394

test/dialect/postgresql/test_enum_name_confilt_with_internal_types.py
zzzeek, line 1: this suite should be in test/dialect/postgresql/test_types.py.   I think there's only one test here that we need and it should be part of ``test_types.py::NamedTypeTest``

There should also be compilation tests for ENUM/DOMAIN DDL .  See existing tests `test/dialect/postgresql/test_compiler.py::CompileTest::test_create_drop_enum`, `test/dialect/postgresql/test_compiler.py::CompileTest::test_domain`.  The reserved name tests can be added as additional test methods or as variations to the existing methods.

test/dialect/postgresql/test_enum_name_confilt_with_internal_types.py
```diff
@@ -0,0 +1,151 @@
...
+        column: str,
+    ) -> str:
+        cols = inspect(connection).get_columns(table)
+        return [c for c in cols if c["name"] == column][0][
+            "type"
+        ].__class__.__name__
+
+    def test_enum_with_builtin_name(
+        self,
+        connection: Connection,
```
zzzeek: also use the metadata fixture here so that objects are dropped automatically in case of failure.

these two tests (which i think should be combined into one) can also exercise an enum.drop() (along with table.drop()), then confirm type is gone, to make sure that part works.

test/dialect/postgresql/test_enum_name_confilt_with_internal_types.py
```diff
@@ -0,0 +1,151 @@
...
+        )
+
+        Table("calendar", metadata, Column("season", enum_type))
+
+        metadata.create_all(connection)
+
+        col_type_name = self._get_column_type_name(
+            connection, "calendar", "season"
+        )
+        assert col_type_name == "ENUM"
```
zzzeek: use `eq_(x, y)` for `assert x==y`

test/dialect/postgresql/test_enum_name_confilt_with_internal_types.py
```diff
@@ -0,0 +1,151 @@
...
+        connection: Connection,
+    ) -> None:
+        metadata = MetaData()
+        enum_type = Enum(
+            "Spring",
+            "Summer",
+            "Autumn",
+            "Winter",
+            name="text",
+            schema="public",
```
zzzeek: this should use one of the test schemas; use the name that comes from `testing.config.test_schema` which you'll see used elsewhere in tests.

test/dialect/postgresql/test_enum_name_confilt_with_internal_types.py
```diff
@@ -0,0 +1,151 @@
...
+        Table("palette", metadata, Column("primary_color", enum_type))
+
+        metadata.create_all(connection)
+
+        col_type_name = self._get_column_type_name(
+            connection, "palette", "primary_color"
+        )
+        assert col_type_name == "ENUM"
+
+    def test_enum_with_explicit_schema_and_builtin_name(
```
zzzeek: the two tests test_enum_with_builtin_name and test_enum_with_explicit_schema_and_builtin_name can be joined into one test using a variation:

```py
@testing.variation("use_explicit_schema", [True, False])
def test_enum_with_builtin_name(self, connection: Connection, metadata: MetaData, use_explicit_schema: testing.Variation):
    if use_explicit_schema:
        schema = testing.config.test_schema
    else:
        schema = None

    # ... etc
```

Also, we can add a second variation and have it create either an ENUM or a DOMAIN with the name/schema combination.

test/dialect/postgresql/test_enum_name_confilt_with_internal_types.py
```diff
@@ -0,0 +1,151 @@
...
+        Table("calendar", metadata, Column("season", enum_type))
+
+        metadata.create_all(connection)
+
+        col_type_name = self._get_column_type_name(
+            connection, "calendar", "season"
+        )
+        assert col_type_name == "ENUM"
+
+    def test_multiple_enums_with_builtin_names(
```
zzzeek: im not sure what this is testing that is new here, this should be covered in our existing enum tests in test_types

test/dialect/postgresql/test_enum_name_confilt_with_internal_types.py
```diff
@@ -0,0 +1,151 @@
...
+            Column("shape", enum1),
+            Column("animal", enum2),
+        )
+
+        metadata.create_all(connection)
+
+        cols = inspect(connection).get_columns("sketchbook")
+        assert all(col["type"].__class__.__name__ == "ENUM" for col in cols)
+
+    def test_enum_with_non_conflicting_name(
```
zzzeek: im not sure what this is testing that is new here, this should be covered in our existing enum tests in test_types

test/dialect/postgresql/test_enum_name_confilt_with_internal_types.py
```diff
@@ -0,0 +1,151 @@
...
+        Table("map", metadata, Column("direction", enum_type))
+
+        metadata.create_all(connection)
+
+        col_type_name = self._get_column_type_name(
+            connection, "map", "direction"
+        )
+        assert col_type_name == "ENUM"
+
+    def test_enum_reuse_across_tables(
```
zzzeek: im not sure what this is testing that is new here, this should be covered in our existing enum tests in test_types

test/dialect/postgresql/test_enum_name_confilt_with_internal_types.py
```diff
@@ -0,0 +1,151 @@
...
+        Table("notebook", metadata, Column("label", enum_type))
+
+        metadata.create_all(connection)
+
+        col_type_name = self._get_column_type_name(
+            connection, "notebook", "label"
+        )
+        assert col_type_name == "ENUM"
+
+    def test_enum_dropped_and_recreated(
```
zzzeek: we definitely test create/drop of enums already

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2775,6 +2785,13 @@ def visit_enum(self, type_, **kw):
...
+
         return identifier_preparer.format_type(type_)
 
     def visit_DOMAIN(self, type_, identifier_preparer=None, **kw):
```
zzzeek: IIUC we need similar logic for DOMAIN

zzzeek (changes requested):
Hi !

Thanks for taking this task on.   There's a lot going on with this one that is a lot for a newcomer to take in, so deep breath, there's a bunch of changes I'm requesting here.   The main things are that we need to ensure we dont hardcode a schema name that we cant be sure of, it's better to raise an error, and then after that there's a bunch of existing tests/conventions we need to follow for the tests.  that's mainly it, but I commented everything individually.   I hope you're able to continue work on this, thanks!

kapildagur:
Morning @zzzeek ,

Thanks for you feedback!!! I'll start working on these changes and notify you once made.

test/dialect/postgresql/test_enum_name_confilt_with_internal_types.py
kapildagur: Test Written inside `test_types.py::NamedTypeTest::test_builtin_name_conflict()`

test/dialect/postgresql/test_enum_name_confilt_with_internal_types.py
kapildagur: Removed Extra Test

kapildagur:
Hi @zzzeek,

Thanks again for providing such constructive feedback! I really appreciate how each of your comments highlighted areas for improvement... it was very helpful. I’ve implemented all the suggested changes. Please let me know if anything else needs adjustment; I’d be happy to address any further suggestions.

Thank you so much!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision d7390d768cfa47b0d9fbd2d4a6c38c4f2700a1ed of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change d7390d768cfa47b0d9fbd2d4a6c38c4f2700a1ed: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2383,6 +2383,19 @@ def define_constraint_ondelete_cascade(self, constraint):
...
+                raise exc.CompileError(
+                    f"ENUM with name '{type_.name}' requires an explicit "
+                    "schema when no default_schema_name is configured"
+                )
+
+            # If schema is not given and name conflicts with a builtin type,
+            # attach it to the dialect's default_schema_name.
+            # This mutation is required so that CREATE TYPE places it
+            # in the correct schema and subsequent reflection sees it there.
+            type_.schema = self.dialect.default_schema_name
```
sqla-tester: **Michael Bayer** (zzzeek) wrote:

we can't mutate the object inside the compilation.  the schema here needs to be passed through into format_type().

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2399,6 +2412,19 @@ def visit_drop_enum_type(self, drop, **kw):
...
+                raise exc.CompileError(
+                    f"DOMAIN with name '{domain.name}' requires an explicit "
+                    "schema when no default_schema_name is configured"
+                )
+
+            # If schema is not given and name conflicts with a builtin type,
+            # attach it to the dialect's default_schema_name.
+            # This mutation is required so that CREATE DOMAIN places it
+            # in the correct schema and subsequent reflection sees it there.
+            domain.schema = self.dialect.default_schema_name
```
sqla-tester: **Michael Bayer** (zzzeek) wrote:

same here, the schema needs to be passed into the format_type() method below as a separate argument

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2775,11 +2801,39 @@ def visit_enum(self, type_, **kw):
     def visit_ENUM(self, type_, identifier_preparer=None, **kw):
         if identifier_preparer is None:
             identifier_preparer = self.dialect.identifier_preparer
+
+        # Assign default schema for builtin names if schema not already set
+        if (
+            hasattr(type_, "schema")
```
sqla-tester: **Michael Bayer** (zzzeek) wrote:

this should just be a type_.schema call.  if the object does not have a .schema attribute at all, that is probably a bug

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2775,11 +2801,39 @@ def visit_enum(self, type_, **kw):
...
+            and not getattr(type_, "schema")
+            and type_.name in ischema_names
+        ):
+            if self.dialect.default_schema_name is None:
+                raise exc.CompileError(
+                    f"ENUM with name '{type_.name}' requires an explicit "
+                    "schema when no default_schema_name is configured"
+                )
+            type_.schema = self.dialect.default_schema_name
```
sqla-tester: **Michael Bayer** (zzzeek) wrote:

same here, no assignment

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2775,11 +2801,39 @@ def visit_enum(self, type_, **kw):
...
+
         return identifier_preparer.format_type(type_)
 
     def visit_DOMAIN(self, type_, identifier_preparer=None, **kw):
         if identifier_preparer is None:
             identifier_preparer = self.dialect.identifier_preparer
+
+        # Assign default schema for builtin names if schema not already set
+        if (
+            hasattr(type_, "schema")
```
sqla-tester: **Michael Bayer** (zzzeek) wrote:

this should just be a type_.schema call. if the object does not have a .schema attribute at all, that is probably a bug

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -3262,9 +3316,15 @@ def __init__(
...
         self._json_serializer = json_serializer
 
+        # Default schema name resolved during initialize(); None until then.
+        self.default_schema_name: Optional[str] = None
+
     def initialize(self, connection):
         super().initialize(connection)
 
+        # Resolve and cache the server current default schema (e.g. "public").
+        self.default_schema_name = self._get_default_schema_name(connection)
```
sqla-tester: **Michael Bayer** (zzzeek) wrote:

this is already done by the super().initialize() step (or if not, it's a bug).

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

lib/sqlalchemy/dialects/postgresql/named_types.py
```diff
@@ -497,6 +497,22 @@ def __init__(
...
         self.create_type = create_type
+
+        # To resolve ArgumentError("Ambiguously setting inherit_schema=True
+        # while ""also passing a non-None schema argument")
+        # we have to set inherit_schema flag with false if
+        # it got a schema implectly via default_dialect_schema
+
+        # Only force inherit_schema=False if schema is truly non-None
+        # i.e. we are explicitly binding to a schema to avoid conflicts
+        if kw.get("schema") is not None and (
```
sqla-tester: **Michael Bayer** (zzzeek) wrote:

remove all this stuff, we should not have to change how the object is constructed to be able to compile its string representation a certain way.

also there are changes coming down the line for these arguments that would conflict here in any case.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

test/dialect/postgresql/test_types.py
```diff
@@ -467,7 +468,7 @@ def test_schema_inheritance(
     ):
         """test #6373"""
 
-        metadata.schema = testing.config.test_schema
+        metadata.schema = testing.config.test_schema_2
         default_schema = testing.config.db.dialect.default_schema_name
```
sqla-tester: **Michael Bayer** (zzzeek) wrote:

how come?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

test/dialect/postgresql/test_types.py
```diff
@@ -1240,6 +1241,121 @@ def test_generic_w_some_other_variant(self, metadata, connection):
...
 
+    def _type_names_for_schema(self, connection, method_name, schema):
+        """Helper Function: Returns List of Names for given inspect method"""
+        return [
+            o["name"]
+            for o in getattr(inspect(connection), method_name)(schema=schema)
+        ]
+
+    @testing.variation("type_kind", ["enum", "domain"])
+    def test_builtin_name_conflict(self, connection, type_kind):
```
sqla-tester: **Michael Bayer** (zzzeek) wrote:

I know you had multiple tests here before that we removed, but in this case, the "case 1 / 2/ 3/ 4" should likely be separate test methods.  The `@testing.variation` can travel along with them and for now each can have that little bit of setup code at the top (there are ways to make it even more succinct but we can start there).

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

sqla-tester (commented):

this is good and the tests are very close. we still need to fix up how the compiler deals with the schema name missing, though, there cant be any mutation of the object inside of compilation.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

lib/sqlalchemy/dialects/postgresql/base.py
kapildagur: I got your point, Mutating the object inside compilation breaks SRP. I’ve refactored `_get_schema_for_type` and `format_type()` so the schema is now passed as an argument, avoiding object mutation. 
Thanks for the clarification.

lib/sqlalchemy/dialects/postgresql/base.py
kapildagur: Same here... object mutation removed, schema is now passed as an argument to `format_type()`

lib/sqlalchemy/dialects/postgresql/base.py
kapildagur: Understood. I initially added `hasattr()` defensively, but I’ll remove it and rely on the object having a .schema attribute as expected.

lib/sqlalchemy/dialects/postgresql/base.py
kapildagur: Same here... object mutation removed, schema is now passed as an argument to `format_type()`

lib/sqlalchemy/dialects/postgresql/base.py
kapildagur: same here... `hasattr()` is defensively

lib/sqlalchemy/dialects/postgresql/base.py
kapildagur: Thanks confirmed that super().initialize() already sets default_schema_name correctly.

lib/sqlalchemy/dialects/postgresql/named_types.py
kapildagur: Agree — removed unnecessary modifications during object construction. Compilation now relies only on schema passed through `format_type()`.

test/dialect/postgresql/test_types.py
kapildagur: I’ll split the previous single test into separate methods, each with its own setup, while retaining `@testing.variation`

kapildagur:
Hey @zzzeek ,

Thanks a lot for patiently reviewing my work and pointing out the needed changes. I made quite a few mistakes (probably more than I’d like to admit 😅), but I’ve fixed them and also tried to clean up a few other things while I was at it.

Really appreciate your guidance and patience!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 87cae1e7b1a8b7b8a38d48d3d877c3f3a65c0654 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 87cae1e7b1a8b7b8a38d48d3d877c3f3a65c0654 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2380,11 +2380,43 @@ def define_constraint_ondelete_cascade(self, constraint):
...
+        enforces explicit schema for builtin names when unresolved.
+        """
+        schema = getattr(type_, "schema", None)
+
+        # Apply schema translation first (if available)
+        if getattr(self, "schema_translate_map", None):
+            schema = self.schema_translate_map.get(schema, schema)
+
+        # Fall back to dialect default schema (omit "public")
+        if schema is None:
```
sqla-tester: **Michael Bayer** (zzzeek) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2393,11 +2425,15 @@ def visit_create_enum_type(self, create, **kw):
 
     def visit_drop_enum_type(self, drop, **kw):
```
sqla-tester: **Michael Bayer** (zzzeek) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2772,15 +2813,49 @@ def visit_enum(self, type_, **kw):
         else:
             return self.visit_ENUM(type_, **kw)
 
+    def _get_schema_for_type(self, type_):
```
sqla-tester: **Michael Bayer** (zzzeek) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2772,15 +2813,49 @@ def visit_enum(self, type_, **kw):
...
+        """Resolve effective schema for ENUM/DOMAIN.
+
+        Applies schema translation maps, falls back to dialect default
+        schema (excluding "public" to preserve legacy behavior), and
+        enforces explicit schema for builtin names when unresolved.
+        """
+        schema = getattr(type_, "schema", None)
+
+        # Apply schema translation first (if available)
+        if getattr(self, "schema_translate_map", None):
```
sqla-tester: **Michael Bayer** (zzzeek) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

lib/sqlalchemy/dialects/postgresql/base.py
sqla-tester: **Michael Bayer** (zzzeek) wrote:

OK it looks like this method is copied to SQLCompiler and TypeCompiler, we definitely can't duplicate all this code it has to be in one place.  but also I think this is wrong, it's looking for "schema_translate_map", that's part of the IdentifierPreparer.

So this method should be in one place and be PGIdentifierPreparer._get_schema_for_type().

Then PGIdentifierPreparer.format_type() doesn't need a new "schema" argument.

Looking at PGIdentifierPreparer.format_type(), in current code it's already calling upon "schema_for_object(type_)".   All of this schema logic should just be there, there's not even a change to SQLCompiler or TypeCompiler needed.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

test/dialect/postgresql/test_types.py
```diff
@@ -1241,6 +1242,111 @@ def test_generic_w_some_other_variant(self, metadata, connection):
...
         ]
 
+    def _setup_type(self, type_kind, name, **kw):
+        """Create ENUM or DOMAIN with given name and args."""
+        if type_kind.enum:
+            return ENUM("a", "b", "c", name=name, **kw)
+        else:
+            return DOMAIN(name=name, data_type=Integer, **kw)
+
+    def _method_for(self, type_kind):
```
sqla-tester, line 1252: **Michael Bayer** (zzzeek) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

sqla-tester (commented):

OK.  It looks like everything done in the compiler should be within PGIdentifierPreparer, it already is doing schema resolution, so the enhanced schema resolution stuff should just be there, and this would be a pretty minimal change.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

* lib/sqlalchemy/dialects/postgresql/base.py (line 2807): Done
* lib/sqlalchemy/dialects/postgresql/base.py (line 3326): Done
* lib/sqlalchemy/dialects/postgresql/named_types.py (line 508): Done

lib/sqlalchemy/dialects/postgresql/base.py
kapildagur: Thanks! I was defiantly sure that unit `_get_schema_for_type()` goes inside `PGIdentifierPreparer` because it violating `DRY` & `SRP`  Principle but just wanted you suggestion :)

kapildagur:
I don't know what's happening in pipeline but jobs failing I had tested locally `pre-commit` & `tox` working properly but Github Action Failing Jobs

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2866,14 +2866,50 @@ def _unquote_identifier(self, value):
...
         if not type_.name:
             raise exc.CompileError(
                 f"PostgreSQL {type_.__class__.__name__} type requires a name."
             )
 
         name = self.quote(type_.name)
-        effective_schema = self.schema_for_object(type_)
+
+        # prefer schema from object, fall back to override
+        effective_schema = self.schema_for_object(
```
zzzeek: doesn't _get_schema_for_type() encapsulate evertything that schema_for_object() does?  why call schema_for_object() first ?

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 63413881132a2e857017dce1f7aca42648699482 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 63413881132a2e857017dce1f7aca42648699482 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

lib/sqlalchemy/dialects/postgresql/base.py
kapildagur: Agree!! `_get_schema_for_type()` doing all stuffs I just forgot to wipe `schema_for_object()` routine call.
Thanks!!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 793cf27be28a37e8882d89ba6b79570c79f936ac of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 793cf27be28a37e8882d89ba6b79570c79f936ac added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

* lib/sqlalchemy/dialects/postgresql/base.py (line 2816): Done

zzzeek:
OK going to try to bring this one home, thanks!

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

* test/dialect/postgresql/test_types.py (line 472): Done

sqla-tester:

OK the scope of this change has shrunk quite a bit more to just the compileerror raise if no schema, and nothing else.   the issue here is broader than just "reserved names", it's about names that are already in visible schemas like pg_catalog, public, etc.   if folks are going to make enums /domains with conflicting names (super bad idea!), they need to put explicit schemas in even if that means "public".  This so that when they go to reflect like `inspector.get_enums()` they will know they have to say `schema="public"` there also; it changes the reflection query.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6110 has been **merged**. Congratulations! :)

kapildagur:
Thanks @zzzeek! 
I had seen recent patches (Patch 5-6) which directly raising `CompileError` if any conflict enum/domain unqualified name passed without schema passing Is something else planned for future???
This has been my best experience ever, and I would like to continue contributing while learning more about these issues.
Thanks for your guidance!!

zzzeek:
> Thanks @zzzeek! I had seen recent patches (Patch 5-6) which directly raising `CompileError` if any conflict enum/domain unqualified name passed without schema passing Is something else planned for future??? This has been my best experience ever, and I would like to continue contributing while learning more about these issues. Thanks for your guidance!!

I dont think so, because i misunderstood this issue originally.  usually when you want to use a name that matches a reserved word in postgresql, you quote it, like `"user"`.   So here I thought if you wanted to name an enum `text`, you'd also quote it, like `"text"` and that would differentiate it from the built in `text` type.

But that's not how PostgreSQL does types, the built in `text` type is a name that comes from a very default namespace `pg_catalog`, whereas our enum named `text` comes from a user-defined default namespace `public`, that does not take precedence over `pg_catalog`.   So given this architecture it makes sense that if someone wants to make a new type that uses the same name as one in the always-default `pg_catalog` namespace, they'd need to schema qualify it.  

The difference between these two methods became apparent once I tried to reflect back an ENUM with the name "text" - it does not come back unless we specify "public" as the schema name, because otherwise, we see the one in `pg_catalog`.   So since an object like `Enum(name="text")` will never come back from `inspect(connection).get_enums()` due to missing schemas, the schema has to be on both sides, `Emum(name="text", schema="public")` and `inspect(connection).get_enums(schema="public")`.

Beyond that I think it's an exceedingly bad idea to use the name of a PostgreSQL built in type as the name of a custom type.

kapildagur:
Thanks again for guiding me through this 🙏 — it’s been such a great learning experience!

Fun fact I picked up digging deeper: whether I try `CREATE TYPE text AS ENUM ...` or `CREATE TYPE "text" AS ENUM ...,` both end up represented inside the pg_type catalog. If I look at the catalog struct (src/include/pg_type.h):

[`L36`](https://github.com/postgres/postgres/blob/master/src%2Finclude%2Fcatalog%2Fpg_type.h#L36): the whole data structure definition for pg_type.

[`L41`](https://github.com/postgres/postgres/blob/master/src%2Finclude%2Fcatalog%2Fpg_type.h#L41): NameData typname (this is where "text" literally gets stored).

[`L44`](https://github.com/postgres/postgres/blob/master/src%2Finclude%2Fcatalog%2Fpg_type.h#L44): Oid typnamespace (this is the schema reference — pg_catalog for built-ins, public for user-space by default).


That’s why the precedence works the way it does:

If I don’t qualify, Postgres resolves from `pg_catalog` first, then `public` (via `search_path`).

If I schema-qualify (`public.text`), it resolves explicitly.

And if I omit schema on creation, it still lands in public, since that’s the default user-space in search_path.

I completely agree that naming a custom type as a built-in is a extremely bad idea and raising `CompileError` is first thought!

PR #12830, opened 2025-08-26
[Fixes #12829] Added support for Annotated Aliases to DeclarativeBase

### LINKED ISSUE #12829: Improve TypeAliasType handling in sql type resolution
### Describe the bug

DeclarativeBase doesn't detect type from Annotated if its created by alias using `type` syntax from Python 3.12. It requires to define entire type (like `PrimaryKey[UUID]`) in type_annotation_map instead of treating this just as `UUID`.

In the past I was using a function:
```py
def sql_primary_key[T](col_type: T) -> Annotated[T, MappedColumn[Any]]:
    return Annotated[col_type, mapped_column(primary_key=True)]

class User(DeclarativeBase):
    id: Mapped[sql_primary_key(UUID)]
```
but ruff noticed that function shouldn't be used in the type context. 

_No response_

### SQLAlchemy Version in Use

2.0.43

psycopg

PostgreSQL 17

### Python Version

3.13

### Operating system

MacOS

### To Reproduce

```python
from typing import Annotated
from uuid import UUID

type PrimaryKey[T] = Annotated[T, mapped_column(primary_key=True)]

class User(DeclarativeBase):
    id: Mapped[PrimaryKey[UUID]]
```

### Error
```
  File "/Users/sk/Projects/settlements/.venv/bin/alembic", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/Users/sk/Projects/settlements/.venv/lib/python3.13/site-packages/alembic/config.py", line 1016, in main
    CommandLine(prog=prog).main(argv=argv)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/Users/sk/Projects/settlements/.venv/lib/python3.13/site-packages/alembic/config.py", line 1006, in main
    self.run_cmd(cfg, options)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/Users/sk/Projects/settlements/.venv/lib/python3.13/site-packages/alembic/config.py", line 940, in run_cmd
    fn(
    ~~^
        config,
        ^^^^^^^
        *[getattr(options, k, None) for k in positional],
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        **{k: getattr(options, k, None) for k in kwarg},
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/sk/Projects/settlements/.venv/lib/python3.13/site-packages/alembic/command.py", line 309, in revision
    script_directory.run_env()
    ~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/sk/Projects/settlements/.venv/lib/python3.13/site-packages/alembic/script/base.py", line 549, in run_env
    util.load_python_file(self.dir, "env.py")
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/Users/sk/Projects/settlements/.venv/lib/python3.13/site-packages/alembic/util/pyfiles.py", line 116, in load_python_file
    module = load_module_py(module_id, path)
  File "/Users/sk/Projects/settlements/.venv/lib/python3.13/site-packages/alembic/util/pyfiles.py", line 136, in load_module_py
    spec.loader.exec_module(module)  # type: ignore
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "<frozen importlib._bootstrap_external>", line 1026, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/Users/sk/Projects/settlements/migrations/env.py", line 6, in <module>
    from app.config import settings
  File "/Users/sk/Projects/settlements/app/__init__.py", line 4, in <module>
    from app.users.models import User  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/sk/Projects/settlements/app/users/models.py", line 9, in <module>
    class User(BaseDbModel):
    ...<4 lines>...
        updated_at: Mapped[datetime_tz] = datetime.now(timezone.utc)
  File "/Users/sk/Projects/settlements/.venv/lib/python3.13/site-packages/sqlalchemy/orm/decl_api.py", line 838, in __init_subclass__
    _as_declarative(cls._sa_registry, cls, cls.__dict__)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/sk/Projects/settlements/.venv/lib/python3.13/site-packages/sqlalchemy/orm/decl_base.py", line 234, in _as_declarative
    return _MapperConfig.setup_mapping(registry, cls, dict_, None, {})
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/sk/Projects/settlements/.venv/lib/python3.13/site-packages/sqlalchemy/orm/decl_base.py", line 315, in setup_mapping
    return _ClassScanMapperConfig(
        registry, cls_, dict_, table, mapper_kw
    )
  File "/Users/sk/Projects/settlements/.venv/lib/python3.13/site-packages/sqlalchemy/orm/decl_base.py", line 552, in __init__
    self._extract_mappable_attributes()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/sk/Projects/settlements/.venv/lib/python3.13/site-packages/sqlalchemy/orm/decl_base.py", line 1548, in _extract_mappable_attributes
    value.declarative_scan(
    ~~~~~~~~~~~~~~~~~~~~~~^
        self,
        ^^^^^
    ...<7 lines>...
        is_dataclass,
        ^^^^^^^^^^^^^
    )
    ^
  File "/Users/sk/Projects/settlements/.venv/lib/python3.13/site-packages/sqlalchemy/orm/properties.py", line 741, in declarative_scan
    self._init_column_for_annotation(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        cls,
        ^^^^
    ...<4 lines>...
        originating_module,
        ^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/sk/Projects/settlements/.venv/lib/python3.13/site-packages/sqlalchemy/orm/properties.py", line 918, in _init_column_for_annotation
    raise orm_exc.MappedAnnotationError(
    ...<3 lines>...
    )
sqlalchemy.orm.exc.MappedAnnotationError: Could not locate SQLAlchemy Core type for Python type PrimaryKey[uuid.UUID] inside the 'id' attribute Mapped annotation
```
### Additional context

Got this error while creating migrations using alembic.

KaliszS:
### Description
After the fix, I've succesfully created an alembic migration. Minimal example:
```py
type PrimaryKey[T] = Annotated[T, mapped_column(primary_key=True)]

class User(BaseDbModel):
    id: Mapped[PrimaryKey[UUID]]
```
```py
# alembic migration

def upgrade() -> None:
    op.create_table('user',
    sa.Column('id', sa.Uuid(), nullable=False),
    sa.PrimaryKeyConstraint('id')
    )
```

What is hidden under the variables from this example:
```py
python_type == PrimaryKey[uuid.UUID]
python_type.__args__ == (<class 'uuid.UUID'>,)
python_type.__value__ == typing.Annotated[T, <sqlalchemy.orm.properties.MappedColumn object at 0x1045e2cb0>]
```

### Checklist
This pull request is:

- [x] A short code fix

### DIFF
--- a/lib/sqlalchemy/orm/decl_api.py
+++ b/lib/sqlalchemy/orm/decl_api.py
@@ -29,6 +29,8 @@ from typing import Type
 from typing import TYPE_CHECKING
 from typing import TypeVar
 from typing import Union
+from typing import Annotated
+from typing import get_origin, get_args
 import weakref
 
 from . import attributes
@@ -1249,6 +1251,10 @@ class registry:
         else:
             python_type_type = python_type  # type: ignore[assignment]
             search = ((python_type, python_type_type),)
+
+            if get_origin(python_type.__value__) is Annotated:
+                type_alias_annotated_value = get_args(python_type)[0]
+                search = ((type_alias_annotated_value, python_type_type),)
 
         for pt, flattened in search:
             # we search through full __mro__ for types.  however...


KaliszS:
I haven't had time to look through the tests yet. Could you please tell me which test module is most suitable to add tests for this type of task? There are a lot of them.

KaliszS:
Fixes #12829

CaselIT:
hi, the current behaviour is as designed since that's what's most aligned with runtime behaviour of type alias type, meaning:
```py
type one = int
type two = int
print(one == two)
```
```
False
```

KaliszS:
> hi, the current behaviour is as designed since that's what's most aligned with runtime behaviour of type alias type, meaning:
> 
> ```python
> type one = int
> type two = int
> print(one == two)
> ```
> 
> ```
> False
> ```

My change shouldn’t affect the behavior of most aliases. It will only impact one case -  Annotated types.
The goal is to enable mostly automated project templates with minimal effort from users who want to generate a project. Adding a requirement to manually adjust the type annotation map feels redundant, especially since it can be automated so easily.

I see this case as special because it improves the user experience with SQLAlchemy models by reducing syntax overhead to a minimum.

CaselIT:
I don't see how it doesn't. as designed type alias types create new types, that are no longer base types, so sqlalchemy cannot assume to know what `PrimaryKey[uuid.UUID]` should be mapped to, since it's an entirely new type.

We already tried automatically resolving type alias types in 2.0 and it's a mess, since the runtime fight you every step of the way

KaliszS:
It doesn't because it reacts only when `python_type.__value__` is bound to `Annotated` class. Do you have any examples on your mind that would break with those changes? It looks like `python_type.__args__` is storing those types it should be mapped to. Of course maybe I'm missing the point, but if so, I would love to learn more about the issue :)

CaselIT:
it seems very inconsistent if `type some_int = int` required to be in the type map but `type some_generic[T] = Annotated[T, 123]` did not. I don't see a difference there

KaliszS:
> it seems very inconsistent if `type some_int = int` required to be in the type map but `type some_generic[T] = Annotated[T]` did not. I don't see a difference there

I know. But type checkers reports such variable as undefined, so it might be confusing for users with Python 3.12+. Its also seems to be a part of the language new design.

CaselIT:
Sorry, I don't understand what type checkers report as undefined. could you clarify?

KaliszS:
> it seems very inconsistent if `type some_int = int` required to be in the type map but `type some_generic[T] = Annotated[T, 123]` did not. I don't see a difference there

In my opinion this is special case because its suggested way from docs to avoid code redundancy: https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html#mapping-whole-column-declarations-to-python-types

Using python 3.12 syntax with Annotated aliases would be more consistent with language design.

> Sorry, I don't understand what type checkers report as undefined. could you clarify?

I wanted to reply in issue, just wrong tab, so deleted that answer here, sorry :)

CaselIT:
but those are not type alias types, that's a different construct.

KaliszS:
> I don't see how it doesn't. as designed type alias types create new types, that are no longer base types, so sqlalchemy cannot assume to know what `PrimaryKey[uuid.UUID]` should be mapped to, since it's an entirely new type.
> 
> We already tried automatically resolving type alias types in 2.0 and it's a mess, since the runtime fight you every step of the way

I'm just curious - why in my case everything looks clear? Might it be that aliases with generics are easier to handle?

CaselIT:
not sure what looks clear?

KaliszS:
> not sure what looks clear?

I mean those generic aliases are seems to be consistent in terms of whats type it should be mapped to.

CaselIT:
It's not really any different than any other type alias type. Nothing prevents someone from doing `type x[T] = Annotated[T|str|bool, ...]`

Your change doesn't take that into consideration, considering only the type of T

KaliszS:
I have checked that again and some examples:
`type PrimaryKey = uuid.UUID` -> `__value__` from PrmaryKey equals to `uuid.UUID`
`type PrimaryKey = uuid.UUID | str` -> `__value__` from PrmaryKey equals to `uuid.UUID | str`
It would only require two more lines in the PR, but it would work perfectly. The first case should work right away since uuid.UUID is already correctly mapped. The second case would just need an explicit mapping, because when it’s UUID | str it ultimately has to be simplified to a single SQL type anyway, so nothing really changes here.

`type x[T] = Annotated[T|str|bool, ...]` -> that case already works with my PR, since it ignores all union types and just takes uuid.UUID as defined in the model. The other possible types don’t matter here.
```py
class User(BaseDbModel):
    id: Mapped[PrimaryKey[UUID]]
```

To be honest, I still don’t understand why we can’t implement it into the library. I’d be happy to volunteer and work on coding the other alias cases if you could share all the scenarios you consider problematic. I believe that together we can help introduce this cool new generic syntax from 3.12 :)

CaselIT:
> I have checked that again and some examples: `type PrimaryKey = uuid.UUID` -> `__value__` from PrmaryKey equals to `uuid.UUID` `type PrimaryKey = uuid.UUID | str` -> `__value__` from PrmaryKey equals to `uuid.UUID | str` It would only require two more lines in the PR, but it would work perfectly. The first case should work right away since uuid.UUID is already correctly mapped. The second case would just need an explicit mapping, because when it’s UUID | str it ultimately has to be simplified to a single SQL type anyway, so nothing really changes here.

That breaks very quickly if you have more complex types, and start combining type alias types inside other type alias types or annotated etc. Again done that and it was a mess.

> `type x[T] = Annotated[T|str|bool, ...]` -> that case already works with my PR, since it ignores all union types and just takes uuid.UUID as defined in the model. The other possible types don’t matter here.
> 
> ```python
> class User(BaseDbModel):
>     id: Mapped[PrimaryKey[UUID]]
> ```

That's wrong, the types should be `uuid|str|bool`

> To be honest, I still don’t understand why we can’t implement it into the library. I’d be happy to volunteer and work on coding the other alias cases if you could share all the scenarios you consider problematic. I believe that together we can help introduce this cool new generic syntax from 3.12 :)

It was already explained why. Making everything works consistently and maintaining that was very time consuming and the results was all but consistent. Also it's also not how the runtime suggests things should regarding type alias type.

KaliszS:
> That's wrong, the types should be uuid|str|bool

Tbh I don't know whats wrong here. It just took concrete type, which is UUID here, so it can be correctly mapped to an SQL type. You can't have union of types in SQL anyway.

> start combining type alias types inside other type alias types or annotated etc

So we really want to resign from modern syntax for such edge cases nobody probably going to use? I mean right now none of them works, including those simple cases we can easily handle. After the change we can have 99% use cases covered.

KaliszS:
> That's wrong, the types should be `uuid|str|bool`

I have also checked this example with the suggested solution:
```py
T = TypeVar("T")
PrimaryKey = Annotated[T | str, mapped_column(primary_key=True)]

class User(BaseDbModel):
    id: Mapped[PrimaryKey[UUID]]
```

```sh
(Pdb) python_type
typing.Annotated[typing.Union[uuid.UUID, str], <sqlalchemy.orm.properties.MappedColumn object at 0x109ea5f30>]
```
```sh
sqlalchemy.orm.exc.MappedAnnotationError: Could not locate SQLAlchemy Core type for Python type typing.Annotated[typing.Union[uuid.UUID, str], <sqlalchemy.orm.properties.MappedColumn object at 0x104192170>] inside the 'id' attribute Mapped annotation
```

What user should do in that case? Map this to sql.String or sql.Uuid? I guess it still should have been always Uuid, because the first type from union comes from conrecte type in Model, but why do we need the second one then? If we would catch Uuid here, we wouldn't need to create any confusion for user when he needs to define custom mappings.

CaselIT:
> > That's wrong, the types should be uuid|str|bool
> 
> Tbh I don't know whats wrong here. It just took concrete type, which is UUID here, so it can be correctly mapped to an SQL type. You can't have union of types in SQL anyway.

The resolved type is `Annotated[UUID|str|bool, ...]`, why should we consider only UUID? It makes it inconsistent with everything else and with how the type checker work.

> So we really want to resign from modern syntax for such edge cases nobody probably going to use? I mean right now none of them works, including those simple cases we can easily handle. After the change we can have 99% use cases covered.

You are making it overly simple. Since it's then the maintainers of sqlalchemy that need to keep maintaining this part for the forceable future, having a limited api is better than a inconsistent one that works seemingly at random and only sometimes.

The best option may be to provide a way for users to implement a callable that would be called to by sqlalchemy. This way one can do whatever they want.

> What user should do in that case? Map this to sql.String or sql.Uuid? I guess it still should have been always Uuid, because the first type from union comes from conrecte type in Model, but why do we need the second one then? If we would catch Uuid here, we wouldn't need to create any confusion for user when he needs to define custom mappings.

This seems an arbitrary decision, next week a new user will come and say they want the second type to be taken. It's not a good api. Also types have no order, these are equivalent `int | str == str | int` -> `True`

KaliszS:
> Also types have no order, these are equivalent int | str == str | int -> True

Interesting, it seems you are right: https://github.com/python/cpython/issues/103749. The issue hasnt'b been active for 2 years now, so if nobody responde, maybe I will try to implement this.

> Since it's then the maintainers of sqlalchemy that need to keep maintaining this part for the forceable future, having a limited api is better than a inconsistent one that works seemingly at random and only sometimes.

Ok, that's the point. You are totally right and I respect that. But what do you think about adding compromise solution? For example user could add specific metadata to Annotated class, like:
```py
type PrimaryKey[T] = Annotated[T, mapped_column(primary_key=True), Literal["unstable-alias"]]
```
```sh
(Pdb) python_type.__value__
typing.Annotated[T, <sqlalchemy.orm.properties.MappedColumn object at 0x113709f30>, typing.Literal['unstable-alias']]

(Pdb) python_type.__value__.__metadata__
(<sqlalchemy.orm.properties.MappedColumn object at 0x113709f30>, typing.Literal['unstable-alias'])
```

We could simply check if the Annotated class contains metadata, then iterate through it and look for our Literal. If it exists, we could trigger my piece of code, which should handle most cases. This could be documented as potentially unstable (with a short explanation why) and unmaintained, so it would be recommended only for users who understand how it works. We could even generate a console warning about its maintenance status.

CaselIT:
if we were to do something like you suggested I would rather provide an way to customise how the type inference worked, and maybe place some use cases in the example folder

zzzeek:
as a compromise solution we are looking into something like this:

```py
_T = TypeVar("_T")
type Type1[_T] = Annotated[_T, mapped_column(primary_key=True)]


class Base(DeclarativeBase):
    type_annotation_map = {Type1 (with or without the _T I dont care): magical_brand_new_sqlalhcemy_feature_to_do_something_here()}

class Foo(Base):
    __tablename__ = "foo"

    id: Mapped[Type1[int]]
```

KaliszS:
> as a compromise solution we are looking into something like this:
> [...]

Tbh it's frustrating when someone joins discussion only for advices like "yeah, have you considered just no to use the pep-695". You know, there are more patterns, like you can also respond to every issue to not use python and try different language or different ORM, why not.


> if we were to do something like you suggested I would rather provide an way to customise how the type inference worked, and maybe place some use cases in the example folder

Alright, understandable.

CaselIT:
it's also frustrating when someone provide solutions claiming that it will solve every issue, while in reality it's inconsistent with current api and breaks other simple cases, like this change doesn't support non generic annotated type alias types, such as `type X = Annotated[int, ...]`.

Overall here we are trying to find a solution that's consistent with the current api while at the same time improving the behaviour in the usecase

KaliszS:
Where have I claimed that it solves every issue? I have even stated that it affects only `Annotated` class. I understand that working here every day, you often encounter recurring issues, many of which stem not from actual problems, but simply from misunderstandings. That naturally makes it easier to get irritated. Still, I hope you’ve noticed that throughout this discussion I’ve tried to remain open to different points of view. I kept asking questions, always provided concrete data and examples to back up my arguments, and whenever you explained something clearly, I was able to respect your perspective without pushing further. Even when I sensed some frustration on your side, I had nothing to complain about in the way you handled the discussion. But when someone just drops rather arrogant comments or copies an example from AI that breaks it and then demands fixes, it becomes quite unpleasant. And I wonder, do we really need that?

> Overall here we are trying to find a solution that's consistent with the current api while at the same time improving the behaviour in the usecase

That's great. If you find an idea of how it should have been handled, you could always share your ideas in the issue. I would love to open another PR then, if your backlog would be already full.

CaselIT:
well you said it should "handle most cases", but I'm not here to nitpick.

Sorry if I seemed arrogant in my responses, English is not my primary language, and that sums to the fact that working with python typing is very frustrating since python changes things all the time and there seem to be no clear plan and things that seem simple are all but.
Like it's full of "it works with way seemingly by chance" as you can see from https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/util/typing.py

Let's see if we can come up with a way to improve things for "simple cases"

KaliszS:
Of course I meant most cases for Annotated classes :) There are issues like Annotated without Generic, as you already have caught one, but I assumed that PR still needed some polishing. And no - as I said, I had nothing to complain about you, so that part about arrogant in responses wasn't about you.

Yeah, I've seen that file, looks like a source of a lot of struggles.

zzzeek:
@KaliszS can you give me a clear and simple explanation for what is insufficient about not using a type alias, as in the case below:

```py

from typing import Annotated
from typing import TypeVar
from uuid import UUID

T = TypeVar("T")

PrimaryKey = Annotated[T, mapped_column(primary_key=True)]

class Base(DeclarativeBase):
    pass

class User(Base):
    __tablename__ = "users"

    id: Mapped[PrimaryKey[UUID]]

    data: Mapped[PrimaryKey[UUID]]

    data1: Mapped[PrimaryKey[UUID] | None]


print(CreateTable(User.__table__))
```

I am not able to detect any changes in typing or behavior inculding if we look at instances of User, compared to `type PrimaryKey[T] = Annotated[T, mapped_column(primary_key=True)]` (using a patch where I've made both versions "work")

KaliszS:
Sure. It’s mainly about syntax (most modern languages use the `[T]` syntax instead of Python’s `TypeVar`) and redundancy (if you want to define a type with multiple independent generic values, you need to declare a separate TypeVar for each one).

zzzeek:
you mean literally the issue is that by using the `type` keyword you dont have to declare the `TypeVar`......OK.  At the moment I can't imagine typing a whole codebase without needing to use `TypeVar` ever, ~~I guess I dont understand pep 695 very well~~

update, OK, they changed the class def syntax also , OK

CaselIT:
yes, PEP696 made python more similar to other languages that offer generics, like java, typescript, c#, rust, etc but with the `[]` syntax instead of the "standard" `<>` syntax

KaliszS:
> you mean literally the issue is that by using the type keyword you dont have to declare the TypeVar......OK. At the moment I can't imagine typing a whole codebase without needing to use TypeVar ever, I guess I dont understand pep 695 very well

We don’t know each other, so maybe I’m misreading the tone, but it feels like every message from you comes across as either ironic or passive-aggressive. Why the tension? I’m just bringing an idea that I believe could make the library better. I’m explaining why I think it’s useful, but I’m not forcing anyone to accept it or asking you to treat it as a priority. Anyway, coming back to the example…

```py
T = TypeVar("T")
U = TypeVar("U")
V = TypeVar("V")

PrimaryKey = Annotated[dict[T, U] | V, ...]
```
vs

```py
type PrimaryKey[T, U, V] = Annotated[dict[T, U] | V, ...]
```

zzzeek:
> > you mean literally the issue is that by using the type keyword you dont have to declare the TypeVar......OK. At the moment I can't imagine typing a whole codebase without needing to use TypeVar ever, I guess I dont understand pep 695 very well
> 
> We don’t know each other, so maybe I’m misreading the tone, but it feels like every message from you comes across as either ironic or passive-aggressive. Why the tension? 

huh?   the message was literally what it said, I knew there was a `type` keyword but I didnt realize until I just looked at the pep that it intends to make `Generic` and `TypeVar` no longer needed.

> I’m just bringing an idea that I believe could make the library better. I’m explaining why I think it’s useful, but I’m not forcing anyone to accept it or asking you to treat it as a priority. Anyway, coming back to the example…

you misread me as being skeptical.   I seek to understand the use case because that drives how we do things.   I receive PRs all the time for one liner changes to make some code possible and if I dont understand the context of what the user is trying to do and why they need that case, it wont be merged until we do.  nothing in SQLAlchemy can be there because "you can also do X here!" without there being a valid reason in a real application that this would be valuable, because we always need to be able to document "here is why you'd want to use this".    There must be a rationale for everything otherwise the little one liner of code that we added is often not actually useful, or implemented inconsistently.   in the case of your PR, it apparently makes an arbitrary choice to ignore `str` if the type is `type PrimaryKey[T] = Annotated[T | str, mapped_column(...)]` for example.

in the case of pep-695, they made a real choice to have two equivalent types mean different things

```
>>> type X = int
>>> type Y = int
>>> X == Y
False
```

that really throws a wrench into the way we are repurposing these types

PR #12878, opened 2025-09-23
WIP: Add tests for free-threading and build wheels

lysnikolaou:
### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**


lysnikolaou:
Whoops, I opened this by accident. I first wanted to open a PR against my own fork to test there. I'll open an upstream PR as soon as I have tested that all workflows work correctly.

test/base/test_core_threading.py
```diff
@@ -0,0 +1,121 @@
...
+                Column("id", Integer, primary_key=True),
+                Column("data", String(50)),
+            )
+
+        metadata.create_all(self.bind)
+
+        errors = []
+        barrier = threading.Barrier(parties=NUM_THREADS)
+
+        def create_table(thread_id):
```
zzzeek, line 100: just a comment as I see how this is happening, it would be good if the general pattern here of "make a barrier, spin up N threads, join" could be turned into a microframework of sorts, so that each test method would have something like:

```
@with_threaded
def create_table()
    # ...

```

See the `@profile_memory` decorator used in https://github.com/sqlalchemy/sqlalchemy/blob/main/test/aaa_profiling/test_memusage.py for a similar example

Also based on our conventions these tests maybe all should be part of the aaa_profiling directory which is where we put all the "test robustness against various runtime concerns" kinds of tests

zzzeek (commented):
I know you closed this but added some suggestions.

I appreciate that this is doing real tests that we can run now.

Mostly I think look into test/aaa_profiling at some of the patterns we use there and seek to have these new suites over there using similar patterns, great work!

test/base/test_core_threading.py
lysnikolaou: Thanks for this! It's very helpful.

Unfortunately, it looks like some tests are failing on CI, so I'm looking into these now (locally all was okay). After I'm done I'll push all the changes (along with your suggestions) to a new PR.

tox.ini
```diff
@@ -312,3 +315,6 @@ deps = {[testenv]deps}
 commands=
   python -m pytest {env:PYTEST_COLOR} {env:WORKERS} {env:SQLITE:} {env:POSTGRESQL:} {env:MYSQL:} {env:ORACLE:} {env:MSSQL:} {env:IDENTS:} {env:PYTEST_EXCLUDES:} {env:COVERAGE:} {posargs}
   oracle,mssql,sqlite_file: python reap_dbs.py db_idents.txt
+
+  py{313t,314t}-sqlite_file: python -m pytest {env:PYTEST_COLOR} {env:SQLITE:} {env:IDENTS:} {posargs}
+  tests/base/test_core_threading.py tests/orm/test_session_threading.py
```
CaselIT, line 320: You could use a marker for this instead of having to list the tests files here

lysnikolaou: How would I go about doing that? I want to explicitly run these files with `--db sqlite_file` which as far as I understand is not run on CI.

lysnikolaou:
Opened #12882.

PR #12911, opened 2025-10-14
_row_cy.py and related cython stuff

albertedwardson:
Hi! I’m going to close this PR and open a new one with a clean commit history and the benchmarks included.

I’m opening this draft mostly to share a rough sketch, get some initial feedback, and document a couple of things I discovered along the way.

### Discovery 1

Unfortunately, Cython can’t emit specialised C APIs based on protocol-style types (like `Sequence`). It requires concrete types instead.
Reference: https://groups.google.com/g/cython-users/c/yzLAkAJEjMA

So if we want to mirror Python 3.11+ behaviour with the specializing interpreter, we need to narrow type hints down to concrete ones (`tuple`, `list`, etc.) and avoid relying on protocol-based annotations.

### Discovery 2

It might be possible to implement a module-level `__getattr__` in `util.cython.py` that satisfies type checkers without affecting runtime behaviour. That could eliminate the need to define no-op fallbacks for every `cython.something` in non-compiled mode.

I experimented with it, but the results were not so good, my relationship with Python’s type “system” is complicated

I’d really appreciate any thoughts on these discoveries or suggestions for alternative approaches — especially if there are patterns already used in the codebase that I might be missing

### DIFF
--- a/lib/sqlalchemy/engine/_row_cy.py
+++ b/lib/sqlalchemy/engine/_row_cy.py
@@ -142,10 +142,17 @@ try:
     from cython.cimports.cpython import PySequence_Length
 except ImportError:
     if not cython.compiled:
-        PyTuple_New = lambda n: [None] * n # actually list
-        def PyTuple_SET_ITEM(tup, idx, item): tup[idx] = item # type: ignore
+
+        def PyTuple_New(n: int) -> list:
+            """Actually produces list if not compiled"""
+            return [None] * n
+
+        def PyTuple_SET_ITEM(tup, idx, item):
+            tup[idx] = item  # type: ignore
+
         PySequence_Length = len
         Py_INCREF = cython._no_op
+
 
 @cython.inline
 @cython.cfunc
--- a/lib/sqlalchemy/engine/_row_cy.py
+++ b/lib/sqlalchemy/engine/_row_cy.py
@@ -135,20 +135,35 @@ class BaseRow:
         return self._data
 
 
+try:
+    from cython.cimports.cpython import PyTuple_New, PyTuple_SET_ITEM, PySequence_Length
+except ImportError:
+    if not cython.compiled:
+        PyTuple_New = lambda n: [] # actually list
+        def PyTuple_SET_ITEM(tup, idx, item): tup[idx] = item
+        PySequence_Length = len
+
 @cython.inline
 @cython.cfunc
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.returns(tuple)
+@cython.locals(res=tuple, proc_size=cython.Py_ssize_t, p=object, value=object)
 def _apply_processors(
     proc: _ProcessorsType, data: Sequence[Any]
 ) -> Tuple[Any, ...]:
-    res: List[Any] = list(data)
-    proc_size: cython.Py_ssize_t = len(proc)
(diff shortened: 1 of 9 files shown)


CaselIT:
> Hi! I’m going to close this PR and open a new one with a clean commit history and the benchmarks included.

it's not really necessary, since we use gerrit that will squash everything anyway. Feel free to do as you prefer

CaselIT:
this appears to include other changes as well, like the changelog file is not right. merge/rebase/or open a new PR (as you prefer)

lib/sqlalchemy/engine/_row_cy.py
```diff
@@ -135,20 +135,46 @@ def _to_tuple_instance(self) -> Tuple[Any, ...]:
...
+            tup[idx] = item
+
+        PySequence_Length = len
+        Py_INCREF = cython._no_op
+
+
 @cython.inline
 @cython.cfunc
+@cython.returns(tuple)
+@cython.locals(res=tuple, proc_size=cython.Py_ssize_t, p=object, value=object)
```
CaselIT, line 160: oh interesting, I did not know about locals, and used the type annotation instead

CaselIT (commented):
the approach looks similar to what I had in mind once you made the suggestion in the discussion the other day, so that looks fine by me.

were you able to see performance differences between the two implementations?

albertedwardson:
> it's not really necessary, since we use gerrit that will squash everything anyway. Feel free to do as you prefer

ok, thanks! I'll make a new one; I've messed up with branches anyway :)

> were you able to see performance differences between the two implementations?

only in theory :) I'll open a new one, only after benchmarking and if there will be speedups not in the range of noise

CaselIT:
> > were you able to see performance differences between the two implementations?
> 
> only in theory :) I'll open a new one, only after benchmarking and if there will be speedups not in the range of noise

I know the feeling, sadly it's not easy to predict performance improvements. Sometimes ones comes up with new impl that should be a lot more efficient in theory but in practice have no impact.. I guess python is a bit strange on this part compared to c.

albertedwardson:
Notes:

creating empty tuple with size n and filling it in loop resulted in slower execution time :) 

mostly likely due to extra branch in loop and/or due to how list(data) can just copy values from data in one, maybe, memcpy call, without loops (assuming)

however, this approach should be lower in memory consumption, but only in theory though :) 

currently playing on how to fill result tuple in one call without loops, and then mutate its values by applying appropriate processors if processor is not None. at this point: no luck. it segfaults or emits strange exceptions 

p.s. I got really sad seeing that cython compiled getattr for row is about x3 times slower than the python one. was tinkering with it too but got no luck too. maybe cython generated c code that firstly tries to search for attr in classvars, `__slots__, __dict__`, and only then falls back to overriden `__getattr__`? idk

CaselIT:
> creating empty tuple with size n and filling it in loop resulted in slower execution time :)

:(

> p.s. I got really sad seeing that cython compiled getattr for row is about x3 times slower than the python one. was tinkering with it too but got no luck too. maybe cython generated c code that firstly tries to search for attr in classvars, __slots__, __dict__, and only then falls back to overriden __getattr__? idk

IIRC I tried having it in cython and got the same results. that's why it's currently defined in python (IIRC)

PR #12919, opened 2025-10-16
perf tuning of `BaseRow`

albertedwardson:
tl;dr: it's now bit faster overall 

Before:
```
Running case BaseRow
Running python     .................. Done
Running cython     .................. Done
                    | python  | cython  | cy / py |
base_row_new        | 1.25334 | 0.16760 | 0.13372 |
row_new             | 1.29170 | 0.20576 | 0.15929 |
base_row_new_proc   | 4.43647 | 2.09866 | 0.47304 |
row_new_proc        | 4.45219 | 2.26768 | 0.50934 |
brow_new_proc_none  | 1.79195 | 0.41670 | 0.23254 |
row_new_proc_none   | 1.87007 | 0.46924 | 0.25092 |
row_dumps           | 0.25029 | 0.32780 | 1.30968 |
row_loads           | 1.48172 | 0.51749 | 0.34925 |
row_values_impl     | 0.26831 | 0.29232 | 1.08948 |
row_iter            | 0.65250 | 0.36315 | 0.55655 |
row_len             | 0.21574 | 0.07232 | 0.33525 |
row_hash            | 0.40309 | 0.24871 | 0.61702 |
getitem             | 0.28383 | 0.16171 | 0.56974 |
getitem_slice       | 0.60968 | 0.34910 | 0.57259 |
get_by_key          | 0.46660 | 0.44169 | 0.94661 |
getattr             | 0.63748 | 1.89029 | 2.96525 |
get_by_key_recreate | 1.92867 | 1.79520 | 0.93079 |
getattr_recreate    | 0.85012 | 1.50871 | 1.77468 |
> mean of values    | —       | —       | 0.76532 |
```

After:

```
Running case BaseRow
Running python     .................. Done
Running cython     .................. Done
                    | python  | cython  | cy / py |
base_row_new        | 1.62721 | 0.20094 | 0.12349 |
row_new             | 1.43012 | 0.29490 | 0.20621 |
base_row_new_proc   | 3.92941 | 1.74143 | 0.44317 |
row_new_proc        | 4.11000 | 1.75141 | 0.42613 |
brow_new_proc_none  | 2.00083 | 0.51876 | 0.25927 |
row_new_proc_none   | 2.09076 | 0.55015 | 0.26313 |
row_dumps           | 0.30078 | 0.47270 | 1.57157 |
row_loads           | 1.48260 | 0.67376 | 0.45444 |
row_values_impl     | 0.33249 | 0.52256 | 1.57165 |
row_iter            | 0.77671 | 0.47556 | 0.61227 |
row_len             | 0.24589 | 0.06300 | 0.25621 |
row_hash            | 0.48600 | 0.22412 | 0.46116 |
getitem             | 0.33727 | 0.16818 | 0.49865 |
getitem_slice       | 0.72768 | 0.38407 | 0.52780 |
get_by_key          | 0.89138 | 0.74737 | 0.83844 |
getattr             | 0.78927 | 0.34847 | 0.44151 |
get_by_key_recreate | 2.79428 | 2.43914 | 0.87290 |
getattr_recreate    | 1.10803 | 0.74660 | 0.67380 |
> mean of values    | —       | —       | 0.58343 |
```

### DIFF
--- a/lib/sqlalchemy/engine/_row_cy.py
+++ b/lib/sqlalchemy/engine/_row_cy.py
@@ -79,14 +79,13 @@ else:
     from cython.cimports.cpython import PyTuple_SET_ITEM
     from cython.cimports.cpython import PySequence_Fast_GET_SIZE
 
-    obj_getattr = object.__getattribute__
 
     @cython.inline
     @cython.cfunc
     @cython.wraparound(False)
     @cython.boundscheck(False)
     @cython.returns(tuple)
-    @cython.locals(res=tuple, proc_size=cython.Py_ssize_t, p=object)
+    @cython.locals(res=tuple, proc_size=cython.Py_ssize_t, p=object, value=object)
     def _apply_processors(proc: object, data: object) -> Tuple[Any, ...]:
         proc_size = PySequence_Fast_GET_SIZE(proc)
         # TODO: would be nice to do this only on the fist row
@@ -95,9 +94,11 @@ else:
         for i in range(proc_size):
             p = proc[i]
             if p is not None:
-                PyTuple_SET_ITEM(res, i, Py_INCREF(p(data[i])))
+                value = p(data[i])
             else:
-                PyTuple_SET_ITEM(res, i, Py_INCREF(data[i]))
+                value = data[i]
+            Py_INCREF(value)
+            PyTuple_SET_ITEM(res, i, value)
         return res
 
     @cython.inline
@@ -105,10 +106,9 @@ else:
     def rowproxy_reconstructor(
         cls: Type[BaseRow], state: Dict[str, Any]
     ) -> BaseRow:
-        obj = cls.__new__(cls)
+        obj = cython.cast(BaseRow, cls.__new__(cls))
         obj.__setstate__(state)
         return obj
-
 
 @cython.cclass
 class BaseRow:
@@ -121,10 +121,6 @@ class BaseRow:
         )
         _data: Tuple[Any, ...] = cython.declare(tuple, visibility="readonly")
 
-        @cython.inline
-        @cython.cfunc
-        def _getstate_impl(self) -> dict:
-            return {"_parent": self._parent, "_data": self._data}
 
     def __init__(
         self,
@@ -134,14 +130,13 @@ class BaseRow:
         data: Sequence[Any],
     ) -> None:
         """Row objects are constructed by CursorResult objects."""
-
         self._set_attrs(
             parent,
             key_to_index,
             (
                 _apply_processors(processors, data)
                 if processors is not None
-                else (data if isinstance(data, tuple) else tuple(data))
+                else data if isinstance(data, tuple) else tuple(data)
             ),
         )
 
@@ -153,27 +148,40 @@ class BaseRow:
         key_to_index: Dict[_KeyType, int],
         data: Tuple[Any, ...],
     ):
-        if cython.compiled:
-            # cython does not use __setattr__
-            self._parent = parent
-            self._key_to_index = key_to_index
-            self._data = data
-        else:
-            # python does, so use object.__setattr__
-            object.__setattr__(self, "_parent", parent)
-            object.__setattr__(self, "_key_to_index", key_to_index)
-            object.__setattr__(self, "_data", data)
+        self._parent = parent
+        self._key_to_index = key_to_index
+        self._data = data
 
     if cython.compiled:
-
+        @cython.inline
+        @cython.cfunc
+        def _getstate_impl(self) -> dict:
+            self = cython.cast(BaseRow, self)
+            return {"_parent": self._parent, "_data": self._data}
+        
         def __getstate__(self) -> Dict[str, Any]:
-            return self._getstate_impl()
+            return cython.cast(BaseRow, self)._getstate_impl()
 
         def __reduce__(self) -> Tuple[Any, Any]:
+            self = cython.cast(BaseRow, self)
             return (
                 rowproxy_reconstructor,
                 (self.__class__, self._getstate_impl()),
             )
+        
+
+        def __getattribute__(self, name: _KeyType) -> object:
+            self = cython.cast(BaseRow, self)
+            if isinstance(name, str):
+                if name == "_data":
+                    return self._data
+                if name == "_key_to_index":
+                    return self._key_to_index
+                if name == "_parent":
+                    return self._parent
+            if name[0] != "_" and name[-1] != "_":
+                return self._get_by_key_impl(name, True)
+            return object.__getattribute__(self, name)
 
     else:
 
@@ -190,13 +198,14 @@ class BaseRow:
             )
 
     def __setstate__(self, state: Dict[str, Any]) -> None:
+        self = cython.cast(BaseRow, self)
         parent = state["_parent"]
         self._set_attrs(parent, parent._key_to_index, state["_data"])
 
     @cython.wraparound(False)
     @cython.boundscheck(False)
     def _values_impl(self) -> List[Any]:
-        return list(self._data)
+        return list(cython.cast(BaseRow, self)._data)
 
     def __iter__(self) -> Iterator[Any]:
         return iter(self._data)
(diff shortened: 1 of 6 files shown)


zzzeek:
just curious are you testing on python 3.14?   that's the latest and greatest interpreter that is supposedly even faster than 3.13.

albertedwardson:
@zzzeek no, I was testing on 3.12

CaselIT:
great work. I'll look into it a bit better later.

From a quick glace I thin we can simplify some of it, in general to avoid too many `if compiled` blocks, it's fine to incur an additional fn call in the pure python code, for example for the get/set state, we could just use the cython version that would use the additional call to _getstate_impl in python.

albertedwardson:
many thanks! 

p.s. maybe we could bring back the old speeds of `get_by_key`, `row_values_impl` and `row_dumps` by somehow telling cython compiler that `BaseRow` object has direct pointers to those functions like it was before. those slowdowns caused by defining ` __getattribute__`, which accelerated arbitrary attribute access speeds in a cost of direct pointers to known methods and attributes.

CaselIT:
maybe we could cast self to baserow? something like 
```py
def _get_by_key_impl(self, ...):
  base_row = cython.cast(BaseRow, self)
  ...
```

albertedwardson:
I think it should be considered to somehow reduce number of function decorators. Since `get_by_key_recreate` went from ~`1.92867` to ~ `2.70708` in pure python mode on my last commit where I added two cython decorators on `_get_by_key_impl`

CaselIT:
strange since the decorators should be mostly no-op that happen at import time, not at runtime.
unless you meat `cython.cast` that has a runtime impact in python since it would still call the no-op cast fn

albertedwardson:
I'm done and satisfied with results :) It's fully ready for review and I will not make any further changes 

Maybe there is actually some simple and clever trick to make `_values_impl` and `__getstate__` faster when compiled, but I lost the eye for details

CaselIT:
the results look nice I'll try taking a look this week!

albertedwardson:
> strange since the decorators should be mostly no-op that happen at import time, not at runtime. unless you meat `cython.cast` that has a runtime impact in python since it would still call the no-op cast fn

yeah, you're right. I wrongly assumed the slowdown was caused by the decorators themselves, thinking about the usual closure and wrapping overhead, but forgot that this only applies at import time, during module initialization


> the results look nice I'll try taking a look this week!

thanks!

lib/sqlalchemy/engine/_row_cy.py
```diff
@@ -52,6 +100,22 @@ class BaseRow:
...
+        def __getattribute__(self, name: _KeyType) -> object:
+            self = cython.cast(BaseRow, self)
+            if isinstance(name, str):
+                if name == "_data":
+                    return self._data
+                if name == "_key_to_index":
+                    return self._key_to_index
+                if name == "_parent":
+                    return self._parent
+            if name[0] != "_" and name[-1] != "_":
```
albertedwardson, line 112: simple heuristic assuming that column names do not start and end with `_`, more likely it is some `_protected_attr`, `__private` or `__dunder__`

lib/sqlalchemy/engine/_row_cy.py
```diff
@@ -60,13 +124,15 @@ def __init__(
...
-            _apply_processors(processors, data)
-            if processors is not None
-            else tuple(data)
+        self._set_attrs(
+            parent,
+            key_to_index,
+            (
+                _apply_processors(processors, data)
+                if processors is not None
+                else data if isinstance(data, tuple) else tuple(data)
```
albertedwardson, line 133: it is tuple already, do not copy it

lib/sqlalchemy/engine/_row_cy.py
```diff
@@ -76,28 +142,24 @@ def _set_attrs(  # type: ignore[no-untyped-def] # cython crashes
...
+    def __getstate__(self) -> Dict[str, Any]:
+        return {"_parent": self._parent, "_data": self._data}
 
     def __reduce__(self) -> Tuple[Any, Any]:
         return (
             rowproxy_reconstructor,
-            (self.__class__, self.__getstate__()),
+            (
+                self.__class__,
+                {"_parent": self._parent, "_data": self._data},
```
albertedwardson, line 157: manual inlining

lib/sqlalchemy/engine/_row_cy.py
```diff
@@ -117,38 +179,34 @@ def __hash__(self) -> int:
         def __getitem__(self, key: Any) -> Any:
             return self._data[key]
 
-    def _get_by_key_impl_mapping(self, key: _KeyType) -> Any:
-        return self._get_by_key_impl(key, False)
+    def _get_by_key_impl_mapping(self, key: _KeyType) -> object:
+        return cython.cast(BaseRow, self)._get_by_key_impl(key, False)
```
albertedwardson, line 183: maybe do not cast in non-cythonized version?

lib/sqlalchemy/engine/row.py
```diff
@@ -274,7 +263,7 @@ def _fields(self) -> Tuple[str, ...]:
             :attr:`.Row._mapping`
 
         """
-        return tuple([k for k in self._parent.keys if k is not None])
+        return tuple(k for k in self._parent.keys if k is not None)
```
albertedwardson: do not create intermediate list, something that I noticed while just passing by

test/aaa_profiling/test_resultset.py
```diff
@@ -266,4 +266,7 @@ def __getitem__(self, item):
             def __iter__(self):
                 return iter(self.data)
 
+            def __len__(self):
```
albertedwardson, line 269: unrelated; sequence protocol defines `__len__` of an object

lib/sqlalchemy/engine/_row_cy.py
```diff
@@ -4,13 +4,16 @@
 #
 # This module is part of SQLAlchemy and is released under
 # the MIT License: https://www.opensource.org/licenses/mit-license.php
-# mypy: disable-error-code="misc"
+# mypy: disable-error-code="misc,no-redef,valid-type,no-untyped-call"
+# mypy: disable-error-code="index,no-any-return,arg-type,assignment"
+# cython: wraparound=False, boundscheck=False
```
albertedwardson, line 9: all the bounds checking in `_apply_processors` is done manually via assert, other methods do not need it

albertedwardson (commented):
Added notes to code

lib/sqlalchemy/engine/row.py
CaselIT: it's usually significantly faster to go list -> tuple vs generator -> tuple. that's why we use that pattern

```
In [2]: %timeit tuple(i for i in range(100))
2.87 µs ± 19.1 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [3]: %timeit tuple([i for i in range(100)])
1.73 µs ± 12 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [4]: %timeit tuple([i for i in range(10000)])
183 µs ± 1.12 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [5]: %timeit tuple(i for i in range(10000))
301 µs ± 3.94 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

let's revert

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision b13204b7416083ccff09a45291b11fd606ff0fbd of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change b13204b7416083ccff09a45291b11fd606ff0fbd: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6264

lib/sqlalchemy/engine/_row_cy.py
```diff
@@ -52,6 +100,22 @@ class BaseRow:
...
+            if isinstance(name, str):
+                if name == "_data":
+                    return self._data
+                if name == "_key_to_index":
+                    return self._key_to_index
+                if name == "_parent":
+                    return self._parent
+            if name[0] != "_" and name[-1] != "_":
+                try:
+                    return self._get_by_key_impl(name, False)
```
CaselIT, line 114: I don't think we can do this, since it would mean that a column that shadows a method would get returned instead of the method :(

CaselIT: found a fix

CaselIT (commented):
Great work. I've run with it a bit and optimized another bit the getattr method.

It's on gerrit

lib/sqlalchemy/engine/_row_cy.py
```diff
@@ -76,28 +142,24 @@ def _set_attrs(  # type: ignore[no-untyped-def] # cython crashes
...
-            # cython does not use __setattr__
-            self._parent = parent
-            self._key_to_index = key_to_index
-            self._data = data
-        else:
-            # python does, so use object.__setattr__
-            object.__setattr__(self, "_parent", parent)
-            object.__setattr__(self, "_key_to_index", key_to_index)
-            object.__setattr__(self, "_data", data)
+        self._parent = parent
```
sqla-tester, line 145: **Federico Caselli** (CaselIT) wrote:

NIT: do `name and name[0] != '_'`

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6264

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6264

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6264

lib/sqlalchemy/engine/_row_cy.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6264

CaselIT:
(note that we will continue on gerrit, no need to push any other changes here)

thanks for improving the implementation

albertedwardson:
Thanks a lot for taking the time to review my changes and for all the feedback! I really appreciate the detailed attention and the opportunity to contribute

Just to confirm, is there anything else needed from my side at this point? It looks like everything in Gerrit is resolved

CaselIT:
> Just to confirm, is there anything else needed from my side at this point? It looks like everything in Gerrit is resolved

no, not at this time. The gerrit will be merged once the review is done. Thanks!

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6264 has been **merged**. Congratulations! :)

Issue #12948, opened 2025-10-30
PostgreSQL 14+: use subscripting operator for HSTORE


loic-simon:
### Describe the use case

Since [PostgreSQL 14.0](https://www.postgresql.org/docs/release/14.0), HSTORE values can be accessed using square brackets in addition to the `->` operator:

```sql
SELECT store['key'] FROM my_table;
SELECT store -> 'key' FROM my_table;  -- exactly equivalent
```

But contrary to `->`, this syntax can be used in UPDATE statements:

```sql
UPDATE my_table SET store['key'] = 'value';  -- ✅ works as expected
UPDATE my_table SET store -> 'key' = 'value';  -- ❌ ERROR: syntax error at or near "->" 
```

SQLAlchemy currently always use `->`, which means that that the following MRE emits invalid SQL:

```py
from sqlalchemy import Column, MetaData, Table, update
from sqlalchemy.dialects.postgresql import HSTORE, dialect

my_table = Table("my_table", MetaData(), Column("store", HSTORE()))

stmt = update(my_table).values({my_table.c.store["key"]: "value"})
print(stmt.compile(dialect=dialect()))  # ❌ UPDATE my_table SET store -> :store_1=:param_1
```

Just updating `HSTORE.Comparator._setup_getitem` to use `operators.getitem` make it works:

```py
from sqlalchemy import Column, MetaData, Table, update
from sqlalchemy.dialects.postgresql import HSTORE, dialect
from sqlalchemy.sql import operators

HSTORE.comparator_factory._setup_getitem = lambda self, ix: (operators.getitem, ix, self.type.text_type)

my_table = Table("my_table", MetaData(), Column("store", HSTORE()))

stmt = update(my_table).values({my_table.c.store["key"]: "value"})
print(stmt.compile(dialect=dialect()))  # ✅ UPDATE my_table SET store[%(store_1)s]=%(param_1)s
```

Would you be open to change this, probably opt-in to not break PostgreSQL <14? (eg via `HSTORE(use_subscription=True)` or similar option)

### Databases / Backends / Drivers targeted

PostgreSQL 14+, all drivers

### Example Use

See MRE above, or using ORM equivalent versions that currently do not work:

```py
session.query(MyTable).update({MyTable.store["key"]: "value"})
session.execute(update(MyTable).values({MyTable.store["key"]: "value"}))
```

zzzeek:
[...]

PR #12950, opened 2025-10-31
PostgreSQL / HSTORE: add parameter to use subscripting operator

loic-simon:
### Description

Add a new opt-in parameter `use_subscripting_operator` to the `dialects.postgresql.hstore.HSTORE` type:

| Mode | `__getitem__` operator | Usable in SELECT | Usable in UPDATE |
|-------|----------|-------------------|-------------------|
| `use_subscripting_operator=False` (default) | `data -> %(key)s` (`dialects.postgresql.operators.GETITEM`) | ✅ | ❌ |
| `use_subscripting_operator=True` |`data[%(key)s]` (`sql.operators.getitem`) | ✅ | ✅ |

I'm not totally sure of the parameter name, totally open to bikeshedding!

### Checklist

This pull request is:

- [ ] A short code fix
	- Issue: #12948 
	- Tests: new tests inheriting of existing HSTORE test classes:
	    - code generation (offline)
	    - round-trip tests (online), including a new test using subscription to update a specific key

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/hstore.py
+++ b/lib/sqlalchemy/dialects/postgresql/hstore.py
@@ -18,6 +18,7 @@ from .operators import HAS_ANY
 from .operators import HAS_KEY
 from ... import types as sqltypes
 from ...sql import functions as sqlfunc
+from ...sql import operators
 from ...types import OperatorClass
 
 __all__ = ("HSTORE", "hstore")
@@ -104,6 +105,7 @@ class HSTORE(sqltypes.Indexable, sqltypes.Concatenable, sqltypes.TypeEngine):
     __visit_name__ = "HSTORE"
     hashable = False
     text_type = sqltypes.Text()
+    use_subscripting_operator = False
 
     operator_classes = (
         OperatorClass.BASE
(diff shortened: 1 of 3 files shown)


lib/sqlalchemy/dialects/postgresql/hstore.py
```diff
@@ -112,15 +114,29 @@ class MyClass(Base):
         | OperatorClass.CONCATENABLE
     )
 
-    def __init__(self, text_type=None):
+    def __init__(self, text_type=None, use_subscripting_operator=False):
```
zzzeek: id rather not have this an opt-in like this, because this becomes a forever parameter that should not be needed, we already know what version of postgresql is in use.   

this should be based on the version of Postgresql that's being used in a way similar to 0f2702caba.   it can use the existing _supports_jsonb_subscripting flag.

zzzeek (changes requested):
hi -

so this is not how I want to do this.  i want it to work like the approach in 0f2702caba, making use of  _supports_jsonb_subscripting inline inside the compile to use the correct operator, it should work automatically.  having a flag floating around implies you should just use a custom comparator so that we don't have extra baggage like this (see below).

this is not really feasible in 2.0 because HSTORE is using a custom op and we don't have compiler hooks for custom ops.   I'm even trying some workarounds and not getting far.   a new capability will need to be added to 2.1. to support this.

so for 2.0, use the recipe I just added at https://github.com/sqlalchemy/sqlalchemy/issues/12948#issuecomment-3474022642

PR #13000, opened 2025-11-22
Add native BOOLEAN type support for Oracle 23c and later versions

### LINKED ISSUE #11633: Support Oracle 23ai BOOLEAN datatype
### Describe the bug

The latest Oracle DB release as of 7/22/24 - DB23AI - Supports native "BOOLEAN" types, whereas the type used to be captured by a SMALLINT. As a result, the documentation and behavior around CamelCase type "Boolean" is now out of date for Oracle DB:

https://support.oracle.com/knowledge/Oracle%20Database%20Products/3002488_1.html
"""
Unlike [String](https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.String), which represents a string datatype that all databases have, not every backend has a real “boolean” datatype; some make use of integers or BIT values 0 and 1, some have boolean literal constants true and false while others dont. For this datatype, [Boolean](https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Boolean) may render BOOLEAN on a backend such as PostgreSQL, BIT on the MySQL backend and _SMALLINT on Oracle_
"""

_No response_

### SQLAlchemy Version in Use

2.0.31

cx_oracle

OracleDB 23

### Python Version

3.10

### Operating system

Linux

### To Reproduce

```python
# create table w/ sqlalchemy
op.create_table('test', sa.Column('col1', sa.Boolean())

# query table in oracle
select col1 from test;

COL1
----
   0

# create table using Boolean in oracle just fine
create table test2 (col2 boolean);
insert into test values (TRUE);

Col2
----
TRUE
```

### Error

The result of the behavior mismatch is we get some nasty error when trying to run queries using the .is_ flag - the .is_(False) feature works just fine on Postgres, and then fails when using Oracle. The fix, for now, I believe is to use BOOLEAN instead of Boolean type.

### Additional context

Oracle 23 only just came out recently, so I don't expect this to be done right away. See also: https://github.com/sqlalchemy/sqlalchemy/issues/10375

ybjeon01:
Oracle Database 23c introduced native BOOLEAN type support. This change adds automatic detection and usage of native BOOLEAN when connected to Oracle 23c or later. For earlier versions, boolean values continue to be emulated using SMALLINT.

Fixes: #11633

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/oracle/base.py
+++ b/lib/sqlalchemy/dialects/oracle/base.py
@@ -640,6 +640,47 @@ E.g. to use this datatype in a :class:`.Table` definition::
         ),
     )
 
+Boolean Support
+---------------
+
+.. versionadded:: 2.1
+
+Oracle Database 23ai introduced native support for the ``BOOLEAN`` datatype.
+The Oracle dialect automatically detects the database version and uses the
+native ``BOOLEAN`` type when available, or falls back to emulation using
+``SMALLINT`` on older Oracle versions.
+
+The standard :class:`_types.Boolean` type can be used in table definitions::
+
+    from sqlalchemy import Boolean, Column, Integer, Table, MetaData
+
+    metadata = MetaData()
+
+    my_table = Table(
+        "my_table",
+        metadata,
+        Column("id", Integer, primary_key=True),
+        Column("flag", Boolean),
+    )
+
+On Oracle 23ai and later, this will generate DDL using the native ``BOOLEAN`` type:
+
+.. code-block:: sql
+
+    CREATE TABLE my_table (
+        id INTEGER NOT NULL,
+        flag BOOLEAN,
+        PRIMARY KEY (id)
+    )
+
+On earlier Oracle versions, it will use ``SMALLINT`` for storage with appropriate
+constraints and conversions.
+
+The :class:`_types.Boolean` type is also available as ``BOOLEAN`` from the Oracle
+dialect for consistency with other type names::
+
+    from sqlalchemy.dialects.oracle import BOOLEAN
+
 DateTime Compatibility
 ----------------------
 
@@ -964,6 +1005,7 @@ from .types import _OracleDate
 from .types import BFILE
 from .types import BINARY_DOUBLE
 from .types import BINARY_FLOAT
+from .types import BOOLEAN
 from .types import DATE
 from .types import FLOAT
 from .types import INTERVAL
@@ -1060,6 +1102,7 @@ ischema_names = {
     "BINARY_DOUBLE": BINARY_DOUBLE,
     "BINARY_FLOAT": BINARY_FLOAT,
     "ROWID": ROWID,
+    "BOOLEAN": BOOLEAN,
     "VECTOR": VECTOR,
 }
 
@@ -1207,7 +1250,10 @@ class OracleTypeCompiler(compiler.GenericTypeCompiler):
         return self.visit_NUMBER(type_, precision=19, **kw)
 
     def visit_boolean(self, type_, **kw):
-        return self.visit_SMALLINT(type_, **kw)
+        if self.dialect.supports_native_boolean:
+            return self.visit_BOOLEAN(type_, **kw)
+        else:
+            return self.visit_SMALLINT(type_, **kw)
 
     def visit_RAW(self, type_, **kw):
         if type_.length:
@@ -1973,6 +2019,7 @@ class OracleDialect(default.DefaultDialect):
     supports_default_metavalue = True
     supports_empty_insert = False
     supports_identity_columns = True
+    supports_native_boolean = True
 
     aggregate_order_by_style = AggregateOrderByStyle.WITHIN_GROUP
 
@@ -2053,6 +2100,7 @@ class OracleDialect(default.DefaultDialect):
             self.colspecs.pop(sqltypes.Interval)
             self.use_ansi = False
 
+        self.supports_native_boolean = self.server_version_info >= (23,)
         self.supports_identity_columns = self.server_version_info >= (12,)
         self._supports_offset_fetch = (
             self.enable_offset_fetch and self.server_version_info >= (12,)
--- a/lib/sqlalchemy/dialects/oracle/types.py
+++ b/lib/sqlalchemy/dialects/oracle/types.py
@@ -23,6 +23,9 @@ if TYPE_CHECKING:
     from ...sql.type_api import _LiteralProcessorType
 
 
+BOOLEAN = sqltypes.BOOLEAN
+
+
 class RAW(sqltypes._Binary):
     __visit_name__ = "RAW"
 
--- a/test/dialect/oracle/test_dialect.py
+++ b/test/dialect/oracle/test_dialect.py
@@ -740,6 +740,17 @@ class CompatFlagsTest(fixtures.TestBase, AssertsCompiledSQL):
         self.assert_compile(Unicode(50), "NVARCHAR2(50)", dialect=dialect)
         self.assert_compile(UnicodeText(), "NCLOB", dialect=dialect)
 
+    def test_native_boolean_flag(self):
+        dialect = self._dialect((19, 0, 0))
+        assert dialect.supports_native_boolean
+        dialect.initialize(Mock())
+        assert not dialect.supports_native_boolean
+
+        dialect = self._dialect((23, 0, 0))
+        assert dialect.supports_native_boolean
+        dialect.initialize(Mock())
+        assert dialect.supports_native_boolean
+
     def test_ident_length_in_13_is_30(self):
         from sqlalchemy import __version__
 
--- a/test/dialect/oracle/test_types.py
+++ b/test/dialect/oracle/test_types.py
@@ -24,6 +24,7 @@ from sqlalchemy import MetaData
 from sqlalchemy import NCHAR
 from sqlalchemy import Numeric
 from sqlalchemy import NVARCHAR
+from sqlalchemy import schema
 from sqlalchemy import select
 from sqlalchemy import SmallInteger
 from sqlalchemy import String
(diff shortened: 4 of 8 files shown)


test/dialect/oracle/test_types.py
```diff
@@ -248,12 +249,162 @@ def test_float_type_compile(self, type_, sql_text):
...
+            eq_(len(result), 1)
+            eq_(result[0][0], 2)
+        finally:
+            # Restore original setting
+            connection.dialect.supports_native_boolean = (
+                original_supports_native
+            )
+
+
+class TypesTestLegacy(fixtures.TestBase):
```
zzzeek: Hmmm what happened here?    I think this class header was added by accident

test/dialect/oracle/test_types.py
```diff
@@ -248,12 +249,162 @@ def test_float_type_compile(self, type_, sql_text):
     def test_interval_literal_processor(self, type_, expected):
         self.assert_compile(type_, expected, literal_binds=True)
 
+    @testing.only_on("oracle>=23.0")
```
zzzeek: this is a compile test, so we dont need to limit the backend

test/dialect/oracle/test_types.py
```diff
@@ -248,12 +249,162 @@ def test_float_type_compile(self, type_, sql_text):
...
+            Column("x", sqltypes.Boolean),
+            Column("y", oracle.BOOLEAN),
+            Column(
+                "z", sqltypes.Boolean().with_variant(oracle.BOOLEAN, "oracle")
+            ),
+        )
+
+        self.assert_compile(
+            schema.CreateTable(t),
+            "CREATE TABLE t (x SMALLINT, y BOOLEAN, z BOOLEAN)",
```
zzzeek, line 288: this is fine even though not a real-world use case, because yes if you use BOOLEAN it should render that no matter what

lib/sqlalchemy/dialects/oracle/types.py
```diff
@@ -23,6 +23,30 @@
     from ...sql.type_api import _LiteralProcessorType
 
 
+class BOOLEAN(sqltypes.BOOLEAN):
```
zzzeek: so we'll note here that this subclass doesnt do anything except provide a docstring.    the convention here is that we don't add a subclass, we just have the "from sqltypes import BOOLEAN as BOOLEAN" and we export it, rather than having an extra subclass like this.   

So I would remove the extra class here, and for the docs I would add a section underneath "FLOAT / DOUBLE Support and Behaviors[¶](https://docs.sqlalchemy.org/en/20/dialects/oracle.html#float-double-support-and-behaviors)"  called "BOOLEAN support", that can describe what happens here for oracle >< 23.

zzzeek (changes requested):
hi, great job.

this PR can be even simpler, as we dont actually need `class BOOLEAN(sqltypes.BOOLEAN)`, just re-export `sqltypes.BOOLEAN` and add a new doc section for "Boolean support" in oracle.rst.

I think our CI is only up to Oracle 18 right now so will have to see if we can get 23 in there.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision a7a3a61e24b417b530eae9ec39c0353e9e222f9d of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change a7a3a61e24b417b530eae9ec39c0353e9e222f9d: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6466

test/dialect/oracle/test_types.py
ybjeon01: You're right, that was added by accident. I'll remove it.

test/dialect/oracle/test_types.py
ybjeon01: My intention was to verify that Boolean compiles to SMALLINT when `supports_native_boolean = False`. Actually, only the first column is really important - the second and third columns are unnecessary. Should I remove them from this test?

lib/sqlalchemy/dialects/oracle/types.py
ybjeon01: Thanks, I've updated the code as you suggested.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 65b8a2c92f539c57ab6b0b1bf045bd23eedb1695 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 65b8a2c92f539c57ab6b0b1bf045bd23eedb1695 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6466

test/dialect/oracle/test_types.py
zzzeek: i think this is fine because it demonstrates that BOOLEAN is always going to render BOOLEAN even if the dialect doesn't support that.   that's the contract of `LowerCase` vs `UPPERCASE` SQL types

lib/sqlalchemy/dialects/oracle/base.py
```diff
@@ -640,6 +640,47 @@
         ),
     )
 
+Boolean Support
+---------------
+
+.. versionadded:: 2.1
```
sqla-tester, line 646: **Federico Caselli** (CaselIT) wrote:

let's move it at the end of the section

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6466

lib/sqlalchemy/dialects/oracle/base.py
```diff
@@ -2053,6 +2099,7 @@ def initialize(self, connection):
             self.colspecs.pop(sqltypes.Interval)
             self.use_ansi = False
 
+        self.supports_native_boolean = self.server_version_info >= (23,)
```
sqla-tester, line 2103: **Federico Caselli** (CaselIT) wrote:

let's also add a class variable with this flag, likely true by default

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6466

test/dialect/oracle/test_dialect.py
```diff
@@ -740,6 +740,15 @@ def test_use_nchar(self):
         self.assert_compile(Unicode(50), "NVARCHAR2(50)", dialect=dialect)
         self.assert_compile(UnicodeText(), "NCLOB", dialect=dialect)
 
+    def test_native_boolean_flag(self):
+        dialect = self._dialect((19, 0, 0))
+        dialect.initialize(Mock())
+        assert not dialect.supports_native_boolean
```
sqla-tester, line 747: **Federico Caselli** (CaselIT) wrote:

this we could the also call before initialize to check that it's true by default

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6466

sqla-tester (commented):

great work, just a couple of minor things

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6466

lib/sqlalchemy/dialects/oracle/base.py
ybjeon01: Hi @CaselIT,

Thank you for the feedback! I want to make sure I understand your suggestion correctly before making the change.

When you mentioned moving the "Boolean Support" section "at the end of the section," I wanted to clarify which section you're referring to. Currently, the "Boolean Support" section is placed right after "FLOAT / DOUBLE Support and Behaviors", which was the original guidance from @zzzeek.

Could you please clarify where you'd like me to move it? For example, should it be moved to the end of all datatype-related sections (after "VECTOR Datatype")?

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6466

lib/sqlalchemy/dialects/oracle/base.py
sqla-tester: **Michael Bayer** (zzzeek) wrote:

i had them put it underneath the FLOAT/DOUBLE section, if at the bottom it's below VECTOR and index options which are long sections and not as commonly used

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6466

lib/sqlalchemy/dialects/oracle/base.py
sqla-tester: **Michael Bayer** (zzzeek) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6466

test/dialect/oracle/test_dialect.py
sqla-tester: **Michael Bayer** (zzzeek) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6466

lib/sqlalchemy/dialects/oracle/base.py
ybjeon01: Hi @zzzeek,

I noticed some code was added to the PR that I didn't write myself. I'm not very familiar with the Gerrit system, but it appears that maintainers can add commits directly to the PR.

I have a concern about the following code that references cx_Oracle.BOOLEAN:

```python
elif isinstance(type_impl, Boolean):
    if self.dialect.supports_native_boolean:
        out_parameters[name] = self.cursor.var(
            cx_Oracle.BOOLEAN, arraysize=len_params
        )
    else:
        out_parameters[name] = self.cursor.var(
            cx_Oracle.NUMBER,
            arraysize=len_params,
            outconverter=bool,
        )
```
As far as I know, cx_Oracle only supports Oracle Database up to version 19c, while native BOOLEAN support was introduced in Oracle 23ai. Additionally, I couldn't find any mention of BOOLEAN support in the cx_Oracle documentation.

zzzeek: hi @ybjeon01 -

Yes I have added final changes to the code to make sure RETURNING is working as well as made sure Boolean with a former SMALLINT type still works.

The variable you see called cx_Oracle is in fact `self.dialect.dbapi`, which means in the case of using the oracledb dialect, it's using oracledb and not cx_Oracle (the oracledb dialect subclasses the cx_Oracle dialect right now).

Also, both drivers run perfectly fine on Oracle 23ai minus the fact that the free version of 23 needs to have connection retry turned on to be stable (that applies to both drivers).     There is of course no requirement that cx_Oracle be installed to use SQLAlchemy, you can have just oracledb and that's it.   but our test CI is still running both drivers on both DBs.

> Additionally, I couldn't find any mention of BOOLEAN support in the cx_Oracle documentation.

it has it

```
>>> import cx_Oracle
>>> cx_Oracle.BOOLEAN
<cx_Oracle.DbType DB_TYPE_BOOLEAN>

```

anyway thanks for this PR it should be ready to merge now

sqla-tester: **Federico Caselli** (CaselIT) wrote:

I mean versionadded. it's not that important though

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6466

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6466

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6466

lib/sqlalchemy/dialects/oracle/base.py
sqla-tester: **Michael Bayer** (zzzeek) wrote:

oh right.   I tend to put those at the top for a new thing with a big paragraph section, since this is often how the main python docs do it.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6466

sqla-tester: **Federico Caselli** (CaselIT) wrote:

ok then, let's merge it then

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6466

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6466

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6466 has been **merged**. Congratulations! :)

PR #13002, opened 2025-11-23
Use UUID for anonymous label construction

### LINKED ISSUE #12990: anon label logic does not guarantee uniqueness
#### for context:

i noticed following annoying issue on a flask web app running on gunicorn, after upgrading to python3.13 *and* changing the container base distro (so its not quite clear which change caused it)...

when the app starts, **theres a very small chance** that some or all endpoints wont work, 
throwing `sqlalchemy.exc.CompileError: Multiple, unrelated CTEs found with the same name: 'anon_1'`  
~~this issue seems to neither appear nor disappear after the app has already started. looks like it boots into a wrong state.~~ no  
*restarting the app will likely fix this issue.*

since its happening **randomly** and i cannot reproduce it locally, i had to inject into the faulty server process with *pyrasite* for some tricky debugging.

#### findings:

> note: codes here have some sensitive infos stripped...

running the following code in the injected python shell:
```python
import sqlalchemy
import unittest.mock


@classmethod
def CTE_construct(
    cls,
    selectable,
    *,
    name,
    **kw,
):
    obj = cls.__new__(cls)
    print("cte object created:", id(obj))
    obj._init(selectable, name=name, **kw)
    return obj


with unittest.mock.patch.object(sqlalchemy.CTE, "_construct", CTE_construct):
    sqlalchemy.select(...).compile()
```
output:
```
cte object created: 139720728185984
cte object created: 139720728186816
cte object created: 139720728192640
cte object created: 139720728195136
cte object created: 139720728187440
cte object created: 139720728191600
cte object created: 139720728197424
cte object created: 139720728197424  # here!
  File "<console>", line 5, in <module>
  ...
  File ".../lib64/python3.13/site-packages/sqlalchemy/sql/compiler.py", line 5382, in visit_join
    join.left._compiler_dispatch(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        self, asfrom=True, from_linter=from_linter, **kwargs
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/.../lib64/python3.13/site-packages/sqlalchemy/sql/visitors.py", line 138, in _compiler_dispatch
    return meth(self, **kw)  # type: ignore  # noqa: E501
  File "/.../lib64/python3.13/site-packages/sqlalchemy/sql/compiler.py", line 4129, in visit_cte
    raise exc.CompileError(
    ...<2 lines>...
    )
sqlalchemy.exc.CompileError: Multiple, unrelated CTEs found with the same name: 'anon_1'
```

notice that there are two `CTE` created with same `id()`, which causes generated names to collide:

https://github.com/sqlalchemy/sqlalchemy/blob/06a4f84c3e2cd2aebf794f7a4805fadd13c06991/lib/sqlalchemy/sql/selectable.py#L1732

but they are obviously *not* the same object. i have confirmed that the inner selectables are not the same. 

i guess this might be related to some aggressive garbage collections causing memory reuse so:

```python
import sqlalchemy
import unittest.mock

leaker = []


@classmethod
def CTE_construct(
    cls,
    selectable,
    *,
    name,
    **kw,
):
    obj = cls.__new__(cls)
    leaker.append(obj)  # keep a ref to `obj` until compiling finished
    print("cte object created:", id(obj))
    obj._init(selectable, name=name, **kw)
    return obj


try:
    with unittest.mock.patch.object(sqlalchemy.CTE, "_construct", CTE_construct):
        sqlalchemy.select(...).compile()
finally:
    leaker.clear()
```
output:

```
cte object created: 139720728192432
cte object created: 139720728197216
cte object created: 139720727279920
cte object created: 139720727280336
cte object created: 139720727270560
cte object created: 139720727273680
cte object created: 139720727272848
cte object created: 139720727275968
<sqlalchemy.sql.compiler.StrSQLCompiler object at 0x7f13446c6350>
```

which seems to bypass this issue.

----

the app works by generating complex deepnested queries on each incoming request, so its expected to  create and destroy many temp statement objects each time.  
im not quite sure whether this is a sqlalchemy bug or a misuse.  
any advices appreciated.

_Originally posted by @lictex in https://github.com/sqlalchemy/sqlalchemy/discussions/12982_

Context:
- https://github.com/sqlalchemy/sqlalchemy/discussions/12982#discussioncomment-14994881
- https://github.com/sqlalchemy/sqlalchemy/discussions/12982#discussioncomment-14998151
- https://github.com/sqlalchemy/sqlalchemy/discussions/12982#discussioncomment-15003376

ashm-dev:
Fixes #12990

### DIFF
--- a/lib/sqlalchemy/sql/selectable.py
+++ b/lib/sqlalchemy/sql/selectable.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 import collections
 from enum import Enum
 import itertools
-from uuid import uuid4
+from os import urandom
 
 from typing import AbstractSet
 from typing import Any as TODO_Any
@@ -1744,7 +1744,7 @@ class AliasedReturnsRows(NoInit, NamedFromClause):
                 name = getattr(selectable, "name", None)
                 if isinstance(name, _anonymous_label):
                     name = None
-            name = _anonymous_label.safe_construct(uuid4().hex, name or "anon")
+            name = _anonymous_label.safe_construct(urandom(8).hex(), name or "anon")
         self.name = name
 
     def _refresh_for_new_column(self, column: ColumnElement[Any]) -> None:
--- a/lib/sqlalchemy/sql/selectable.py
+++ b/lib/sqlalchemy/sql/selectable.py
@@ -16,6 +16,8 @@ from __future__ import annotations
 import collections
 from enum import Enum
 import itertools
+from uuid import uuid4
+
 from typing import AbstractSet
 from typing import Any as TODO_Any
 from typing import Any
@@ -1742,7 +1744,7 @@ class AliasedReturnsRows(NoInit, NamedFromClause):
                 name = getattr(selectable, "name", None)
                 if isinstance(name, _anonymous_label):
                     name = None
-            name = _anonymous_label.safe_construct(id(self), name or "anon")
+            name = _anonymous_label.safe_construct(uuid4().hex, name or "anon")
         self.name = name
 
     def _refresh_for_new_column(self, column: ColumnElement[Any]) -> None:


CaselIT (changes requested):
I think using os.urandom is better peformance wise

also could we make it so that only cte has this change?

PR #13038, opened 2025-12-16
fix typos

fardyn:
### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**


CaselIT:
wow thanks!

I'll take a look and merge this week!

CaselIT (approved):
thanks for the fixes! will merge later

fardyn:
No problem, 
I also found some words in docs, comments and code need rewording like "non-useful" which is simply "useless", or "deactive" which isn't standard English should be "deactivate",
and also some "doesnt, wont, cant" missing apostrophes

some of them are in the code itself, which is potentially sensitive, like:
DEACTIVE could be DEACTIVATE
```
class SessionTransactionState(_StateChangeState):
    ACTIVE = 1
    PREPARED = 2
    COMMITTED = 3
    DEACTIVE = 4
    CLOSED = 5
    PROVISIONING_CONNECTION = 6

```
or 
delimeter should be delimiter

```
   if order_by is not None:
            return (
                f"group_concat({expr._compiler_dispatch(self, **kw)} "
                f"ORDER BY {order_by._compiler_dispatch(self, **kw)} "
                f"SEPARATOR "
                f"{delimeter._compiler_dispatch(self, **literal_exec)})"
            )

```
can I change them? or just stick to docs and comments?

CaselIT:
So as a quick guideline I would say
- variable names are likely fine, if they do not change the semantic of the code
- argument names are also likely fine in the internal api, while likely not in the public api, or at least they need to be weighted on a case by case basis. 
- for constants such as SessionTransactionState and public api, it's likely best to avoid changing them in bulk, but we may consider them on a case by case basis. 

In the case of SessionTransactionState the proper word is likely inactive there, since DEACTIVATE is not a great sounding state name. it has likely been like that since forever, and since it's mostly in the internal internal api, the effort to change the name is likely not worth the hassle.

fardyn:
> In the case of SessionTransactionState the proper word is likely inactive there, since DEACTIVATE is not a great sounding state name. it has likely been like that since forever, and since it's mostly in the internal internal api, the effort to change the name is likely not worth the hassle.

Yes, INACTIVE is fine and more appropriate.
According to the guideline, I’ll go ahead and edit the code.
Thanks for the quick response.

PR #13042, opened 2025-12-17
improved immutabledict.merge_with and union

### LINKED ISSUE #13043: improve immutabledict.merge_with and union
### Discussed in https://github.com/sqlalchemy/sqlalchemy/discussions/13041

<sup>Originally posted by **LucasMalor** December 17, 2025</sup>
As @zzzeek pointed out [here](https://discuss.python.org/t/pep-814-add-frozendict-built-in-type/104854/109), `merge_with` and `union` methods of `immutabledict` can return an `immutabledict` unchanged when it's possible -- that is, only one of the dicts to be merged is not empty and it's an `immutabledict`.</div>

LucasMalor:
Now if `immutabledict.merge`_with have only one `immutabledict`, it returns that.
Also, `union` returns `other` if `self` is empty and `other` is an `immutabledict`.

Fixes: #13043


lib/sqlalchemy/ext/asyncio/engine.py
```diff
@@ -590,7 +591,8 @@ async def stream(
             statement,
             parameters,
             execution_options=util.EMPTY_DICT.merge_with(
-                execution_options, {"stream_results": True}
+                execution_options,
+                immutabledict(stream_results=True),
```
CaselIT, line 595: why are these changes needed?

lib/sqlalchemy/util/_immutabledict_cy.py
```diff
@@ -146,11 +146,13 @@ def __repr__(self) -> str:
...
     @cython.annotation_typing(False)  # avoid cython crash from generic return
-    def union(
-        self, other: Optional[Mapping[_KT, _VT]] = None, /
-    ) -> immutabledict[_KT, _VT]:
+    def union(self, other: Mapping[_KT, _VT], /) -> immutabledict[_KT, _VT]:
         if not other:
             return self
+
+        if not self and type(other) == immutabledict:
+            return self
```
CaselIT: this should likely return other

lib/sqlalchemy/util/_immutabledict_cy.py
```diff
@@ -163,25 +165,46 @@ def union(
 
     @cython.annotation_typing(False)  # avoid cython crash from generic return
     def merge_with(
```
CaselIT, line 167: the change here doesn't seem correct

CaselIT (changes requested):
Thanks for getting the ball rolling, I've left a few comments

lib/sqlalchemy/ext/asyncio/engine.py
LucasMalor: Because if the execution_options is empty, `merge_with` can't return the other dict as-is if it's not an `immutabledict`.

Side note: the kwargs constructor is used since this way we can avoid the creation of an intermediate `dict`.

lib/sqlalchemy/util/_immutabledict_cy.py
LucasMalor: Yes, of course :) I'll fix it.

lib/sqlalchemy/util/_immutabledict_cy.py
LucasMalor: Well, I suppose that if you use `merge_with`, you want to pass at least one arg. On the contrary, you can simply use `copy`.

lib/sqlalchemy/ext/asyncio/engine.py
CaselIT: it's kinda the same though, since we still have to construct the a new immutable dict. whatever it's not here or in the merge_with it doesn't change much

lib/sqlalchemy/util/_immutabledict_cy.py
CaselIT: ok I guess i misread the proposed change. it could likely be improved.

Some unitest for this should likely be added here for the new behaviours https://github.com/sqlalchemy/sqlalchemy/blob/7194b5ba3de050815ea20aafb80d262b6fcd1d07/test/base/test_utils.py#L362

lib/sqlalchemy/ext/asyncio/engine.py
LucasMalor: > we still have to construct the a new immutable dict

Previously we constructed a `dict`. Since `immutabledict` inherits from `dict`, this new creation has the same speed of `dict(k=v)` (that I suppose is fast enough). If you want, I can do some benchs.

lib/sqlalchemy/util/_immutabledict_cy.py
LucasMalor: Done.

lib/sqlalchemy/ext/asyncio/engine.py
CaselIT: Ok since the idea is that in most cases no other execution options are provided. I guess it makes sense.
I guess at this point we could have the stream only as module variable

CaselIT:
let me know if you plan on continuing working on this. I can take over and make it mergeable

LucasMalor:
Yes, I have only to find the time :)

LucasMalor:
As requested, I added 3 unit tests and modified one already existing. This way they will cover the new code added for merge_with and union.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision bd53e488432edd5986c28f196b0363b976b26b04 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change bd53e488432edd5986c28f196b0363b976b26b04: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

CaselIT:
I've done a pass on gerrit with a few other improvements. I also tried using a single for, but the complexity was not worth it:
```py
@cython.annotation_typing(False)  # avoid cython crash from generic return
    def merge_with(
        self,
        *dicts: Optional[Mapping[_KT, _VT]],
    ) -> immutabledict[_KT, _VT]:
        size: int = len(dicts)
        if size==0:
            return self
        if size == 1:
            return self.union(dicts[0])
        
        result: Optional[immutabledict[_KT, _VT]] = None
        pos_one = -1 if self else -2
        pos_two = -2
        for i in range(size):
            d = dicts[i]
            if not d:
                continue
            if result is not None:
                _update(result, d)
            else:
                if pos_one == -2:
                    pos_one = i
                elif pos_two == -2:
                    pos_two = i
                else:
                    result = immutabledict()
                    if pos_one == -1:
                        PyDict_Update(result, self)
                    else:
                        _update(result, dicts[pos_one])
                    _update(result, dicts[pos_two])
                    _update(result, d)

        if result is not None:
            return result
        if pos_one == -2:
            return self # self is empty
        if pos_one == -1:
            if pos_two == -2:
                return self
            else:
                return self.union(dicts[pos_two])

        if pos_two == -2:
            d= dicts[pos_one]
            if isinstance(d, immutabledict):
                return d
            else:
                result = immutabledict()
                _update(result, d)
                return result
        else:
            d1 = dicts[pos_one]
            d2 = dicts[pos_two]
            result = immutabledict()
            _update(result, d1)
            _update(result, d2)
            return result
```

it's not significantly faster

lib/sqlalchemy/util/_immutabledict_cy.py
```diff
@@ -163,25 +165,46 @@ def union(
...
+                continue
+
+            if type(d) == immutabledict:
+                if only_one is None:
+                    only_one = d
                 else:
-                    dict.update(result, d)
+                    only_one = None
+                    break
+            else:
```
sqla-tester, line 187: **LucasMalor** (LucasMalor) wrote:

I think it's better to cycle over dicts instead of the index, it's more pythonic. If we need the index, we can use enumerate.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

sqla-tester (commented):
**LucasMalor** (LucasMalor) wrote:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

* lib/sqlalchemy/util/_immutabledict_cy.py (line 162): It's possible to simplify it. You need only another if and no else, since there's a return in every branch.

sqla-tester:

Thanks for the review. Will update

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

lib/sqlalchemy/util/_immutabledict_cy.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

Sure, but this gets compiled in cython and a for over indexes is generally faster

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

lib/sqlalchemy/util/_immutabledict_cy.py
```diff
@@ -146,11 +146,13 @@ def __repr__(self) -> str:
         return f"immutabledict({dict.__repr__(self)})"
 
     @cython.annotation_typing(False)  # avoid cython crash from generic return
-    def union(
-        self, other: Optional[Mapping[_KT, _VT]] = None, /
-    ) -> immutabledict[_KT, _VT]:
+    def union(self, other: Mapping[_KT, _VT], /) -> immutabledict[_KT, _VT]:
         if not other:
```
sqla-tester, line 150: **Michael Bayer** (zzzeek) wrote:

if we're coming in and trying to improve the implementation, why does union() and merge_with() need different implementations at all?    Per the discussion re: frozendict it looks like we should have simply one method union() that accepts multiple mappings

i mean cython is supposed to be somewhat close to C code, so certainly we can make just one implementation that performs great?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

sqla-tester (commented):

why dont we make this a little simpler and unify merge_with() and union() ?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

sqla-tester:

will update

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

lib/sqlalchemy/util/_immutabledict_cy.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

yes makes sense. I'll update

- merge and union will be synonym
- they both call into a single impl

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

lib/sqlalchemy/ext/asyncio/engine.py
LucasMalor: > I guess at this point we could have the stream only as module variable

I missed that. What do you mean?

CaselIT: It's updated on Gerrit

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

* lib/sqlalchemy/util/_immutabledict_cy.py (line 162): Done

lib/sqlalchemy/util/_immutabledict_cy.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

lib/sqlalchemy/util/_immutabledict_cy.py
```diff
@@ -163,25 +165,46 @@ def union(
...
+            if type(d) == immutabledict:
+                if only_one is None:
+                    only_one = d
                 else:
-                    dict.update(result, d)
+                    only_one = None
+                    break
+            else:
+                only_one = None
+                break
```
sqla-tester, line 189: **Michael Bayer** (zzzeek) wrote:

I think there's still a path here where we make a new dict that we dont need to (this is also a common case):

      d = util.EMPTY_DICT.merge_with(util.EMPTY_DICT, util.EMPTY_DICT)
      
we start with only_one=None, self_is_empty=True.  then into the loop through "others", we never modify "only_one" because "if not d: continue"

now we're out of the loop, only_one=None.   we make an immutabledict().   when we should be returning self.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

sqla-tester (commented):

can you confirm my reading of the logic?   I think this creates a new dict:

   d = util.EMPTY_DICT.merge_with(util.EMPTY_DICT)
   
or if not, the logic is at least pretty hard to read then :)

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

* lib/sqlalchemy/util/_immutabledict_cy.py (line 162): perfect (there's just one implementation)

sqla-tester:

drat, it works fine. why can't I read this logic?

```
>>> d = util.EMPTY_DICT.merge_with(util.EMPTY_DICT)
>>> d is util.EMPTY_DICT
True
```

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

sqla-tester:

oh no, i'm right.  it doesnt work.  will add a test

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

lib/sqlalchemy/util/_immutabledict_cy.py
sqla-tester: **Michael Bayer** (zzzeek) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

sqla-tester:

OK, I've fixed that, let's stare at this very hard

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

sqla-tester:

claude gave it the thumbs up after much doubting

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

sqla-tester:

Yes it wasn't token into consideration. Your impl seeems fine

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

lib/sqlalchemy/util/_immutabledict_cy.py
```diff
@@ -163,25 +165,46 @@ def union(
...
 
-        return self if result is None else result
+        if only_one is not None:
```
sqla-tester, line 191: **LucasMalor** (LucasMalor) wrote:

It's a bit hard for me to follow the changes on Gerrit, but I think there's a problem here. If you have, for example, `x = EMPTY_DICT.merge_with(immd, {})`, it seems to me that `x is immd` will be False.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

sqla-tester (commented):
**LucasMalor** (LucasMalor) wrote:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

lib/sqlalchemy/util/_immutabledict_cy.py
sqla-tester: **Michael Bayer** (zzzeek) wrote:

this test passes:

    @methods
    def test_only_one_non_empty_no_copy(self, method):
        d = util.immutabledict()
        e = util.immutabledict({1: 2})
        d2 = method(d, e, {})
        eq_(d2, {1: 2})
        is_(d2, e)
        


steps:

1. self resolves to to false, so only_one = False, self_is_empty = True
2. loop begins through others
3. first element in the loop is "immd" (or "e" in my example above)
4. if only_one is False and isinstance(d, immutabledict) -> True
5. only_one = d  (which in this case is "immd" / "e")
6. second element in the loop is empty plain dictionary
7. "if not d: continue" -> loop continues on next element.   
8. no more elements, out of loop, only_one == the single non-empty dictionary
9. if only_one is False -> False
10. if only_one is not None: -> True
11. only_one is returned


the "if not d:" part is what prevents us from considering any other immutable / plain dict that is empty

this same test is present in the "def test_no_change_everyone_empty" test case

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6532 has been **merged**. Congratulations! :)

PR #13053, opened 2025-12-24
Add `deprecated` parameter to SQLAlchemy Column

zifo-10:
- Introduce an optional `deprecated` flag to the Column class to mark columns that are planned for removal.
- `deprecated=True` signals the team to avoid using the column.

### Description
This pull request adds a new optional parameter `deprecated` to the SQLAlchemy `Column` class.

- `deprecated` can be set to `True` or a string message describing why the column is deprecated.
- Columns marked as deprecated serve as a signal to developers to avoid using them.
- This feature is metadata-only and does not affect database behavior.
- Existing Column usage remains fully backward compatible.

Example usage:

```python
from sqlalchemy import Column, Integer, String

Base = declarative_base()

class User(Base):
    __tablename__ = "users"

    name = Column(String(255))
    old_id = Column(Integer, deprecated=True)  # deprecated column

### DIFF
--- a/lib/sqlalchemy/sql/compiler.py
+++ b/lib/sqlalchemy/sql/compiler.py
@@ -2671,6 +2671,13 @@ class SQLCompiler(Compiled):
         if name is None:
             name = self._fallback_column_name(column)
 
+        if getattr(column, "deprecated", False):
+            if isinstance(column.deprecated, str):
+                msg = column.deprecated
+            else:
+                msg = f"The column '{column.name}' is deprecated."
+            util.warn_deprecated(msg, "2.1")
+
         is_literal = column.is_literal
         if not is_literal and isinstance(name, elements._truncated_label):
             name = self._truncated_identifier("colident", name)
--- a/lib/sqlalchemy/sql/schema.py
+++ b/lib/sqlalchemy/sql/schema.py
@@ -1666,6 +1666,7 @@ class Column(DialectKWArgs, SchemaItem, ColumnClause[_T]):
         system: bool = False,
         comment: Optional[str] = None,
         insert_sentinel: bool = False,
+        deprecated: Optional[Union[bool, str]] = None,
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by zzzeek** [comment]
hey there-

`Column` has an existing collection [.info](https://docs.sqlalchemy.org/en/20/orm/mapping_api.html#sqlalchemy.orm.column_property.params.info) where application-specific things like this should be placed:

```
Column("mycolumn", Integer, info={"deprecated": True})
```

additionally, `util.warn_deprecated` is specific to things that are deprecated **in sqlalchemy**, not in your application, so for warnings, you'd want to use your own `warnings.warn("message", DeprecationWarning)`.     `Column` objects can also be used in lots of ways, not just in the `visit_column()` of compiler, so that seems like a pretty arbitrary place to put it (for example, if a mapped column referred to it, then someone did `select(MyClass)`, they get that warning, even though they never referred to that column!     There's not really a great place to put a deprecation warning for this kind of thing, but you can get your compiler version of it using [compiles](https://docs.sqlalchemy.org/en/20/core/compiler.html)

here's how this should be done

```py

import warnings

from sqlalchemy import Column
from sqlalchemy import select
from sqlalchemy.ext.compiler import compiles


@compiles(Column)
def _warn_for_column(element, compiler, **kw):
    if element.info.get("deprecated", False):
        warnings.warn(
            f"Column {element.name!r} is deprecated", DeprecationWarning
        )
    return compiler.visit_column(element, **kw)


class Base(DeclarativeBase):
    pass


class A(Base):
    __tablename__ = "a"

    data: Mapped[str]
    old: Mapped[str] = mapped_column(info={"deprecated": True})


e = create_engine("sqlite://", echo=True)
Base.metadata.create_all(e)

with Session(e) as sess:
    sess.execute(select(A))

```

that will get you the warning "DeprecationWarning: Column 'old' is deprecated" you're looking for.

hope this helps

PR #13074, opened 2026-01-12
correct mariadb sequence behavior when cycle=False - fixes #13073

### LINKED ISSUE #13073: Sequence are wrongly created for MariaDB when using cycle=False
### Describe the bug

Sequence are wrongly created for MariaDB when using cycle=False, creating SQL like 
"CREATE SEQUENCE ab_permission_id_seq INCREMENT BY 1 START WITH 1 MINVALUE 1 NO CYCLE"
resulting in Error
sqlalchemy.exc.ProgrammingError: (pymysql.err.ProgrammingError) (1064, "You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'CYCLE' at line 1")

A PR will be created just after, just ignoring cycle=False (like cycle=None) for MariaDB

_No response_

### SQLAlchemy Version in Use

2.0.45

pymysql

MariaDB 12.2

### Python Version

3.11

### Operating system

ubuntu

### To Reproduce

```python

class User(Base):
    __tablename__ = 'users'

    id = Column(Integer, Sequence('user_id_seq', cycle=False), primary_key=True)
    name = Column(String(50))
    email = Column(String(100))
```

### Error

```
# Copy the complete stack trace and error message here, including SQL log output if applicable.
``` 

### Additional context

_No response_

rusher:
correct mariadb sequence behavior when cycle=False is set, default implementation add "NO CYCLE"  that doesn't exists in mariadb

Fixes: #13073

### Description
provide dedicated get_identity_options method for mariadb 

This pull request is:

- [X] A short code fix


lib/sqlalchemy/dialects/mysql/mariadb.py
```diff
@@ -64,13 +65,36 @@ def native(self) -> bool:  # type: ignore[override]
...
-    def bind_processor(self, dialect: MariaDBDialect) -> Optional[_BindProcessorType[_UUID_RETURN]]:  # type: ignore[override] # noqa: E501
+    def bind_processor(self, dialect: "MariaDBDialect") -> Optional[_BindProcessorType[_UUID_RETURN]]:  # type: ignore[override] # noqa: E501
         if not dialect.supports_native_uuid or not dialect._allows_uuid_binds:
             return super().bind_processor(dialect)  # type: ignore[return-value] # noqa: E501
         else:
             return None
 
 
+class MariaDBDDLCompiler(MySQLDDLCompiler):
+    def get_identity_options(self, identity_options):
```
zzzeek, line 76: this is repeating just about all the code from the original method.   I think it would be better to do this:

```py
text = super().get_identity_options(identity_options)
text = text.sub("NO CYCLE", "")
return text
```

zzzeek (changes requested):
suggested a shorter version of the change.

also this needs tests for mariadb, as im not sure what other databases support CYCLE on sequences this can be mariadb-only for now, a round trip test in test/dialect/mysql/test_query.py should do it (test with all supported flags on and off, make sure sequence created successfully)

CaselIT (changes requested):
This needs to be similar to oracle: https://github.com/sqlalchemy/sqlalchemy/blob/cc0cef70d142a030d79ce08bf98b38d9ea3689cf/lib/sqlalchemy/dialects/oracle/base.py#L1919

replace `NO CYCLE` with `NOCYCLE`: https://mariadb.com/docs/server/reference/sql-structure/sequences/create-sequence

All other sequence options seem ok

Also please revert the other changes

rusher:
in fact, yes, it must be similar to oracle, i'll test it and modify then.
for the other correction, i'll revert it even if i prefer having good type.

lib/sqlalchemy/dialects/mysql/mariadb.py
```diff
@@ -101,7 +126,7 @@ def initialize(self, connection: Connection) -> None:
         )
 
 
-def loader(driver: str) -> Callable[[], type[MariaDBDialect]]:
+def loader(driver: str) -> type[MariaDBDialect]:
```
CaselIT, line 112: this one seems indeed fine. For the other we don't really run type checks on this module

rusher:
corrected as indicated, added simple test case

rusher:
ah, since it's 2.1 beta version, must i create a separate patch in 2.0 ?

CaselIT:
> ah, since it's 2.1 beta version, must i create a separate patch in 2.0 ?

no, we usually do that in gerrit directly. thanks

CaselIT (approved):
thanks for the update!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision c91ef953f4363a754beada8d76f88ada71b4581c of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change c91ef953f4363a754beada8d76f88ada71b4581c: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6554

rusher:
ah, one thing remains, the runtime mariadb check that is not update since some time, creating difference compare to using "mariadb" prefix directly. 

since it contains correction for sequence as well, must i add this patch 
[patch.patch](https://github.com/user-attachments/files/24593641/Add_MariaDB-specific_compiler_and_UUID_support_to_runtime_detection__When_MariaDB_is_detec.patch)
directly or create another issue, and just add the sequence patch here for now ?

zzzeek:
> ah, one thing remains, the runtime mariadb check that is not update since some time, creating difference compare to using "mariadb" prefix directly.
> 
> since it contains correction for sequence as well, must i add this patch [patch.patch](https://github.com/user-attachments/files/24593641/Add_MariaDB-specific_compiler_and_UUID_support_to_runtime_detection__When_MariaDB_is_detec.patch) directly or create another issue, and just add the sequence patch here for now ?

That looks like a big new change / problem, I'd need to look at what's happening there, why is UUID involved here?

This seems to be the fact that MariadbTypeCompiler, and then this new MariadbDDLCompiler, only takes effect if the URL is "mariadb+driver://".       So there is precedent for setting these things up in _is_mariadb() but if these are bugs, they need to be exercised in tests that show the incorrect behavior right now; test_runtime_mariadb_detection_complete is not bad in theory but doesnt really show what problem is being solved.    We have UUID tests for mariadb right now.   Why do they pass?

Additionally there's other things wrong with that patch, imports are in the wrong place, MariaDB tests dont go into test/sql/test_sequences.py.     

Overall that patch points to things that should be looked at but it turns this all into a much bigger change.

CaselIT:
the TLDR is, please open another issue to explain the problem :)

rusher:
here is the issue desription : https://github.com/sqlalchemy/sqlalchemy/issues/13076

test/sql/test_sequences.py
```diff
@@ -772,3 +772,45 @@ def test_drop_ordering_single_table(self):
                 ),
             )
         )
+
+
+class MariaDBSequenceDDLTest(fixtures.TestBase, testing.AssertsCompiledSQL):
```
sqla-tester: **Michael Bayer** (zzzeek) wrote:

these tests should be in test/dialect/mysql/

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6554

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6554

rusher:
i've move the test, and change so connection using mysql:// prefix runtime detection get that change too.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision ead18a04018db6d574a3bc4bd71f21c23256737c of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset ead18a04018db6d574a3bc4bd71f21c23256737c added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6554

sqla-tester:

this is in good shape, I'm going to adapt it for main/2.0 in a common way, then for main put https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6559 on top of it which is a more complete code separation of the mariadb/mysql things.   it does feature transparent switching, as it would be too disruptive to try to separate these dialects fully.   thanks for the PR !

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6554

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6554

* test/sql/test_sequences.py (line 777): Done

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6554 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6561 has been **merged**. Congratulations! :)

Issue #13076, opened 2026-01-13
Maridb legacy behavior with 'mysql' prefix differ from directly setting 'mariadb' prefix


rusher:
### Describe the bug

# Runtime MariaDB Detection (`_set_mariadb()`) Does Not Configure All MariaDB-Specific Features

## Summary

When connecting to MariaDB using the `mysql://` URL prefix (which triggers runtime MariaDB detection via `_set_mariadb()`), several critical MariaDB-specific features are not properly configured. This results in **actual functional limitations** where MariaDB-specific types cannot be used and DDL syntax differs from the proper MariaDB dialect.

## Environment

- **SQLAlchemy Version**: 2.0.46 + https://github.com/sqlalchemy/sqlalchemy/pull/13074
- **MariaDB Version**: 12.3.0
- **Driver**: mariadbconnector
- **Python Version**: 3.12

## Problem Description

The `MySQLDialect._set_mariadb()` method is called when MariaDB is detected at connection time (when using `mysql://` prefix). However, this method does **not** set up all the MariaDB-specific components that `MariaDBDialect` provides, leading to missing functionality.

### Missing Components in `_set_mariadb()`:

1. ❌ **MariaDBTypeCompiler** - Not set (stays as MySQLTypeCompiler)
2. ❌ **MariaDBDDLCompiler** - Not set (stays as MySQLDDLCompiler)  
3. ❌ **supports_native_uuid** - Not set (stays False even for MariaDB 10.7+)
4. ❌ **UUID colspec** - Not updated (missing _MariaDBUUID for mariadbconnector)

### Critical Issue: INET4/INET6 Types Cannot Be Used

When using `mysql://` prefix with MariaDB, attempting to use MariaDB-specific types (INET4, INET6) results in a compilation error:

```python
from sqlalchemy import create_engine, Column, Integer, Table, MetaData
from sqlalchemy.dialects.mysql import mariadb

# Using mysql:// prefix
engine = create_engine("mysql+mariadbconnector://root@localhost/testdb")
metadata = MetaData()

table = Table(
    'test_inet4',
    metadata,
    Column('id', Integer, primary_key=True),
    Column('ip_addr', mariadb.INET4())
)

# This FAILS with:
# CompileError: Compiler <sqlalchemy.dialects.mysql.base.MySQLTypeCompiler> 
# can't render element of type INET4
metadata.create_all(engine)
```

**The same code works perfectly with `mariadb://` prefix.**

[...]

PR #13077, opened 2026-01-14
Configure MariaDB-specific type compiler and UUID support in _set_mariadb() Fixes: #13076

rusher:
### Description
When connecting to MariaDB using the mysql:// URL prefix (which triggers runtime MariaDB detection via _set_mariadb()), several critical MariaDB-specific features are not properly configured. This results in actual functional limitations where MariaDB-specific types cannot be used and DDL syntax differs from the proper MariaDB dialect.

This ensures that MariaDB-specific types work correctly regardless of whether the connection uses mysql:// or mariadb:// URL prefix

### Checklist

This pull request is:

- [x] A short code fix

### DIFF
--- a/lib/sqlalchemy/dialects/mysql/base.py
+++ b/lib/sqlalchemy/dialects/mysql/base.py
@@ -1135,6 +1135,7 @@ from ...sql import roles
 from ...sql import sqltypes
 from ...sql import util as sql_util
 from ...sql import visitors
+from ...sql.sqltypes import Uuid
 from ...sql.compiler import InsertmanyvaluesSentinelOpts
 from ...sql.compiler import SQLCompiler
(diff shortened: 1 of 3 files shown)


zzzeek:
hi -

As mentioned, I did a more deep reaching approach in https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6559 that changes the inheritance architecture of the MySQL classes.    but the end effect is the same, it doesn't really matter if the URL is `mysql+` or `mariadb+` now.    if and when the two DBs really begin to break inter-compatibility it should be easier to revisit.

the NOCYCLE thing is in 2.1 / 2.0 so will be in the next 2.0 release.  thanks for your help.

PR #13117, opened 2026-02-10
Add frame exclusion support for window functions

### LINKED ISSUE #11671: Workaround/support for frame exclusion in PostgreSQL's window function
### Describe the use case

frame exclusion reference: https://www.postgresql.org/docs/current/sql-expressions.html#:~:text=The%20frame_exclusion%20option%20allows%20rows,ordering%20peers%20from%20the%20frame.

I had a hard time trying to figure out how to implement PostgreSQL's frame exclusion when using window function, without resorting to building raw sql string. This is roughly what I've been working with (In my use case, `text` construct are used to build fragments of a query): 

```python
def demo(db):
    window_args = {"partition_by": text("group_id"), "order_by": text("created_at"), "range_": (None, 0)}
    query = sa.select(
        func.sum(sa.text("price")).over(**window_args),
        # this translates to -> SUM(price) PARTITION BY group_id ORDER BY created_at RANGE UNBOUNDED PRECEDING AND CURRENT ROW
        # but I want to it to be   .....RANGE UNBOUNDED PRECEDING EXCLUDE CURRENT ROW"
        sa.column("other_column"),
    )
    for row in db.execute(query):
        print(row)

```
Currently,  it looks like the parameters in `over()` does not support frame exclusion.  What would be a reliable workaround for this ?  Patching `SQLCompiler._format_frame_clause` may work, but I want to seek out other "safer" options first.

Postgres 11 or higher

### Example Use

If it will be supported in future, it may look something like this:
```python
column("foo").over(exclude="CURRENT ROW") # other values:  GROUP, TIES
```

### Additional context

_No response_

veeceey:
## Summary

Adds the `exclude` parameter to the `Over` construct and all `.over()` methods, enabling SQL standard frame exclusion clauses (`EXCLUDE CURRENT ROW`, `EXCLUDE GROUP`, `EXCLUDE TIES`, `EXCLUDE NO OTHERS`) in window functions.

Closes #11671

## Changes

- **`lib/sqlalchemy/sql/elements.py`**: Added `exclude` attribute to `Over` class with `dp_string` in `_traverse_internals` for proper cache key generation. Added validation in `__init__` that checks the value against a regex and requires a frame spec (`rows`, `range_`, or `groups`). Added `exclude` param to `AggregateOrderBy.over()` and `FunctionFilter.over()`.
- **`lib/sqlalchemy/sql/compiler.py`**: Added `WINDOW_EXCLUDE` regex for compile-time validation (matching the `FK_ON_DELETE`/`FK_ON_UPDATE` pattern). Updated `visit_over` to append `EXCLUDE <value>` after the frame clause.
- **`lib/sqlalchemy/sql/functions.py`**: Added `exclude` param to `FunctionElement.over()`.
- **`lib/sqlalchemy/sql/_elements_constructors.py`**: Added `exclude` param to standalone `over()` function with docstring.
- **`test/sql/test_compiler.py`**: Added `test_over_frame_exclude` (5 compile tests covering rows/range_/groups/NO OTHERS/case insensitivity) and `test_over_frame_exclude_invalid` (invalid value + missing frame spec).

## Example usage

```python
func.sum(column("price")).over(
    partition_by=text("group_id"),
    order_by=text("created_at"),
    range_=(None, 0),
    exclude="current row",
)
# -> SUM(price) OVER (PARTITION BY group_id ORDER BY created_at
#    RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW EXCLUDE CURRENT ROW)
```

## Implementation notes

- Follows the `ForeignKey` `ondelete`/`onupdate` pattern: accepts a plain string, validates against a regex at construction time, and renders directly in the compiler.
- The `exclude` value is part of `_traverse_internals` as `dp_string`, ensuring distinct cache keys for different exclusion values.
- Input is case-insensitive (normalized to uppercase on construction).
- Requires a frame specification, raising `ArgumentError` if `exclude` is provided without `rows`, `range_`, or `groups`.

## Test plan

- [x] All 5 new compile tests pass for each valid exclusion type
- [x] Error tests pass for invalid values and missing frame spec
- [x] All 422 existing tests in `test_compiler.py` pass (no regressions)
- [x] All 182 tests in `test_external_traversal.py` pass (cache key integrity)
- [x] All 148 tests in `test_functions.py` pass

### DIFF
--- a/lib/sqlalchemy/sql/_elements_constructors.py
+++ b/lib/sqlalchemy/sql/_elements_constructors.py
@@ -1621,6 +1621,7 @@ def over(
     range_: _FrameIntTuple | FrameClause | None = None,
     rows: _FrameIntTuple | FrameClause | None = None,
     groups: _FrameIntTuple | FrameClause | None = None,
+    exclude: Optional[str] = None,
 ) -> Over[_T]:
     r"""Produce an :class:`.Over` object against a function.
 
@@ -1725,6 +1726,14 @@ def over(
 
      .. versionadded:: 2.0.40
 
+    :param exclude: optional string for the frame exclusion clause.
+     This is a string value which can be one of ``CURRENT ROW``,
+     ``GROUP``, ``TIES``, or ``NO OTHERS`` and will render an
+     EXCLUDE clause within the window frame specification.  Requires
+     that one of ``rows``, ``range_``, or ``groups`` is also specified.
+
+     .. versionadded:: 2.1
+
     This function is also available from the :data:`~.expression.func`
     construct itself via the :meth:`.FunctionElement.over` method.
 
@@ -1737,7 +1746,10 @@ def over(
         :func:`_expression.within_group`
 
     """  # noqa: E501
-    return Over(element, partition_by, order_by, range_, rows, groups)
+    return Over(
+        element, partition_by, order_by,
+        range_, rows, groups, exclude=exclude,
+    )
 
 
 @_document_text_coercion("text", ":func:`.text`", ":paramref:`.text.text`")
--- a/lib/sqlalchemy/sql/_elements_constructors.py
+++ b/lib/sqlalchemy/sql/_elements_constructors.py
@@ -1621,7 +1621,7 @@ def over(
     range_: _FrameIntTuple | FrameClause | None = None,
     rows: _FrameIntTuple | FrameClause | None = None,
     groups: _FrameIntTuple | FrameClause | None = None,
-    exclude: Optional[str] = None,
+    exclude: str | None = None,
 ) -> Over[_T]:
     r"""Produce an :class:`.Over` object against a function.
 
@@ -1747,8 +1747,13 @@ def over(
 
     """  # noqa: E501
     return Over(
-        element, partition_by, order_by,
-        range_, rows, groups, exclude=exclude,
+        element,
+        partition_by,
+        order_by,
+        range_,
+        rows,
+        groups,
+        exclude,
     )
 
 
--- a/lib/sqlalchemy/sql/compiler.py
+++ b/lib/sqlalchemy/sql/compiler.py
@@ -2991,9 +2991,7 @@ class SQLCompiler(Compiled):
             range_ = None
 
         if range_ is not None and over.exclude is not None:
-            range_ += " EXCLUDE %s" % self.preparer.validate_sql_phrase(
-                over.exclude, _WINDOW_EXCLUDE_RE
-            )
+            range_ += f" EXCLUDE {self.preparer.validate_sql_phrase(over.exclude, _WINDOW_EXCLUDE_RE)}"
 
         return "%s OVER (%s)" % (
             text,
--- a/lib/sqlalchemy/sql/compiler.py
+++ b/lib/sqlalchemy/sql/compiler.py
@@ -262,6 +262,9 @@ FK_ON_UPDATE = re.compile(
     r"^(?:RESTRICT|CASCADE|SET NULL|NO ACTION|SET DEFAULT)$", re.I
 )
 FK_INITIALLY = re.compile(r"^(?:DEFERRED|IMMEDIATE)$", re.I)
+WINDOW_EXCLUDE = re.compile(
+    r"^(?:CURRENT ROW|GROUP|TIES|NO OTHERS)$", re.I
+)
 BIND_PARAMS = re.compile(r"(?<![:\w\$\x5c]):([\w\$]+)(?![:\w\$])", re.UNICODE)
 BIND_PARAMS_ESC = re.compile(r"\x5c(:[\w\$]*)(?![:\w\$])", re.UNICODE)
 
@@ -2986,6 +2989,9 @@ class SQLCompiler(Compiled):
             range_ = f"GROUPS BETWEEN {self.process(over.groups, **kwargs)}"
         else:
             range_ = None
+
+        if range_ is not None and over.exclude is not None:
+            range_ += f" EXCLUDE {over.exclude}"
 
         return "%s OVER (%s)" % (
             text,
--- a/lib/sqlalchemy/sql/compiler.py
+++ b/lib/sqlalchemy/sql/compiler.py
@@ -262,7 +262,7 @@ FK_ON_UPDATE = re.compile(
     r"^(?:RESTRICT|CASCADE|SET NULL|NO ACTION|SET DEFAULT)$", re.I
 )
 FK_INITIALLY = re.compile(r"^(?:DEFERRED|IMMEDIATE)$", re.I)
-WINDOW_EXCLUDE = re.compile(
+_WINDOW_EXCLUDE_RE = re.compile(
     r"^(?:CURRENT ROW|GROUP|TIES|NO OTHERS)$", re.I
 )
 BIND_PARAMS = re.compile(r"(?<![:\w\$\x5c]):([\w\$]+)(?![:\w\$])", re.UNICODE)
@@ -2991,7 +2991,9 @@ class SQLCompiler(Compiled):
             range_ = None
 
         if range_ is not None and over.exclude is not None:
-            range_ += f" EXCLUDE {over.exclude}"
+            range_ += " EXCLUDE %s" % self.preparer.validate_sql_phrase(
+                over.exclude, _WINDOW_EXCLUDE_RE
+            )
 
         return "%s OVER (%s)" % (
             text,
--- a/lib/sqlalchemy/sql/compiler.py
+++ b/lib/sqlalchemy/sql/compiler.py
@@ -2991,7 +2991,9 @@ class SQLCompiler(Compiled):
             range_ = None
 
         if range_ is not None and over.exclude is not None:
-            range_ += f" EXCLUDE {self.preparer.validate_sql_phrase(over.exclude, _WINDOW_EXCLUDE_RE)}"
+            range_ += " EXCLUDE " + self.preparer.validate_sql_phrase(
+                over.exclude, _WINDOW_EXCLUDE_RE
+            )
 
         return "%s OVER (%s)" % (
             text,
--- a/lib/sqlalchemy/sql/elements.py
+++ b/lib/sqlalchemy/sql/elements.py
@@ -4575,6 +4575,14 @@ class Over(ColumnElement[_T]):
 
         self.exclude = exclude
 
+        if exclude is not None and (
+            range_ is None and rows is None and groups is None
+        ):
+            raise exc.ArgumentError(
+                "'exclude' requires that one of 'rows', "
+                "'range_', or 'groups' is also specified"
+            )
+
     if not TYPE_CHECKING:
 
         @util.memoized_property
--- a/lib/sqlalchemy/sql/elements.py
+++ b/lib/sqlalchemy/sql/elements.py
@@ -4528,6 +4528,7 @@ class Over(ColumnElement[_T]):
         ("range_", InternalTraversal.dp_clauseelement),
         ("rows", InternalTraversal.dp_clauseelement),
         ("groups", InternalTraversal.dp_clauseelement),
+        ("exclude", InternalTraversal.dp_string),
     ]
 
     order_by: Optional[ClauseList] = None
@@ -4540,6 +4541,7 @@ class Over(ColumnElement[_T]):
     range_: FrameClause | None
     rows: FrameClause | None
     groups: FrameClause | None
+    exclude: str | None
 
     def __init__(
         self,
@@ -4549,6 +4551,7 @@ class Over(ColumnElement[_T]):
         range_: _FrameIntTuple | FrameClause | None = None,
         rows: _FrameIntTuple | FrameClause | None = None,
         groups: _FrameIntTuple | FrameClause | None = None,
+        exclude: Optional[str] = None,
     ):
         self.element = element
         if order_by is not None:
@@ -4569,6 +4572,23 @@ class Over(ColumnElement[_T]):
             self.range_ = FrameClause._parse(range_, coerce_int=False)
             self.rows = FrameClause._parse(rows, coerce_int=True)
             self.groups = FrameClause._parse(groups, coerce_int=True)
+
+        if exclude is not None:
+            exclude = exclude.upper()
+            from .compiler import WINDOW_EXCLUDE
+
+            if not WINDOW_EXCLUDE.match(exclude):
+                raise exc.ArgumentError(
+                    "Invalid window frame exclusion '%s'. Valid values "
+                    "are CURRENT ROW, GROUP, TIES, NO OTHERS" % exclude
+                )
+            if self.range_ is None and self.rows is None \
+                    and self.groups is None:
+                raise exc.ArgumentError(
+                    "exclude requires that rows, range_ or groups "
+                    "is specified"
+                )
+        self.exclude = exclude
 
     if not TYPE_CHECKING:
 
@@ -4804,6 +4824,7 @@ class AggregateOrderBy(WrapsColumnExpression[_T]):
         rows: _FrameIntTuple | FrameClause | None = None,
         range_: _FrameIntTuple | FrameClause | None = None,
         groups: _FrameIntTuple | FrameClause | None = None,
+        exclude: Optional[str] = None,
     ) -> Over[_T]:
         """Produce an OVER clause against this :class:`.WithinGroup`
         construct.
@@ -4819,6 +4840,7 @@ class AggregateOrderBy(WrapsColumnExpression[_T]):
             range_=range_,
             rows=rows,
             groups=groups,
+            exclude=exclude,
         )
 
     @overload
@@ -4948,6 +4970,7 @@ class FunctionFilter(Generative, ColumnElement[_T]):
         range_: _FrameIntTuple | FrameClause | None = None,
         rows: _FrameIntTuple | FrameClause | None = None,
         groups: _FrameIntTuple | FrameClause | None = None,
+        exclude: Optional[str] = None,
     ) -> Over[_T]:
         """Produce an OVER clause against this filtered function.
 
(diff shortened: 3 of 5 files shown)


lib/sqlalchemy/sql/elements.py
```diff
@@ -4570,6 +4573,23 @@ def __init__(
             self.rows = FrameClause._parse(rows, coerce_int=True)
             self.groups = FrameClause._parse(groups, coerce_int=True)
 
+        if exclude is not None:
+            exclude = exclude.upper()
+            from .compiler import WINDOW_EXCLUDE
```
CaselIT: I don't think it makes much sense to have the regexp in compiler if it's used only here.

lib/sqlalchemy/sql/elements.py
```diff
@@ -4570,6 +4573,23 @@ def __init__(
             self.rows = FrameClause._parse(rows, coerce_int=True)
             self.groups = FrameClause._parse(groups, coerce_int=True)
 
+        if exclude is not None:
+            exclude = exclude.upper()
```
CaselIT: I would avoid changing the case

lib/sqlalchemy/sql/elements.py
```diff
@@ -4570,6 +4573,23 @@ def __init__(
...
+            from .compiler import WINDOW_EXCLUDE
+
+            if not WINDOW_EXCLUDE.match(exclude):
+                raise exc.ArgumentError(
+                    "Invalid window frame exclusion '%s'. Valid values "
+                    "are CURRENT ROW, GROUP, TIES, NO OTHERS" % exclude
+                )
+            if self.range_ is None and self.rows is None \
```
CaselIT: please run the formatters. there is a pre-commit file that will run them for you

test/sql/test_compiler.py
```diff
@@ -3300,6 +3300,94 @@ def test_over_framespec(self):
             checkparams={"param_1": 1, "param_2": 3},
         )
 
+    def test_over_frame_exclude(self):
```
CaselIT: so for newer test we try to have single tests or to use variations/combinations 
but I see that many of the test in this file are like this, so it's likely fine as is

CaselIT (commented):
looks like a good start, thanks. I've left a few initial comments, will look better later

lib/sqlalchemy/sql/_elements_constructors.py
```diff
@@ -1725,6 +1726,14 @@ def over(
 
      .. versionadded:: 2.0.40
 
+    :param exclude: optional string for the frame exclusion clause.
+     This is a string value which can be one of ``CURRENT ROW``,
+     ``GROUP``, ``TIES``, or ``NO OTHERS`` and will render an
+     EXCLUDE clause within the window frame specification.  Requires
+     that one of ``rows``, ``range_``, or ``groups`` is also specified.
+
+     .. versionadded:: 2.1
```
zzzeek, line 1735: issue #11671 targets this to the 2.0.x series since it's a small use case

lib/sqlalchemy/sql/compiler.py
```diff
@@ -2991,7 +2988,7 @@ def visit_over(self, over, **kwargs):
             range_ = None
 
         if range_ is not None and over.exclude is not None:
-            range_ += f" EXCLUDE {over.exclude}"
+            range_ += f" EXCLUDE {over.exclude.upper()}"
```
CaselIT: we don't need to upper case here either I think.

@zzzeek do you have a preference here?

lib/sqlalchemy/sql/_elements_constructors.py
```diff
@@ -1621,6 +1621,7 @@ def over(
     range_: _FrameIntTuple | FrameClause | None = None,
     rows: _FrameIntTuple | FrameClause | None = None,
     groups: _FrameIntTuple | FrameClause | None = None,
+    exclude: Optional[str] = None,
```
zzzeek: we should use `str | None` for consistency here.    the 2.0 backport will need to adjust it

lib/sqlalchemy/sql/_elements_constructors.py
```diff
@@ -1737,7 +1746,7 @@ def over(
         :func:`_expression.within_group`
 
     """  # noqa: E501
-    return Over(element, partition_by, order_by, range_, rows, groups)
+    return Over(element, partition_by, order_by, range_, rows, groups, exclude=exclude)
```
zzzeek: for consistency pass exclude positionally like the others

lib/sqlalchemy/sql/elements.py
```diff
@@ -4549,6 +4555,7 @@ def __init__(
         range_: _FrameIntTuple | FrameClause | None = None,
         rows: _FrameIntTuple | FrameClause | None = None,
         groups: _FrameIntTuple | FrameClause | None = None,
+        exclude: Optional[str] = None,
```
zzzeek: use `str | None`

lib/sqlalchemy/sql/elements.py
```diff
@@ -4508,6 +4508,10 @@ def _gen_cache_key(
 
 _FrameIntTuple = tuple[int | None, int | None]
 
+_WINDOW_EXCLUDE_RE = re.compile(
```
zzzeek: this is part of compiler.py, not elements.py - see compiler.py -> FK_ON_DELETE, FK_ON_UPDATE etc. for examples

lib/sqlalchemy/sql/elements.py
```diff
@@ -4570,6 +4577,23 @@ def __init__(
             self.rows = FrameClause._parse(rows, coerce_int=True)
             self.groups = FrameClause._parse(groups, coerce_int=True)
 
+        if exclude is not None:
+            if not _WINDOW_EXCLUDE_RE.match(exclude):
```
zzzeek: remove this block, this happens in compiler

lib/sqlalchemy/sql/elements.py
```diff
@@ -4804,6 +4828,7 @@ def over(
         rows: _FrameIntTuple | FrameClause | None = None,
         range_: _FrameIntTuple | FrameClause | None = None,
         groups: _FrameIntTuple | FrameClause | None = None,
+        exclude: Optional[str] = None,
```
zzzeek: str | None

lib/sqlalchemy/sql/elements.py
```diff
@@ -4948,6 +4974,7 @@ def over(
         range_: _FrameIntTuple | FrameClause | None = None,
         rows: _FrameIntTuple | FrameClause | None = None,
         groups: _FrameIntTuple | FrameClause | None = None,
+        exclude: Optional[str] = None,
```
zzzeek: str | None

lib/sqlalchemy/sql/functions.py
```diff
@@ -460,6 +460,7 @@ def over(
         rows: _FrameIntTuple | FrameClause | None = None,
         range_: _FrameIntTuple | FrameClause | None = None,
         groups: _FrameIntTuple | FrameClause | None = None,
+        exclude: Optional[str] = None,
     ) -> Over[_T]:
```
zzzeek, line 464: str | None

lib/sqlalchemy/sql/compiler.py
```diff
@@ -2987,6 +2987,9 @@ def visit_over(self, over, **kwargs):
         else:
             range_ = None
 
+        if range_ is not None and over.exclude is not None:
+            range_ += f" EXCLUDE {over.exclude.upper()}"
```
zzzeek: write this as:

        range_ += f"EXCLUDE {self.preparer.validate_sql_phrase(over.exclude, _WINDOW_EXCLUDE_RE)}"

zzzeek (changes requested):
this will have a 2.0 backport but let's keep the annotations using the same `|` format as the others that surround it here.    we'll do the backport from here once this one is good to go.

lib/sqlalchemy/sql/compiler.py
zzzeek: as i mentioned in my review this should come from validate_sql_phrase

lib/sqlalchemy/sql/elements.py
CaselIT: I asked to change that, since the validation is on creation not on compile

I think it makes more sense to validate when constructing the element

lib/sqlalchemy/sql/elements.py
CaselIT: I disagree, I think it's more useful to validate it here. This way you get the exact place where something did not work fine

zzzeek: third party dialects can support additional keywords here and there's precedent for doing this in the compiler.   I dont think there's any downside to it being in the compiler, if someone uses an unknown keyword and it raises, it's not hard to see where it originates from in user code

veeceey: Done - moved the validation to the compiler using `validate_sql_phrase` with `_WINDOW_EXCLUDE_RE`, following the same pattern as FK_ON_DELETE/FK_ON_UPDATE. Also addressed the other feedback (str | None, positional exclude arg, ran formatters).

lib/sqlalchemy/sql/elements.py
CaselIT: of course I was not thinking about the fact that dialects may want to change what's supported.

zzzeek: it's more important to me that it's consistent with how we do other kinds of "keywords inside of string fields", if someday we wanted to enhance the system to either somehow work at construction time or provide better tracing, or use some kind of new enumeration thing, all the different cases would be improved in a consistent fashion.

zzzeek: maybe we should add a pep-695 type to indicate strings like this.   like a `type COMPILER_CHECKED_PHRASE = str`.   probably not yet since pep-695 is not universal yet

veeceey:
Hi @zzzeek @CaselIT, I've addressed all the feedback from the review — moved the validation to the compiler using validate_sql_phrase with _WINDOW_EXCLUDE_RE (following the FK_ON_DELETE pattern), switched to str | None annotations, made exclude positional, and ran the formatters. Let me know if anything else needs adjusting\!

test/sql/test_compiler.py
zzzeek: never too late to start, i'd see if we can use combinations here

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 8733837013599e17130c6b1a484e9d72b8a46123 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 8733837013599e17130c6b1a484e9d72b8a46123: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6597

test/sql/test_compiler.py
```diff
@@ -3300,73 +3300,69 @@ def test_over_framespec(self):
...
-        )
-
-        # exclude with groups
-        self.assert_compile(
-            select(
-                func.row_number().over(
-                    order_by=expr, groups=(-1, 1), exclude="TIES"
-                )
-            ),
+            {},
```
zzzeek, line 3322: nice

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision c38c634174708b0fa29f5546a976fb4e85bdc4ca of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset c38c634174708b0fa29f5546a976fb4e85bdc4ca added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6597

zzzeek:
let me know if you're able to iterate on failing test collection

```
test/sql/test_compiler.py::SelectTest::test_over_frame_exclude: in "parametrize" the number of names (4):
  ['frame_kwargs', 'exclude', 'expected_sql', 'checkparams']
must be equal to the number of values (5):
  ('rows_current_row', {'rows': (None, 0)}, 'CURRENT ROW', 'SELECT row_number() OVER (ORDER BY mytable.myid ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW EXCLUDE CURRENT ROW) AS anon_1 FROM mytable', {})
```

veeceey:
Thanks for flagging that @zzzeek — I see the parametrize mismatch. The test tuple has 5 values but the parameter names list only has 4. I'll fix the test parametrization and push an update shortly.

veeceey:
Fixed the parametrize mismatch — added `id_="iaaaa"` to the `@testing.combinations` decorator so the first string element in each tuple is used as the test ID rather than passed as a parameter. Should match the 4-parameter function signature now. CI should re-run with this push.

zzzeek:
can you install [pre-commit](https://pre-commit.com/) which will ensure most pep8 is flagged on the next commit

veeceey:
@zzzeek Sure thing, I'll get pre-commit installed and run it across the files I've touched. Will push a formatting cleanup commit shortly.

veeceey:
Addressed the latest feedback -- refactored the tests to use `testing.combinations` with `id_="iaaaa"` so each tuple's first element serves as the test ID. Also installed pre-commit and ran the formatters across the touched files (latest commit). Let me know if anything else needs adjusting\!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 76e0d5ca0b6271537c4ed6e481f5589c574e8edf of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 76e0d5ca0b6271537c4ed6e481f5589c574e8edf added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6597

lib/sqlalchemy/sql/_elements_constructors.py
sqla-tester, line 1735: **Federico Caselli** (CaselIT) wrote:

do we want to have this in 2.0 or keep it for 2.1?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6597

lib/sqlalchemy/sql/elements.py
```diff
@@ -4570,6 +4573,8 @@ def __init__(
             self.rows = FrameClause._parse(rows, coerce_int=True)
             self.groups = FrameClause._parse(groups, coerce_int=True)
 
+        self.exclude = exclude
```
sqla-tester, line 4576: **Federico Caselli** (CaselIT) wrote:

q: should we validate that one of range/rows/groups was specified when exclude is used?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6597

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6597

lib/sqlalchemy/sql/elements.py
veeceey: Good point -- per the SQL spec, EXCLUDE is only valid within a window frame clause (i.e. when ROWS/RANGE/GROUPS is specified). I can add a validation check for that if you'd like. Probably just raise an ArgumentError if exclude is set but none of rows/range/groups are. Let me know if you want me to go ahead with that.

CaselIT: yes, let's do that, since we validate all other cases. It would be strange to allow an exclude but then not render anything since there is now frame clause

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 23f9d34dd80c45dff68ecc8c08331acf22b82279 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 23f9d34dd80c45dff68ecc8c08331acf22b82279 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6597

veeceey:
hey @CaselIT, pushed the validation — now raises ArgumentError if exclude is specified without a frame clause (rows/range_/groups). Added test coverage for that case too. Let me know if anything else needs adjusting on gerrit!

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6597

lib/sqlalchemy/sql/elements.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6597

CaselIT:
> hey @CaselIT, pushed the validation — now raises ArgumentError if exclude is specified without a frame clause (rows/range_/groups). Added test coverage for that case too. Let me know if anything else needs adjusting on gerrit!

the change seems fine, I'm gonna finish it. No further action should be needed from your part, thanks!

lib/sqlalchemy/sql/_elements_constructors.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

kept in 2.1 since 2.0 is significantly differnt

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6597

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6597

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6597 has been **merged**. Congratulations! :)

PR #13118, opened 2026-02-10
Add overload for `CursorResult` return type for `Session.execute`

soceanainn:
When running update/delete statements without `returning` clauses using `Session.execute()`, developers are typically validating success by accessing `rowcount` on the returned result, which causes type checkers to complain.

This change intends to add an overload to `Session.execute()` to return `CursorResult` in these situations, which includes rowcount.

### Description

My understanding from debugging this at runtime / exploring code is that:
1. Any `UpdateBase` derived statement with a `returning` clause, will not be impacted by this change. They will fall under the previous overload for `execute` by matching `statement: TypedReturnsRows[Unpack[_Ts]]` and as a result will continue to return a type of `Result[Unpack[_Ts]]`.
2. All other `UpdateBase` derived statements should actually return `CursorResult[Unpack[TupleAny]]` (and possibly even this could be more explicitly typed as `CursorResult[None]`). This assumption is based on observations at runtime, as opposed to any concrete guarantees observed in code (as `_execute_internal` is quite dynamic here).

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/orm/session.py
+++ b/lib/sqlalchemy/orm/session.py
@@ -130,6 +130,7 @@ if typing.TYPE_CHECKING:
     from ..sql._typing import _TypedColumnClauseArgument as _TCCA
     from ..sql.base import Executable
     from ..sql.base import ExecutableOption
+    from ..sql.dml import UpdateBase
     from ..sql.elements import ClauseElement
     from ..sql.roles import TypedColumnsClauseRole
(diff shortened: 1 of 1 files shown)


CaselIT:
Hi,

This was removed intentionally since it's not universally true. See #12813

PR #13123, opened 2026-02-13
Allow opting specific mapped classes out of bind config for Session

soceanainn:
Draft status while working on updating tests.

### Description

We have a use case where we want to set our sessionmaker as follows:
```python3
class M1(DeclarativeBase):
    pass

class M2(DeclarativeBase):
    pass

Session = sessionmaker(
    bind=engine,
    binds={M1: engine, M2: None}
)
session = Session()
```

The goal is:
1. Custom non-ORM statements (e.g. for controlling custom scoped extended session timeout values) will continue to work by accessing underlying bind.
2. ORM statements accessing models created by subclassing M1 will work with these sessions.
3. ORM statements accessing models created by subclassing M2 will not work with these sessions.

This kind of works today - attempting to use an M2 based ORM model with this session will result in an AttributeError once `conn = bind.connect()` is called in [_connection_for_bind](https://github.com/soceanainn/sqlalchemy/blob/9314ab0a5bdde1b8407134cc676b176f73689a60/lib/sqlalchemy/orm/session.py#L1188).

I think the behaviour is more consistent to raise `UnboundExecutionError`  if we encounter a value of `None` within binds - this is consistent with encountering an explicit `None` value for `bind`.

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

This prompts me to create an issue, but when I try to create an issue it prompts me to open a discussion instead.

Started a discussion here:
https://github.com/sqlalchemy/sqlalchemy/discussions/13122

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/orm/session.py
+++ b/lib/sqlalchemy/orm/session.py
@@ -2843,6 +2843,20 @@ class Session(_SessionClassMethods, EventTarget):
         else:
             inspected_mapper = None
 
+        def _generate_unbound_execution_error() -> (
+            sa_exc.UnboundExecutionError
+        ):
+            context = []
+            if inspected_mapper is not None:
+                context.append(f"mapper {inspected_mapper}")
+            if clause is not None:
+                context.append("SQL expression")
+
+            return sa_exc.UnboundExecutionError(
+                f"Could not locate a bind configured on "
+                f'{", ".join(context)} or this Session.'
+            )
+
         # match up the mapper or clause in the __binds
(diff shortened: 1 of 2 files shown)


zzzeek:
hi -

this is not a PR we want to accept; the get_bind() method should be overridden by a subclass that's local to your project.

soceanainn:
Why wouldn't the behaviour of `bind` and `binds` for sessionmaker be consistent for null values?

PR #13133, opened 2026-02-19
fix: MySQLDDLCompiler create index falls back to default compiler args

TiansuYu:
### Description

Custom DDL compilers that inherit from `MySQLDDLCompiler` (e.g. third-party dialects like StarRocks) cannot use the default `visit_create_index` method when their dialect name differs from `"mysql"`.

The issue is that `visit_create_index` resolves index options — `prefix`, `length`, `using`, `with_parser` — by looking them up exclusively under `{dialect_name}_<arg>` / `dialect_options[dialect_name]`. When a subclass has a different dialect name (e.g. `"starrocks"`), these lookups return `None` even though the user specified them via the standard `mysql_*` kwargs, since the options are stored under the `"mysql"` key.

This change makes each lookup fall back to the `"mysql"` key when the dialect-specific key yields no result, so subclasses of `MySQLDDLCompiler` work out of the box without having to reimplement `visit_create_index` just to fix option resolution.

### Changes

- **`lib/sqlalchemy/dialects/mysql/base.py`** — For `prefix`, `length`, `with_parser`, and `using`, try `dialect_options[self.dialect.name]` first, then fall back to `dialect_options["mysql"]`.
- **`test/dialect/mysql/test_compiler.py`** — Added `MySQLBasedCustomDialectDDLTest` covering `CREATE INDEX` compilation through a simulated custom dialect: simple, length, prefix, using, parser, unique, composite, and expression-based indexes.

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/mysql/base.py
+++ b/lib/sqlalchemy/dialects/mysql/base.py
@@ -2215,7 +2215,9 @@ class MySQLDDLCompiler(
         if index.unique:
             text += "UNIQUE "
 
-        index_prefix = index.kwargs.get("%s_prefix" % self.dialect.name, None)
+        index_prefix = index.kwargs.get(
+            "%s_prefix" % self.dialect.name) or index.kwargs.get("mysql_prefix")
+
         if index_prefix:
             text += index_prefix + " "
 
@@ -2224,7 +2226,7 @@ class MySQLDDLCompiler(
             text += "IF NOT EXISTS "
         text += "%s ON %s " % (name, table)
 
-        length = index.dialect_options[self.dialect.name]["length"]
+        length = index.dialect_options[self.dialect.name].get("length") or index.dialect_options["mysql"]["length"]
         if length is not None:
             if isinstance(length, dict):
                 # length value can be a (column_name --> integer value)
@@ -2252,11 +2254,15 @@ class MySQLDDLCompiler(
             columns_str = ", ".join(columns)
         text += "(%s)" % columns_str
 
-        parser = index.dialect_options["mysql"]["with_parser"]
+        parser = index.dialect_options[self.dialect.name].get(
+            "with_parser") or index.dialect_options["mysql"]["with_parser"]
+
         if parser is not None:
             text += " WITH PARSER %s" % (parser,)
 
-        using = index.dialect_options["mysql"]["using"]
+        using = index.dialect_options[self.dialect.name].get(
+            "using") or index.dialect_options["mysql"]["using"]
+
         if using is not None:
             text += " USING %s" % (preparer.quote(using))
 
--- a/test/dialect/mysql/test_compiler.py
+++ b/test/dialect/mysql/test_compiler.py
@@ -683,6 +683,105 @@ class CompileTest(ReservedWordFixture, fixtures.TestBase, AssertsCompiledSQL):
         )
 
 
+class MySQLBasedCustomDialectDDLTest(
+    fixtures.TestBase, AssertsCompiledSQL
+):
+    """Test that a custom dialect subclassing MySQLDialect correctly
+    compiles CREATE INDEX via the inherited MySQLDDLCompiler.
+
+    This simulates third-party dialects (e.g. StarRocks) that inherit
+    from MySQLDialect with a different dialect name but rely on
+    MySQLDDLCompiler.visit_create_index without overrides.
+    """
+
+    @testing.fixture(autouse=True)
+    def _register_custom_dialect(self):
+        from sqlalchemy.dialects import registry
+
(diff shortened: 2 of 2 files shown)


zzzeek:
hi , sorry for the close here but please read and follow the instructions for "short code fix" PRs so that we can review what's being proposed according to our workflow, thanks!

TiansuYu:
@zzzeek I was about to post the reproducible behavior here. But you closed it before I can. Also even following the suggested workflow: opening a bug report before open the PR: the issue workflow also suggests open a behavior discussion before considering a bug. My question is: does the contribution process has to be so sophisticated? And where would be a good starting case for my specific case?

TiansuYu:
# Bug: StarRocks dialect `KeyError: 'length'` when creating an Index with `mysql_using`

## Root Cause

The StarRocks SQLAlchemy dialect (`construct_arguments`) only registers options for
`Table` and `Column`, but not for `Index`. When Alembic's `op.create_index()` compiles
the DDL, the inherited MySQL `visit_create_index` reads
`index.dialect_options[self.dialect.name]["length"]` — since `self.dialect.name` is
`"starrocks"` and no Index construct args are registered, it raises `KeyError: 'length'`.

Additionally, the MySQL `visit_create_index` hardcodes `dialect_options["mysql"]` for
`using` and `with_parser` (lines 2236-2242 in `sqlalchemy/dialects/mysql/base.py`),
so even if the `KeyError` is fixed, `starrocks_using="BITMAP"` would be ignored —
only `mysql_using="BITMAP"` would work.

## Dependencies

```bash
pip install sqlalchemy alembic starrocks
```

## Minimal Reproducible Example

```python
"""example.py: StarRocks dialect KeyError: 'length' on Index creation."""

from sqlalchemy import BigInteger, Column, Index, MetaData, String, Table
from sqlalchemy.schema import CreateIndex
from starrocks.dialect import StarRocksDialect

metadata = MetaData()

demo = Table(
    "demo",
    metadata,
    Column("id", BigInteger, primary_key=True),
    Column("status", String(32)),
    Index("idx_status_bitmap", "status", mysql_using="BITMAP"),
    starrocks_primary_key="id",
    starrocks_distributed_by="HASH(id)",
)

dialect = StarRocksDialect()
idx = list(demo.indexes)[0]

try:
    compiled = CreateIndex(idx).compile(dialect=dialect)
    print(f"SUCCESS: {compiled}")
except KeyError as e:
    print(f"FAILED: {e}")
```

## Command

```bash
python example.py
```

## Actual Output

```
FAILED: KeyError: 'length'
```

## Expected Output (after fix)

```
SUCCESS: CREATE INDEX idx_status_bitmap ON demo (`status`) USING `BITMAP`
```

zzzeek:
> @zzzeek I was about to post the reproducible behavior here. But you closed it before I can. Also even following the suggested workflow: opening a bug report before open the PR: the issue workflow also suggests open a behavior discussion before considering a bug. My question is: does the contribution process has to be so sophisticated? And where would be a good starting case for my specific case?

the bug template only suggests a discussion since the vast majority of reports are new users who are confused by existing APIs.    It doesn't prohibit opening a bug report and dialect authors can assume it's fine to open bug reports.

the reason for a bug report prior to the PR is that we need to vet the scope of the problem and determine the best course of action.  the PR you proposed may or may not be the best way to do this.       In some ways the "open a bug before the PR" request is analogous to the "open a discussion before the bug" request;  end users report things as "bugs" that are not bugs all the time, developers post PRs that are not always the way we prefer to approach th eproblem.

This issue points to a larger issue in the dialect options space, and I also am not sure why "mysql" was hardcoded in some spots and not others (is it trying to sidestep "mariadb" for example).   So I really prefer to understand the problem fully and do some thinking on the general architectural issue, where we can then agree on the approach.    It's quite jarring to get code first for an issue I've not been thinking about already.

CaselIT:
BTW in general I would expect that an user using starrocks would use `starrocks_using` etc, not the mysql ones, so the change would likely be the opposite of the one proposed here, so always using the dialect name.

> is it trying to sidestep "mariadb" for example

this is always complicating stuff :(

TiansuYu:
Thanks for the detailed explanation. I will open a bug report first, then we can go from there.

TiansuYu:
> BTW in general I would expect that an user using starrocks would use starrocks_using etc, not the mysql ones, so the change would likely be the opposite of the one proposed here, so always using the dialect name.

The bug happens regardless of `mysql_using` or `starrocks_using` due to many hard-coded references

CaselIT:
agree, I just meant that I would not hardcode mysql as the fallback there

PR #13161, opened 2026-03-09
Update _NamingSchemaCallable to support Index

embecka:
### Description

According to [the documentation](https://docs.sqlalchemy.org/en/21/core/metadata.html#sqlalchemy.schema.MetaData.params.naming_convention), the values associated with user-defined “token” keys in `naming_convention` should be callables of the form `fn(constraint, table)`, which accepts the constraint/index object and Table. However, the type alias `_NamingSchemaCallable` accepts only constraint in the first argument. I propose to update `_NamingSchemaCallable` to accept also an index.

### Example
```python
import sqlalchemy


def include_0_N_name(
		schema_item: sqlalchemy.Index | sqlalchemy.Constraint,
		table: sqlalchemy.Table,
) -> str:
	tokens = []

	for column in schema_item.dialect_options.get('postgresql', {}).get('include', []):
		tokens.append("_")
		tokens.append(column.name)

	return "".join(tokens)


metadata = sqlalchemy.MetaData(
	naming_convention={
		"include_0_N_name": include_0_N_name,
		"ix": "%(table_name)s_%(column_0_N_name)s%(include_0_N_name)s_idx",
		"uq": "%(table_name)s_%(column_0_N_name)s%(include_0_N_name)s_key",
		"fk": "%(table_name)s_%(column_0_N_name)s%(include_0_N_name)s_fkey",
		"pk": "%(table_name)s_pkey",
	},
)
```

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

### DIFF
--- a/lib/sqlalchemy/sql/schema.py
+++ b/lib/sqlalchemy/sql/schema.py
@@ -5616,7 +5616,10 @@ class Index(
         )
 
 
-_NamingSchemaCallable = Callable[[Constraint, Table], str]
+_NamingSchemaCallable = Union[
+    Callable[[Constraint, Table], str],
+    Callable[[Index, Table], str],
+]
 _NamingSchemaDirective = Union[str, _NamingSchemaCallable]
 
 
--- a/test/typing/plain_files/sql/schema.py
+++ b/test/typing/plain_files/sql/schema.py
@@ -66,3 +66,20 @@ NAMING_CONVENTIONS_STR = {
 }
 
 MetaData(naming_convention=NAMING_CONVENTIONS_STR)
+
+
+def index_only(index: Index, table: Table) -> str:
+    return "index_only"
+
+
+def constraint_only(constraint: Constraint, table: Table) -> str:
+    return "constraint_only"
+
+
+# constraint-only callable or index-only callable
+MetaData(
+    naming_convention={
+        "ix": index_only,
+        "uq": constraint_only,
+    }
+)


lib/sqlalchemy/sql/schema.py
```diff
@@ -5616,7 +5616,10 @@ def __repr__(self) -> str:
         )
 
 
-_NamingSchemaCallable = Callable[[Constraint, Table], str]
+_NamingSchemaCallable = Union[
+    Callable[[Constraint, Table], str],
+    Callable[[Index, Table], str],
```
CaselIT, line 5621: could we use `Callable[[Union[Constraint, Index], Table], str]`?

CaselIT (commented):
thanks for the PR.

Can you add a simple typing test for this here https://github.com/sqlalchemy/sqlalchemy/blob/main/test/typing/plain_files/sql/schema.py?

there is already a test

lib/sqlalchemy/sql/schema.py
embecka: You can write a function:
1. `Callable[[Constraint, Table], str]` to name only constraints,
2.  `Callable[[Index, Table], str]` to name only indices, or
3. `Callable[[Union[Constraint, Index], Table], str]` to name both constraints and indices.

All three options match the annotation `Union[Callable[[Constraint, Table], str], Callable[[Index, Table], str]]`, but only the third option matches the annotation `Callable[[Union[Constraint, Index], Table], str]`.

CaselIT: I guess if we were to be 100% correct we would assign different annotations to the different keys of the naming schema.

let's keep your version.

Can you add a typing test with first 2 options, since it seems that currently we have one only for 3rd?

CaselIT (approved):

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision cf53bd722932741bceaba1dd16a52ba93ff579cc of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change cf53bd722932741bceaba1dd16a52ba93ff579cc: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6616

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6616 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6617 has been **merged**. Congratulations! :)

Issue #6008, opened 2021-03-06
select distinct raises NosuchColumnError on SQLAlchemy 1.4.0b3


bsdz:
**Describe the bug**

Selecting distinct from table with schema raises NosuchColumnError on SQLAlchemy 1.4.0b3 but works on older 1.3.23.

**Expected behavior**

It should not raise an exception and behave like in previous versions.

**To Reproduce**

Create database in postgresql:

```sql
CREATE DATABASE foo;

\connect foo;

CREATE SCHEMA bar;

CREATE TABLE bar.table1 (
    col1 integer NOT NULL,
    ADD CONSTRAINT table1_pk PRIMARY KEY (col1);
);
```

Then run following script:

```python
from sqlalchemy.ext.automap import automap_base
from sqlalchemy import create_engine
from sqlalchemy import distinct
from sqlalchemy.orm import sessionmaker, scoped_session

Base = automap_base()


class Table1(Base):
    __tablename__ = "table1"
    __table_args__ = {"schema": "bar"}


engine = create_engine("postgresql+psycopg2://docker:docker@localhost:5400/foo")
engine.echo = True
Base.prepare(engine, schema="bar", reflect=True)

session = scoped_session(sessionmaker(autocommit=False, autoflush=False, bind=engine))

# runs okay on SQLA 1.3.23
# fails on SQLA 1.4.0b3 with NoSuchColumnError: "Could not locate column in row for column 'DISTINCT bar.table1.col1'"
recs = session.query(distinct(Table1.col1)).select_from(Table1).all()
```

**Error**

```
Traceback (most recent call last):
  File "/home/blair/scratch/sqla/.venv/lib/python3.8/site-packages/sqlalchemy/engine/cursor.py", line 639, in _index_for_key
    rec = self._keymap[key]
KeyError: <sqlalchemy.sql.elements.UnaryExpression object at 0x7fc2f99d3790>

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/blair/.vscode/extensions/ms-python.python-2021.2.582707922/pythonFiles/lib/python/debugpy/__main__.py", line 45, in <module>
    cli.main()
[...]

PR #13162, opened 2026-03-10
Compile-time warning feedback for unary distinct outside aggregate function

### LINKED ISSUE #11526: Warn when `distinct()` is used outside of aggregated functions
Discussed in #11518

The alternatives could be:
- at compile time: you could do it at compile time by passing a kw flag that "distinct" looks for , indicating the appropriate context. if flag not there, boom. would have to find a way to make it OK for str() compile -- https://github.com/sqlalchemy/sqlalchemy/discussions/11518#discussioncomment-9840468
- at coercion time: does "distinct()" only go inside of a SQL function ? That's all we'd have to look for. we can put "distinct()" into a special coercion category based on the roles, like FunctionDistinctRole -- https://github.com/sqlalchemy/sqlalchemy/discussions/11518#discussioncomment-9841617

the second option is likely better since it will give better feedback to the user (since it will happen at statement creation time, instead of when it will be compiled)

bekapono:
### Description

Adds a warning when unary `distinct()` is used outside an aggregate 
function. Misuse can cause runtime errors on some dialects (e.g. MSSQL via 
pyodbc) or silently produce incorrect SQL on others.

The warning check is implemented via boolean flag that gets set to TRUE  in `visit_function()`.  
So when we get to `visit_unary()` the flag is checked. In `visit_unary()` we check for if the 
operator is distinct_op and if the flag is TRUE, if the flag is FALSE or doesn't exists, and 
operator is distinct_op, the warning is issued.

Warning issued: 
                "Column-expression-level unary distinct() "
                "should not be used outside of an aggregate "
                "function. For general 'SELECT DISTINCT' support"
                "use select().distinct()."

Fixes: #11526

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/sql/coercions.py
+++ b/lib/sqlalchemy/sql/coercions.py
@@ -1138,13 +1138,6 @@ class ColumnsClauseImpl(_SelectIsNotFrom, _CoerceLiterals, RoleImpl):
 
     _guess_straight_column = re.compile(r"^\w\S*$", re.I)
 
-    _aggregate_functions = {
-        # Basic Aggregate function listed from GenericFunctions
-        'count', 'sum', 'avg',
-
-        # other aggregate functions
-    }
-
     def _raise_for_expected(
         self, element, argname=None, resolved=None, *, advice=None, **kw
     ):
@@ -1157,41 +1150,6 @@ class ColumnsClauseImpl(_SelectIsNotFrom, _CoerceLiterals, RoleImpl):
         return super()._raise_for_expected(
             element, argname=argname, resolved=resolved, advice=advice, **kw
         )
-
-    @staticmethod
-    def _warn_unary_expression_outside_aggregate():
-        util.warn(
-            "column-expression-level unary DISTINCT clause used outside aggregate;"
-            " use select().distinct() for full-statement DISTINCT."
-        )
-
-    def _valid_unary_hierarchy(self, ent: ColumnElement[_T], op_name: str) -> bool:
-        seen_agg_func = False
-
-        if not isinstance(ent, elements.ClauseElement):
-            return True
-
-        for item in visitors.iterate(ent):
-            operator = getattr(item, 'operator', None)
-
-            if (
-                isinstance(item, elements.UnaryExpression)
-                and getattr(operator, "__name__", None) == op_name
-                and seen_agg_func is False
-            ):
-                return False
-
-            if getattr(type(item), '__name__', None) in self._aggregate_functions:
-                seen_agg_func = True
-
-        return True
-
-    def _post_coercion(
-            self, resolved, *, original_element, argname=None, **kw
-    ) -> Any:
-        if not self._valid_unary_hierarchy(original_element, op_name='distinct_op'):
-            self._warn_unary_expression_outside_aggregate()
-        return resolved
 
     def _text_coercion(self, element, argname=None):
         element = str(element)
--- a/lib/sqlalchemy/sql/coercions.py
+++ b/lib/sqlalchemy/sql/coercions.py
@@ -1138,6 +1138,13 @@ class ColumnsClauseImpl(_SelectIsNotFrom, _CoerceLiterals, RoleImpl):
 
     _guess_straight_column = re.compile(r"^\w\S*$", re.I)
 
+    _aggregate_functions = {
+        # Basic Aggregate function listed from GenericFunctions
+        'count', 'sum', 'avg',
+
+        # other aggregate functions
+    }
+
     def _raise_for_expected(
         self, element, argname=None, resolved=None, *, advice=None, **kw
     ):
@@ -1150,6 +1157,41 @@ class ColumnsClauseImpl(_SelectIsNotFrom, _CoerceLiterals, RoleImpl):
         return super()._raise_for_expected(
             element, argname=argname, resolved=resolved, advice=advice, **kw
         )
+
+    @staticmethod
+    def _warn_unary_expression_outside_aggregate():
+        util.warn(
+            "column-expression-level unary DISTINCT clause used outside aggregate;"
+            " use select().distinct() for full-statement DISTINCT."
+        )
+
+    def _valid_unary_hierarchy(self, ent: ColumnElement[_T], op_name: str) -> bool:
+        seen_agg_func = False
+
+        if not isinstance(ent, elements.ClauseElement):
+            return True
+
+        for item in visitors.iterate(ent):
+            operator = getattr(item, 'operator', None)
+
+            if (
+                isinstance(item, elements.UnaryExpression)
+                and getattr(operator, "__name__", None) == op_name
+                and seen_agg_func is False
+            ):
+                return False
+
+            if getattr(type(item), '__name__', None) in self._aggregate_functions:
+                seen_agg_func = True
+
+        return True
+
+    def _post_coercion(
+            self, resolved, *, original_element, argname=None, **kw
+    ) -> Any:
+        if not self._valid_unary_hierarchy(original_element, op_name='distinct_op'):
+            self._warn_unary_expression_outside_aggregate()
+        return resolved
 
     def _text_coercion(self, element, argname=None):
         element = str(element)
--- a/test/dialect/mssql/test_compiler.py
+++ b/test/dialect/mssql/test_compiler.py
@@ -445,25 +445,6 @@ class CompileTest(fixtures.TestBase, AssertsCompiledSQL):
             "tbl_row_count FROM sometable",
         )
 
-    # test-case 1 of 2 for mssql dialect - issue#11526
-    def test_unary_distinct_with_limit(self):
-        t = table("sometable", column("somecolumn"))
-        with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
-            self.assert_compile(
-                select(t.c.somecolumn.distinct()).limit(1),
-                "SELECT TOP __[POSTCOMPILE_param_1] DISTINCT sometable.somecolumn "
-                "FROM sometable"
-            )
-
-    # test-case 2 of 2 for mssql dialect - issue#11526
-    def test_selectable_distinct_with_limit(self):
-        t = table("sometable", column("somecolumn"))
-        self.assert_compile(
-            select(t.c.somecolumn).distinct().limit(1),
-            "SELECT DISTINCT TOP __[POSTCOMPILE_param_1] sometable.somecolumn "
-            "FROM sometable"
-        )
-
     def test_noorderby_insubquery(self):
         """test "no ORDER BY in subqueries unless TOP / LIMIT / OFFSET"
         present"""
--- a/test/dialect/mssql/test_compiler.py
+++ b/test/dialect/mssql/test_compiler.py
@@ -30,7 +30,7 @@ from sqlalchemy.sql import column
 from sqlalchemy.sql import quoted_name
 from sqlalchemy.sql import table
 from sqlalchemy.sql.ddl import CreateView
-from sqlalchemy.testing import assert_raises_message
+from sqlalchemy.testing import assert_raises_message, expect_warnings
 from sqlalchemy.testing import AssertsCompiledSQL
 from sqlalchemy.testing import eq_
 from sqlalchemy.testing import eq_ignore_whitespace
@@ -443,6 +443,25 @@ class CompileTest(fixtures.TestBase, AssertsCompiledSQL):
             t.count(),
             "SELECT count(sometable.somecolumn) AS "
             "tbl_row_count FROM sometable",
+        )
+
+    # test-case 1 of 2 for mssql dialect - issue#11526
+    def test_unary_distinct_with_limit(self):
+        t = table("sometable", column("somecolumn"))
+        with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
+            self.assert_compile(
+                select(t.c.somecolumn.distinct()).limit(1),
+                "SELECT TOP __[POSTCOMPILE_param_1] DISTINCT sometable.somecolumn "
+                "FROM sometable"
+            )
+
+    # test-case 2 of 2 for mssql dialect - issue#11526
+    def test_selectable_distinct_with_limit(self):
+        t = table("sometable", column("somecolumn"))
+        self.assert_compile(
+            select(t.c.somecolumn).distinct().limit(1),
+            "SELECT DISTINCT TOP __[POSTCOMPILE_param_1] sometable.somecolumn "
+            "FROM sometable"
         )
 
     def test_noorderby_insubquery(self):
--- a/test/dialect/mssql/test_compiler.py
+++ b/test/dialect/mssql/test_compiler.py
@@ -30,7 +30,7 @@ from sqlalchemy.sql import column
 from sqlalchemy.sql import quoted_name
 from sqlalchemy.sql import table
 from sqlalchemy.sql.ddl import CreateView
-from sqlalchemy.testing import assert_raises_message, expect_warnings
+from sqlalchemy.testing import assert_raises_message
 from sqlalchemy.testing import AssertsCompiledSQL
 from sqlalchemy.testing import eq_
 from sqlalchemy.testing import eq_ignore_whitespace
--- a/test/orm/test_query.py
+++ b/test/orm/test_query.py
@@ -4631,21 +4631,23 @@ class DistinctTest(QueryTest, AssertsCompiledSQL):
         # issue 6008.  the UnaryExpression now places itself into the
         # result map so that it can be matched positionally without the need
         # for any label.
-        with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
-            q = fixture_session().query(distinct(User.id)).order_by(User.id)
-            self.assert_compile(
-                q, "SELECT DISTINCT users.id FROM users ORDER BY users.id"
-            )
-            eq_([(7,), (8,), (9,), (10,)], q.all())
+
+        # pr 11362
+        # removed column-unary-level distinct() for query.distinct()
+        # in response to column-unary-level distinct() warning system.
+        q = fixture_session().query(User.id).distinct().order_by(User.id)
+        self.assert_compile(
+            q, "SELECT DISTINCT users.id AS users_id FROM users ORDER BY users.id"
+        )
+        eq_([(7,), (8,), (9,), (10,)], q.all())
 
     def test_standalone_w_subquery(self):
         User = self.classes.User
-        with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
-            q = fixture_session().query(distinct(User.id))
+        q = fixture_session().query(User.id).distinct()
 
-            subq = q.subquery()
-            q = fixture_session().query(subq).order_by(subq.c[0])
-            eq_([(7,), (8,), (9,), (10,)], q.all())
+        subq = q.subquery()
+        q = fixture_session().query(subq).order_by(subq.c[0])
+        eq_([(7,), (8,), (9,), (10,)], q.all())
 
     def test_no_automatic_distinct_thing_w_future(self):
         User = self.classes.User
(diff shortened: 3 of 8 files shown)


bekapono:
## Warning feedback for unary distinct outside aggregate function

### Overview
Address misuse of distinct functionality, and how that can result in runtime errors or unintended SQL statement results. The solution proposed focuses on a coercion time check that will issue a warning if misuse is detected.

### Context
User (tboddyspargo) reported #11526 where the SQL statement produced with MSSQL via pyodbc was incorrect, causing a runtime error. Maintainer (zzzeek) notified the user that they misused column-level `distinct()` and should instead use statement-level `distinct()`. Misuse of `distinct()` can cause runtime errors on certain dialects, or produce silent incorrect SQL statements on others. In #11518, two approaches were considered: a compile-time check and a coercion-time check. Discussion concluded with CaselIT opening #11526 for tracking. Column-level distinct is intended to be wrapped within an aggregate function, whereas statement-level `distinct()` is intended to be used for the whole statement i.e. `SELECT DISTINCT …`

### Goals & Non-goals
**Goals:**
- Detection of column-level `distinct()` used outside an aggregate function
- Feedback when users use column-level `distinct()` incorrectly

**Non-goals:**
- Existing behavior will not be broken — warning is advisory and does not raise an exception
- Will not address other `UnaryExpression` misuse patterns

### Proposed Solution
The warning check is implemented in `ColumnsClauseImpl._post_coercion`. By the time `_post_coercion` is triggered, the full column expression tree is constructed bottom-up.

To detect column-level `distinct()` used outside of an aggregate, `visitors.iterate(ent)` walks the expression tree top to bottom, checking for `UnaryExpression+distinct_op` and an aggregate function. If an aggregate function is not seen while `UnaryExpression+distinct_op` is present, a warning is issued. This also covers cases like `select(distinct("x") + 1)` where `UnaryExpression` is not the top level expression.

This logic is dependent on visitors.iterate() to traverse top-down. Doing so, we are checking
if an aggregate function has been seen prior to arriving to UnaryExpression+distinct_op

**Warning issued:**
> "column-expression-level unary DISTINCT clause used outside aggregate; use select().distinct() for full-statement DISTINCT."

### Alternatives Considered

**Trigger coercion check in `implicit_coercion` vs `post_coercion`:**
- Cons:  Would require introducing a new role purposefully excluded from the MRO — felt like a hack that could interfere with existing coercion implementations
- Pros: I felt it would of followed SQLAlchemy culture more closely based on other `implicit_coercion` patterns

**Top level `UnaryExpression+distinct_op` coercion check (best alternative):**
- Con: Does not validate that column-level `distinct()` is wrapped within an aggregate function
- Pro: Faster — no need to traverse the full expression tree
- Pro: Satisfies the original issue of column-level `distinct()` used in place of statement-level `distinct()`

**Creating a new role `FunctionDistinctRole`:**
- Con: Expression trees are constructed bottom-up, so parent context hasn't been constructed yet when `FunctionDistinctImpl` is triggered, making aggregate detection impossible

**Creating a new subclass of `UnaryExpression`:**
- Con: Without full knowledge of the codebase and future intentions for `distinct()`, this felt too invasive
- Pro: Would simplify detection by checking class type directly rather than inspecting `UnaryExpression` for its operator

test/dialect/mssql/test_compiler.py
```diff
@@ -445,6 +445,25 @@ def test_count(self):
             "tbl_row_count FROM sometable",
         )
 
+    # test-case 1 of 2 for mssql dialect - issue#11526
+    def test_unary_distinct_with_limit(self):
+        t = table("sometable", column("somecolumn"))
+        with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
```
zzzeek: these tests should all be modified to show correct use of column.distinct(), rather than adding a warning catch

test/orm/test_query.py
```diff
@@ -4631,19 +4631,21 @@ def test_basic_standalone(self):
         # issue 6008.  the UnaryExpression now places itself into the
         # result map so that it can be matched positionally without the need
         # for any label.
-        q = fixture_session().query(distinct(User.id)).order_by(User.id)
-        self.assert_compile(
-            q, "SELECT DISTINCT users.id FROM users ORDER BY users.id"
-        )
-        eq_([(7,), (8,), (9,), (10,)], q.all())
+        with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
+            q = fixture_session().query(distinct(User.id)).order_by(User.id)
```
zzzeek, line 4637: fix SQL to be correct, remove expect_warnings

test/orm/test_query.py
```diff
@@ -4631,19 +4631,21 @@ def test_basic_standalone(self):
...
+            self.assert_compile(
+                q, "SELECT DISTINCT users.id FROM users ORDER BY users.id"
+            )
+            eq_([(7,), (8,), (9,), (10,)], q.all())
 
     def test_standalone_w_subquery(self):
         User = self.classes.User
-        q = fixture_session().query(distinct(User.id))
+        with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
```
zzzeek: fix SQL to be correct, remove expect_warnings

test/sql/test_compiler.py
```diff
@@ -1941,32 +1941,44 @@ def test_true_short_circuit(self):
         )
 
     def test_distinct(self):
-        self.assert_compile(
-            select(table1.c.myid.distinct()),
-            "SELECT DISTINCT mytable.myid FROM mytable",
-        )
+        with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
+            self.assert_compile(
```
zzzeek: fix SQL to be correct, remove expect_warnings

test/sql/test_compiler.py
```diff
@@ -1941,32 +1941,44 @@ def test_true_short_circuit(self):
...
+                select(table1.c.myid.distinct()),
+                "SELECT DISTINCT mytable.myid FROM mytable",
+            )
 
-        self.assert_compile(
-            select(distinct(table1.c.myid)),
-            "SELECT DISTINCT mytable.myid FROM mytable",
-        )
+        with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
+            self.assert_compile(
```
zzzeek: fix SQL to be correct, remove expect_warnings

test/sql/test_compiler.py
```diff
@@ -1941,32 +1941,44 @@ def test_true_short_circuit(self):
...
+            )
 
-        self.assert_compile(
-            select(distinct(table1.c.myid)).set_label_style(
-                LABEL_STYLE_TABLENAME_PLUS_COL
-            ),
-            "SELECT DISTINCT mytable.myid FROM mytable",
-        )
+        with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
+            self.assert_compile(
```
zzzeek: fix SQL to be correct, remove expect_warnings

test/sql/test_compiler.py
```diff
@@ -1941,32 +1941,44 @@ def test_true_short_circuit(self):
...
         # "SELECT anon_2.anon_1 FROM (SELECT distinct mytable.myid
         # FROM mytable) AS anon_2"
-        self.assert_compile(
-            select(select(distinct(table1.c.myid)).subquery()),
-            "SELECT anon_2.anon_1 FROM (SELECT "
-            "DISTINCT mytable.myid AS anon_1 FROM mytable) AS anon_2",
-        )
+        with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
+            self.assert_compile(
+                select(select(distinct(table1.c.myid)).subquery()),
```
zzzeek, line 1965: fix SQL to be correct, remove expect_warnings

test/sql/test_compiler.py
```diff
@@ -1941,32 +1941,44 @@ def test_true_short_circuit(self):
...
+                "SELECT anon_2.anon_1 FROM (SELECT "
+                "DISTINCT mytable.myid AS anon_1 FROM mytable) AS anon_2",
+            )
+
+        # test-case 1 of 1 for issue#11526
+        with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
```
zzzeek: fix SQL to be correct, remove expect_warnings

test/sql/test_compiler.py
```diff
@@ -1941,32 +1941,44 @@ def test_true_short_circuit(self):
...
-            "DISTINCT mytable.myid AS anon_1 FROM mytable) AS anon_2",
-        )
+        with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
+            self.assert_compile(
+                select(select(distinct(table1.c.myid)).subquery()),
+                "SELECT anon_2.anon_1 FROM (SELECT "
+                "DISTINCT mytable.myid AS anon_1 FROM mytable) AS anon_2",
+            )
+
+        # test-case 1 of 1 for issue#11526
```
zzzeek: this isn't where we'd have tests for 11526.  test_query is for legacy orm.Query use, the coercion of distinct() should be tested individually in test_operators

test/sql/test_operators.py
```diff
@@ -5321,19 +5321,32 @@ def test_wraps_named_column_heuristic(self, py_op, sql_op, named):
...
             assert isinstance(expr, UnaryExpression)
 
-            self.assert_compile(
-                select(expr),
-                f"SELECT {sql_op}q",
-            )
-
+            if py_op is operators.distinct_op:
+                with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
+                    self.assert_compile(
```
zzzeek: id rather use valid SQL here than catch a warning

test/sql/test_operators.py
```diff
@@ -5321,19 +5321,32 @@ def test_wraps_named_column_heuristic(self, py_op, sql_op, named):
...
             expr = py_op(literal("x", Integer))
             assert isinstance(expr, UnaryExpression)
 
-            self.assert_compile(
-                select(expr),
-                f"SELECT {sql_op}:param_1 AS anon_1",
-            )
+            if py_op is operators.distinct_op:
+                with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
+                    self.assert_compile(
```
zzzeek: id rather use valid SQL here than catch a warning

test/sql/test_operators.py
```diff
@@ -5344,10 +5357,17 @@ def test_wraps_named_column_heuristic(self, py_op, sql_op, named):
             else:
                 assert isinstance(expr, UnaryExpression)
 
-            self.assert_compile(
-                select(expr),
-                f"SELECT {sql_op}:param_1 AS z",
-            )
+            if py_op is operators.distinct_op:
+                with testing.expect_warnings("column-expression-level unary.*DISTINCT"):
+                    self.assert_compile(
```
zzzeek: id rather use valid SQL here than catch a warning

test/sql/test_resultset.py
```diff
@@ -978,11 +978,12 @@ def test_column_accessor_unary(self, connection):
         connection.execute(users.insert(), dict(user_id=1, user_name="john"))
 
         # unary expressions
-        r = connection.execute(
-            select(users.c.user_name.distinct()).order_by(users.c.user_name)
-        ).first()
-        eq_(r._mapping[users.c.user_name], "john")
-        eq_(r.user_name, "john")
+        with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
+            r = connection.execute(
```
zzzeek: use valid SQL

test/sql/test_types.py
```diff
@@ -3943,12 +3943,13 @@ def test_expression_typing(self):
 
     def test_distinct(self, connection):
         test_table = self.tables.test
+        with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
+            s = select(distinct(test_table.c.avalue))
```
zzzeek: use valid SQL

test/sql/test_types.py
```diff
@@ -3943,12 +3943,13 @@ def test_expression_typing(self):
...
+            eq_(connection.execute(s).scalar(), 25)
 
-        s = select(distinct(test_table.c.avalue))
-        eq_(connection.execute(s).scalar(), 25)
-
-        s = select(test_table.c.avalue.distinct())
-        eq_(connection.execute(s).scalar(), 25)
+        with testing.expect_warnings("column-expression-level unary.*DISTINCT.*outside aggregate"):
+            s = select(test_table.c.avalue.distinct())
```
zzzeek: use valid SQL

lib/sqlalchemy/sql/coercions.py
```diff
@@ -1151,6 +1158,41 @@ def _raise_for_expected(
...
         )
 
+    @staticmethod
+    def _warn_unary_expression_outside_aggregate():
+        util.warn(
+            "column-expression-level unary DISTINCT clause used outside aggregate;"
+            " use select().distinct() for full-statement DISTINCT."
+        )
+
+    def _valid_unary_hierarchy(self, ent: ColumnElement[_T], op_name: str) -> bool:
```
zzzeek: I havent checked closely, but it looks like _valid_unary_hierarchy would be called for every element of a select() that is coercing on ColumnsClauseImpl, that is:

    select(a, b, c)

would call _valid_unary_hierarchy three times, no matter what a, b, c were, and would then call into at least two isinstance calls, as well as a whole visitors.iterate() call, etc.

unfortunately this approach cannot be taken here, this would add an order of magnitude of performance overhead to virtually all SQLAlchemy applications.

I have to apologize because I wasn't really thinking through a complete solution here when I said this:

> does "distinct()" only go inside of a SQL function ? That's all we'd have to look for. we can put "distinct()" into a special coercion category based on the roles, like FunctionDistinctRole.

Sorry, I think I was incorrect there, since the main thing we need to catch is when `distinct()` is used *outside* of a function and warn for that, and to do this at coercion would mean `distinct()` would have to be taken outside of the whole `UnaryExpressionElement` hierarchy so that it could be caught directly.   So I apologize if this caused a lot of time to be spent on the wrong approach.

When we need to warn for aspects of a SQL construct that depend on traversal of the hierarchy in context, this has to happen in the compiler, since that's where we're already traversing.  we also have a little bit more performance wiggle room there since compilations are cached (coercions are not).

That said here's a simple patch to get this behavior (needs polish):

```diff
diff --git a/lib/sqlalchemy/sql/compiler.py b/lib/sqlalchemy/sql/compiler.py
index f0e608e74..09cf3d196 100644
--- a/lib/sqlalchemy/sql/compiler.py
+++ b/lib/sqlalchemy/sql/compiler.py
@@ -3111,6 +3111,8 @@ class SQLCompiler(Compiled):
 
         text: str
 
+        kwargs["allow_distinct_op"] = True
+
         if disp:
             text = disp(func, **kwargs)
         else:
@@ -3258,6 +3260,13 @@ class SQLCompiler(Compiled):
             disp = self._get_operator_dispatch(
                 unary.operator, "unary", "operator"
             )
+            if unary.operator is operators.distinct_op and not kw.get(
+                "allow_distinct_op", False
+            ):
+                util.warn(
+                    "can't use distinct() outside of an aggregate function "
+                    "TODO improve this message"
+                )
             if disp:
                 return disp(unary, unary.operator, **kw)
             else:
```

Im not sure if I want to make this "allow_distinct_op" or "inside_aggregate_function", but this is a starting point

lib/sqlalchemy/sql/coercions.py
```diff
@@ -1138,6 +1138,13 @@ class ColumnsClauseImpl(_SelectIsNotFrom, _CoerceLiterals, RoleImpl):
 
     _guess_straight_column = re.compile(r"^\w\S*$", re.I)
 
+    _aggregate_functions = {
+        # Basic Aggregate function listed from GenericFunctions
+        'count', 'sum', 'avg',
```
zzzeek: we should allow DISTINCT inside of *all* functions for now, we can't predict what kinds of aggregate functions may exist on all platforms, and requiring functions to indicate "aggregate OK" is overkill.   we are just trying to prevent the simple case of `select(distinct(col))`.

zzzeek (changes requested):
hi -

thanks for this PR!    I think this is a good start, however I think we need to change the approach to use the compiler, rather than coercion.   the former should be much simpler and not have any performance issues.

test/sql/test_resultset.py
```diff
@@ -977,9 +977,8 @@ def test_column_accessor_unary(self, connection):
 
         connection.execute(users.insert(), dict(user_id=1, user_name="john"))
 
-        # unary expressions
```
zzzeek, line 980: this test wants to test unary expressions, so it should do that:

    select(func.max(users.c.user_name).distinct())

then the return result would be different since we changed the SQL

test/sql/test_types.py
```diff
@@ -3943,11 +3943,13 @@ def test_expression_typing(self):
 
     def test_distinct(self, connection):
         test_table = self.tables.test
+        s = select(test_table.c.avalue).distinct()
```
zzzeek, line 3946: this one's good.  tests all three varieties of distinct()

test/sql/test_compiler.py
```diff
@@ -1942,35 +1942,30 @@ def test_true_short_circuit(self):
...
-            "DISTINCT mytable.myid AS anon_1 FROM mytable) AS anon_2",
+            select(select(table1.c.myid).distinct().subquery()),
+            "SELECT anon_1.myid FROM (SELECT "
+            "DISTINCT mytable.myid AS myid FROM mytable) AS anon_1",
         )
 
         self.assert_compile(
-            select(table1.c.myid).distinct(),
-            "SELECT DISTINCT mytable.myid FROM mytable",
+            select(func.sum(distinct(table1.c.myid))),
```
zzzeek, line 1981: OK I was a little concerned we were losing the use of the distinct() standalone function in some of these tests but you added it here, however I dont see any tests at all for the "expect_warnings" case.

I think we should do this:


1. this function, `test_distinct()` gets renamed to `test_distinct_select_modifier()`
2. all statements in this function use `select().distinct()`
3. add a new method `test_distinct_function_6008()`, test issue #6008 exactly as it was, include the expect_warnings
4. add another new method `test_distinct_function()` that includes:
  a. tests for func.some_aggregate().distinct()
  b. tests for distinct(some_column) which include the expect_warnings

that way everything is covered here

test/sql/test_compiler.py
```diff
@@ -1942,35 +1942,30 @@ def test_true_short_circuit(self):
...
-            "SELECT DISTINCT mytable.myid FROM mytable",
+            "SELECT DISTINCT mytable.myid AS mytable_myid FROM mytable",
         )
 
         # the bug fixed here as part of #6008 is the same bug that's
         # in 1.3 as well, producing
         # "SELECT anon_2.anon_1 FROM (SELECT distinct mytable.myid
         # FROM mytable) AS anon_2"
         self.assert_compile(
-            select(select(distinct(table1.c.myid)).subquery()),
```
zzzeek, line 1966: OK for this one, in #6008, the test was **explicitly** for the standalone distinct function, so for this one, I'd actually leave it exactly as it was and add in the "expect_warnings" that we are using

zzzeek (changes requested):
hi  -

better, but we are still missing coverage for:

1. the #6008 test case (that likely emits a warning now)
2. an assortment of func.some_aggregate() with distinct() that does not warn
3. an assortment of some_column with distinct() that does warn

thanks!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision c6e8be7a49fecb1402dc249c84e7f982c1c34821 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change c6e8be7a49fecb1402dc249c84e7f982c1c34821: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6632

sqla-tester:

It looks ok to me.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6632

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6632 has been **merged**. Congratulations! :)

PR #13165, opened 2026-03-11
feat: add Prometheus metrics support for AsyncAdaptedQueuePool

benksih-code:
## Changes:
- Add optional Prometheus metrics support for AsyncAdaptedQueuePool (async connection pool)
- Metrics included:
  * : Number of idle connections in the pool
  * : Number of active (checked out) connections
  * : Number of requests waiting for a connection
  * : Total number of connections created
  * : Total number of connections destroyed
- Metrics are disabled by default, enabled via  pool parameter
- Requires  installed to use, no hard dependency
- Fully backwards compatible, no breaking changes to existing API

### DIFF
--- a/lib/sqlalchemy/pool/impl.py
+++ b/lib/sqlalchemy/pool/impl.py
@@ -23,6 +23,13 @@ from typing import TYPE_CHECKING
 from typing import Union
 import weakref
 
+try:
+    from prometheus_client import Gauge, Counter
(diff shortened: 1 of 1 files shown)


CaselIT:
Hi, 

Sorry but this is not something that we would accept, expecially without a proper issue or discussion first. 

Also I think the event system allows the addition of such metrics without needing to change the pool implementation

PR #13168, opened 2026-03-13
Drop the mutable default in inspect_formatargspec

tejas-ae:
Upon reviewing lib/sqlalchemy/util/compat.py, I noticed an opportunity for improvement. Drop the mutable default in inspect_formatargspec. Mutable defaults are shared across calls and usually turn into state leaks. I kept the patch small and re-ran syntax checks after applying it.

Backward-compat note: This changes a public function signature by replacing mutable defaults with None guards. Call sites stay source-compatible, but callers introspecting defaults will see None now.

### DIFF
--- a/lib/sqlalchemy/util/compat.py
+++ b/lib/sqlalchemy/util/compat.py
@@ -193,8 +193,8 @@ def inspect_formatargspec(
     varkw: Optional[str] = None,
     defaults: Optional[Sequence[Any]] = None,
     kwonlyargs: Optional[Sequence[str]] = (),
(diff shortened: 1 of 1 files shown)


CaselIT:
It seems like the api cold just be changed to accept none and default to that. No other code chanfe is needed in the function since it already create a new value on falsy value

zzzeek:
> Mutable defaults are shared across calls and usually turn into state leaks

only if they are actually mutated, which is not the case here

this code is ported from cpython standard lib

https://github.com/python/cpython/blob/5c02a39a0b31a330e06b4d6f44835afb205dc7cc/Lib/inspect.py#L1207-L1214

PR #13183, opened 2026-03-18
ci: add riscv64 wheel builds

gounthar:
Add riscv64 to the wheel build matrix using QEMU emulation on ubuntu-22.04.

## Changes

- Add `riscv64` to `linux_archs` matrix in `create-wheels.yaml`
- Exclude riscv64 from non-Linux runners (Windows, macOS, ARM)
- Add QEMU setup step for riscv64
- Add `riscv64` to `archs` in `pyproject.toml` cibuildwheel config

Built successfully from source on native riscv64 hardware (BananaPi F3, SpacemiT K1).

### DIFF
--- a/.github/workflows/create-wheels.yaml
+++ b/.github/workflows/create-wheels.yaml
@@ -35,6 +35,7 @@ jobs:
           # this is only meaningful on linux. windows and macos ignore exclude all but one arch
           - "aarch64"
           - "x86_64"
+          - "riscv64"
 
         include:
           # create pure python build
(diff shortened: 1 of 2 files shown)


.github/workflows/create-wheels.yaml
```diff
@@ -54,12 +55,26 @@ jobs:
...
+            linux_archs: "riscv64"
+          - os: "ubuntu-22.04-arm"
+            linux_archs: "riscv64"
 
       fail-fast: false
 
     steps:
       - uses: actions/checkout@v4
 
+      - name: Set up QEMU
```
CaselIT, line 72: so we are emulating risc on x64, like we had to for arm before arm runners?

I guess it should be fine, at worst only the risc wheel will fail so damage should be limited

CaselIT (approved):

CaselIT:
@zzzeek anything against merging this mike?

zzzeek:
this is just the create wheels so fine by me

zzzeek:
but we want to backport for 2.0 right?

CaselIT:
> but we want to backport for 2.0 right?

i think it would be fine to have this on 2.1 only

CaselIT:
yes, 2.0 is still supporting 3.7. Let's not

gounthar:
Thanks @CaselIT and @zzzeek for the quick merge! Great to see riscv64 wheels coming to SQLAlchemy.

.github/workflows/create-wheels.yaml
gounthar: @CaselIT Exactl, same pattern as ARM before native runners were available. The QEMU build will be slower but as you said, worst case only the riscv64 wheel fails.

If the build time becomes a concern, [RISE](https://github.com/apps/rise-risc-v-runners) is starting to provide free native riscv64 runners for open source projects. They already support numpy, llama.cpp, and pytorch. Happy to help set that up down the line if needed.

CaselIT:
2.1b2 should be close to be ready, so we will try it then

PR #13200, opened 2026-03-27
fix: PostgreSQL check constraint reflection strips brackets naively

### LINKED ISSUE #13157: PostgreSQL check constraint reflection strips brackets naively
### Describe the bug

_No response_

### SQLAlchemy Version in Use

2.0.48

psycopg2

PostgreSQL 17

### Python Version

3.14

### Operating system

macOS

### To Reproduce

```python
from sqlalchemy import (
    CheckConstraint,
    Column,
    Integer,
    MetaData,
    Table,
    create_engine,
    inspect,
)


def main() -> None:

    foo = Table(
        "foo",
        MetaData(),
        Column("id", Integer),
        Column("x", Integer),
        Column("y", Integer),
        CheckConstraint(
            name="broken",
        ),
    )

        inspector = inspect(conn)
        print(broken["sqltext"])


if __name__ == "__main__":
    main()
```

### Error

output:

```
```

### Additional context

_No response_

ViceSaber:
## Description
Fix PostgreSQL CHECK constraint reflection that naively strips brackets from constraint expressions, producing syntactically invalid SQL.

## Problem
When reflecting CHECK constraints from PostgreSQL, the code uses a greedy regex `(.+)` to strip outer parentheses. This regex naively matches the first `(` and last `)` in the expression, ignoring balanced inner parentheses.

For a constraint like `(x IS NULL OR y IS NULL) AND (x IS NULL OR y IS NULL)`, the regex strips the first `(` and last `)`, producing `x IS NULL OR y IS NULL) AND (x IS NULL OR y IS NULL` — syntactically invalid SQL.

## Solution
Replace the greedy regex with a proper balanced-parenthesis stripping function (`_strip_outer_parens`) that:
1. Only removes outer parentheses when the closing paren at depth 0 matches the opening paren
2. Preserves inner parenthesized subexpressions intact

## Tests
Verified with the exact case from the issue:
- Input: `(x IS NULL OR y IS NULL) AND (x IS NULL OR y IS NULL)`
- Before (buggy): `x IS NULL OR y IS NULL) AND (x IS NULL OR y IS NULL`
- After (fixed): `(x IS NULL OR y IS NULL) AND (x IS NULL OR y IS NULL)` ✅

Also verified:
- `((a > 1) AND (a < 5))` → `(a > 1) AND (a < 5)` ✅
- `(a > 1)` → `a > 1` ✅
- `a > 1` → `a > 1` (no parens, unchanged) ✅

Closes: #13157

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -1831,6 +1831,30 @@ from ...types import TEXT
 from ...types import UUID as UUID
 from ...types import VARCHAR
 
+
+def _strip_outer_parens(text):
+    """Strip only balanced outer parentheses from constraint text.
+
+    Unlike a naive greedy regex, this correctly handles expressions
+    like '(a > 1) AND (b < 2)' where inner parens must be preserved.
+    """
+    text = text.strip()
+    if not text.startswith("(") or not text.endswith(")"):
+        return text
+    depth = 0
+    for i, ch in enumerate(text):
+        if ch == "(":
+            depth += 1
+        elif ch == ")":
+            depth -= 1
+        if depth == 0:
+            # Found the closing paren matching the first opening paren
+            if i == len(text) - 1:
+                return text[1:-1].strip()
+            return text
+    return text
+
+
 IDX_USING = re.compile(r"^(?:btree|hash|gist|gin|[\w_]+)$", re.I)
 
 RESERVED_WORDS = {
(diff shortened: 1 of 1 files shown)


CaselIT (changes requested):
Hi,

This does not take into consideration string literals and has no tests.

I've looked around and there is already some piece of code that does this:
https://github.com/sqlalchemy/sqlalchemy/blob/a49a4f021ad9c9ff69f5482881b00f831d523a22/lib/sqlalchemy/dialects/sqlite/base.py#L2880-L2902

So the proposed solution here could be:
- extract that code into a function to place into `util` package
- use that in pg and sqlite
- see if there are already test for this in sqlite, if so we could reuse them also in pg. if not it may make sense to add them

PR #13225, opened 2026-04-08
postgresql: harden two-phase xid SQL literal quoting

c4tzzz:
### Description

  This PR introduces a defensive hardening change for PostgreSQL two-phase transaction SQL string construction.

  It replaces direct %s interpolation of xid in:

  - PREPARE TRANSACTION
  - COMMIT PREPARED
  - ROLLBACK PREPARED

  with SQL-quoted literals using existing internal helpers.

  Updated locations:

  - PostgreSQL base dialect twophase methods
  - PostgreSQL psycopg twophase command composition
  - PostgreSQL provision cleanup path (ROLLBACK PREPARED)

  Tests added:

  - verifies single-quote escaping in twophase xid literal rendering
  - verifies generated SQL command strings for base dialect and psycopg dialect paths

  This is intended as a hardening/defensive coding improvement; no behavior change is expected for normal valid xid
  values.

  ### Checklist

  This pull request is:

  - [x] A short code fix
      - Related to defensive hardening requested by maintainer feedback
      - Includes tests for command construction and quoting behavior

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -3687,19 +3687,21 @@ class PGDialect(default.DefaultDialect):
         self.do_begin(connection.connection)
 
     def do_prepare_twophase(self, connection, xid):
-        connection.exec_driver_sql("PREPARE TRANSACTION '%s'" % xid)
+        xid_literal = self._twophase_xid_literal(xid)
+        connection.exec_driver_sql(f"PREPARE TRANSACTION {xid_literal}")
 
     def do_rollback_twophase(
         self, connection, xid, is_prepared=True, recover=False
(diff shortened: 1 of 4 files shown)


CaselIT:
Hi,

We require an issue for these kind of PRs, that also shows the issue.
Furthermore I don't think it makes much sense to only fix this for postgresql, since similar issues may also be present in other dialects

c4tzzz:
If this isn't relevant, I'll close the request. Have a good day.

CaselIT:
it's not that it's not relevant, it's that we have a process that should be followed for contributions that aren't just simple typo/docs/etc fixes. This doesn't classify as a short code fix.

In any case I'll open an issue as reminder so that a better analysis could be done.

PR #13237, opened 2026-04-15
Fix 'compatiblity' typo in sqlite dialect docstring

MukundaKatta:
Docstring-only change: `cross-compatiblity` -> `cross-compatibility` in `lib/sqlalchemy/dialects/sqlite/base.py`.


CaselIT:
thanks!

PR #13239, opened 2026-04-16
Add mypy ignore comments for Python 3.14 module imports

### LINKED ISSUE #13240: mypy errors on Python 3.14 version-gated imports in compat.py
mypy reports `import-not-found` for `annotationlib` and `string.templatelib` when running on Python <3.14, because the version guard uses the `py314` variable instead of `sys.version_info`. mypy does not narrow on plain boolean variables — it checks both branches, so the 3.14-only imports are flagged as missing.

Adding `type: ignore` comments is not a solution either — on 3.14 they become `unused-ignore`.

**Expected:** mypy should pass cleanly on all supported Python versions without spurious errors on version-gated imports.

ashm-dev:
Fixes #13240

Use `sys.version_info >= (3, 14)` instead of the `py314` variable for
the py314-only imports block (`annotationlib`, `string.templatelib`).

mypy has built-in support for `sys.version_info` guards and only
type-checks the reachable branch, while a plain variable like `py314`
causes mypy to check both branches — leading to `import-not-found` on
<3.14 and `unused-ignore` on 3.14 if `type: ignore` comments are added.

### DIFF
--- a/lib/sqlalchemy/util/compat.py
+++ b/lib/sqlalchemy/util/compat.py
@@ -31,7 +31,6 @@ from typing import Sequence
 from typing import Set
 from typing import Tuple
 from typing import Type
-from typing import TYPE_CHECKING
 
 py314b1 = sys.version_info >= (3, 14, 0, "beta", 1)
 py314 = sys.version_info >= (3, 14)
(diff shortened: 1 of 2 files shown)


lib/sqlalchemy/util/compat.py
```diff
@@ -51,11 +50,23 @@
 
 dottedgetter = operator.attrgetter
 
-if py314 or TYPE_CHECKING:
+
+if sys.version_info >= (3, 14):
```
CaselIT, line 55: let's leave a not so it doesn't get reverted to py14.

something like `use version_info to make type-checker happy`

CaselIT (commented):
makes sense, i did notice those errors on 3.13 but I had not investigated.
Can you update the PR title and apply the suggestion?

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision f58bfc11694c71020b98ed69d8400a0173be096b of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change f58bfc11694c71020b98ed69d8400a0173be096b: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6655

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6655 has been **merged**. Congratulations! :)

PR #13255, opened 2026-04-22
Fix PostgreSQL check constraint reflection stripping parentheses incorrectly

### LINKED ISSUE #13157: PostgreSQL check constraint reflection strips brackets naively

armorbreak001:
## Problem

In `PGDialect.get_multi_check_constraints()`, the regex used to strip outer parentheses from CHECK constraint text uses a greedy match that corrupts nested expressions.

For example, a constraint defined as:
```sql
CHECK (((x IS NULL OR y IS NULL) AND (x IS NULL OR y IS NULL)))
```

gets reflected as:
```
x IS NULL OR y IS NULL) AND (x IS NULL OR y IS_NULL
```

The leading `(` and trailing content are dropped because the greedy `(.+)` captures from the first inner `(` to the last inner `)`, not the actual outer delimiters.

## Fix

Replace the single regex substitution with a depth-tracking loop that walks the string character-by-character. It only strips outer parens when they genuinely balance around the entire expression, and repeats to handle multiple wrapping levels (e.g. `((a > 1))` → `a > 1`).

Fixes #13157

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -5505,44 +5505,36 @@ class PGDialect(default.DefaultDialect):
             # "CHECK (((a\n < 1)\n OR\n (a\n >= 5))\n)"
             # "CHECK (a NOT NULL) NO INHERIT"
             # "CHECK (a NOT NULL) NO INHERIT NOT VALID"
+            # "CHECK (c != 'hi')"
 
-            m = re.match(
-                r"^CHECK *\((.+)\)( NO INHERIT)?( NOT VALID)?$",
-                src,
-                flags=re.DOTALL,
-            )
-            if not m:
+            _suffix = ""
+            if not src.startswith("CHECK ("):
                 util.warn("Could not parse CHECK constraint text: %r" % src)
                 sqltext = ""
             else:
-                sqltext = m.group(1)
-                # Strip outer parens iteratively; regex greedy match
-                # breaks on nested expressions like ((a) AND (b)).
-                while True:
-                    s = sqltext.strip()
-                    if len(s) >= 2 and s[0] == "(" and s[-1] == ")":
-                        depth = matched = 0
-                        for i, c in enumerate(s):
-                            depth += (c == "(") - (c == ")")
-                            if depth == 0:
-                                if i == len(s) - 1:
-                                    sqltext = s[1:-1]
-                                    matched = 1
-                                break
-                        if not matched:
-                            break
-                    else:
-                        break
+                try:
+                    end = util.find_parentheses_end(src, len("CHECK (") - 1)
+                except ValueError:
+                    util.warn(
+                        "Could not parse CHECK constraint text: %r" % src
+                    )
+                    sqltext = ""
+                else:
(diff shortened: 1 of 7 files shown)


CaselIT:
Hi,

There is already #13200 that has a comment. this has exactly the same issue.
That on seems stalled so if you want to apply the [comment](https://github.com/sqlalchemy/sqlalchemy/pull/13200#pullrequestreview-4022937474) in the other review feel free to continue

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -5505,29 +5505,36 @@ def get_multi_check_constraints(
...
+            _suffix = ""
+            if not src.startswith("CHECK ("):
                 util.warn("Could not parse CHECK constraint text: %r" % src)
                 sqltext = ""
             else:
-                sqltext = re.compile(
-                    r"^[\s\n]*\((.+)\)[\s\n]*$", flags=re.DOTALL
-                ).sub(r"\1", m.group(1))
+                try:
+                    end = util.find_parentheses_end(src, len("CHECK (") - 1)
```
zzzeek: it would be better if there was a single function "consume_parenthesized_expression()" that did both of these steps at once.

lib/sqlalchemy/dialects/sqlite/base.py
```diff
@@ -2876,31 +2876,21 @@ def get_check_constraints(self, connection, table_name, schema=None, **kw):
...
-                                {"sqltext": sqltext, "name": constraint_name}
-                            )
-                            break
+
+            try:
+                close_end = util.find_parentheses_end(
+                    table_data, start - 1
+                )
+            except ValueError:
+                continue
```
zzzeek, line 2889: what does ValueError mean?

zzzeek (changes requested):
I would want this to be cleaned up, it looks a little vibe-coded at the moment

armorbreak001:
Hi @zzzeek, thanks for the review. I've addressed all three points in 84caeda:

1. **Single function**: Added `consume_parenthesized_expression(text, start)` that combines finding the matching close paren and stripping outer wrapping into one call. The PG dialect now uses this instead of the two-step `find_parentheses_end` + `strip_parentheses` approach.

2. **ValueError**: Added a comment in the SQLite dialect explaining that ValueError means unbalanced parentheses (no matching close paren found), in which case we skip that constraint.

3. **Cleanup**: Removed the two-step logic in favor of the single function call, making the PG dialect code more straightforward.

PR #13258, opened 2026-04-23
fix: remove parentheses around SELECT with LIMIT/OFFSET in SQLite UNION ALL

### LINKED ISSUE #5564: Parentheses Inserted for UNION ALL Causes Syntax Error in SQLite
**Describe the bug**
SQLAlchemy produces unwanted parentheses `()` when using `UNION ALL` which causes syntax errors in SQLite.

**Expected behavior**
`UNION ALL` clause produced without extra/wrapping parentheses.

**To Reproduce**

```py
q = select([0]).union_all(select([0]).limit(30))
```
Produces:
```sql
SELECT ? UNION ALL (SELECT 0 LIMIT ? OFFSET ?)
```
(which leads to the error below). 

The following SQL **should** be produced instead:
```sql
SELECT ? UNION ALL SELECT ? LIMIT ? OFFSET ?
```

**Error**

```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) near "(": syntax error
```

**Versions.**
 - OS: macOS v10.15.6
 - Python: 3.8
 - SQLAlchemy: v1.3.19
 - Database: SQLite v3.31.1
 - DBAPI: SQLite

**Additional context**
- SQLAlchemy does not generate parentheses if no `LIMIT` clause is provided.

namitjain11:
Fixes #5564

## Problem
SQLite does not allow parentheses around SELECT statements in a compound 
select (UNION, UNION ALL) when they have LIMIT or OFFSET clauses.

This produced invalid SQL like:
SELECT 0 UNION ALL (SELECT 0 LIMIT 30)

## Fix
Override visit_compound_select in SQLiteCompiler to strip the unwanted 
parentheses that are added by the default compiler.

Now produces correct SQL:
SELECT 0 UNION ALL SELECT 0 LIMIT 30

## Tests
Added test_union_all_with_limit_no_parens to test_compiler.py
All 57 existing tests pass.

### DIFF
--- a/lib/sqlalchemy/dialects/sqlite/base.py
+++ b/lib/sqlalchemy/dialects/sqlite/base.py
@@ -1712,6 +1712,27 @@ class SQLiteCompiler(compiler.SQLCompiler):
         and_ = self._generate_generic_binary(binary, " & ", **kw)
         return f"({or_} - {and_})"
 
+    def visit_compound_select(
+        self, cs, asfrom=False, compound_index=None, **kwargs
+    ):
+        text = super().visit_compound_select(
+            cs, asfrom=asfrom, compound_index=compound_index, **kwargs
+        )
+
+        # SQLite does not allow parentheses around SELECT statements in a
+        # compound select (UNION, etc.) when they have LIMIT/OFFSET.
+        # Strip the wrapping parentheses from such statements.
+        # Match (SELECT ... LIMIT/OFFSET ...) and remove outer parens
+        text = re.sub(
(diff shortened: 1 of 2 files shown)


CaselIT (changes requested):
Hi,

I may be wrong here but I don't the two expressions are equivalent. See https://sqlite.org/lang_select.html#compound_select_statements

the first one applies the limit to the `SELECT 0 LIMIT 30` while the second one applies the limit to the union. the two are not equivalent

Issue #10673, opened 2023-11-22
Try supporting protocols that include `Mapped` with type union


CaselIT:
The following code reports errors:
```py
from typing import Protocol
from uuid import UUID, uuid4

import sqlalchemy as sa
from sqlalchemy.orm import DeclarativeBase, Mapped, declared_attr


class ModelBase(DeclarativeBase):
    pass


class CompareProtocol(Protocol):
    id: Mapped[int | UUID]  # Or `Mapped[int] | Mapped[UUID]`


class CompareMixin:
    def compare(self: CompareProtocol, other: CompareProtocol) -> bool:
        return self.id == other.id  # Simplified example


class IntIdMixin:
    @declared_attr
    def id(cls) -> Mapped[int]:
        return sa.orm.mapped_column(sa.Integer, primary_key=True)


class UuidIdMixin:
    @declared_attr
    def id(cls) -> Mapped[UUID]:
        return sa.orm.mapped_column(sa.UUID, primary_key=True, default=uuid4)


class MyModel(CompareMixin, IntIdMixin, ModelBase):
    pass


m1 = MyModel()
m2 = MyModel()
m1.compare(m2)  # Mypy reports two errors here
```

This errors with both mypy and pylance.

jace:
Related Mypy bug: https://github.com/python/mypy/issues/16709

Working solution:

1. Use `declared_attr[type]` in the Protocol, not `Mapped[type]`.
2. Don't use `@declared_attr` with the decorator form. Instead, use it to create an attribute:
    ```python
    class MyMixin:
        @classmethod
        def __foo(cls) -> Mapped[...]: ...
        foo = declared_attr(__foo)
        del __foo
    ```

Possible resolutions in SQLAlchemy:

1. Make `declared_attr` generic on a covariant, matching the change to `Mapped` and related types in #10288 and #10289 (where `declared_attr` was missed).
2. Make `declared_attr` a sub-class of `Mapped` so it will match in protocols. Possibly also needed for `hybrid_property` since that's the other decorator, and decorators can't be typed like attributes.
3. Suggest the non-decorator form in documentation as a workaround when needing compatibility with Protocol. I'm not optimistic about the Mypy issue being resolved in the short term because this problem is present in all versions from 0.7 to 1.8 and appears to be fairly deep into how Mypy works. Pyright doesn't have this problem.

CaselIT:
[...]

PR #13266, opened 2026-04-28
Fixes: #10673: make declared_attr covariant

luizfneves404:
### Description

I made declared_attr covariant as suggested in #10673. mypy didn't seem to complain. Added a regression test for the use case that was asked for. Unfortunately, it seems like using `Mapped[int | UUID]` directly in the Protocol won't work:

```python
class CompareProtocol(Protocol):
    id: Mapped[int | UUID]
```

Because mypy will see this as a settable variable and not as a SQLAlchemy descriptor. Using `@property` instead seems to work and it's what I used in the test (perhaps it should be documented as the way to achieve this?):

 
```python
class CompareProtocol(Protocol):
    @property
    def id(self) -> Mapped[int | UUID]: ...
```

### Checklist

This pull request is:

- [X] A short code fix
	- please include tests.


doc/build/changelog/unreleased_20/10673.rst
```diff
@@ -0,0 +1,11 @@
+.. change::
```
CaselIT: we don't usually add changenotes for typing changes.

let's see what mike prefers before removing. cc @zzzeek

lib/sqlalchemy/orm/decl_api.py
```diff
@@ -96,6 +96,7 @@
     from ..util.typing import _MatchedOnType
 
 _T = TypeVar("_T", bound=Any)
+_T_co = TypeVar("_T_co", bound=Any, covariant=True)
```
CaselIT, line 99: maybe as some point we could use the shared typevars, but it's fine for now

CaselIT (commented):
looks fine, thanks!

doc/build/changelog/unreleased_20/10673.rst
luizfneves404: I see, thanks for pointing it out! It's my first PR ever so I'm not used to contributing to SQLAlchemy. happy to adjust

CaselIT: asked mike, let's remove this

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 8edd2841f4bbe61f8bb9bc15a7a57e0560698779 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 8edd2841f4bbe61f8bb9bc15a7a57e0560698779: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6669

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6669 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6671 has been **merged**. Congratulations! :)

PR #13288, opened 2026-05-12
fix(pool): use time.monotonic() instead of time.time() for connection timestamps

### LINKED ISSUE #13169: Connection pool soft invalidation uses `time.time()` comparison that can silently fail on low-resolution clocks (e.g. Windows ~16 ms)
In `lib/sqlalchemy/pool/base.py`, `_ConnectionRecord.get_connection` detects soft invalidation via:

```python
# base.py, line 852
elif self._soft_invalidate_time > self.starttime:
    do_connect = True
```

`_soft_invalidate_time` is set with `time.time()` (line 812) and `starttime` is also set with `time.time()` (line 893 in `__connect`). The code already acknowledges that `time.time()` can have ~16 ms granularity on Windows (comments at lines 820–831), but this caveat applies equally to soft invalidation.

**Scenario where this can fail:**
1. Connection `A` is created — `starttime = time.time()` → say `T`
2. `soft_invalidate()` is called in the same timer quantum → `_soft_invalidate_time = time.time()` → also `T`
3. The check `T > T` is `False` → the connection is **not** recycled even though it should be

This is the same class of issue as the hard invalidation check, but `_soft_invalidate_time` is only set to `time.time()` directly (no `+1` or fuzz), whereas some backends use monotonic clocks or add epsilon values for robustness.

**Questions:**
1. Should `_soft_invalidate_time` be set to `time.time() + 1e-6` (or similar) to guarantee `>` rather than `==` when created in the same timer tick as the connection?
2. Would `time.monotonic()` be more appropriate here to avoid wall-clock regression issues in addition to the precision concern?
3. Is there a test that exercises soft invalidation with a mocked clock to catch this edge case?

`lib/sqlalchemy/pool/base.py`, lines 812, 852–858, 893.

algojogacor:
### Description

Replaced all `time.time()` calls in `pool/base.py` with `time.monotonic()` for connection record timestamp tracking. Also updated the comment block that documented the Windows precision caveat.

`time.time()` on Windows has approximately 16 ms granularity, which can cause soft invalidation checks to fail when `_soft_invalidate_time` and `starttime` are captured in the same timer quantum. `time.monotonic()` is monotonic by definition and provides sub-millisecond precision on all platforms, so the `>` comparisons between timestamps work correctly regardless of platform timer resolution.

Four sites changed (all in `lib/sqlalchemy/pool/base.py`):
- `Pool._invalidate_time` assignment (line 407)
- `_ConnectionRecord._soft_invalidate_time` assignment (line 812)
- Connection recycle time check in `get_connection` (line 838)
- `_ConnectionRecord.starttime` assignment in `__connect` (line 893)

All comparisons are relative (timestamp vs timestamp or timestamp delta vs timeout), so switching to `time.monotonic()` preserves existing behavior while fixing the precision issue.

- [x] A short code fix
  - Fixes: #13169
  - Existing pool invalidation tests cover the comparison logic

### DIFF
--- a/lib/sqlalchemy/pool/base.py
+++ b/lib/sqlalchemy/pool/base.py
@@ -404,7 +404,7 @@ class Pool(log.Identified, event.EventTarget):
         """
         rec = getattr(connection, "_connection_record", None)
         if not rec or self._invalidate_time < rec.starttime:
-            self._invalidate_time = time.time()
+            self._invalidate_time = time.monotonic()
         if _checkin and getattr(connection, "is_valid", False):
             connection.invalidate(exception)
 
@@ -809,7 +809,7 @@ class _ConnectionRecord(ConnectionPoolEntry):
             )
 
         if soft:
-            self._soft_invalidate_time = time.time()
+            self._soft_invalidate_time = time.monotonic()
         else:
             self.__close(terminate=True)
             self.dbapi_connection = None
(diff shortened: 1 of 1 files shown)


lib/sqlalchemy/pool/base.py
```diff
@@ -809,33 +809,28 @@ def invalidate(
...
-        # within 16 milliseconds accuracy, so unit tests for connection
-        # invalidation need a sleep of at least this long between initial start
-        # time and invalidation for the logic below to work reliably.
+        # NOTE: the various comparisons here compare time.monotonic()
+        # values which are guaranteed to be monotonic and have
+        # sub-millisecond precision across all platforms. This resolves
+        # the issue where time.time() could have ~16 ms granularity
+        # on Windows, causing comparisons like _soft_invalidate_time >
+        # starttime to fail when both timestamps were captured within
+        # the same timer quantum.
```
zzzeek, line 826: I dont want to lose the note about how >= would eliminate the issue, but would run a theoretical risk of constant re-connects if "time" were not incrementing.

zzzeek (changes requested):
I want to do what I propose [here](https://github.com/sqlalchemy/sqlalchemy/issues/13169#issuecomment-4060550530) which is dont hardcode time functions that are used in this way:

1. note that we also use `time.time()` in [queue.py](https://github.com/sqlalchemy/sqlalchemy/blob/1e1c0084b1804eaae8b7f089435241a2b8f4be60/lib/sqlalchemy/util/queue.py#L27) ; [modern cpython correctly uses monotonic](https://github.com/python/cpython/blob/f5fb491341e566bbaf17d9bf3e4ec3af4a56bb3f/Lib/queue.py#L7)
2. other places we use time() for cache invalidation: https://github.com/sqlalchemy/sqlalchemy/blob/1e1c0084b1804eaae8b7f089435241a2b8f4be60/lib/sqlalchemy/testing/util.py#L522-L534 and three places in asyncpg such as https://github.com/sqlalchemy/sqlalchemy/blob/1e1c0084b1804eaae8b7f089435241a2b8f4be60/lib/sqlalchemy/dialects/postgresql/asyncpg.py#L777

steps:

1. add new library function `sqlalchemy.util.compat.highres_clock()` (or whatever name, but the point is it should be named for the use case, not the implementation)
2. apply to pool/base.py, postgresql/asyncpg.py, util/queue.py, testing/util.py
3. this will be for 2.1 and i think we should have a short changelog
4. thanks!

zzzeek:
no activity here in response to my review so at some point ill get around to what i wanted to do for this issue

PR #13289, opened 2026-05-12
fix: raise clear error when default_factory/default passed to WriteOnlyMapped/DynamicMapped relationship

### LINKED ISSUE #13227: WriteOnlyMapped: default_factory=list broken in 2.1 due to new declarative_scan validation
### Describe the bug

The new `default_factory` validation introduced in 
commit [95ad9fe](https://github.com/sqlalchemy/sqlalchemy/commit/95ad9fe42b888d7cdea60b7a71c833cbb8667887)
(related to #12168) breaks `WriteOnlyMapped` relationships that use `default_factory=list`
in Declarative Dataclass Mapping.

_No response_

### SQLAlchemy Version in Use

2.1.0b1

asyncpg

PostgreSQL 18

### Python Version

3.14.2

### Operating system

Linux

### To Reproduce

```python
from sqlalchemy.orm import DeclarativeBase, Mapped, MappedAsDataclass, WriteOnlyMapped, mapped_column, relationship


class Base(DeclarativeBase, MappedAsDataclass):
    pass


class Parent(Base):
    __tablename__ = "parent"

    children: WriteOnlyMapped["Child"] = relationship(
        back_populates="parent",
        default_factory=list,  # ← raises ArgumentError in 2.1
        cascade="all, delete-orphan",
        passive_deletes=True,
    )


class Child(Base):
    __tablename__ = "child"
    parent_id: Mapped[int] = mapped_column(foreign_key="parent.id", ondelete="CASCADE")
    parent: Mapped[Parent] = relationship(back_populates="children")
```

### Error

  File "/home/.../project/sqla_relationship_bug.py", line 8, in <module>
    class Parent(Base):
    ...<8 lines>...
        )
  File "/home/.../project/.venv/lib/python3.14/site-packages/sqlalchemy/orm/decl_api.py", line 867, in __init_subclass__
    _ORMClassConfigurator._as_declarative(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        cls._sa_registry, cls, cls.__dict__
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/.../project/.venv/lib/python3.14/site-packages/sqlalchemy/orm/decl_base.py", line 289, in _as_declarative
    return _DeclarativeMapperConfig(registry, cls_, dict_)
  File "/home/.../project/.venv/lib/python3.14/site-packages/sqlalchemy/orm/decl_base.py", line 1013, in __init__
    self._extract_mappable_attributes()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/.../project/.venv/lib/python3.14/site-packages/sqlalchemy/orm/decl_base.py", line 1590, in _extract_mappable_attributes
    value.declarative_scan(
    ~~~~~~~~~~~~~~~~~~~~~~^
        self,
        ^^^^^
    ...<7 lines>...
        is_dataclass,
        ^^^^^^^^^^^^^
    )
    ^
  File "/home/.../project/.venv/lib/python3.14/site-packages/sqlalchemy/orm/relationships.py", line 1920, in declarative_scan
    raise sa_exc.ArgumentError(
    ...<3 lines>...
    )
sqlalchemy.exc.ArgumentError: For relationship Parent.children using dataclass options, default_factory must be exactly None

### Additional context

## Root Cause

The new validation in relationships.py (added in commit 95ad9fe) checks:

```python
if (
    self._attribute_options.dataclasses_default_factory
    is not _NoArg.NO_ARG
    and self._attribute_options.dataclasses_default_factory
    is not self.collection_class  # ← None for WriteOnlyMapped
):
    raise sa_exc.ArgumentError(...)
```

For `WriteOnlyMapped`, `collection_class` is `None` because Write-Only collections
define their collection type through the annotation itself, not through `collection_class`.
This causes `default_factory=list` to always fail the `is not self.collection_class` check.

The same pattern is already correctly handled just a few lines above for the `uselist` check:

```python
if (
    self.collection_class is None
    and not is_write_only   # ← already guards write-only here
    and not is_dynamic
):
    self.uselist = False
```

## Proposed Fix

Add the same `is_write_only` and `is_dynamic` guards to the new validation:

```python
if (
    self._attribute_options.dataclasses_default_factory
    is not _NoArg.NO_ARG
    and self._attribute_options.dataclasses_default_factory
    is not self.collection_class
    and not is_write_only   # ← add this
    and not is_dynamic      # ← add this
):
    raise sa_exc.ArgumentError(...)
```

algojogacor:
### Description

Fixes: #13227

When `default_factory=list` or `default` is passed to a `WriteOnlyMapped` or `DynamicMapped` relationship, it previously caused cryptic crashes (e.g., `TypeError: LoaderCallableStatus object is not iterable` in `_WriteOnlyAttributeImpl.set()`). This happens because these mapped types have no in-memory collection, so `default_factory` and `default` are not meaningful.

This PR adds an explicit `ArgumentError` in `RelationshipProperty.__init__()` that fires when `default_factory` or `default` is set on a write-only or dynamic relationship, with a clear message:

> WriteOnlyMapped / DynamicMapped relationship {name} does not support 'default' or 'default_factory' properties. Use 'init=False' to exclude this attribute from \_\_init\_\_.

**Changes:**
- `lib/sqlalchemy/orm/relationships.py`: Added check for `is_write_only or is_dynamic` with `dataclasses_default_factory`/`dataclasses_default` raising `ArgumentError`. Also bypassed early default rejection in `_RelationshipDeclared` for write_only/dynamic lazy so the specific error fires.
- `test/orm/declarative/test_typed_mapping.py`: Added tests for both WriteOnlyMapped and DynamicMapped with `default_factory` and `default`.

### Checklist

- [x] Tests added for WriteOnlyMapped + DynamicMapped with both default_factory and default
- [x] Single consolidated error message per review feedback
- [x] Please include: `Fixes: #13227` in the commit message

### DIFF
--- a/lib/sqlalchemy/orm/relationships.py
+++ b/lib/sqlalchemy/orm/relationships.py
@@ -1849,29 +1849,20 @@ class RelationshipProperty(
             if (
                 self._attribute_options.dataclasses_default_factory
                 is not _NoArg.NO_ARG
-            ):
-                raise sa_exc.ArgumentError(
-                    f"For relationship "
-                    f"{self._format_as_string(cls, key)}: "
-                    "'default_factory' is not supported for "
-                    "WriteOnlyMapped relationships, which have "
-                    "no in-memory collection. Use 'init=False' "
-                    "to exclude this attribute from __init__ "
-                    "and add items after construction via "
-                    "direct assignment or "
-                    "WriteOnlyCollection.add_all()."
-                )
-            if (
+            ) or (
                 self._attribute_options.dataclasses_default
                 is not _NoArg.NO_ARG
             ):
                 raise sa_exc.ArgumentError(
                     f"For relationship "
                     f"{self._format_as_string(cls, key)}: "
-                    "'default' is not supported for "
+                    "'default' and 'default_factory' are not supported for "
                     "WriteOnlyMapped relationships, which have "
                     "no in-memory collection. Use 'init=False' "
-                    "to exclude this attribute from __init__."
+                    "to exclude this attribute from __init__ "
+                    "and add items after construction via "
+                    "direct assignment or "
+                    "WriteOnlyCollection.add_all()."
                 )
 
         argument = de_optionalize_union_types(argument)
--- a/lib/sqlalchemy/orm/relationships.py
+++ b/lib/sqlalchemy/orm/relationships.py
@@ -1845,6 +1845,35 @@ class RelationshipProperty(
         else:
             is_write_only = is_dynamic = False
 
+        if is_write_only:
+            if (
+                self._attribute_options.dataclasses_default_factory
+                is not _NoArg.NO_ARG
+            ):
+                raise sa_exc.ArgumentError(
+                    f"For relationship "
+                    f"{self._format_as_string(cls, key)}: "
+                    "'default_factory' is not supported for "
+                    "WriteOnlyMapped relationships, which have "
+                    "no in-memory collection. Use 'init=False' "
+                    "to exclude this attribute from __init__ "
+                    "and add items after construction via "
+                    "direct assignment or "
+                    "WriteOnlyCollection.add_all()."
+                )
+            if (
+                self._attribute_options.dataclasses_default
+                is not _NoArg.NO_ARG
+            ):
+                raise sa_exc.ArgumentError(
+                    f"For relationship "
+                    f"{self._format_as_string(cls, key)}: "
+                    "'default' is not supported for "
+                    "WriteOnlyMapped relationships, which have "
+                    "no in-memory collection. Use 'init=False' "
+                    "to exclude this attribute from __init__."
+                )
+
         argument = de_optionalize_union_types(argument)
 
         if hasattr(argument, "__origin__"):
--- a/lib/sqlalchemy/orm/relationships.py
+++ b/lib/sqlalchemy/orm/relationships.py
@@ -464,10 +464,11 @@ class RelationshipProperty(
             _NoArg.NO_ARG,
             None,
         ):
-            raise sa_exc.ArgumentError(
-                "Only 'None' is accepted as dataclass "
-                "default for a relationship()"
-            )
+            if lazy not in ("write_only", "dynamic"):
+                raise sa_exc.ArgumentError(
+                    "Only 'None' is accepted as dataclass "
+                    "default for a relationship()"
+                )
 
         self.post_update = post_update
         self.viewonly = viewonly
@@ -1845,7 +1846,7 @@ class RelationshipProperty(
         else:
             is_write_only = is_dynamic = False
 
-        if is_write_only:
+        if is_write_only or is_dynamic:
             if (
                 self._attribute_options.dataclasses_default_factory
                 is not _NoArg.NO_ARG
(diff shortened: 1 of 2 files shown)


lib/sqlalchemy/orm/relationships.py
```diff
@@ -1845,6 +1845,35 @@ def declarative_scan(
         else:
             is_write_only = is_dynamic = False
 
+        if is_write_only:
```
CaselIT: @zzzeek do default / default_factory make sense with dynamic? if I'm not mistaken it should be similar to write only

lib/sqlalchemy/orm/relationships.py
```diff
@@ -1845,6 +1845,35 @@ def declarative_scan(
...
+            if (
+                self._attribute_options.dataclasses_default_factory
```
CaselIT, line 1851: I think we could use a single error, there isn't really a need to duplicate them here.

Something like

> WriteOnlyMapped relationship {self._format_as_string(cls, key)} does not support 'default' or 'default_factory' properties. Use 'init=False' to exclude this attribute from __init__

test/orm/declarative/test_typed_mapping.py
```diff
@@ -4572,6 +4572,27 @@ class A(decl_base):
 
         self._assertions(A, B, "write_only")
 
+    def test_write_only_default_factory_raises(self, decl_base):
```
CaselIT, line 4575: would be nice to also test the default

CaselIT (changes requested):
thanks, left a few suggestions. let's also wait mike input on whatever we should include also dynamic here

algojogacor:
Thanks for the review! Addressed both suggestions:
- Combined the `default` and `default_factory` error checks into a single `ArgumentError` with a unified message
- Added `test_write_only_default_raises` covering the `default=[]` case

Waiting for Mike's input on the dynamic question.

lib/sqlalchemy/orm/relationships.py
zzzeek: dynamic is not *exactly* the same as write_only but yes, these are descriptors that probably shouldnt be considered within the dataclass regime, unless someone can show us a use case for that

lib/sqlalchemy/orm/relationships.py
```diff
@@ -1845,6 +1845,26 @@ def declarative_scan(
...
+                is not _NoArg.NO_ARG
+            ) or (
+                self._attribute_options.dataclasses_default
+                is not _NoArg.NO_ARG
+            ):
+                raise sa_exc.ArgumentError(
+                    f"For relationship "
+                    f"{self._format_as_string(cls, key)}: "
+                    "'default' and 'default_factory' are not supported for "
+                    "WriteOnlyMapped relationships, which have "
```
zzzeek: yeah would mention dynamic here also

lib/sqlalchemy/orm/relationships.py
```diff
@@ -1845,6 +1845,26 @@ def declarative_scan(
...
+                    "no in-memory collection. Use 'init=False' "
+                    "to exclude this attribute from __init__ "
+                    "and add items after construction via "
+                    "direct assignment or "
```
zzzeek: this message is a bit wordy.  if there's no way to make it more succinct consider linking to errors.rst (see the "code" parameter of the ArgumentError class hierarchy)

test/orm/declarative/test_typed_mapping.py
```diff
@@ -4572,6 +4572,48 @@ class A(decl_base):
...
```
zzzeek, line 4575: I'd like to move these tests.  test_typed_mapping is not specific to dataclasses nor writeonly/dynamic.  

these tests should be either in:

test_dc_transforms.py

or

test_dynamic.py

and there should be tests for both the writeonly and dynamic cases

zzzeek (changes requested):

zzzeek:
so there's still a lot unaddressed here and no activity for the past three months, we can address this locally so closing

Issue #9468, opened 2023-03-10
Implementation for get_unique_constraints for MS SQL Server


GeorgeSchneelochVA:
### Describe the use case

It would be useful to have the method [Inspector.get_unique_constraints](https://github.com/sqlalchemy/sqlalchemy/blob/rel_2_0_5/lib/sqlalchemy/engine/interfaces.py#L1562) implemented for SQL Server. My immediate use case is that alembic [uses this method](https://github.com/sqlalchemy/alembic/blob/rel_1_10_2/alembic/autogenerate/compare.py#L540) to detect unique constraints, and it would be useful to be able to detect these changes. 

### Databases / Backends / Drivers targeted

SQL Server 2019

### Example Use

This SQL lists all unique constraints for a table along with column names. You could filter on the table name and schema to get a list that could be turned into a dictionary similar to how [postgres does it](https://github.com/sqlalchemy/sqlalchemy/blob/rel_2_0_5/lib/sqlalchemy/dialects/postgresql/base.py#L4382).

    SELECT
      indexes.name,
      OBJECT_NAME(indexes.object_id) AS table_name,
      OBJECT_SCHEMA_NAME(indexes.object_id) AS 'schema',
      STRING_AGG(COL_NAME(index_columns.object_id, index_columns.column_id), ', ') AS column_names,
      type_desc,
      is_unique,
      is_unique_constraint,
      is_primary_key
    FROM sys.indexes AS indexes
    JOIN sys.index_columns AS index_columns
    ON indexes.object_id = index_columns.object_id
    AND indexes.index_id = index_columns.index_id
    WHERE is_unique_constraint = 1
    GROUP BY
      indexes.object_id,
      indexes.name,
      type_desc,
      is_unique,
      is_unique_constraint,
      is_primary_key

zzzeek:
I've not had a good understanding of this issue as it's not clear how distinct unique constraints are from unique indexes in SQL server or if there is even such a distinct concept.

can you try out get_indexes() and see if there is something we are missing?

zzzeek:
cc @CaselIT  @gordthompson

GeorgeSchneelochVA:
As far as I can tell the feature set of unique indexes and unique constraints is the same. However they are technically separate concepts in the database and alembic ignores a `UniqueConstraint` while it detects a unique `Index`. It would be useful to support both so that if a person adds a `UniqueConstraint` to a model, it can be detected via `flask db check` or `flask db migrate`.

[...]

PR #13297, opened 2026-05-14
Implement native multi-table reflection API for the mssql dialect

### LINKED ISSUE #8430: Implement multi-reflection api in MSSQL
When creating the multi reflection api in db08a699489c9b0259579d7ff7fd6bf3496ca3a2 only postgresql and oracle where reworked to implement the new api.
The other internal dialect gained support using the general fallback, but mssql could probably support a native implementation that should be faster that the fallback. (mysql/mariadb and sqlite use describe so would be poor candidates for a native implementation)

bewithgaurav:
### Description

Adds 5 native `get_multi_*` reflection methods (columns, pk, fk, indexes, table_comment) for the MSSQL dialect, replacing the per-table loop in `_default_multi_reflect`. Single-table methods now delegate to the multi versions (PG/Oracle pattern); legacy per-table SQL is retained as `_internal_get_*` helpers, used only for tempdb reflection.

Not implemented here: `get_multi_unique_constraints`, `get_multi_check_constraints`, `get_multi_table_options` -MSSQL has no single-table counterparts to delegate from. Happy to add as a follow-up.

### Performance

Measured with `test/perf/many_table_reflection.py` against SQL Server 2022 (Docker, localhost, pyodbc + ODBC Driver 18) on a 250-table fixture, 15-50 cols, with PKs/FKs/indexes/comments:

| | single | multi | speedup |
|---|---|---|---|
| `get_columns` | 1.33s | 0.35s | 3.8x |
| `get_pk_constraint` | 1.41s | 0.08s | 18x |
| `get_foreign_keys` | 7.07s | 0.19s | 37x |
| `get_indexes` | 0.80s | 0.09s | 8.6x |
| `get_table_comment` | 0.71s | 0.05s | 14x |
| **MetaData.reflect** | **12.62s** | **1.15s** | **11x** |

### Checklist

This pull request is:

	- Fixes: #8430
	- Tests added in `test/dialect/mssql/test_reflection.py`
	- Changelog entry: `doc/build/changelog/unreleased_21/8430.rst`

### DIFF
--- a/doc/build/changelog/unreleased_21/8430.rst
+++ b/doc/build/changelog/unreleased_21/8430.rst
@@ -1,0 +1,17 @@
+.. change::
+    :tags: performance, mssql, reflection
+    :tickets: 8430
+
+    Implemented native multi-table reflection methods for the SQL Server
+    dialect, providing :meth:`.MSDialect.get_multi_columns`,
+    :meth:`.MSDialect.get_multi_pk_constraint`,
+    :meth:`.MSDialect.get_multi_foreign_keys`,
+    :meth:`.MSDialect.get_multi_indexes` and
+    :meth:`.MSDialect.get_multi_table_comment`.  Previously the SQL Server
+    dialect relied on :meth:`.DefaultDialect._default_multi_reflect`,
+    which calls the per-table methods in a loop; the new implementations
+    issue a single bulk query per object type against the ``sys.*``
+    catalog views, avoiding the per-table round trips.  The single-table
+    reflection methods are now thin wrappers over the multi-table ones,
+    matching the pattern used by the PostgreSQL and Oracle dialects.
+    Pull request courtesy Gaurav Sharma.
--- a/lib/sqlalchemy/dialects/mssql/base.py
+++ b/lib/sqlalchemy/dialects/mssql/base.py
@@ -982,6 +982,8 @@ from ... import text
 from ... import util
 from ...engine import cursor as _cursor
 from ...engine import default
+from ...engine import ObjectKind
+from ...engine import ObjectScope
 from ...engine import reflection
 from ...engine.reflection import ReflectionDefaults
 from ...sql import coercions
@@ -2967,6 +2969,24 @@ def _db_plus_owner(fn):
     return update_wrapper(wrap, fn)
 
 
+def _db_plus_owner_multi(fn):
+    def wrap(dialect, connection, schema=None, **kw):
+        dbname, owner = _owner_plus_db(dialect, schema)
+        return _switch_db(
+            dbname,
+            connection,
+            fn,
+            dialect,
+            connection,
+            dbname,
+            owner,
+            schema,
+            **kw,
+        )
+
+    return update_wrapper(wrap, fn)
+
+
 def _switch_db(dbname, connection, fn, *arg, **kw):
     if dbname:
         current_db = connection.exec_driver_sql("select db_name()").scalar()
@@ -3673,6 +3693,97 @@ order by
                 % tablename
             ) from ne
 
+    def _parse_column_info(
+        self,
+        name,
+        type_,
+        nullable,
+        maxlen,
+        numericprec,
+        numericscale,
+        default,
+        collation,
+        definition,
+        is_persisted,
+        is_identity,
+        identity_start,
+        identity_increment,
+        comment,
+        base_type=None,
+    ):
+        # Try to resolve the user type first (e.g., "sysname"),
+        # then fall back to the base type (e.g., "nvarchar").
+        # base_type may be None for CLR types (geography, geometry,
+        # hierarchyid) which have no corresponding base type.
+        coltype = self.ischema_names.get(type_, None)
+        if coltype is None and base_type is not None and base_type != type_:
+            coltype = self.ischema_names.get(base_type, None)
+        kwargs = {}
+
+        if coltype in (MSBinary, MSVarBinary, sqltypes.LargeBinary):
+            kwargs["length"] = maxlen if maxlen != -1 else None
+        elif coltype in (MSString, MSChar, MSText):
+            kwargs["length"] = maxlen if maxlen != -1 else None
+            if collation:
+                kwargs["collation"] = collation
+        elif coltype in (MSNVarchar, MSNChar, MSNText):
+            kwargs["length"] = maxlen // 2 if maxlen != -1 else None
+            if collation:
+                kwargs["collation"] = collation
+
+        if coltype is None:
+            if base_type is not None and base_type != type_:
+                util.warn(
+                    "Did not recognize type '%s' (user type) or '%s' "
+                    "(base type) of column '%s'" % (type_, base_type, name)
+                )
+            else:
+                util.warn(
+                    "Did not recognize type '%s' of column '%s'"
+                    % (type_, name)
+                )
+            coltype = sqltypes.NULLTYPE
+        else:
+            if issubclass(coltype, sqltypes.NumericCommon):
+                kwargs["precision"] = numericprec
+                if not issubclass(coltype, sqltypes.Float):
+                    kwargs["scale"] = numericscale
+            coltype = coltype(**kwargs)
+
+        cdict = {
+            "name": name,
+            "type": coltype,
+            "nullable": nullable,
+            "default": default,
+            "autoincrement": is_identity is not None,
+            "comment": comment,
+        }
+
+        if definition is not None and is_persisted is not None:
+            cdict["computed"] = {
+                "sqltext": definition,
+                "persisted": is_persisted,
+            }
+
+        if is_identity is not None:
+            if identity_start is None or identity_increment is None:
+                cdict["identity"] = {}
+            else:
+                if isinstance(
+                    coltype, (sqltypes.BigInteger, sqltypes.Integer)
+                ):
+                    start = int(identity_start)
+                    increment = int(identity_increment)
+                else:
+                    start = identity_start
+                    increment = identity_increment
+                cdict["identity"] = {
+                    "start": start,
+                    "increment": increment,
+                }
+
+        return cdict
+
     @reflection.cache
     @_db_plus_owner
     def get_columns(self, connection, tablename, dbname, owner, schema, **kw):
@@ -3793,115 +3904,25 @@ order by
 
         cols = []
         for row in c.mappings():
-            name = row[sys_columns.c.name]
-            type_ = row[sys_types.c.name]
-            base_type = row["base_type"]
-            nullable = row[sys_columns.c.is_nullable] == 1
-            maxlen = row[sys_columns.c.max_length]
-            numericprec = row[sys_columns.c.precision]
-            numericscale = row[sys_columns.c.scale]
-            default = row[sys_default_constraints.c.definition]
-            collation = row[sys_columns.c.collation_name]
-            definition = row[computed_definition]
-            is_persisted = row[computed_cols.c.is_persisted]
-            is_identity = row[identity_cols.c.is_identity]
-            identity_start = row[identity_cols.c.seed_value]
-            identity_increment = row[identity_cols.c.increment_value]
-            comment = row[extended_properties.c.value]
-
-            # Try to resolve the user type first (e.g., "sysname"),
-            # then fall back to the base type (e.g., "nvarchar").
-            # base_type may be None for CLR types (geography, geometry,
-            # hierarchyid) which have no corresponding base type.
-            coltype = self.ischema_names.get(type_, None)
-            if (
-                coltype is None
-                and base_type is not None
-                and base_type != type_
-            ):
-                coltype = self.ischema_names.get(base_type, None)
-
-            kwargs = {}
-
-            if coltype in (
-                MSBinary,
-                MSVarBinary,
-                sqltypes.LargeBinary,
-            ):
-                kwargs["length"] = maxlen if maxlen != -1 else None
-            elif coltype in (
-                MSString,
-                MSChar,
-                MSText,
-            ):
-                kwargs["length"] = maxlen if maxlen != -1 else None
-                if collation:
-                    kwargs["collation"] = collation
-            elif coltype in (
-                MSNVarchar,
-                MSNChar,
-                MSNText,
-            ):
-                kwargs["length"] = maxlen // 2 if maxlen != -1 else None
-                if collation:
-                    kwargs["collation"] = collation
-
-            if coltype is None:
-                if base_type is not None and base_type != type_:
-                    util.warn(
-                        "Did not recognize type '%s' (user type) or '%s' "
-                        "(base type) of column '%s'" % (type_, base_type, name)
-                    )
-                else:
-                    util.warn(
-                        "Did not recognize type '%s' of column '%s'"
-                        % (type_, name)
-                    )
-                coltype = sqltypes.NULLTYPE
-            else:
-                if issubclass(coltype, sqltypes.NumericCommon):
-                    kwargs["precision"] = numericprec
-
-                    if not issubclass(coltype, sqltypes.Float):
-                        kwargs["scale"] = numericscale
-
-                coltype = coltype(**kwargs)
-            cdict = {
-                "name": name,
-                "type": coltype,
-                "nullable": nullable,
-                "default": default,
-                "autoincrement": is_identity is not None,
-                "comment": comment,
-            }
-
-            if definition is not None and is_persisted is not None:
-                cdict["computed"] = {
-                    "sqltext": definition,
-                    "persisted": is_persisted,
-                }
-
-            if is_identity is not None:
-                # identity_start and identity_increment are Decimal or None
-                if identity_start is None or identity_increment is None:
-                    cdict["identity"] = {}
-                else:
-                    if isinstance(coltype, sqltypes.BigInteger):
-                        start = int(identity_start)
-                        increment = int(identity_increment)
-                    elif isinstance(coltype, sqltypes.Integer):
-                        start = int(identity_start)
-                        increment = int(identity_increment)
-                    else:
-                        start = identity_start
-                        increment = identity_increment
-
-                    cdict["identity"] = {
-                        "start": start,
-                        "increment": increment,
-                    }
-
-            cols.append(cdict)
+            cols.append(
+                self._parse_column_info(
+                    name=row[sys_columns.c.name],
+                    type_=row[sys_types.c.name],
+                    base_type=row["base_type"],
+                    nullable=row[sys_columns.c.is_nullable] == 1,
+                    maxlen=row[sys_columns.c.max_length],
+                    numericprec=row[sys_columns.c.precision],
+                    numericscale=row[sys_columns.c.scale],
+                    default=row[sys_default_constraints.c.definition],
+                    collation=row[sys_columns.c.collation_name],
+                    definition=row[computed_definition],
+                    is_persisted=row[computed_cols.c.is_persisted],
+                    is_identity=row[identity_cols.c.is_identity],
+                    identity_start=row[identity_cols.c.seed_value],
+                    identity_increment=row[identity_cols.c.increment_value],
+                    comment=row[extended_properties.c.value],
+                )
+            )
 
         if cols:
             return cols
@@ -3967,14 +3988,9 @@ order by
                 **kw,
             )
 
-    @reflection.cache
-    @_db_plus_owner
-    def get_foreign_keys(
-        self, connection, tablename, dbname, owner, schema, **kw
-    ):
-        # Foreign key constraints
-        s = (
-            text("""\
+    @staticmethod
+    def _fk_query_sql(fk_info_where, extra_cols=""):
+        return """\
 WITH fk_info AS (
     SELECT
         ischema_ref_con.constraint_schema,
@@ -3992,11 +4008,11 @@ WITH fk_info AS (
         INFORMATION_SCHEMA.REFERENTIAL_CONSTRAINTS ischema_ref_con
         INNER JOIN
         INFORMATION_SCHEMA.KEY_COLUMN_USAGE ischema_key_col ON
-            ischema_key_col.table_schema = ischema_ref_con.constraint_schema
+            ischema_key_col.table_schema =
+                ischema_ref_con.constraint_schema
             AND ischema_key_col.constraint_name =
-            ischema_ref_con.constraint_name
-    WHERE ischema_key_col.table_name = :tablename
-        AND ischema_key_col.table_schema = :owner
+                ischema_ref_con.constraint_name
+    WHERE %(fk_info_where)s
 ),
 constraint_info AS (
     SELECT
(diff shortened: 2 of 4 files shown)


sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 4cce2bc926c8c00cd3ae15ac5e0dc7b263425886 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 4cce2bc926c8c00cd3ae15ac5e0dc7b263425886: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

zzzeek:
@bewithgaurav question - when we make even minor changes in MSSQL reflection, we get a lot of regression reports .   if we put this in 2.1 right up front are you around to help with regressions?

bewithgaurav:
@zzzeek - sure, happy to commit to that.
I can usually turn around triage/repro within a day or two and a fix PR within a week.
I've subscribed to the repo - will keep an eye out for issues with mssql/reflection tags.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision a852ecd6fe37a132fead1edc1687f3f38d87ae01 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset a852ecd6fe37a132fead1edc1687f3f38d87ae01 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

CaselIT:
thanks, i'll try to review this weekend

> Not implemented here: get_multi_unique_constraints, get_multi_check_constraints, get_multi_table_options -MSSQL has no single-table counterparts to delegate from. Happy to add as a follow-up.

if you want to sure, feel free to provide a followup once this on is merged, thanks! The issue is #9468. I'm not sure if there are issues for the other missing features, but we can update that one or create new ones

> @zzzeek - sure, happy to commit to that.

thanks! we can ping you if needed

CaselIT:
sorry for the delay, I'll try looking at it this week

lib/sqlalchemy/dialects/mssql/base.py
zzzeek, line 3669: it looks like we are keeping this version of the query around to suit the "temp tables" use case.   that's fine, but this should be expanded to also do the "multi" version.

lib/sqlalchemy/dialects/mssql/base.py
```diff
@@ -3673,9 +3695,101 @@ def _get_internal_temp_table_name(self, connection, tablename):
...
+                cdict["identity"] = {
+                    "start": start,
+                    "increment": increment,
+                }
+
+        return cdict
+
     @_db_plus_owner
-    def get_columns(self, connection, tablename, dbname, owner, schema, **kw):
+    def _internal_get_columns(
```
zzzeek: the form here looks to be "_internal_get_XYZ" and "get_multi_XYZ".  This creates a lot of duplication that I'd rather not have.   There should only be one method for each thing that does a SQL query, and if the query needs special limitations to suit reading for a temp table, it should be a variation of that same SQL query.

lib/sqlalchemy/dialects/mssql/base.py
```diff
@@ -4154,3 +4189,714 @@ def get_foreign_keys(
...
+            rp = connection.execute(
+                sql.text("""
+SELECT
+    o.name AS table_name,
+    kc.name AS constraint_name,
+    c.name AS column_name,
+    ic.key_ordinal,
+    i.type AS index_type
+FROM sys.key_constraints kc
+JOIN sys.objects o ON kc.parent_object_id = o.object_id
```
zzzeek: we also have a really nice Core Expression query here in "_internal_get_columns()", if we're using LLMs to help us anyway it would be very nice for this query to be written in the same way

lib/sqlalchemy/dialects/mssql/base.py
zzzeek: is there an actual reason we can't use the "multi" form for tempdb?    in any case there should only be one "get columns" query that we can build up for the main / tempdb cases (that is, these can still be queried separately, but i dont want to maintain two totally different SQL queries /methods that do 99% the exact same thing)

zzzeek (changes requested):
hey there -

sorry for the delay in reviewing.  There's two big things I would need with this:

1. eliminate the bulk of the duplication between _internal_get_XYZ and get_multi_XYZ, as this is too much duplication to maintain.    the SQL queries done for a "single" and a "multi" run should not be that different
2. where the get_multi_XYZ method is replacing a previous get_XYZ method where we had a Core SQL expression in place (e.g. select(...)) please keep the query as a core select and not a text().

Is there a real reason why the tempdb queries cannot query across multiple names at once?    I'm fine doing main / temp in two passes but i dont see why we need a "single" version at least for get_columns (which is the big one).

lib/sqlalchemy/dialects/mssql/base.py
bewithgaurav: agreed, will reshape this same select to handle the multi case too

lib/sqlalchemy/dialects/mssql/base.py
bewithgaurav: agreed, 5 of the "_internal_get_XYZ" and "get_multi_XYZ" pairs run near-identical sql, will collapse to one builder per method 
temp variation will become an execution option on the same query rather than separate code path

lib/sqlalchemy/dialects/mssql/base.py
bewithgaurav: nice - thanks for the suggestion, will rewrite as Core sys.* select matching the "_internal_get_columns" pattern
this will need a few new Table aliases in information_schema.py ("sys_key_constraints", "sys_indexes", "sys_index_columns", "sys_objects", "sys_schemas")

the existing "_internal_get_pk_constraint" uses INFORMATION_SCHEMA with an objectproperty workaround & the new sys.* version will get that from sys.indexes.type in a much cleaner way

lib/sqlalchemy/dialects/mssql/base.py
bewithgaurav: no real reason
the upcoming revised changes with unified builder (as mentioned in prev comments) will handle tempdb via "schema_translate_map" (same thing "_internal_get_columns" uses today) - no per-temp loop

bewithgaurav:
@zzzeek - thanks for your review, will take a day or two to add the revised set of changes

sqla-tester (commented):
Hi, this is **sqla-tester** and I see you've pinged me for review. However, user **bewithgaurav** is not authorized to initiate CI jobs.  Please wait for a project member to do this!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision b389bc7f84d64a3a6928237fd8772f2085ddfb62 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset b389bc7f84d64a3a6928237fd8772f2085ddfb62 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision b389bc7f84d64a3a6928237fd8772f2085ddfb62 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset b389bc7f84d64a3a6928237fd8772f2085ddfb62 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

lib/sqlalchemy/dialects/mssql/base.py
```diff
@@ -2967,6 +2969,24 @@ def wrap(dialect, connection, tablename, schema=None, **kw):
     return update_wrapper(wrap, fn)
 
 
+def _db_plus_owner_multi(fn):
```
sqla-tester, line 2972: **Michael Bayer** (zzzeek) wrote:

im wondering if it might be time to use a different approach for this part.   the function signatures are all forced to change when it's like this which upsets pyright and eventually mypy if we turn it on here.  options include changing the required function signature to use required KW args, or maybe using a context manager rather than a decorator.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

lib/sqlalchemy/dialects/mssql/base.py
```diff
@@ -4074,15 +3845,528 @@ def get_foreign_keys(
...
+        connection,
+        dbname,
+        owner,
+        schema,
+        filter_names,
+        scope,
+        kind,
+        **kw,
+    ):
+        run_bulk, regular_names, temp_names = self._partition_filter_names(
```
sqla-tester: **Michael Bayer** (zzzeek) wrote:

I notice that all get_multi_XYZ methods call upon _partition_filter_names(), then all do the identical logic, "if run_bulk: names = self._get_multi_object_names()".  Both _partition_filter_names() and _get_multi_object_names() are **only** called this way.

This would suggest the two methods should be merged, so code here in get_multi_columns() would look something like:


```py
(
   run_bulk, regular_names, multi_object_names, temp_names
) = self._partition_filter_names(
    filter_names, scope, kind
)

if run_bulk and multi_object_names:
    name_map = self._multi_name_map(regular_names)
    self._fetch_multi_columns(..., names=multi_object_names, ...)

```

does that seem right?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

sqla-tester (commented):

Can we make the change regarding partition_filter_names and get_multi_object_names?   The other two comments at the moment are just for discussion.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

lib/sqlalchemy/dialects/mssql/base.py
```diff
@@ -4136,21 +4419,451 @@ def get_foreign_keys(
...
+                schema=schema,
+                name_map=temp_name_map,
+                result=result,
+                exec_opts={"schema_translate_map": {"sys": "tempdb.sys"}},
             )
+
+    # --- Single-table reflection wrappers (delegate to multi) ---
+
+    @reflection.cache
+    def get_columns(self, connection, table_name, schema=None, **kw):
```
sqla-tester, line 4823: **Michael Bayer** (zzzeek) wrote:

I want to do a followup patch that puts these single-to-multi wrappers in a common mixin as pg, oracle have the identical code, should not be needed to be repeated.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

zzzeek:
send me a ping if you want to do that last change or i can just finish up, thanks

lib/sqlalchemy/dialects/mssql/base.py
bewithgaurav: makes sense, the four-tuple unpacking reads cleaner and removes the repeated `if run_bulk: names = ...` block from all 5 call sites will fold `get_multi_object_names` into `_partition_filter_names`
I will make this change

lib/sqlalchemy/dialects/mssql/base.py
bewithgaurav: sure - will re-request one last review once I make push the last commit so you can take it forward

lib/sqlalchemy/dialects/mssql/base.py
bewithgaurav: agree the decorator is awkward, the wrap function injects dbname/owner as positional args

if the goal is making pyright/mypy work well, kw-only args via the decorator just rename the issue, dbname/owner would still be required params on the override
the other option - a context manager around `_switch_db` (already a try/finally setup/teardown) removes the extra params from the method signature entirely
ideally one shared context manager covering `_db_plus_owner`, `_db_plus_owner_listing`, and `_db_plus_owner_multi` rather than just the multi variant

I can help out with this one as well if the approach looks okay

zzzeek: sure if you've got an idea that minimizes duplication then be my guest

bewithgaurav: thanks - noted, will open a separate PR for this after the current one lands

bewithgaurav:
@zzzeek - I have committed the suggested change, thanks for your inputs

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 2c6c69f159225f85312e39d99222279ef8cbadd1 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 2c6c69f159225f85312e39d99222279ef8cbadd1 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

sqla-tester:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

lib/sqlalchemy/dialects/mssql/base.py
sqla-tester: **Michael Bayer** (zzzeek) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

sqla-tester: **Michael Bayer** (zzzeek) wrote:

let's leave it for now here as enough is already changing, thanks

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

lib/sqlalchemy/dialects/mssql/base.py
```diff
@@ -4074,15 +3845,537 @@ def get_foreign_keys(
...
+                f"{schema}.{table}" if schema else table
+            ) from None
+
+    @_db_plus_owner_multi
+    def get_multi_columns(
+        self,
+        connection,
+        dbname,
+        owner,
+        schema,
```
sqla-tester, line 3983: **Michael Bayer** (zzzeek) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

sqla-tester (commented):

great work.

I've left a few comment but I'll fix them and merge. Thank you for the work!

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

doc/build/changelog/unreleased_21/8430.rst
```diff
@@ -0,0 +1,17 @@
...
+    :tags: performance, mssql, reflection
+    :tickets: 8430
+
+    Implemented native multi-table reflection methods for the SQL Server
+    dialect, providing :meth:`.MSDialect.get_multi_columns`,
+    :meth:`.MSDialect.get_multi_pk_constraint`,
+    :meth:`.MSDialect.get_multi_foreign_keys`,
+    :meth:`.MSDialect.get_multi_indexes` and
+    :meth:`.MSDialect.get_multi_table_comment`.  Previously the SQL Server
+    dialect relied on :meth:`.DefaultDialect._default_multi_reflect`,
```
sqla-tester, line 11: **Federico Caselli** (CaselIT) wrote:

there is no need to be this specific. let's just say something along the lines of "relied on the default dialect default implementation which calls ..."

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

lib/sqlalchemy/dialects/mssql/base.py
```diff
@@ -3673,40 +3541,121 @@ def _get_internal_temp_table_name(self, connection, tablename):
...
+        Returns a ``select()`` that includes ``table_name`` and ``owner``
+        in the result so callers can group rows by table. The caller
+        applies the appropriate WHERE clause for either the single-table
+        case (``sys_columns.c.object_id == X``) or the multi-table case
+        (``sys_schemas.c.name == owner AND sys_objects.c.name.in_(names)``).
+
+        Used by :meth:`.get_multi_columns` (and indirectly by
+        :meth:`.get_columns` via the multi delegation). Replaces the
+        previous duplication between ``_internal_get_columns`` and
+        ``get_multi_columns``.
```
sqla-tester, line 3648: **Federico Caselli** (CaselIT) wrote:

> Replaces the previous duplication between ``_internal_get_columns`` and ``get_multi_columns``.

we can drop this since there is no _internal_get_columns in the history once merged

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

lib/sqlalchemy/dialects/mssql/base.py
```diff
@@ -4074,15 +3845,537 @@ def get_foreign_keys(
...
+            regular_names = []
+            run_bulk = include_default
+
+        if not run_bulk:
+            return (run_bulk, regular_names, [], temp_names)
+
+        type_filter = []
+        if ObjectKind.TABLE in kind:
+            type_filter.append("'U'")
+        if ObjectKind.VIEW in kind:
```
sqla-tester, line 3918: **Federico Caselli** (CaselIT) wrote:

shouldn't we also check for materialized views here? From a quick search it seems like there is no easy way of filtering them, they seem to still be V here
https://learn.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-objects-transact-sql?view=sql-server-ver17

I think it's likely fine to threat them as synonym in mssql, so we could just use 

`if ObjectKind.VIEW in kind or ObjectKind.MATERIALIZED_VIEW in kind:`

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

lib/sqlalchemy/dialects/mssql/base.py
```diff
@@ -4074,15 +3845,537 @@ def get_foreign_keys(
...
+
+        return result.items()
+
+    @lru_cache()
+    def _pk_constraint_select(self):
+        """Build the unified Core sys.* select for PK constraint reflection.
+
+        Returns a ``select()`` that includes ``table_name`` and ``owner``
+        in the result so callers can group rows by table. Replaces the
+        previous duplication between ``_internal_get_pk_constraint``
```
sqla-tester, line 4180: **Federico Caselli** (CaselIT) wrote:

here too could be removed

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

lib/sqlalchemy/dialects/mssql/base.py
```diff
@@ -4074,15 +3845,537 @@ def get_foreign_keys(
...
+
+        return final.items()
+
+    def _fetch_multi_foreign_keys(
+        self, connection, dbname, owner, names, schema, name_map, final
+    ):
+        """Execute the unified FK query for the main-DB pass.
+
+        Builds the same INFORMATION_SCHEMA + sys.* CTE that
+        ``_fk_query_sql`` produces, restricted to the provided owner and
```
sqla-tester, line 4357: **Federico Caselli** (CaselIT) wrote:

we likely could update to say "it uses the _fk_query_sql query, restricted to the provided ..."

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

lib/sqlalchemy/dialects/mssql/base.py
```diff
@@ -4136,21 +4428,453 @@ def get_foreign_keys(
...
+        self, connection, owner, names, schema, name_map, result, exec_opts
+    ):
+        """Execute the unified indexes queries for one catalog pass.
+
+        Runs two Core sys.* selects (index metadata + index columns),
+        grouped by table. Used for both the main-DB pass and the
+        tempdb pass (with ``schema_translate_map={"sys": "tempdb.sys"}``
+        applied).
+        """
+        meta_q = self._indexes_metadata_select().where(
```
sqla-tester, line 4582: **Federico Caselli** (CaselIT) wrote:

if you prefer creating a single query that returns both index data and columns we could also do that, but having the two is fine

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

doc/build/changelog/unreleased_21/8430.rst
sqla-tester: **Federico Caselli** (CaselIT) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

lib/sqlalchemy/dialects/mssql/base.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

lib/sqlalchemy/dialects/mssql/base.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

actually it does not seem to have materialized views, it's just the mssql docs that's not clear, since if you don't notice the "Applies to" it would seem it's support my mssql https://learn.microsoft.com/en-us/sql/t-sql/statements/create-materialized-view-as-select-transact-sql?view=azure-sqldw-latest&viewFallbackFrom=sql-server-ver17

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

lib/sqlalchemy/dialects/mssql/base.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

lib/sqlalchemy/dialects/mssql/base.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

lib/sqlalchemy/dialects/mssql/base.py
sqla-tester: **Federico Caselli** (CaselIT) wrote:

will keep two separate queries. if someone wants to unify we can do that later

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6689 has been **merged**. Congratulations! :)

bewithgaurav:
thanks for the reviews and landing it @zzzeek @CaselIT :)

CaselIT:
I've also updated https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6704 to add has-multi-table also for mssql

PR #13312, opened 2026-05-20
test: add coverage for delete with nested alias exists

zifo-10:
### Description

Adds test coverage for compiling a `DELETE` statement with a correlated `EXISTS` subquery that uses an aliased table.

This verifies that SQLAlchemy correctly renders the aliased table inside the nested `EXISTS` clause while correlating it against the outer `DELETE` target.

### Checklist

This pull request is:

- [ ] A short code fix

### DIFF
--- a/test/sql/test_delete.py
+++ b/test/sql/test_delete.py
@@ -145,6 +145,25 @@ class DeleteTest(_DeleteTestBase, fixtures.TablesTest, AssertsCompiledSQL):
             ")",
         )
 
+    def test_delete_with_nested_alias_exists(self):
+        table1, table2 = self.tables.mytable, self.tables.myothertable
+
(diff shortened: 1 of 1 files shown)


CaselIT:
hi,

what's the motivation for this test? it's not that it's wrong, it just seems kinda random

zifo-10:
@CaselIT the intention was to add coverage for a correlated `EXISTS` subquery using an aliased table inside a `DELETE` statement, since there are already nearby tests covering correlation and alias handling independently.

I wanted to ensure the aliased table is rendered correctly while still correlating against the outer delete target.

That said, I agree it’s not tied to a specific regression or issue at the moment, so I’m happy to either refine it into a more targeted edge case or close the PR if it doesn’t add meaningful coverage

PR #13321, opened 2026-05-21
Fix yield_per leaking into selectinload queries

### LINKED ISSUE #13301: Setting lazy in a relationship with do_orm_execute causes error
```python

from collections.abc import Iterator

import sqlalchemy as sa
from sqlalchemy import select
from sqlalchemy.orm import DeclarativeBase, ORMExecuteState, selectinload, sessionmaker
from sqlalchemy.orm import foreign
from sqlalchemy.orm import MappedAsDataclass
from sqlalchemy.orm import remote as sa_remote
from sqlalchemy import event
from sqlalchemy import select


class Base(DeclarativeBase):
    pass


class Message(MappedAsDataclass, Base, kw_only=True):
    __tablename__ = "message"

    pk: Mapped[int] = mapped_column(primary_key=True, init=False)

    id: Mapped[str | None] = mapped_column(default=None)
    correction_id: Mapped[str | None] = mapped_column(default=None)

    corrections: Mapped[list[Message]] = relationship(
        lazy="selectin",
        init=False,
        primaryjoin=foreign(id) == sa_remote(correction_id),
        join_depth=1,
        viewonly=True,
        uselist=True,
    )


def _do_orm_execute(orm_execute_state: ORMExecuteState) -> None:
    return


event.listen(Session, "do_orm_execute", _do_orm_execute)


def search(session: Session) -> Iterator[Message]:
    stmt = (
        select(Message).where(Message.correction_id.is_(None))
        .execution_options(yield_per=5)
    )

    return session.scalars(stmt)


def create_session(engine: sa.engine.Engine) -> Session:
    return sessionmaker(expire_on_commit=False, autoflush=False, bind=engine)()


if __name__ == "__main__":
    engine = sa.create_engine(f"sqlite://")

    messages: list[Message] = []
    for num in range(10):
        messages.append(Message(id=str(num), correction_id=None))

    corrections: list[Message] = []
    for num in range(11, 13):
        corrections.append(Message(id=str(num), correction_id="1"))

    with create_session(engine) as sess:
        sess.add_all(messages)
        sess.add_all(corrections)
        sess.commit()

    with create_session(engine) as sess:
        for row in search(sess):
            print(row.id, "corrections:", len(row.corrections))
```

_Originally posted by @lovetox in https://github.com/sqlalchemy/sqlalchemy/discussions/13296#discussioncomment-16932784_

commenting out the listen or the lazy makes the query work again

puneetdixit200:
Fixes #13301.

### Summary
- keep top-level `yield_per` from being inherited by loader-generated ORM statements
- add a selectinload regression with a no-op `do_orm_execute` listener

### Testing
- `python -m pytest test/orm/test_selectin_relations.py::SelfReferentialTest::test_yield_per_orm_execute_event -q`
- `python -m pytest test/orm/test_selectin_relations.py::SelfReferentialTest test/orm/test_selectin_relations.py::ChunkingTest::test_yield_per test/orm/test_query.py::ExecutionOptionsTest::test_execution_options_to_load_options -q`
- `python -m pytest test/orm/test_selectin_relations.py -q`
- `python -m pytest test/orm/test_query.py::YieldTest test/orm/test_query.py::YieldIterationTest test/orm/test_query.py::ExecutionOptionsTest -q`
- `python -m ruff check lib/sqlalchemy/orm/context.py test/orm/test_selectin_relations.py`
- `git diff --check`

### DIFF
--- a/lib/sqlalchemy/orm/context.py
+++ b/lib/sqlalchemy/orm/context.py
@@ -534,8 +534,21 @@ class _ORMCompileState(_AbstractORMCompileState):
 
         if "sa_top_level_orm_context" in execution_options:
             ctx = execution_options["sa_top_level_orm_context"]
-            execution_options = ctx.query._execution_options.merge_with(
-                ctx.execution_options, execution_options
+            top_level_execution_options = (
+                ctx.query._execution_options.merge_with(ctx.execution_options)
+            )
+            if "yield_per" in top_level_execution_options:
+                # loader-generated statements shouldn't inherit the
+                # top-level result batching size
+                top_level_execution_options = util.immutabledict(
+                    {
+                        key: value
+                        for key, value in top_level_execution_options.items()
+                        if key != "yield_per"
+                    }
+                )
+            execution_options = top_level_execution_options.merge_with(
+                execution_options
             )
 
         if not execution_options:
--- a/lib/sqlalchemy/orm/context.py
+++ b/lib/sqlalchemy/orm/context.py
@@ -100,6 +100,9 @@ _path_registry = PathRegistry.root
 
 LABEL_STYLE_LEGACY_ORM = SelectLabelStyle.LABEL_STYLE_LEGACY_ORM
 
+# Result batching applies only to the statement being consumed.
+_LOADER_EXEC_OPTION_EXCLUSIONS = frozenset(["yield_per"])
+
 
 class QueryContext:
     __slots__ = (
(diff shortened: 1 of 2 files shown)


zzzeek (commented):
hi there -

thanks for this.   I see the approach here is just the most direct, "this one option is problematic, take it out".   

I'd have to look at this code all over again to see how it flows - is this modification indeed local to loaders?   Why would we hardcode a big rule for just this one option (why is "yield_per" in the list up above? should we take it out of that list?)

overall I have doubts the best solution would look this way because it's one big, verbose hardcoded rule.  arent there other options that might fall under this category?  etc.

puneetdixit200:
I reworked this after your note. The current branch now strips top-level loader execution options through `_LOADER_EXEC_OPTION_EXCLUSIONS` before merging them into loader SELECTs, so `yield_per` stays on the primary statement but does not leak into the selectinload query.

I kept it as a small exclusion set because I do not know of another option in this category yet, but it gives us one clear place to add one if another result-batching-style option shows up.

Local check: `.venv/bin/python -m pytest -q test/orm/test_selectin_relations.py::SelfReferentialTest`

zzzeek:
hi Puneet -

thanks but unfortunately I've come up with a more foundational approach to this issue which you can see at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6705 .   this particular area of the library (like many) is kind of an architecturally heavy area of concern, that the calling of the event hook had side effects at all was the issue.

PR #13323, opened 2026-05-21
Apply tuple bind expressions to expanding IN values

### LINKED ISSUE #8992: tuple expanding params do not render bind_expression() for elements
discovered during #8989, this is simply not implemented at all and the bind expression is ignored when it's an element of a tuple.    Also, the impl of a TupleType() is not getting the internal types propagated along, which is part of what would be needed for this to work

```py
from sqlalchemy import Column
from sqlalchemy import func
from sqlalchemy import MetaData
from sqlalchemy import select
from sqlalchemy import String
from sqlalchemy import Table
from sqlalchemy import tuple_
from sqlalchemy.engine import default


class MyString(String):
    def bind_expression(self, bindvalue):
        return func.lower(bindvalue)

    def column_expression(self, col):
        return func.lower(col)


test_table = Table(
    "test_table", MetaData(), Column("x", String), Column("y", MyString())
)

stmt1 = select(test_table).where(test_table.c.y.in_(["y1", "y2", "y3"]))
print(stmt1.compile(compile_kwargs={"render_postcompile": True}))

stmt2 = select(test_table).where(
    tuple_(test_table.c.x, test_table.c.y).in_(
        [
            ("x1", "y1"),
            ("x2", "y2"),
            ("x3", "y3"),
        ]
    )
)

print(stmt2.compile(compile_kwargs={"render_postcompile": True}))

expr = test_table.c.y.in_(["y1", "y2", "y3"])
impl = expr.right.type.dialect_impl(default.DefaultDialect())

print(f"right type of expr: {expr.right.type}")
print(f"right impl type of expr: {impl}")


expr = tuple_(test_table.c.x, test_table.c.y).in_(
    [
        ("x1", "y1"),
        ("x2", "y2"),
        ("x3", "y3"),
    ]
)
impl = expr.right.type.dialect_impl(default.DefaultDialect())


print(f"right type of expr: {expr.right.type}")
print(f"right impl type of expr: {impl}")

```

```
# non-tuple - correct
SELECT test_table.x, lower(test_table.y) AS y 
FROM test_table 
WHERE test_table.y IN (lower(:y_1_1), lower(:y_1_2), lower(:y_1_3))

# tuple - incorrect
SELECT test_table.x, lower(test_table.y) AS y 
FROM test_table 
WHERE (test_table.x, test_table.y) IN ((:param_1_1_1, :param_1_1_2), (:param_1_2_1, :param_1_2_2), (:param_1_3_1, :param_1_3_2))

# non-tuple - VARCHAR for outer, inner, correct
right type of expr: VARCHAR
right impl type of expr: VARCHAR

# tuple - outer is correct, inner is incorrect
right type of expr: TupleType(String(), MyString())

# incorrect, should be something like  TupleType(VARCHAR, MyString)
right impl type of expr: TupleType()
```

puneetdixit200:
### Description

Tuple-valued expanding parameters now apply each element type's `bind_expression()` when rendering expanded tuples. This matches scalar `IN` behavior for custom types and covers both `render_postcompile` and `literal_binds`.

Fixes: #8992

### Checklist

This pull request is:

- [x] A short code fix
  - Fixes: #8992
  - includes regression coverage in `test/sql/test_type_expressions.py`

Local checks:
- `uv run --group tests pytest test/sql/test_type_expressions.py -q`
- `uv run --group tests pytest test/sql/test_compiler.py -q -k "tuple_expanding_in or expanding_parameter"`
- `uv run --group lint flake8 lib/sqlalchemy/sql/compiler.py test/sql/test_type_expressions.py`
- `uv run --group lint black --check lib/sqlalchemy/sql/compiler.py test/sql/test_type_expressions.py`

### DIFF
--- a/lib/sqlalchemy/sql/compiler.py
+++ b/lib/sqlalchemy/sql/compiler.py
@@ -3427,11 +3427,27 @@ class SQLCompiler(Compiled):
             and isinstance(values[0], collections_abc.Sequence)
             and not isinstance(values[0], (str, bytes))
         ):
-            if typ_dialect_impl._has_bind_expression:
-                raise NotImplementedError(
-                    "bind_expression() on TupleType not supported with "
-                    "literal_binds"
-                )
+            if typ_dialect_impl._is_tuple_type:
+
+                def _render_tuple_literal(value, typ):
+                    rendered = self.render_literal_value(value, typ)
+                    impl = typ.dialect_impl(self.dialect)
+                    if not impl._has_bind_expression:
+                        return rendered
+
+                    replacement = "__SQLA_BIND_EXPR__"
+                    bind_expression = impl.bind_expression(
+                        elements.literal_column(replacement, type_=typ)
+                    )
+                    return self.process(
+                        bind_expression, skip_bind_expression=True
+                    ).replace(replacement, rendered)
+
+                tuple_types = parameter.type.types
+
+            else:
+                _render_tuple_literal = None
+                tuple_types = ()
 
             replacement_expression = (
                 "VALUES " if self.dialect.tuple_in_values else ""
@@ -3439,10 +3455,14 @@ class SQLCompiler(Compiled):
                 "(%s)"
                 % (
                     ", ".join(
-                        self.render_literal_value(value, param_type)
-                        for value, param_type in zip(
-                            tuple_element, parameter.type.types
+                        (
+                            _render_tuple_literal(value, tuple_types[idx])
+                            if _render_tuple_literal
+                            else self.render_literal_value(
+                                value, parameter.type
+                            )
                         )
+                        for idx, value in enumerate(tuple_element)
                     )
                 )
                 for i, tuple_element in enumerate(values)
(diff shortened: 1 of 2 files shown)


lib/sqlalchemy/sql/compiler.py
```diff
@@ -3522,6 +3542,28 @@ def _render_bindtemplate(name):
...
+                def _render_tuple_bindtemplate(name, typ):
+                    impl = typ.dialect_impl(dialect)
+                    if not impl._has_bind_expression:
+                        return _render_bindtemplate(name)
+
+                    replacement = "__SQLA_BIND_EXPR__"
+                    bind_expression = impl.bind_expression(
+                        elements.literal_column(replacement, type_=typ)
+                    )
+                    return self.process(
```
Sanjays2402, line 3557: `bind_expression()` can introduce its own bind params (for example, `func.coalesce(bindvalue, "fallback")`). Calling `self.process()` from postcompile expansion then mutates `self.binds` while `_process_parameters_for_postcompile()` is iterating it, so this tuple fails compilation with `RuntimeError: dictionary changed size during iteration`.

PR #13331, opened 2026-05-27
Use Trusted Publishing to upload to PyPI

nsoranzo:
### Description

Fix https://github.com/sqlalchemy/sqlalchemy/discussions/13324 .

Split building from upload and sdist/pure-Python packages from platform wheels into 4 jobs to allow fast publication on PyPI.

### Checklist

This pull request is:

- [ ] A short code fix

### DIFF
--- a/.github/workflows/create-wheels.yaml
+++ b/.github/workflows/create-wheels.yaml
@@ -13,7 +13,61 @@ on:
 #   TWINE_REPOSITORY_URL: https://test.pypi.org/legacy/
 
 jobs:
-  build_wheels:
+  build_source_dist_and_pure_python_wheel:
+    name: Build source dist and pure-Python wheel
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


CaselIT (changes requested):
As mentioned in the discussion, I think we should focus on other things first

zzzeek:
hey there -

I'm planning to put the patch at https://github.com/sqlalchemy/sqlalchemy/discussions/13324#discussioncomment-17116704 into main and try doing a release with that which only works with the wheels.    For now, the solution for the .tar.gz would be on hold since that involves redoing my internal scripts to be compatible with that.

nsoranzo:
This is not building the sdist any more per your request, just using the `pypa/gh-action-pypi-publish` action in separate jobs to quickly upload the platform-independent wheel and (later) the platform wheels.

zzzeek:
right....i was confusing pure python wheel,  sorry.

still i think we will do much better with calling pypi-attestations and twine directly.

PR #13335, opened 2026-05-29
fix(result): SimpleResultMetaData raises for ambiguous columns after freeze()

### LINKED ISSUE #9427: Frozen result does not raise for ambiguous column name
The following code illustrates the issue:
```py
import sqlalchemy as sa
engine = sa.create_engine("sqlite://", echo=True)

q = sa.select(sa.literal(4).label('x'), sa.literal(2).label('x'))
    r = conn.execute(q)
    print(r.freeze()().mappings().first()['x'])
    r = conn.execute(q)
    print(r.mappings().first()['x'])
    # sqlalchemy.exc.InvalidRequestError: Ambiguous column name 'x' in result set column descriptions
```

A CursorResultMetaData correctly raised if the key is ambiguous, while SimpleResultMetaData does not

me-saurabhkohli:
Fixes #9427.

## Problem

`result.freeze()` produces a `FrozenResult` backed by `SimpleResultMetaData`. When the original `CursorResultMetaData` contained duplicate/ambiguous column names (keys with `MD_INDEX = None`), those ambiguity markers were lost during `_for_freeze()` conversion. Subsequent key-based lookups on the thawed result silently returned the *last* matched value instead of raising `InvalidRequestError`, diverging from the live-cursor behaviour.

## Root cause

`CursorResultMetaData._for_freeze()` (cursor.py) builds a plain `SimpleResultMetaData` from the raw key list without preserving any ambiguity information. `SimpleResultMetaData._index_for_key` had no ambiguity check.

## Fix

* **`lib/sqlalchemy/engine/result.py`** – Added an `_ambiguous_keys: Optional[Sequence[str]]` parameter to `SimpleResultMetaData.__init__`. When supplied, those keys are marked with `index=None` in the internal `_keymap`. Added `_raise_for_ambiguous_column_name()` (matching the cursor-side helper) and updated `_index_for_key` to call it when `rec[0] is None`.
* **`lib/sqlalchemy/engine/cursor.py`** – Updated `_for_freeze()` to collect the ambiguous keys from the cursor keymap and forward them.
* `result_tuple()` (ORM helper) is **unaffected** — it passes no `_ambiguous_keys`, so intentional duplicate labels continue to work.

## Tests

* Added `test_ambiguous_column_raises_in_simple_result_metadata` – directly constructs `SimpleResultMetaData` with `_ambiguous_keys` and asserts the raise.
* Added `test_frozen_ambiguous_column_raises` – end-to-end: real engine query with duplicate column names, `freeze()`, thaw, assert `InvalidRequestError`.
* All 733 `test/base/` tests pass (3 skipped, env-related).

### DIFF
--- a/lib/sqlalchemy/engine/cursor.py
+++ b/lib/sqlalchemy/engine/cursor.py
@@ -183,9 +183,17 @@ class CursorResultMetaData(ResultMetaData):
         return key in self._keymap
 
     def _for_freeze(self) -> ResultMetaData:
+        # collect any key names whose index is None (ambiguous duplicates)
+        # so that SimpleResultMetaData can preserve the ambiguity flag
+        ambiguous = [
+            key
+            for key in self._keys
+            if self._keymap.get(key, (1,))[MD_INDEX] is None
+        ]
         return SimpleResultMetaData(
             self._keys,
             extra=[self._keymap[key][MD_OBJECTS] for key in self._keys],
+            _ambiguous_keys=ambiguous if ambiguous else None,
         )
 
     def _make_new_metadata(
--- a/lib/sqlalchemy/engine/cursor.py
+++ b/lib/sqlalchemy/engine/cursor.py
@@ -183,17 +183,18 @@ class CursorResultMetaData(ResultMetaData):
         return key in self._keymap
 
     def _for_freeze(self) -> ResultMetaData:
-        # collect any key names whose index is None (ambiguous duplicates)
-        # so that SimpleResultMetaData can preserve the ambiguity flag
-        ambiguous = [
-            key
-            for key in self._keys
-            if self._keymap.get(key, (1,))[MD_INDEX] is None
-        ]
+        # Collect string key names that are marked ambiguous (MD_INDEX is None)
+        # by scanning the full keymap; this catches all aliases/objects that
+        # CursorResultMetaData marks ambiguous, not just those in _keys.
+        ambiguous = {
+            rec[MD_LOOKUP_KEY]
+            for rec in self._keymap.values()
+            if rec[MD_INDEX] is None
+        }
         return SimpleResultMetaData(
             self._keys,
             extra=[self._keymap[key][MD_OBJECTS] for key in self._keys],
-            _ambiguous_keys=ambiguous if ambiguous else None,
+            _ambiguous_keys=list(ambiguous) if ambiguous else None,
         )
 
     def _make_new_metadata(
--- a/lib/sqlalchemy/engine/result.py
+++ b/lib/sqlalchemy/engine/result.py
@@ -271,6 +271,7 @@ class SimpleResultMetaData(ResultMetaData):
                 Sequence[Optional[Callable[[Any], Any]]],
             ]
         ] = None,
+        _ambiguous_keys: Optional[Sequence[str]] = None,
     ):
         self._keys = list(keys)
         self._tuplefilter = _tuplefilter
@@ -292,6 +293,16 @@ class SimpleResultMetaData(ResultMetaData):
             ]
 
         self._keymap = {key: rec for keys, rec in recs_names for key in keys}
+
+        # If the caller supplies _ambiguous_keys (a set of key names that are
+        # duplicated and therefore ambiguous), mark those entries with
+        # index=None so that key-based lookups raise InvalidRequestError,
+        # matching the CursorResultMetaData behaviour (see gh#9427).
+        if _ambiguous_keys:
+            for name in _ambiguous_keys:
+                if name in self._keymap:
+                    rec = self._keymap[name]
+                    self._keymap[name] = (None,) + rec[1:]  # type: ignore[assignment]  # noqa: E501
 
         self._processors = _processors
 
@@ -347,7 +358,17 @@ class SimpleResultMetaData(ResultMetaData):
         except KeyError as ke:
             rec = self._key_fallback(key, ke, raiseerr)
 
+        if rec[0] is None:
+            self._raise_for_ambiguous_column_name(rec)
         return rec[0]  # type: ignore[no-any-return]
+
+    def _raise_for_ambiguous_column_name(
+        self, rec: _KeyMapRecType
+    ) -> NoReturn:
+        raise exc.InvalidRequestError(
+            "Ambiguous column name '%s' in "
+            "result set column descriptions" % rec[1]
+        )
 
     def _indexes_for_keys(self, keys: Sequence[Any]) -> Sequence[int]:
         return [self._keymap[key][0] for key in keys]
--- a/lib/sqlalchemy/engine/result.py
+++ b/lib/sqlalchemy/engine/result.py
@@ -254,6 +254,7 @@ class SimpleResultMetaData(ResultMetaData):
         "_translated_indexes",
         "_create_unique_filters",
         "_key_to_index",
+        "_ambiguous_keys",
     )
 
     _keys: Sequence[str]
@@ -306,6 +307,8 @@ class SimpleResultMetaData(ResultMetaData):
 
         self._processors = _processors
 
+        self._ambiguous_keys = list(_ambiguous_keys) if _ambiguous_keys else None
+
         self._key_to_index = self._make_key_to_index(self._keymap, 0)
 
     def _has_key(self, key: object) -> bool:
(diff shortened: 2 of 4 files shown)


test/base/test_result.py
```diff
@@ -994,8 +994,59 @@ def test_scalars_freeze(self):
...
+            lambda: row._mapping["x"],
+        )
+
+    def test_frozen_ambiguous_column_raises(self):
+        """After freeze(), the thawed SimpleResultMetaData should also
+        raise for ambiguous column names (regression for #9427).
+        """
+        import sqlalchemy as sa
+
+        engine = sa.create_engine("sqlite://", echo=False)
```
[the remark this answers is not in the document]
zzzeek: this test needs to be in test/sql/test_resultset.py

test/base/test_result.py
```diff
@@ -994,8 +994,59 @@ def test_scalars_freeze(self):
...
+            exc.InvalidRequestError,
+            "Ambiguous column name 'x'",
+            lambda: row._mapping["x"],
+        )
+
+    def test_frozen_ambiguous_column_raises(self):
+        """After freeze(), the thawed SimpleResultMetaData should also
+        raise for ambiguous column names (regression for #9427).
+        """
+        import sqlalchemy as sa
```
zzzeek: dont put imports inside of tests

test/base/test_result.py
```diff
@@ -994,8 +994,59 @@ def test_scalars_freeze(self):
...
+
+        assert_raises_message(
+            exc.InvalidRequestError,
+            "Ambiguous column name 'x'",
+            lambda: row["x"],
+        )
+
+
 
-class MergeResultTest(fixtures.TestBase):
```
zzzeek, line 998: this needs to be restored

lib/sqlalchemy/engine/cursor.py
```diff
@@ -183,9 +183,17 @@ def _has_key(self, key: Any) -> bool:
         return key in self._keymap
 
     def _for_freeze(self) -> ResultMetaData:
+        # collect any key names whose index is None (ambiguous duplicates)
+        # so that SimpleResultMetaData can preserve the ambiguity flag
+        ambiguous = [
+            key
+            for key in self._keys
+            if self._keymap.get(key, (1,))[MD_INDEX] is None
+        ]
```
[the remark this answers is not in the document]
zzzeek: yup

lib/sqlalchemy/engine/result.py
```diff
@@ -293,6 +294,16 @@ def __init__(
...
 
+        # If the caller supplies _ambiguous_keys (a set of key names that are
+        # duplicated and therefore ambiguous), mark those entries with
+        # index=None so that key-based lookups raise InvalidRequestError,
+        # matching the CursorResultMetaData behaviour (see gh#9427).
+        if _ambiguous_keys:
+            for name in _ambiguous_keys:
+                if name in self._keymap:
+                    rec = self._keymap[name]
+                    self._keymap[name] = (None,) + rec[1:]  # type: ignore[assignment]  # noqa: E501
```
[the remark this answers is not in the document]
zzzeek: yup

lib/sqlalchemy/engine/result.py
```diff
@@ -347,8 +358,18 @@ def _index_for_key(self, key: Any, raiseerr: bool = True) -> int:
...
+    def _raise_for_ambiguous_column_name(
+        self, rec: _KeyMapRecType
+    ) -> NoReturn:
+        raise exc.InvalidRequestError(
+            "Ambiguous column name '%s' in "
+            "result set column descriptions" % rec[1]
+        )
+
     def _indexes_for_keys(self, keys: Sequence[Any]) -> Sequence[int]:
         return [self._keymap[key][0] for key in keys]
```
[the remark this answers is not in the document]
zzzeek: it's not wrong but the same issue seems to exist in cursor.py as well.   these methods are only used by the ORM and they are not used with string keys so this part can be skipped

zzzeek (changes requested):
thanks for this.    some corrections

me-saurabhkohli (commented):
Thanks for the review! Addressed all the points in the latest commit:

- **Tests moved** to `test/sql/test_resultset.py` as a new `AmbiguousColumnFreezeTest` class — no inline imports, uses the `connection` fixture throughout
- **`class MergeResultTest`** restored in `test/base/test_result.py`
- **`cursor._for_freeze`** now scans the full `_keymap` for entries where `MD_INDEX is None` using `MD_LOOKUP_KEY` (index 3), so all ambiguous aliases are captured rather than just the string names in `_keys`
- **`SimpleResultMetaData`** now stores `_ambiguous_keys` as a slot attribute and propagates it through `_for_freeze` and `__getstate__`/`__setstate__`, so double-freeze and pickle/unpickle paths also preserve the guard

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision c03904ece298493ca69bf6e9cbdae23c7fb6a7b0 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change c03904ece298493ca69bf6e9cbdae23c7fb6a7b0: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6713

zzzeek:
thanks.  it was quicker for me to make the remaining changes directly, feel free to browse https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6713 for where I'm going with it

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6713

* lib/sqlalchemy/engine/result.py (line 352): i think this could just be `_ambiguous_keys-state.get("_ambiguous_keys")`

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6713

* lib/sqlalchemy/engine/result.py (line 352): ah right

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6713 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6714 has been **merged**. Congratulations! :)

lib/sqlalchemy/engine/result.py
```diff
@@ -347,8 +364,18 @@ def _index_for_key(self, key: Any, raiseerr: bool = True) -> int:
         except KeyError as ke:
             rec = self._key_fallback(key, ke, raiseerr)
 
+        if rec[0] is None:
+            self._raise_for_ambiguous_column_name(rec)
         return rec[0]  # type: ignore[no-any-return]
```
jayaddison, line 370: @me-saurabhkohli @zzzeek 

The `MD_INDEX` constant is currently defined in `cursor.py` -- but it might help to make the intent of the code here more readable, if we could reuse it?

```suggestion
        except KeyError as ke:
            rec = self._key_fallback(key, ke, raiseerr)

        if rec[MD_INDEX] is None:
            self._raise_for_ambiguous_column_name(rec)
        return rec[MD_INDEX]  # type: ignore[no-any-return]

```

jayaddison:
@zzzeek why is the `pickle`-related test removed in 4fb459aaf05dd9c31ce3ece57c1bbf81ca9855de (the commit that closed this PR)?  From an initial browse of the diff I'd continue to expect an `InvalidRequestError` about ambiguous column names to be raised when an access to an ambiguous column name is attempted on the deserialized resultset.  But I might misunderstand.

Ref: https://github.com/sqlalchemy/sqlalchemy/commit/4fb459aaf05dd9c31ce3ece57c1bbf81ca9855de#diff-04d46ba6672faf0ebc2ccc148351857bb2e2f4e76ea05487cff3f524b9efd76dL1285-L1290

zzzeek:
the new test there still uses pickle?

lib/sqlalchemy/engine/result.py
zzzeek: sure, though too trivial to put by itself, maybe in a future refactoring

jayaddison:
@zzzeek 
> the new test there still uses pickle?

D'oh, my mistake.  I had noticed the `scenario.pickled` conditional, but not the raises-assertion after it.  Thanks!

lib/sqlalchemy/engine/result.py
jayaddison: Reading a bit more into the code, I'm not sure it's so trivial?  The cursor-related metadata constants (`MD_*`) seem to relate to `_CursorKeyMapRecType` -- and I'm not sure they all generalise to the parent `_KeyMapRecType`.

Maybe there is a cleanup possible here but I think it might require a few steps of refactoring.

PR #13341, opened 2026-06-01
Perf/conditional unique selectinload

### LINKED ISSUE #13339: performance: selectinload skips .unique() when no nested joinedload on collection inflates rows
### Describe the use case

`selectinload` calls `.unique()` on every inner `Result`, even when the inner query cannot produce duplicate rows. The uniqueness pass only matters when a nested `joinedload` on a collection inflates rows (one row per `(child, grandchild)` instead of one per child). In the common leaf-load case — 1:N, M2O, and M2M omit-join paths — rows are unique by construction and the per-row hashing is pure overhead.

`loading.instances()` already signals when uniquing is needed: it sets `Result._unique_filter_state` to a `require_unique` guard when the inner compile state has `multi_row_eager_loaders=True`. Making the `.unique()` call conditional on that attribute eliminates the overhead when it provides no benefit.

All databases. This is a Python-level ORM optimization in `_SelectInLoader` — no SQL dialect or DBAPI is involved.

### Example Use

```python
# Any selectinload without a nested joinedload on a collection benefits —
# this is the typical case.

stmt = select(User).options(selectinload(User.addresses))
users = session.scalars(stmt).all()

# nested joinedload on a collection — .unique() is still called (correct):
stmt = select(User).options(
    selectinload(User.addresses).joinedload(Address.dingalings)
)
users = session.scalars(stmt).all()
```

### Additional context

In-memory SQLite bench, 2000 parents × 5 children + M:N tags, 3-iter mean:

| Workload | before | after | delta |
|---|---|---|---|
| selectin children (1:N) | 74.9 ms | ~61 ms | **−18%** |
| selectin tags M2M | 30.3 ms | ~26 ms | **−14%** |
| joined children | 46.5 ms | 46.5 ms | ~0% |
| plain parent load | 3.75 ms | 3.7 ms | ~0% |

Behaviour change: `yield_per` set via `do_orm_execute` for a relationship load no longer raises `InvalidRequestError("Can't use yield_per in conjunction with unique")` for `selectinload` without a nested collection joinedload. `immediateload` is unaffected.

ollz272:
## Description

`_SelectInLoader._load_via_parent` and `_load_via_child` currently call
`.unique()` unconditionally on the inner `Result`. The uniqueness pass is only
required when a nested `joinedload` on a collection is in effect — in that case
the `JOIN` inflates rows (one row per `(child, grandchild)` instead of one per
child) and the outer `groupby` would produce duplicated entries without dedup.

`loading.instances()` already signals exactly this condition: it sets
`Result._unique_filter_state` to a `require_unique` guard when the inner
compile state has `multi_row_eager_loaders=True`. When that flag is unset (no
nested collection `joinedload`), `_unique_filter_state` stays `None`, meaning
the inner query produces unique rows by construction:

- `omit_join` 1:N — one row per child
- `omit_join` M2O — one row per parent
- `omit_join` M2M — one row per (parent, entity)
- `load_with_join` (non-omit) — one row per (parent, child)

This PR makes the `.unique()` call conditional, via a new `_has_unique_filter`
property on `Result` that exposes this state without reaching into the private
`_unique_filter_state` attribute directly:

```python
if result._has_unique_filter:
    result = result.unique()
```

Per-row hashing in `_iterator_getter` is avoided in the common leaf-load case.
On an in-memory SQLite bench (2000 parents × 5 children + M:N tags, Python
3.14, n=1000 iterations):

| Workload | before avg | after avg | delta | before sd | after sd |
|---|---|---|---|---|---|
| selectin children (1:N) | 56.0 ms | 41.4 ms | **−26%** | 5.2 ms | 1.9 ms |
| selectin tags M2M | 25.6 ms | 20.0 ms | **−22%** | 4.2 ms | 3.1 ms |
| joined children (1:N) | 42.9 ms | 37.6 ms | ~0% | — | — |
| plain parent load | 4.2 ms | 3.6 ms | ~0% | — | — |

**Behaviour change:** `yield_per` set in a `do_orm_execute` event for a
relationship load no longer raises
`InvalidRequestError("Can't use yield_per in conjunction with unique")` for
`selectinload` without a nested collection `joinedload` — because `.unique()`
is no longer called in that path. `immediateload` is unaffected (it still calls
`.unique()` unconditionally).

## Checklist

This pull request is:

  Good to go, no issue or tests are needed
- [x] A short code fix  
  please include the issue number, and create an issue if none exists, which
  must include a complete example of the issue. one line code fixes without an
  Please include: `Fixes: #<issue number>` in the commit message  
  please include tests. one line code fixes without tests will not be accepted.
  please include the issue number, and create an issue if none exists, which
  must include a complete example of how the feature would look.  
  Please include: `Fixes: #13339` in the commit message  
  please include tests.

Fixes: #13339

Have a nice day!

### DIFF
--- a/lib/sqlalchemy/engine/result.py
+++ b/lib/sqlalchemy/engine/result.py
@@ -664,6 +664,11 @@ class Result(_WithKeys, ResultInternal[Row[Unpack[_Ts]]]):
         self._unique_filter_state = (set(), strategy)
         return self
 
+    @property
+    def has_unique_filter(self) -> bool:
+        """True if unique filtering has been requested on this result."""
+        return self._unique_filter_state is not None
+
     def columns(self, *col_expressions: _KeyIndexType) -> Self:
         r"""Establish the columns that should be returned in each row.
 
--- a/lib/sqlalchemy/engine/result.py
+++ b/lib/sqlalchemy/engine/result.py
@@ -664,11 +664,6 @@ class Result(_WithKeys, ResultInternal[Row[Unpack[_Ts]]]):
         self._unique_filter_state = (set(), strategy)
         return self
 
-    @property
-    def has_unique_filter(self) -> bool:
-        """True if unique filtering has been requested on this result."""
-        return self._unique_filter_state is not None
-
     def columns(self, *col_expressions: _KeyIndexType) -> Self:
         r"""Establish the columns that should be returned in each row.
 
@@ -1945,6 +1940,7 @@ class ChunkedIteratorResult(IteratorResult[Unpack[_Ts]]):
         self.raw = raw
         self.iterator = itertools.chain.from_iterable(self.chunks(None))
         self.dynamic_yield_per = dynamic_yield_per
+        self.context: Any = None
 
     @_generative
     def yield_per(self, num: int) -> Self:
--- a/lib/sqlalchemy/engine/result.py
+++ b/lib/sqlalchemy/engine/result.py
@@ -1924,6 +1924,8 @@ class ChunkedIteratorResult(IteratorResult[Unpack[_Ts]]):
 
     """
 
+    context: Any = None
+
     def __init__(
         self,
         cursor_metadata: ResultMetaData,
@@ -1940,7 +1942,6 @@ class ChunkedIteratorResult(IteratorResult[Unpack[_Ts]]):
         self.raw = raw
         self.iterator = itertools.chain.from_iterable(self.chunks(None))
         self.dynamic_yield_per = dynamic_yield_per
-        self.context: Any = None
 
     @_generative
     def yield_per(self, num: int) -> Self:
--- a/lib/sqlalchemy/engine/result.py
+++ b/lib/sqlalchemy/engine/result.py
@@ -1819,6 +1819,7 @@ class IteratorResult(Result[Unpack[_Ts]]):
 
     _hard_closed = False
     _soft_closed = False
+    context: Any = None
 
     def __init__(
         self,
@@ -1923,8 +1924,6 @@ class ChunkedIteratorResult(IteratorResult[Unpack[_Ts]]):
     .. versionadded:: 1.4
 
     """
-
-    context: Any = None
 
     def __init__(
         self,
--- a/lib/sqlalchemy/orm/strategies.py
+++ b/lib/sqlalchemy/orm/strategies.py
@@ -3425,12 +3425,7 @@ class _SelectInLoader(_PostLoader, util.MemoizedSlots):
                 },
                 execution_options=execution_options,
             )
-            # the inner query only produces duplicate rows when a nested
-            # joinedload on a collection is in effect, in which case
-            # `instances()` will have already flagged the result for uniquing
-            # (see loading.instances). uniquing rows has measurable cost
-            # (hashing each Row), so skip the call when not required.
-            if result.has_unique_filter:
+            if result.context is not None and result.context.requires_uniquing:
                 result = result.unique()
             data = {k: v for k, v in result}
 
@@ -3479,10 +3474,7 @@ class _SelectInLoader(_PostLoader, util.MemoizedSlots):
                 params={"primary_keys": primary_keys},
                 execution_options=execution_options,
             )
-            # uniquing only needed when a nested joinedload on a collection
-            # inflates rows; otherwise `instances()` leaves the result
-            # without a unique filter and per-row hashing is pure overhead.
-            if result.has_unique_filter:
+            if result.context is not None and result.context.requires_uniquing:
                 result = result.unique()
             data = collections.defaultdict(list)
             for k, v in itertools.groupby(result, lambda x: x[0]):
--- a/lib/sqlalchemy/orm/strategies.py
+++ b/lib/sqlalchemy/orm/strategies.py
@@ -3420,8 +3420,7 @@ class _SelectInLoader(_PostLoader, util.MemoizedSlots):
                 q,
                 params={
                     "primary_keys": [
-                        key[0] if query_info.zero_idx else key
-                        for key in chunk
+                        key[0] if query_info.zero_idx else key for key in chunk
                     ]
                 },
                 execution_options=execution_options,
--- a/lib/sqlalchemy/orm/strategies.py
+++ b/lib/sqlalchemy/orm/strategies.py
@@ -3470,15 +3470,18 @@ class _SelectInLoader(_PostLoader, util.MemoizedSlots):
                 for key, state, state_dict, overwrite in chunk
             ]
 
+            result = context.session.execute(
+                q,
+                params={"primary_keys": primary_keys},
+                execution_options=execution_options,
+            )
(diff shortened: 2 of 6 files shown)


lib/sqlalchemy/orm/strategies.py
```diff
@@ -3416,19 +3416,24 @@ def _load_via_child(
...
+                },
+                execution_options=execution_options,
+            )
+            # the inner query only produces duplicate rows when a nested
+            # joinedload on a collection is in effect, in which case
+            # `instances()` will have already set `_unique_filter_state` on
+            # the result as a guard (see loading.instances). uniquing rows
+            # has measurable cost (hashing each Row), so skip the call when
+            # the result didn't request it.
+            if result._unique_filter_state is not None:
```
CaselIT: I think it should use the flag for this, that's `compile_state.multi_row_eager_loaders`

zzzeek: maybe QueryContext should communicate this instead though, rather than digging into compile_state.

ollz272: The context here is the outer query's execute state; the inner selectin query has its own compile state that instances() already surfaces via _unique_filter_state. Using context.compile_state.multi_row_eager_loaders would incorrectly skip .unique() when the outer query lacks joined collections but the inner selectin query has a nested joinedload — see test_secondary_w_criteria_caching[joinedload-selectinload]

zzzeek: im suggesting adding a new attribute to QueryContext that communicates exactly whether or not loading.py has required uniquing

zzzeek: within the selectinload part.  the QueryContext that's local to the selectinload

ollz272: hm, interesting. Can see where to add to QueryContext - but the context passed is from the outer query, so where would we access the QueryContext thats local to the selectinload?

zzzeek: ah indeed, we'd need to add an accessor to ChunkedIteratorResult

my only argument is not to reach into a private attribute here

ollz272: Updated to have a public accessor.

lib/sqlalchemy/engine/result.py
```diff
@@ -664,6 +664,11 @@ def unique(self, strategy: Optional[_UniqueFilterType] = None) -> Self:
         self._unique_filter_state = (set(), strategy)
         return self
 
+    @property
+    def has_unique_filter(self) -> bool:
```
zzzeek: this is not enough, because the caller is still assuming things based on an implementation detail of how loading.py works.

we can't add ORM-specific accessors to ChunkedIteratorResult so I propose we add "context" that's any `object`, selectinloader casts it to QueryContext and we get `context.requires_uniquing` from that.

ollz272: Thanks @zzzeek, think i've done it right..

lib/sqlalchemy/engine/result.py
```diff
@@ -1940,6 +1940,7 @@ def __init__(
         self.raw = raw
         self.iterator = itertools.chain.from_iterable(self.chunks(None))
         self.dynamic_yield_per = dynamic_yield_per
+        self.context: Any = None
```
CaselIT: let's move this to classlevel so by default the instances will access the none on the class

test/orm/test_selectin_relations.py
```diff
@@ -4227,6 +4227,45 @@ def go():
...
+                    "WHERE a_b.a_id IN "
+                    "(__[POSTCOMPILE_primary_keys])",
+                    {"primary_keys": [1, 2]},
+                ),
+            )
+
+        eq_(sorted(b.id for b in results[0].bs), [1, 2])
+        eq_(sorted(b.id for b in results[1].bs), [1])
+
+        # belt-and-suspenders: no duplicates in either collection
```
CaselIT: can remove this comment

test/orm/test_selectin_relations.py
```diff
@@ -4227,6 +4227,45 @@ def go():
...
+        for a in results:
+            b_ids = [b.id for b in a.bs]
+            assert len(b_ids) == len(
```
CaselIT: nit: we could use eq_ like just above

test/orm/test_selectin_relations.py
```diff
@@ -4412,6 +4451,142 @@ def go():
...
+                        id=3,
+                        email="a3@example.com",
+                        dingalings=[Dingaling(id=3, data="d3")],
+                    )
+                ],
+            )
+        )
+        sess.commit()
+
+    def test_selectin_with_nested_joined_collection_still_dedupes(self):
```
CaselIT: I'm thinking if it could make sense to parametrize the inner loader to have also a selectin in addition to a joinedload, but it's likely already tested somewhere else

test/orm/test_selectin_relations.py
```diff
@@ -4412,6 +4451,142 @@ def go():
...
+            return (
+                sess.query(User)
+                .options(
+                    selectinload(User.addresses).joinedload(Address.dingalings)
+                )
+                .order_by(User.id)
+                .all()
+            )
+
+        # assert exactly 2 SQL queries: 1 for User, 1 for the selectin
```
CaselIT: maybe we could dry up or remove these comments. these are already above

test/orm/test_selectin_relations.py
```diff
@@ -4412,6 +4451,142 @@ def go():
...
+
+        eq_(len(users), 2)
+
+        u1 = users[0]
+        # Address 1 has 2 dingalings; without .unique() the selectin result
+        # would produce 2 rows for address 1 and loading.require_unique would
+        # raise InvalidRequestError.  The assertions below are belt-and-
+        # suspenders for if the conditional logic changes.
+        address_ids = [a.id for a in u1.addresses]
+        assert len(address_ids) == len(
```
CaselIT: nit: use eq_, below too

CaselIT (commented):
i think it's mostly there, the comment I left are mostly stylistic preferences 

Thanks!

CaselIT:
@zzzeek I think a changelog could make sense here, what do you think?

ollz272:
Thanks @CaselIT, applied those comments.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 980af0383b174271bb602a890f595c26f78ddcee of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 980af0383b174271bb602a890f595c26f78ddcee: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6717

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6717 has been **merged**. Congratulations! :)

PR #13348, opened 2026-06-04
fix(orm): make immediateload work with polymorphic entities

goingforstudying-ctrl:
Fixes #13204

When using immediateload on polymorphic entities (e.g. with_polymorphic() or as a nested sub-option on an inheritance chain), create_row_processor is invoked once per concrete subclass mapper. The first invocation registered the post-load callable with PASSIVE_OFF flags, but subsequent invocations would overwrite it with weaker PASSIVE_NO_FETCH_RELATED flags because _setup_for_recursion returned run_loader=False after seeing the path already registered.

The weaker flags lack RELATED_OBJECT_OK, causing _emit_lazyload to return PASSIVE_NO_RESULT without issuing SQL, so the eager load silently fell back to lazy load on attribute access.

Fix by skipping registration when run_loader=False and a loader is already registered for the same path/token. This preserves the stronger eager loader registration.

Includes tests covering both cases from the issue: immediateload at level 1 on with_polymorphic, and immediateload as a chained sub-option.

### DIFF
--- a/lib/sqlalchemy/orm/strategies.py
+++ b/lib/sqlalchemy/orm/strategies.py
@@ -1413,6 +1413,17 @@ class _ImmediateLoader(_PostLoader):
             # "use get" load.   the "_RELATED" part means it may return
             # instance even if its expired, since this is a mutually-recursive
             # load operation.
+            #
+            # if a loader was already registered for this path/token,
+            # don't overwrite it with the weaker PASSIVE_NO_FETCH_RELATED
+            # flags.  This can occur with polymorphic mappers where
+            # create_row_processor is invoked for each subclass and the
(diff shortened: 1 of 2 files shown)


sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision d4ec7a3aa9fd9ff671333130049c5e9d6f79619a of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change d4ec7a3aa9fd9ff671333130049c5e9d6f79619a: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6730

PR #13353, opened 2026-06-05
Fix check constraint reflection stripping nested parentheses incorrectly

### LINKED ISSUE #13157: PostgreSQL check constraint reflection strips brackets naively

goingforstudying-ctrl:
## Summary

Fixes #13157

When reflecting PostgreSQL check constraints, the regex used to strip outermost parentheses from the constraint expression used greedy matching `(.+)`, which caused it to match from the first opening paren to the LAST closing paren. This incorrectly stripped parentheses from expressions with nested groups.

## The Bug

For example, `CHECK (((x IS NULL OR y IS NULL) AND (x IS NULL OR y IS NULL)))` was reflected as:
```
```

Instead of the correct:
```
(x IS NULL OR y IS NULL) AND (x IS NULL OR y IS NULL)
```

The resulting expression is syntactically invalid because the outermost opening paren and the last closing paren are stripped, but the inner parens remain unmatched.

## The Fix

Replaces the greedy regex `^[\s\n]*\((.+)\)[\s\n]*$` with explicit paren-counting logic that only strips the outermost pair when the entire expression is wrapped in a single pair of parentheses.

The new logic:
1. Checks if the expression starts with `(` and ends with `)`
2. Counts parentheses depth from left to right
3. Only strips if the first `(` matches the last `)` (i.e., depth reaches 0 exactly at the final character)
4. If there are unmatched parens in the middle, leaves the expression unchanged

## Changes

- `lib/sqlalchemy/dialects/postgresql/base.py`: Replaced regex-based paren stripping with counting-based logic
- `test/dialect/postgresql/test_reflection.py`: Added `test_reflect_check_constraint_nested_parens()` with 4 test cases covering nested parens, single parens, no-extra parens, and complex nested expressions

## Verification

All test cases pass with the new logic, and existing mock-based tests for check constraint reflection remain compatible.

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -5538,9 +5538,25 @@ class PGDialect(default.DefaultDialect):
                 util.warn("Could not parse CHECK constraint text: %r" % src)
                 sqltext = ""
             else:
-                sqltext = re.compile(
-                    r"^[\s\n]*\((.+)\)[\s\n]*$", flags=re.DOTALL
-                ).sub(r"\1", m.group(1))
+                # Strip outermost matching parentheses, preserving
+                # inner grouping parentheses.  Use paren-counting
+                # instead of regex to avoid greedy/non-greedy pitfalls.
+                # e.g. "((a) AND (b))" -> "(a) AND (b)"
+                # e.g. "(a) AND (b)" -> "(a) AND (b)" (unchanged)
+                sqltext = m.group(1)
+                sqltext_stripped = sqltext.strip()
+                if sqltext_stripped.startswith("(") and sqltext_stripped.endswith(")"):
+                    depth = 0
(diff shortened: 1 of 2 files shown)


CaselIT:
You are now banned

PR #13360, opened 2026-06-10
Improve performance of selectinload result handling

ollz272:
The "SELECT .. IN" query emitted by selectinload now selects a single-column key directly rather than wrapping it in a Bundle, avoiding construction of a nested Row object per result row.  The loader consumes interim row tuples via Result._raw_row_iterator() when no uniquing is required, skipping per-row Row construction, and groups rows with a plain append loop in place of itertools.groupby.  Many-to-one selectin loads use a fast path that reads foreign key values directly from the parent instance dict, falling back to attribute-level access for expired attributes. subqueryload now groups its results on plain tuples rather than Row slices, which would incur Row construction and Row.__eq__ per row.

Adds tests pinning behavior of the rewritten result handling: the uselist=False multiple-rows warning, and many-to-one loads where the foreign key value matches no row or is NULL.

Benchmarked on an in-memory SQLite database (median of 100 runs, ms):

| case | before | after | Δ |
|---|---|---|---|
| selectin o2m (500×10) | 20.09 ms | 15.11 ms | −25% |
| selectin o2m, few big collections (20×500) | 35.88 ms | 26.73 ms | −26% |
| selectin m2m (500×10) | 13.48 ms | 8.87 ms | −34% |
| selectin nested (50×10×10) | 20.09 ms | 14.62 ms | −27% |
| selectin m2o (5000→200) | 17.34 ms | 15.64 ms | −10% |
| subqueryload o2m (500×10) | 23.92 ms | 21.10 ms | −12% |
| plain loads / joinedload | — | — | unchanged |

### DIFF
--- a/lib/sqlalchemy/orm/strategies.py
+++ b/lib/sqlalchemy/orm/strategies.py
@@ -13,7 +13,6 @@ implementations, and related MapperOptions."""
 from __future__ import annotations
 
 import collections
-import itertools
 from typing import Any
 from typing import Dict
 from typing import Literal
@@ -56,6 +55,7 @@ from .. import inspect
 from .. import log
 from .. import sql
 from .. import util
+from ..engine.result import IteratorResult
 from ..sql import util as sql_util
 from ..sql import visitors
 from ..sql.selectable import LABEL_STYLE_TABLENAME_PLUS_COL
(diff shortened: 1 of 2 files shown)


ollz272:
One test failure:
```
FAILED test/aaa_profiling/test_orm.py::SelectInEagerLoadTest_sqlite+pysqlite_3_37_2::test_round_trip_results - AssertionError: Adjusted function call count 246505 not within 5.0% of expected 271805, platform x86_64_linux_cpython_3.13_sqlite_pysqlite_dbapiunicode_cextensions. Rerun with --write-profiles to regenerate this callcount.
```
I cant do this rerun as im using an arm mac and therefore it won't update. However call count seems to have decreased which would have been the desired effect!

CaselIT:
Hi, 

Like I said in the other pr, there is a template for the PR. For these types of changes an issue is required first

ollz272:
@CaselIT sorry... as before discussion here: https://github.com/sqlalchemy/sqlalchemy/discussions/13362

PR #13361, opened 2026-06-10
Process ORM result rows as plain tuples without Row construction

ollz272:
Adds Result._all_interim_rows(), which returns the remaining rows as processed plain tuples, applying result processors and tuple filters but skipping Row object construction.  ORM loading uses this for its row fetch; its row getters are position-based itemgetters that accept any tuple-like row.  Results that require row logging or have scalar sources fall back to Row construction.


CaselIT:
Hi 

What's the issue it's trying to solve? We can't accept PR without a clear use case. 

The templeate for the pr is there for a reason

ollz272:
@CaselIT Sorry... discussion here: https://github.com/sqlalchemy/sqlalchemy/discussions/13362

PR #13371, opened 2026-06-12
fix catastrophic backtracking in hstore literal parser

### LINKED ISSUE #13370: hstore literal parser backtracks catastrophically on malformed input
### Describe the bug

The regular expression used to parse hstore literal strings, `HSTORE_PAIR_RE` in `lib/sqlalchemy/dialects/postgresql/hstore.py`, has an ambiguous alternation for the quoted key and value: `(\\ . | [^"])*`. A backslash matches both branches (`\\ .` consuming two characters, or `[^"]` consuming one), so for a run of backslashes the number of ways to split the run grows exponentially. When the surrounding pattern then fails to match, for example an opened quote with no closing quote, the engine explores all those partitions and the match time roughly doubles for every extra backslash.

`_parse_hstore` runs this regex in the pure-Python result processor, which is the active code path whenever the hstore type is used with `use_native_hstore=False` (the only option for drivers without a native hstore typecaster). Text reaching that parser that is not a well-formed hstore can therefore hang the interpreter.

### To Reproduce

```python
import time
from sqlalchemy.dialects.postgresql.hstore import HSTORE_PAIR_RE

for n in (20, 24, 26, 28):
    s = '"' + "\\" * n + "!"   # opened quote, run of backslashes, never closes
    start = time.time()
    HSTORE_PAIR_RE.match(s)
    print(n, round(time.time() - start, 4))
```

### Error

```
20 0.0023
24 0.0162
26 0.0405
28 0.1080
```

The time doubles per two added backslashes; around 40 backslashes it is already in the tens of seconds.

### Versions

- OS: macOS
- Python: 3.11
- SQLAlchemy: 2.1.0b3
- Database: PostgreSQL (pure-Python hstore path, `use_native_hstore=False`)
- DBAPI: n/a (parser-level)

### Additional context

Excluding the backslash from the negated class (`[^"\\]`) gives each character a single parse and removes the ambiguity, while still matching the same valid hstore output, since a backslash in valid hstore is always followed by another character and so is taken by the `\\ .` branch.

dxbjavid:
### Description

`HSTORE_PAIR_RE` parses quoted hstore keys and values with `(\\ . | [^"])*`, where a backslash matches both the `\\ .` branch and the `[^"]` branch. For a run of backslashes the engine can split it many ways, so when the rest of the pattern fails to match (an opened quote with no closing quote) it backtracks through all of them and the time roughly doubles per extra backslash. `_parse_hstore` runs this in the pure-Python result processor used when `use_native_hstore=False`, so malformed text on that path can hang. Excluding the backslash from the negated class gives each character one parse and matches the same valid hstore, since a backslash there is always followed by another character. Full reproduction is in #13370.

### Checklist

This pull request is:

- [x] A short code fix
	- Fixes: #13370
	- regression test added in `test/dialect/postgresql/test_types.py::HStoreTest::test_result_deserialize_malformed_no_backtracking`, which times out on the current parser and passes with the fix

**Have a nice day!**

### DIFF
--- a/doc/build/changelog/unreleased_20/13370.rst
+++ b/doc/build/changelog/unreleased_20/13370.rst
@@ -1,0 +1,11 @@
+.. change::
+    :tags: bug, postgresql
+    :tickets: 13370
+
+    Fixed issue where the regular expression used by the pure Python hstore
+    result processor, in use when ``use_native_hstore=False`` is set, could
+    exhibit catastrophic backtracking on malformed hstore text containing an
+    unterminated quoted segment followed by a run of backslashes, with
+    parsing time roughly doubling for each additional backslash.  Such input
+    is now rejected immediately; parsing of valid hstore values is
+    unchanged.  Pull request courtesy dxbjavid.
--- a/lib/sqlalchemy/dialects/postgresql/hstore.py
+++ b/lib/sqlalchemy/dialects/postgresql/hstore.py
@@ -319,12 +319,12 @@ class _HStoreMatrixFunction(sqlfunc.GenericFunction[Any]):
 HSTORE_PAIR_RE = re.compile(
     r"""
 (
-  "(?P<key> (\\ . | [^"])* )"       # Quoted key
+  "(?P<key> (\\ . | [^"\\])* )"     # Quoted key
 )
 [ ]* => [ ]*    # Pair operator, optional adjoining whitespace
 (
     (?P<value_null> NULL )          # NULL value
-  | "(?P<value> (\\ . | [^"])* )"   # Quoted value
+  | "(?P<value> (\\ . | [^"\\])* )" # Quoted value
 )
 """,
     re.VERBOSE,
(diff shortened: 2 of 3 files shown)


CaselIT:
thanks!

could you also add a changelog note in the 20 folder https://github.com/sqlalchemy/sqlalchemy/tree/main/doc/build/changelog/unreleased_20? (the number is the issue)

dxbjavid:
done, added 13370.rst to the unreleased_20 folder following the format of the other notes there.

CaselIT:
Thanks!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision f5eddae11c435c78f326b70a15357cbaf6d09337 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change f5eddae11c435c78f326b70a15357cbaf6d09337: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6732

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6732 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6735 has been **merged**. Congratulations! :)

dxbjavid:
Thank you for the review and merge. Much appreciated

Issue #11250, opened 2024-04-09
Login fails when odbc_connect PWD= includes a plus sign


gordthompson:
### Describe the bug

When `odbc_connect` is given a properly-formed ODBC connection string with a password that contains a plus sign, the plus sign is converted to a space and the login fails.

### SQLAlchemy Version in Use

2.1.0b1.dev0

### DBAPI (i.e. the database driver)

pyodbc

### Database Vendor and Major Version

MS SQL Server

### Python Version

3.8

### Operating system

all

### To Reproduce

```python
import pyodbc
import sqlalchemy as sa

connection_string = (
    "Driver=ODBC Driver 17 for SQL Server;"
    "Server=192.168.0.199;"
    "Database=test;"
    "UID=howie;"
    "PWD=ab+cd;"
)

# ___ this works ___

cnxn = pyodbc.connect(connection_string)
crsr = cnxn.cursor()
result = crsr.execute("SELECT 'Connected.' AS foo").fetchval()
print(result)

# ___ but this fails ___

connection_url = sa.URL.create(
    "mssql+pyodbc", query=dict(odbc_connect=connection_string)
)
engine = sa.create_engine(connection_url)
with engine.begin() as conn:
    result = conn.exec_driver_sql("SELECT 'Connected.' AS foo").scalar()
    print(result)
```

### Error

```
Traceback (most recent call last):
  File "/home/gord/git/sqla-gerrit/lib/sqlalchemy/engine/base.py", line 150, in __init__
    self._dbapi_connection = engine.raw_connection()
  File "/home/gord/git/sqla-gerrit/lib/sqlalchemy/engine/base.py", line 3298, in raw_connection
    return self.pool.connect()
  File "/home/gord/git/sqla-gerrit/lib/sqlalchemy/pool/base.py", line 449, in connect
    return _ConnectionFairy._checkout(self)
  File "/home/gord/git/sqla-gerrit/lib/sqlalchemy/pool/base.py", line 1263, in _checkout
    fairy = _ConnectionRecord.checkout(pool)
  File "/home/gord/git/sqla-gerrit/lib/sqlalchemy/pool/base.py", line 712, in checkout
    rec = pool._do_get()
  File "/home/gord/git/sqla-gerrit/lib/sqlalchemy/pool/impl.py", line 180, in _do_get
    self._dec_overflow()
[...]

PR #13379, opened 2026-06-16
quote driver name and pass-through keys in pyodbc connect string

### LINKED ISSUE #13380: ODBC connection string injection via driver name and pass-through parameter keys in pyodbc connector
### Describe the bug

The pyodbc connector's `check_quote()` function brace-quotes parameter values containing `;` or leading `{` (added in #8062), but two other parts of the connection string are emitted unquoted:

1. The driver name in `DRIVER={...}` — an embedded `}` closes the brace early
2. The keys of pass-through query parameters — a `;` starts a new ODBC attribute

When any part of a connection URL is built from untrusted input, this allows injection of arbitrary ODBC connection attributes.

The mssql-python connector at `dialects/mssql/mssqlpython.py` has the same pass-through key issue.

### To Reproduce

```python
from sqlalchemy.dialects.mssql import pyodbc
from sqlalchemy.engine import url

# --- Case 1: closing brace in driver name ---
# A "}" in the driver value escapes the DRIVER={...} wrapper.
# %7D = "}", %3D = "="
u = url.make_url(
    "mssql+pyodbc://user:pass@host/db"
    "?driver=Evil%7DUID%3Dhacker"
)
dialect = pyodbc.dialect()
cargs, cparams = dialect.create_connect_args(u)
print("Connection string:", cargs[0])
# CURRENT (broken):
#   DRIVER={Evil}UID=hacker};Server=host;Database=db;UID=user;PWD=pass
#
#   The } closes the brace early; UID=hacker becomes a
#   separate connection attribute, overriding the real UID.

# EXPECTED (fixed):
#   DRIVER={Evil}}UID=hacker};Server=host;Database=db;UID=user;PWD=pass
#
#   Doubled }} is treated as a literal } inside the braces.


# --- Case 2: semicolon in parameter key ---
# A ";" in a query-string key name breaks out of the token.
# %3B = ";"
u2 = url.make_url(
    "mssql+pyodbc://user:pass@host/db"
    "?driver=ODBC+Driver+17&foo%3BUID=hacker"
)
cargs2, _ = dialect.create_connect_args(u2)
print("Connection string:", cargs2[0])
# CURRENT (broken):
#   ...;foo;UID=hacker
#
#   The ; in the key starts a new attribute, so UID=hacker is
#   parsed as a separate connection parameter.

# EXPECTED (fixed):
#   ...;{foo;UID}=hacker
#
#   Brace-quoted key keeps the semicolon inside a single token.
```

### Versions

- SQLAlchemy: all current versions
- DBAPI: pyodbc, mssql-python
- Database: MSSQL (any)
- Python: all

### Additional context

This is the same class of issue as #8062 (which added `check_quote` for values). The fix is to apply the same escaping to the driver name (`}` → `}}`) and to parameter keys (via `check_quote`).

Pull request: #13379

dxbjavid:
### Description

Fixes: #13380

The pyodbc connector assembles the ODBC connection string by joining `keyword=value` tokens with `;`. Parameter values are brace-quoted by `check_quote` when they contain a `;` or a leading `{`, which is the handling added in #8062 / #11250. Two tokens skip that quoting though: the driver name, which is always wrapped in `DRIVER={...}` without escaping an embedded `}`, and the names of pass-through parameters, which are emitted as-is. A `}` in the driver value closes the brace early, and a `;` in a parameter name starts a new attribute, so the remainder gets read by the driver manager as further connection attributes.

For example `mssql+pyodbc://username:password@hostspec/database?driver=foob%7DUID%3Devil` builds, on the unpatched connector:

```
DRIVER={foob}UID=evil};Server=hostspec;Database=database;UID=username;PWD=password
```

where `UID=evil` has escaped the braces. Similarly a parameter named `foo;UID` lands as a bare `;foo;UID=evil`, adding a `UID` attribute. After the change the driver's `}` is doubled and the keys pass through the same `check_quote`, so each stays a single token (`DRIVER={foob}}UID=evil}` and `{foo;UID}=evil`).

This is reachable wherever any part of the URL (the driver, or extra ODBC keywords) is built from untrusted input. The mssql-python connector shares the pass-through-key handling and is fixed alongside. Tests are added next to the existing `test_pyodbc_token_injection` cases.

### Checklist

This pull request is:

- [x] A short code fix
	- Fixes #13380
	- builds on the ODBC token-quoting from #8062 / #11250
	- tests included in `test/dialect/mssql/test_engine.py`

**Have a nice day!**

### DIFF
--- a/test/dialect/mssql/test_engine.py
+++ b/test/dialect/mssql/test_engine.py
@@ -294,6 +294,7 @@ class ParseConnectTest(fixtures.TestBase):
                 "somehost%3BPORT%3D50001",
                 "somedb%3BPORT%3D50001",
             ),
+            "driver=foob",
             (
                 "DRIVER={foob};Server=somehost%3BPORT%3D50001;"
                 "Database={somedb;PORT=50001};UID={someuser;PORT=50001};"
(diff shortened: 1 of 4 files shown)


test/dialect/mssql/test_engine.py
```diff
@@ -329,6 +329,48 @@ def test_pyodbc_token_injection(self, tokens, connection_string):
             connection,
         )
 
+    def test_pyodbc_driver_token_injection(self):
```
zzzeek: can these tests be moved to be additional combinations on top of the existing test_pyodbc_token_injection test?   the existing cases on test_pyodbc_token_injection would need to be amended to also include the query string (for the two existing cases, it would be `"driver=foob"`)

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 03376dd72468e52e0149a61d253216211c3a28e7 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 03376dd72468e52e0149a61d253216211c3a28e7: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6736

test/dialect/mssql/test_engine.py
dxbjavid: done, folded both into test_pyodbc_token_injection as two extra combinations and added the query string as a parametrised arg, so the two existing cases now carry driver=foob explicitly. the new driver_brace_injection and pass_through_key_injection cases pass the query through the same way.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 815e8fb7515b9662c924799a0efb2727003debe0 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 815e8fb7515b9662c924799a0efb2727003debe0 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6736

zzzeek:
needs pep8, which you can get by installing pre-commit

dxbjavid:
good catch, that one line was over what black wanted. ran black and flake8 over the touched files and pushed the reformat, token injection cases still pass.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 34433408bff86a9044be5d431b31688ab6d7578d of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 34433408bff86a9044be5d431b31688ab6d7578d added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6736

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6736 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6740 has been **merged**. Congratulations! :)

Issue #13389, opened 2026-06-21
Validate test provisioning follower identifiers before composing DDL


cjc0013:
## Describe the bug

SQLAlchemy's test provisioning path generates follower database identifiers with a narrow expected shape, such as `test_` plus 12 lowercase hex characters. Some backend provisioning helpers assume that shape when composing temporary create/drop DDL for the test suite.

If a malformed follower identifier is passed directly to the provisioning helper, or is read from a provisioning reaper idents file, the helper currently accepts the value and formats it into the DDL string instead of failing before SQL is emitted.

This is limited to SQLAlchemy's own test provisioning helpers. It does not affect application query construction or runtime SQLAlchemy query behavior.

## Optional link from https://docs.sqlalchemy.org which documents the behavior that is expected

Not applicable. This is an internal test provisioning guard.

## SQLAlchemy Version in Use

main branch at `ddf3b6589fd6ccb2affbaea5d4f400a8c1ad02d8`

## DBAPI

Not applicable; the reproducer below uses a fake connection and does not require a database driver.

## Database Vendor and Major Version

Not applicable for the reproducer. The example exercises the MySQL test provisioning helper offline.

## Python Version

Python 3.11

## Operating system

Windows; the reproducer is OS-independent.

## To Reproduce

```python
from sqlalchemy.testing.provision import create_db
from sqlalchemy.dialects.mysql import provision as mysql_provision  # noqa: F401


class CaptureConn:
    def __init__(self):
        self.sql = []

    def exec_driver_sql(self, statement, *args, **kw):
        self.sql.append(statement)

    def begin(self):
        return self

    def __enter__(self):
        return self

    def __exit__(self, *exc):
        return False


class CaptureEngine:
    def __init__(self):
        self.conn = CaptureConn()

    def begin(self):
        return self.conn


engine = CaptureEngine()
ident = "test_012345abcdef; SELECT 1; --"

try:
    create_db.fns["mysql"](None, engine, ident)
    print("exception=none")
except Exception as err:
    print("exception=%s: %s" % (type(err).__name__, err))

print("captured_sql_count=%d" % len(engine.conn.sql))
for statement in engine.conn.sql:
    print(statement)
```

## Error

Current behavior on main:

```text
exception=none
captured_sql_count=3
CREATE DATABASE test_012345abcdef; SELECT 1; -- CHARACTER SET utf8mb4
[...]

PR #13390, opened 2026-06-21
Validate test provisioning follower identifiers before composing DDL

cjc0013:
### Description

This adds validation for SQLAlchemy's test provisioning follower identifiers before backend helpers compose create/drop DDL for temporary test databases, schemas, or users.

The normal pytest-xdist path already generates identifiers shaped like `test_` plus 12 lowercase hex characters. The new helper makes that expected shape explicit, and only allows known derived provisioning suffixes (`_test_schema`, `_test_schema_2`, `_ts1`, `_ts2`) where those derived names are intentionally used.

This is a test/provisioning hardening change. It does not change SQLAlchemy runtime query behavior.

The regression tests use fake connections and do not require a live database. Before the guard, the MySQL provisioning helper accepts malformed text and composes it into DDL strings. After the guard, the same input raises `ValueError` before any SQL is emitted.

Changes:

- Add `validate_follower_ident()` in `sqlalchemy.testing.provision`.
- Validate shared create/drop/setup/reap provisioning entry points.
- Validate backend create/drop/configure/reap helpers before they compose follower DDL.
- Add offline tests for accepted generated IDs and rejected malformed IDs.

Verification:

```text
python -m pytest test/dialect/test_provision.py -q
# 5 passed

python -m compileall -q lib/sqlalchemy test/dialect/test_provision.py
# passed

git diff --check
# passed
```

### Checklist

This pull request is:

- [x] A short code fix
  - Issue: `Fixes: #13389`
  - Tests included: `test/dialect/test_provision.py`

Issue: https://github.com/sqlalchemy/sqlalchemy/issues/13389

### DIFF
--- a/lib/sqlalchemy/dialects/mssql/provision.py
+++ b/lib/sqlalchemy/dialects/mssql/provision.py
@@ -26,6 +26,7 @@ from ...testing.provision import normalize_sequence
 from ...testing.provision import post_configure_engine
 from ...testing.provision import run_reap_dbs
 from ...testing.provision import temp_table_keyword_args
+from ...testing.provision import validate_follower_ident
 
 
 @post_configure_engine.for_db("mssql")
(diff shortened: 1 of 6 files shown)


zzzeek:
please dont initiate bug reports + PRs on an automated basis

PR #13405, opened 2026-06-30
# Cython batch instance processor — ~10–30% on most loads (collection joinedload −5%)

### LINKED ISSUE #13407: Cython batch instance processor — ~10–30% on most loads
### Discussed in https://github.com/sqlalchemy/sqlalchemy/discussions/13378

<sup>Originally posted by **ollz272** June 16, 2026</sup>

Add `orm/_loading_cy.py` with `_InstancesBatch`, a whole-chunk row processor
built by `_instance_processor()` for the non-refresh, non-polymorphic load path.
It replaces the per-row `[proc(row) for row in fetch]` comprehension in
`instances()`, hoisting all per-query-constant conditions out of the row loop:

- identity-map lookups go against `WeakInstanceDict._dict` directly;
- `InstanceState` construction is inlined for the default
  `ClassManager` / `InstanceState` combination;
- for tuple rows, the primary key and quick column populators are applied by
  position, with itemgetter indexes recovered by probing the getters against a
  `range` object.

Rows for instances already in the identity map delegate to `_instance()`, which
keeps the partial-population, populate_existing and version-check paths. Runs as
plain Python when the extension is not compiled. Most of the speedup is the
batch *restructuring* itself (it wins uncompiled too); `-O3` compilation adds a
smaller further gain (~2–6%), to weigh against the cost of a new compiled module.

## Benchmarks

Python 3.14, in-memory SQLite, 50 repeats, extension compiled at `-O3`.

| case                 | before | after  |     Δ |
|----------------------|-------:|-------:|------:|
| subquery_o2m         | 15.725 | 11.066 | −29.6% |
| selectin_o2m         | 14.760 | 11.920 | −19.2% |
| selectin_o2m_few_big | 25.880 | 20.967 | −19.0% |
| selectin_nested      | 14.908 | 12.148 | −18.5% |
| selectin_m2o         | 13.849 | 11.354 | −18.0% |
| plain_wide           |  6.794 |  5.884 | −13.4% |
| joined_m2o | 12.551 | 11.109 | −11.5% |
| plain_small          |  2.130 |  1.929 |  −9.4% |
| selectin_m2m         |  9.681 |  9.538 |  −1.5% |
| joined_o2m   | 15.860 | 16.721 |  +5.4% |

Big improvements across the board, though a slight regression on joined loads in the o2m case. I believe this will be due to some double counting going on so could potentially gate this somehow, but also o2m is not recommended for joinedloads _anyway_ so could be a regression worth taking when the m2o case improves.</div>

ollz272:
### Description

Add `orm/_loading_cy.py` with `_InstancesBatch`, a whole-chunk row processor
built by `_instance_processor()` for the non-refresh, non-polymorphic load path.
It replaces the per-row `[proc(row) for row in fetch]` comprehension in
`instances()`, hoisting all per-query-constant conditions out of the row loop:

- identity-map lookups go against `WeakInstanceDict._dict` directly;
- `InstanceState` construction is inlined for the default
  `ClassManager` / `InstanceState` combination;
- for tuple rows, the primary key and quick column populators are applied by
  position, with itemgetter indexes recovered by probing the getters against a
  `range` object.

Rows for instances already in the identity map delegate to `_instance()`, which
keeps the partial-population, populate_existing and version-check paths. Runs as
plain Python when the extension is not compiled. Most of the speedup is the
batch *restructuring* itself (it wins uncompiled too); `-O3` compilation adds a
smaller further gain (~2–6%), to weigh against the cost of a new compiled module.

Fixes: #13407

## Benchmarks

`min` ms, vs `main`. Python 3.14, in-memory SQLite, 50 repeats, extension
compiled at `-O3`.

| case                 | before | after  |     Δ |
|----------------------|-------:|-------:|------:|
| subquery_o2m         | 15.725 | 11.066 | −29.6% |
| selectin_o2m         | 14.760 | 11.920 | −19.2% |
| selectin_o2m_few_big | 25.880 | 20.967 | −19.0% |
| selectin_nested      | 14.908 | 12.148 | −18.5% |
| selectin_m2o         | 13.849 | 11.354 | −18.0% |
| plain_wide           |  6.794 |  5.884 | −13.4% |
| joined_m2o (scalar)  | 12.551 | 11.109 | −11.5% |
| plain_small          |  2.130 |  1.929 |  −9.4% |
| selectin_m2m         |  9.681 |  9.538 |  −1.5% |
| joined_o2m (coll.)   | 15.860 | 16.721 |  +5.4% |

## Note: scalar joinedload improves, collection joinedload regresses

The two joinedload shapes behave oppositely:

- **`joined_m2o`** (scalar — the common joinedload usage, m2o/o2o): **~12%
  faster**. Each row drives a distinct new entity, which the batch handles
  directly.
- **`joined_o2m`** (collection): **~5% slower**. joinedload-o2m rows are ~90%
  repeat-parent rows: the batch builds an identity key, finds the parent already
  present, then delegates to `_instance()`, which rebuilds the same key — a
  double lookup. I can try investigate a way around this, though believe
  this is a bit of an anti pattern anyway.

Full `test/orm` suite passes (10,862) both compiled and uncompiled.
`test/profiles.txt` call-counts will need regeneration on supported platforms.

### Checklist

This pull request is:

- [ ] A short code fix
- [ x ] A new feature implementation
	- Please include: `Fixes: #13407` in the commit message
	- please include tests.

**Have a nice day!**

Note - Created a discussion for this but no issue was made - but was told that they were waiting for a patch. Tried working with gerrit again but couldn't get it working - sorry!.


CaselIT:
Thanks for looking into this.

I've created an issue, it's #13407

zzzeek:
wow this is pretty out there.    ill have to find time to get my head around this.

zzzeek:
just to give you an update, since i know these are well thought out PRs, I'm basically trying to work through the PR backlog as I have time.   because people submit PRs now at a high rate, including lots of people submitting PRs for the same issue (likely because everyone's claude finds the same ones), we have a lot of work to do with these.

your stuff has value here so I do want to get to looking at this but it might be awhile.

zzzeek:
if you do have time, it would be handy if you could rebase this as it seems to be conflicting with more recent changes

zzzeek:
another question....what's the background behind this only addressing the "single instance" use case?    wouldn't it be simpler if we had just one implementation of `_instance()` rather than two ?

PR #13411, opened 2026-07-04
fix: Pool._invalidate lock and _finalize_fairy connection leak on BaseException

webdevdot:
## Summary

Fixes two bugs in `lib/sqlalchemy/pool/base.py`:

- **#24** `Pool._invalidate`: adds `_invalidate_lock` (`threading.Lock`) to `Pool.__init__` and acquires it around the `_invalidate_time` read-compare-write, preventing concurrent calls from silently discarding a newer invalidation timestamp
- **#26** `_finalize_fairy`: moves `connection_record.checkin()` into the `finally` block of the inner `try/except` so that a re-raised `BaseException` (e.g. `KeyboardInterrupt`, `SystemExit`) no longer permanently leaks the connection back to the pool; the original post-block call is kept as a no-op fallback for the `dbapi_connection is None` path

## Test plan
- [ ] `python -m pytest test/engine/test_pool.py -x -q`
- [ ] Verify GC-checkin tests (`test_checkin_event_gc`) still pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/sqlalchemy/pool/base.py
+++ b/lib/sqlalchemy/pool/base.py
@@ -280,6 +280,7 @@ class Pool(log.Identified, event.EventTarget):
         self._creator = creator
         self._recycle = recycle
         self._invalidate_time = 0
+        self._invalidate_lock = threading.Lock()
         self._pre_ping = pre_ping
         self._reset_on_return = util.parse_user_argument_for_enum(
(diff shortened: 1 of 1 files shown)


CaselIT:
Oh its a bot. Blocked

PR #13417, opened 2026-07-06
Add type annotations to sqlalchemy.ext.serializer

mahdiaghtaee:
## Summary

Adds focused type annotations to the legacy `sqlalchemy.ext.serializer`
extension.

## Details

This keeps the change focused on typing without changing serializer runtime
behavior.

- Removes the module-level `mypy: ignore-errors` marker.
- Adds precise annotations for serializer and deserializer internals.
- Uses small structural protocols for metadata and scoped-session context.
- Uses `object` for values passed to serialization.
- Keeps `Any` only at the public deserialization return boundary, where the
  resulting runtime type cannot be determined statically.
- Accounts for scoped sessions bound to either `Engine` or `Connection`.
- Adds typing coverage for the documented `MetaData` and `scoped_session`
  usage.

### DIFF
--- a/lib/sqlalchemy/ext/serializer.py
+++ b/lib/sqlalchemy/ext/serializer.py
@@ -4,7 +4,6 @@
 #
 # This module is part of SQLAlchemy and is released under
 # the MIT License: https://www.opensource.org/licenses/mit-license.php
-# mypy: ignore-errors
 
 """Serializer/Deserializer objects for usage with SQLAlchemy query structures,
 allowing "contextual" deserialization.
@@ -68,12 +67,20 @@ needed for:
 
 """
 
+from __future__ import annotations
+
 from io import BytesIO
 import pickle
 import re
+from typing import Any
+from typing import cast
+from typing import IO
+from typing import Mapping
+from typing import Protocol
 
 from .. import Column
 from .. import Table
+from ..engine import Connection
 from ..engine import Engine
 from ..orm import class_mapper
 from ..orm.interfaces import MapperProperty
@@ -85,9 +92,27 @@ from ..util import b64encode
 __all__ = ["Serializer", "Deserializer", "dumps", "loads"]
 
 
+class _HasTables(Protocol):
+    @property
+    def tables(self) -> Mapping[str, Table]: ...
+
+
+class _HasBind(Protocol):
+    @property
+    def bind(self) -> Engine | Connection | None: ...
+
+
+class _ScopedSession(Protocol):
+    def __call__(self) -> _HasBind: ...
+
+
+class _HasClauseElement(Protocol):
+    def __clause_element__(self) -> object: ...
+
+
 class Serializer(pickle.Pickler):
 
-    def persistent_id(self, obj):
+    def persistent_id(self, obj: object) -> str | None:
         # print "serializing:", repr(obj)
         if isinstance(obj, Mapper):
             id_ = "mapper:" + b64encode(pickle.dumps(obj.class_))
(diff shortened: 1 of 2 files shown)


ashm-dev (changes requested):
Sprinkling `Any` to bypass complex signatures is essentially disabling static analysis for this path. If we are going to type this, let's design proper generics/protocols instead of resorting to type erasure.

mahdiaghtaee:
Thanks for the feedback. That makes sense.

I kept the first version intentionally small, but I agree that adding broad Any annotations does not really improve static analysis much.

I’ll rework this to either use more precise types where practical, or narrow the change to a smaller part of the module where the typing can be meaningful without type erasure.

mahdiaghtaee:
I’ve reworked the typing based on the review feedback. The broad `Any` annotations have been replaced with more precise types, small structural `Protocol`s for metadata/session behavior, and targeted `cast()` calls only where deserialization is inherently dynamic. The module-level `mypy: ignore-errors` marker remains removed.

When convenient, could you please take another look at the updated patch?

lib/sqlalchemy/ext/serializer.py
```diff
@@ -124,61 +144,75 @@ def persistent_id(self, obj):
...
+                assert self.scoped_session is not None
                 return self.scoped_session()
             elif type_ == "engine":
                 return self.get_engine()
             else:
                 raise Exception("Unknown token: %s" % type_)
 
 
-def dumps(obj, protocol=pickle.HIGHEST_PROTOCOL):
+def dumps(obj: _T, protocol: int = pickle.HIGHEST_PROTOCOL) -> bytes:
```
zzzeek: what's T here?  this seems unnecessary

mahdiaghtaee: You're right — `_T` only appeared in the input position, so it did not
establish a useful type relationship. I've removed it and changed `dumps()`
to accept `object`.

I also tightened the surrounding annotations: the pickle protocol now accepts
`int | None`, session binds account for both `Engine` and `Connection`, and
`Any` is limited to the `loads()` return boundary where deserialization is
inherently dynamic. I added a typing regression case for the documented
`MetaData` / `scoped_session` usage.

lib/sqlalchemy/ext/serializer.py
```diff
@@ -124,61 +149,79 @@ def persistent_id(self, obj):
...
         self.scoped_session = scoped_session
         self.engine = engine
 
-    def get_engine(self):
-        if self.engine:
+    def get_engine(self) -> Engine | Connection | None:
+        if self.engine is not None:
             return self.engine
-        elif self.scoped_session and self.scoped_session().bind:
+        elif self.scoped_session is not None:
```
CaselIT, line 167: this is a runtime change but it's equivalent. i guess it's fine

lib/sqlalchemy/ext/serializer.py
```diff
@@ -124,61 +149,79 @@ def persistent_id(self, obj):
...
+    def persistent_load(self, id_: object) -> object:
         m = our_ids.match(str(id_))
         if not m:
             return None
         else:
             type_, args = m.group(1, 2)
             if type_ == "attribute":
                 key, clsarg = args.split(":")
-                cls = pickle.loads(b64decode(clsarg))
+                cls = cast(type[object], pickle.loads(b64decode(clsarg)))
```
CaselIT, line 180: since loads returns any, we can just use `cls: type[object]` avoiding the call to cast. same in the othres

lib/sqlalchemy/ext/serializer.py
```diff
@@ -124,61 +149,79 @@ def persistent_id(self, obj):
...
+                    _HasClauseElement, class_mapper(cls)
+                ).__clause_element__()
             elif type_ == "mapperprop":
                 mapper, keyname = args.split(":")
-                cls = pickle.loads(b64decode(mapper))
+                cls = cast(type[object], pickle.loads(b64decode(mapper)))
                 return class_mapper(cls).attrs[keyname]
             elif type_ == "table":
-                return self.metadata.tables[args]
+                metadata = cast(_HasTables, self.metadata)
```
CaselIT, line 195: remove the cast, there is no need for it. if you want place an `assert self.metadata is not none`. same below for the session branch

lib/sqlalchemy/ext/serializer.py
```diff
@@ -124,61 +149,79 @@ def persistent_id(self, obj):
 
 class Deserializer(pickle.Unpickler):
 
-    def __init__(self, file, metadata=None, scoped_session=None, engine=None):
+    def __init__(
+        self,
+        file: IO[bytes],
+        metadata: _HasTables | None = None,
+        scoped_session: _ScopedSession | None = None,
+        engine: Engine | None = None,
```
CaselIT, line 157: I'm not sure here if we actually need to have protocols for metadata and for scoped session, they are not wrong, just unsure of the need.

what do you think @zzzeek ? should we replace `_HasTables ` with `MetaData` and `_ScopedSession` for `scoped_session[Any]`?

CaselIT (changes requested):
left some comments and an open question also for the other maintainers.

thanks for the change so far!

mahdiaghtaee:
Thank you for the detailed reviews and guidance. I’m closing this pull request rather than continuing to iterate on the typing design in this legacy module.

PR #13430, opened 2026-07-13
escape single quotes in postgresql nextval() identifier rendering

### LINKED ISSUE #13429: postgresql nextval() rendering does not escape single quotes in sequence/schema names

```python

```

runs as its own SQL.

```python
# get_insert_default emits:
```

breaks out.

dxbjavid:
### Description

The PostgreSQL dialect renders `nextval('...')` in a few places by dropping the
sequence identifier straight into a single-quoted SQL string literal, and it
never escapes the single quote. A `'` in a sequence, table or schema name
therefore closes the literal early and whatever follows in the name is read as
SQL. `visit_sequence` reproduces with no database (full example in #13429), and
`get_insert_default` is reachable through a `schema_translate_map` schema name,
which in schema-per-tenant setups is quite often derived from request context.

The affected spots are `visit_sequence`, `fire_sequence` and
`get_insert_default` in `dialects/postgresql/base.py`. The first two already
quote the identifier via `format_sequence` but leave the surrounding quote
alone; `get_insert_default` did not route the name through the identifier
preparer at all, unlike its sibling `fire_sequence`. I've escaped the quote in
the rendered literal and made `get_insert_default` quote the name the same way
`fire_sequence` does. Ordinary unquoted names compile exactly as before, and
the added test covers the escaping.

### Checklist

This pull request is:

- [x] A short code fix
	- Fixes: #13429, which has a complete runnable example
	- regression test added in `test/dialect/postgresql/test_compiler.py`

**Have a nice day!**

### DIFF
--- a/doc/build/changelog/unreleased_20/13429.rst
+++ b/doc/build/changelog/unreleased_20/13429.rst
@@ -10,6 +10,8 @@
     ``schema_translate_map`` supplying such a schema name, or an explicit
     :class:`.Sequence` whose name carries a quote, previously produced a
     malformed statement in which the remainder of the name was interpreted as
-    SQL.  ``get_insert_default`` now also routes the sequence name through the
-    identifier preparer, matching the sibling ``fire_sequence`` method.  Pull
-    request courtesy dxbjavid.
+    SQL.  The sequence name for an implicit SERIAL sequence, which has no
+    :class:`.Sequence` object behind it, is now rendered through the identifier
+    preparer as well, matching the sibling ``fire_sequence`` method; as a
+    result an ordinary lower case name is no longer redundantly double quoted.
+    Pull request courtesy dxbjavid.
--- a/doc/build/changelog/unreleased_20/13429.rst
+++ b/doc/build/changelog/unreleased_20/13429.rst
@@ -1,0 +1,15 @@
+.. change::
+    :tags: bug, postgresql
+    :tickets: 13429
+
+    Fixed quoting in the PostgreSQL ``nextval()`` rendering used for sequence
+    default expressions and for server-side autoincrement pre-execution, so
+    that a single quote occurring in a sequence, table or schema name is
+    escaped rather than terminating the surrounding string literal.  In
+    particular :meth:`_engine.Connection.execution_options` with a
+    ``schema_translate_map`` supplying such a schema name, or an explicit
+    :class:`.Sequence` whose name carries a quote, previously produced a
+    malformed statement in which the remainder of the name was interpreted as
+    SQL.  ``get_insert_default`` now also routes the sequence name through the
+    identifier preparer, matching the sibling ``fire_sequence`` method.  Pull
+    request courtesy dxbjavid.
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -2295,8 +2295,8 @@ class PGCompiler(compiler.SQLCompiler):
         return f"power{self.function_argspec(fn)}"
 
     def visit_sequence(self, seq, **kw):
-        return "nextval('%s')" % self.preparer.format_sequence(seq).replace(
-            "'", "''"
+        return "nextval(%s)" % self.preparer.format_sequence_string_literal(
+            seq
         )
 
     def limit_clause(self, select, **kw):
@@ -3156,8 +3156,42 @@ class PGTypeCompiler(compiler.GenericTypeCompiler):
         return "JSONPATH"
 
 
+class _CompilerSequence:
+    """Minimal stand-in for :class:`.Sequence`.
+
+    Used for the implicit sequence behind a SERIAL column, where no
+    :class:`.Sequence` object exists but a name still has to be rendered
+    through :meth:`.IdentifierPreparer.format_sequence`.
+
+    """
+
+    __slots__ = ("name", "schema")
+
+    # the schema handed to us has already been resolved against the
+    # schema translate map, so don't let format_sequence() translate it
+    # a second time
+    _use_schema_map = False
+
+    def __init__(self, name, schema=None):
+        self.name = name
+        self.schema = schema
+
+
 class PGIdentifierPreparer(compiler.IdentifierPreparer):
     reserved_words = RESERVED_WORDS
+
+    def format_sequence_string_literal(self, sequence, use_schema=True):
+        """Render a sequence name as a SQL string literal.
+
+        ``nextval()`` and friends take the sequence as a string rather than
+        as an identifier, so the quoted identifier produced by
+        :meth:`.format_sequence` has to be escaped a second time for the
+        enclosing literal.
+
+        """
+        return "'%s'" % self.format_sequence(sequence, use_schema).replace(
+            "'", "''"
+        )
 
     def _unquote_identifier(self, value):
         if value[0] == self.initial_quote:
@@ -3360,10 +3394,8 @@ class PGExecutionContext(default.DefaultExecutionContext):
     def fire_sequence(self, seq, type_):
         return self._execute_scalar(
             (
-                "select nextval('%s')"
-                % self.identifier_preparer.format_sequence(seq).replace(
-                    "'", "''"
-                )
+                "select nextval(%s)"
+                % self.identifier_preparer.format_sequence_string_literal(seq)
             ),
             type_,
         )
@@ -3400,17 +3432,12 @@ class PGExecutionContext(default.DefaultExecutionContext):
                 else:
                     effective_schema = None
 
-                if effective_schema is not None:
-                    ident = "%s.%s" % (
-                        self.identifier_preparer.quote_identifier(
-                            effective_schema
-                        ),
-                        self.identifier_preparer.quote_identifier(seq_name),
+                exc = (
+                    "select nextval(%s)"
+                    % self.identifier_preparer.format_sequence_string_literal(
+                        _CompilerSequence(seq_name, effective_schema)
                     )
-                else:
-                    ident = self.identifier_preparer.quote_identifier(seq_name)
-
-                exc = "select nextval('%s')" % ident.replace("'", "''")
+                )
 
                 return self._execute_scalar(exc, column.type)
 
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -2295,7 +2295,9 @@ class PGCompiler(compiler.SQLCompiler):
         return f"power{self.function_argspec(fn)}"
 
     def visit_sequence(self, seq, **kw):
-        return "nextval('%s')" % self.preparer.format_sequence(seq)
+        return "nextval('%s')" % self.preparer.format_sequence(seq).replace(
+            "'", "''"
+        )
 
     def limit_clause(self, select, **kw):
         text = ""
@@ -3359,7 +3361,9 @@ class PGExecutionContext(default.DefaultExecutionContext):
         return self._execute_scalar(
             (
                 "select nextval('%s')"
-                % self.identifier_preparer.format_sequence(seq)
+                % self.identifier_preparer.format_sequence(seq).replace(
+                    "'", "''"
+                )
             ),
             type_,
         )
@@ -3397,12 +3401,16 @@ class PGExecutionContext(default.DefaultExecutionContext):
                     effective_schema = None
 
                 if effective_schema is not None:
-                    exc = 'select nextval(\'"%s"."%s"\')' % (
-                        effective_schema,
-                        seq_name,
+                    ident = "%s.%s" % (
+                        self.identifier_preparer.quote_identifier(
+                            effective_schema
+                        ),
+                        self.identifier_preparer.quote_identifier(seq_name),
                     )
                 else:
-                    exc = "select nextval('\"%s\"')" % (seq_name,)
+                    ident = self.identifier_preparer.quote_identifier(seq_name)
+
+                exc = "select nextval('%s')" % ident.replace("'", "''")
 
                 return self._execute_scalar(exc, column.type)
 
--- a/test/dialect/postgresql/test_compiler.py
+++ b/test/dialect/postgresql/test_compiler.py
@@ -160,7 +160,7 @@ class SequenceTest(fixtures.TestBase, AssertsCompiledSQL):
         # the schema is resolved by the execution context before it reaches
         # _CompilerSequence, so it must not be translated a second time
         preparer = postgresql.dialect().identifier_preparer
-        preparer = preparer._with_schema_translate({"foo": "bar"})
+        preparer = preparer._with_schema_translate({"bar": "foo"})
         eq_(
             preparer.format_sequence_string_literal(
                 _CompilerSequence("my_seq", "bar")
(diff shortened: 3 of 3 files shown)


lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -2295,7 +2295,9 @@ def visit_pow_func(self, fn, **kw):
         return f"power{self.function_argspec(fn)}"
 
     def visit_sequence(self, seq, **kw):
-        return "nextval('%s')" % self.preparer.format_sequence(seq)
+        return "nextval('%s')" % self.preparer.format_sequence(seq).replace(
```
CaselIT: I think format sequence should handle this, behaving like format_table

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -3397,12 +3401,16 @@ def get_insert_default(self, column):
                     effective_schema = None
 
                 if effective_schema is not None:
-                    exc = 'select nextval(\'"%s"."%s"\')' % (
-                        effective_schema,
-                        seq_name,
+                    ident = "%s.%s" % (
+                        self.identifier_preparer.quote_identifier(
```
CaselIT: use format_schema

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -3397,12 +3401,16 @@ def get_insert_default(self, column):
...
+                            effective_schema
+                        ),
+                        self.identifier_preparer.quote_identifier(seq_name),
```
CaselIT: format_sequence

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -3397,12 +3401,16 @@ def get_insert_default(self, column):
                     effective_schema = None
 
                 if effective_schema is not None:
```
CaselIT: @zzzeek any reason why this isn't done by format_sequence?

format_table seems to be handling the schema, so it may make sense for sequence to behave the same

this could be a 2.1 only change though

CaselIT (changes requested):
If thanks for filing the issue and starting the PR, but I think the implementation needs to be different

lib/sqlalchemy/dialects/postgresql/base.py
dxbjavid: the snag with doing the doubling inside format_sequence is that its output doesn't always land in a single-quoted literal. create/drop sequence, oracle's seq.nextval and mssql's NEXT VALUE FOR all embed it straight into the statement, so doubling the quote there would turn a name like se'q into se''q in those cases. the '' escape is really a property of the nextval('...') wrapping rather than of the identifier itself, so i think it wants to stay where we build that literal. happy to fold the three sites into a small helper if the repetition is the concern.

lib/sqlalchemy/dialects/postgresql/base.py
dxbjavid: can do. one thing worth flagging: format_schema goes through quote(), so it only quotes when the name needs it, whereas this branch currently always wraps in double quotes. moving to it lines get_insert_default up with fire_sequence but does drop the redundant quotes for ordinary names (nextval('t_id_seq') rather than nextval('"t_id_seq"')). fine by me if that's the intent.

lib/sqlalchemy/dialects/postgresql/base.py
dxbjavid: there's no Sequence object in this branch, only the raw seq_name string built up from the column, so format_sequence doesn't quite fit here. i can route it through the preparer's quote() (which is what format_schema and format_table use underneath) to land on the same result.

lib/sqlalchemy/dialects/postgresql/base.py
dxbjavid: for what it's worth format_sequence does already qualify with the schema (via quote_schema), so this branch is partly reinventing that. the reason it doesn't just call format_sequence is that get_insert_default never has a Sequence instance here, only the derived name. i'm happy to target 2.1 and rework this path so it goes through format_sequence properly if that's the direction you'd prefer. just say which way and i'll push.

lib/sqlalchemy/dialects/postgresql/base.py
CaselIT: I'm not suggesting to use replace, I'm suggesting to do like format_table, that calls `.quote()`

lib/sqlalchemy/dialects/postgresql/base.py
CaselIT: I don't think it's an issue, since if the schema does not need quoting using it or not is equivalent

lib/sqlalchemy/dialects/postgresql/base.py
CaselIT: will take a second look, but let's also see what mike replies

lib/sqlalchemy/dialects/postgresql/base.py
zzzeek: oh agree 100%, no way we should go through all the trouble to call format_sequence() and then have to modify the string further.   if it is truly necessary that format_sequence() produce two different forms of SQL then there would be some kind of boolean passed as an optional second argument.  but i can't really imagine that would be the case

lib/sqlalchemy/dialects/postgresql/base.py
```diff
@@ -3359,7 +3361,9 @@ def fire_sequence(self, seq, type_):
         return self._execute_scalar(
             (
                 "select nextval('%s')"
-                % self.identifier_preparer.format_sequence(seq)
+                % self.identifier_preparer.format_sequence(seq).replace(
+                    "'", "''"
+                )
```
zzzeek: same here.   this is definitely not how to do this

lib/sqlalchemy/dialects/postgresql/base.py
zzzeek: well in this codepath we are working from a PostgreSQL SERIAL datatype where we don't actually have an explicit Sequence object.

so options here include factoring preparer.format_sequence() into a super-method that's like `format_sequence_name(self, name, schema=None)`, or change the PG logic to actually create either a `Sequence` object or maybe a "shim" of some kind (an object `_CompilerSequence` that extends from `CompilerElement` and can use `__slots__` to make it cheap) that we pass into `format_sequence()` on the fly.   that latter version may be fine here since this isn't super perf critical.

lib/sqlalchemy/dialects/postgresql/base.py
zzzeek: as long as the generated SQL does the right thing when interpreted by the database

lib/sqlalchemy/dialects/postgresql/base.py
zzzeek: let's look into making a lightweight `_CompilerSequence` that's local to the postrgresql dialect so that we can just pass it as though it's a `Sequence`

zzzeek (changes requested):

lib/sqlalchemy/dialects/postgresql/base.py
dxbjavid: agreed, the post-hoc replace on format_sequence output should go. the wrinkle is that the escaping here isn't really a property of the identifier, it's a property of the wrapper: nextval's argument is a single-quoted string literal, and format_sequence gives back a double-quoted identifier that has to sit inside that literal, so a single quote in the name needs doubling at the literal layer. format_table never hits this because its callers drop the name straight into the statement, never inside quotes.

so rather than reach back into the format_sequence string, i think the clean version is to render the argument as an actual string literal: nextval(%s) with render_literal_value(format_sequence(seq), String()). that leaves format_sequence completely untouched and lets the String literal processor do the '' doubling (and the %% doubling, which the current hand-rolled version quietly misses). fire_sequence gets the same shape. does that line up with what you had in mind?

lib/sqlalchemy/dialects/postgresql/base.py
dxbjavid: same fix as visit_sequence, just on the execution-context side. one thing to note here: fire_sequence and get_insert_default don't have render_literal_value the way the compiler does, so i'd route the argument through String().literal_processor(self.dialect) (or a tiny helper) to get the same '' and %% escaping. flagging in case you'd rather it lived somewhere shared.

lib/sqlalchemy/dialects/postgresql/base.py
dxbjavid: the _CompilerSequence shim works for me, and it lets this branch drop its hand-rolled quote_identifier calls and go through format_sequence like fire_sequence does. one thing to settle so it doesn't bite us: format_sequence runs schema_for_object on whatever it's handed, and this branch has already resolved effective_schema through the connection's schema_translate_map, so if the shim carries that resolved schema and still opts into the map it'd get translated a second time. cleanest is to build the shim so it doesn't re-trigger the map (or hand format_sequence the raw name/schema and let it do the translation once). the outer nextval literal then needs the same string-literal escaping as the other two sites.

on format_schema vs always double-quoting: happy to let ordinary names render unquoted since you both confirmed that's equivalent. i'll target 2.1 for this path per Federico's note. want me to push the _CompilerSequence version, or would you prefer the format_sequence_name(name, schema=None) factoring you floated as the other option?

CaselIT: let's try with the _CompilerSequence  if it doesn't causes issues with the schema transalte map

lib/sqlalchemy/dialects/postgresql/base.py
dxbjavid: pushed. the escaping lives in a preparer method now, format_sequence_string_literal, so nothing patches up format_sequence output at the call site any more.

i did try the render_literal_value(String()) route i floated earlier and it doesn't hold up here: the String literal processor doubles percent signs, and _escape_identifier has already done that on the identifier, so a sequence named s%pct would come out as %%%%. keeping both escapes in the preparer keeps the identifier layer and the literal layer in step.

lib/sqlalchemy/dialects/postgresql/base.py
dxbjavid: done, gone with _CompilerSequence. it's a __slots__ shim local to the pg dialect carrying just name and schema, and get_insert_default hands it to format_sequence like the other two sites, so the hand-rolled quote_identifier calls are gone.

on the schema translate map worry: the shim sets _use_schema_map = False. get_insert_default has already resolved effective_schema through the connection, so without that opt-out format_sequence would hand back a __[SCHEMA_x] symbol for a name that had already been translated once, and _render_schema_translates would then have nothing sensible to do with it. added a test that renders through a translating preparer to pin it down.

lib/sqlalchemy/dialects/postgresql/base.py
dxbjavid: this branch goes through format_sequence now, which uses quote_schema underneath, so the unconditional double quoting is gone and an ordinary lower case name renders bare. matches what you and Mike both said above.

lib/sqlalchemy/dialects/postgresql/base.py
dxbjavid: same preparer method here now, no post-processing of the format_sequence result.

dxbjavid:
reworked along the lines you both suggested and pushed.

the escaping is now a single preparer method on PGIdentifierPreparer rather than a replace() at each call site, and get_insert_default builds a small _CompilerSequence shim so it can go through format_sequence like visit_sequence and fire_sequence do. the shim opts out of the schema translate map, since the execution context has already resolved the schema by that point. replies with the detail are on the individual threads.

one loose end: i've left the changelog under unreleased_20 for now. the escaping part looks like a plain bug fix to me, but the get_insert_default path does change rendering slightly for ordinary names, so happy to move it to unreleased_21 if you'd rather keep 2.0 untouched.

test/dialect/postgresql/test_compiler.py
```diff
@@ -138,6 +139,35 @@ def test_nextval_quote_escaping(self):
...
+                _CompilerSequence(name, schema_)
+            ),
+            expected,
+        )
+
+    def test_compiler_sequence_ignores_schema_translate(self):
+        # the schema is resolved by the execution context before it reaches
+        # _CompilerSequence, so it must not be translated a second time
+        preparer = postgresql.dialect().identifier_preparer
+        preparer = preparer._with_schema_translate({"foo": "bar"})
```
CaselIT: invert the mapping (bar->foo) otherwise I don't think this test tests much

CaselIT (commented):
Thanks, I think it's fine now, just a suggestion on a test

test/dialect/postgresql/test_compiler.py
dxbjavid: good point, the old mapping never hit the schema so it would have passed either way. flipped it to bar->foo and pushed, so the test now fails if the shim opts back into the map (it renders __[SCHEMA_bar] instead of bar).

CaselIT (approved):

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 0c19bcbd85c69b492542c43a693600595cbd0916 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 0c19bcbd85c69b492542c43a693600595cbd0916: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6838

doc/build/changelog/unreleased_20/13429.rst
```diff
@@ -0,0 +1,17 @@
...
+    :tags: bug, postgresql
+    :tickets: 13429
+
+    Fixed quoting in the PostgreSQL ``nextval()`` rendering used for sequence
+    default expressions and for server-side autoincrement pre-execution, so
+    that a single quote occurring in a sequence, table or schema name is
+    escaped rather than terminating the surrounding string literal.  In
+    particular :meth:`_engine.Connection.execution_options` with a
+    ``schema_translate_map`` supplying such a schema name, or an explicit
+    :class:`.Sequence` whose name carries a quote, previously produced a
```
sqla-tester, line 11: **Michael Bayer** (zzzeek) wrote:

im going to reduce the verbosity here,the LLMs do a big essay about how they made the change and that's not relevant here

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6838

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6838

* /COMMIT_MSG (line 18): clean up the form here, just the changelog message is fine
* lib/sqlalchemy/dialects/postgresql/base.py (line 3273): looks like use_schema is always true
* lib/sqlalchemy/dialects/postgresql/base.py (line 3273): probably this should be underscored for now

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6838 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6845 has been **merged**. Congratulations! :)

PR #13434, opened 2026-07-13
Fix bare except clauses in util/langhelpers.py

koteshyelamati:
Replace bare `except:` with `except Exception:` in `lib/sqlalchemy/util/langhelpers.py`.

Two locations fixed:
1. `sys._getframe()` fallback: the bare `except:` after `except ValueError:` handles a "weird interpreter issue" where `_getframe` doesn't work. Changed to `except Exception:` to avoid masking `SystemExit`/`KeyboardInterrupt`.
2. `once_fn` retry logic: bare `except:` re-raises after optionally reinserting the function. Changed to `except Exception:` for the same reason.

### DIFF
--- a/lib/sqlalchemy/util/langhelpers.py
+++ b/lib/sqlalchemy/util/langhelpers.py
@@ -1937,7 +1937,7 @@ def _warnings_warn(
         # being called from less than 3 (or given) stacklevels, weird,
         # but don't crash
         stacklevel = 0
-    except:
+    except Exception:
         # _getframe() doesn't work, weird interpreter issue, weird,
         # ok, but don't crash
         stacklevel = 0
@@ -1992,7 +1992,7 @@ def only_once(
             once_fn = once.pop()
             try:
                 return once_fn(*arg, **kw)
-            except:
+            except Exception:
                 if retry_on_exception:
                     once.insert(0, once_fn)
                 raise


lib/sqlalchemy/util/langhelpers.py
```diff
@@ -1992,7 +1992,7 @@ def go(*arg: Any, **kw: Any) -> Optional[_T]:
             once_fn = once.pop()
             try:
                 return once_fn(*arg, **kw)
-            except:
+            except Exception:
```
zzzeek, line 1995: this breaks the state of the decorator.  we've already popped out `once_fn`.  we absolutely definitely need to put once_fn back in the list if there is any exception and that is for BaseExceptions also.     more correct might be that `once.insert(0, once_fn)` is called **unconditionally** for a BaseException, compared to an Exception that would continue testing the retry_on_exception flag; since for a BaseException the once_fn() didn't really "raise"

the reason we need to maintain program state even if BaseException is that  BaseExceptions do not automatically imply program exit (e.g. CancelledError)

i.e.:

```py
once_fn = once.pop()
try:
    return once_fn(*arg, **kw)
except Exception:
    if retry_on_exception:
        once.insert(0, once_fn)
    raise
except BaseException:
    # fn() was interrupted, didn't really "raise"; restore unconditionally
    once.insert(0, once_fn)
    raise
```

the model came up with an additional spot that resembles this same thing, which we could potentially change as well here:

> Worth noting too: the same principle implicates the sibling _exec_once_impl in lib/sqlalchemy/event/attr.py:453-461, which sets self._exec_once = True in finally even on a BaseException — by your reasoning that marks the event "executed" when it was actually interrupted. Out of scope for this PR, but it's the same bug shape.

lib/sqlalchemy/util/langhelpers.py
```diff
@@ -1992,7 +1992,7 @@ def go(*arg: Any, **kw: Any) -> Optional[_T]:
...
                 if retry_on_exception:
                     once.insert(0, once_fn)
                 raise
```
zzzeek, line 1998: it's re-raised right here and this is not optional, so this code does not swallow KeyboardInterrupt

lib/sqlalchemy/util/langhelpers.py
```diff
@@ -1937,7 +1937,7 @@ def _warnings_warn(
         # being called from less than 3 (or given) stacklevels, weird,
         # but don't crash
         stacklevel = 0
-    except:
+    except Exception:
```
zzzeek, line 1940: this is fine, if only because _getframe() can't really raise KeyboardInterrupt /SystemExit / GeneratorExit etc. anyway

zzzeek (changes requested):
did manual review and also ran through a model for extra checks

the second one is incorrect.  it's best to just revert the second one but I also proposed a strategy to potentially make it more "correct" though it's likely not worth the complexity.

CaselIT:
I'm -1 on this. what are why trying to fix?

zzzeek:
agree there's not really anything being fixed here

PR #13437, opened 2026-07-16
fix: escape single quotes in PostgreSQL nextval() rendering

### LINKED ISSUE #13429: postgresql nextval() rendering does not escape single quotes in sequence/schema names

mangodxd:
## What does this PR do?

Fixes SQL injection vulnerability in PostgreSQL 
extval() rendering where single quotes in sequence/schema names are not escaped, allowing an attacker to break out of the SQL string literal.

## What issues does this PR fix?

Fixes #13429

## Problem

The PostgreSQL dialect renders 
extval('...') in three places by interpolating sequence identifiers into single-quoted SQL string literals without escaping single quotes:

- PGCompiler.visit_sequence - renders the 
extval() default expression
- PGExecutionContext.fire_sequence - executes the sequence
- PGExecutionContext.get_insert_default - executes the sequence for SERIAL primary keys

A single quote in a sequence or schema name breaks out of the literal, allowing SQL injection. This is especially dangerous in get_insert_default where effective_schema comes from schema_translate_map, which is often derived from request context in multi-tenant setups.

## Fix

Escape single quotes by replacing ' with '' in the rendered identifier before embedding in the literal, matching PostgreSQL's standard string literal escaping.

## Test

Added 	est_visit_sequence_escapes_single_quotes to verify:
1. Single quote in sequence name is properly escaped
2. Complex SQL injection attempts are neutralized
3. Normal sequences (without single quotes) still work

All 313 existing PostgreSQL compiler tests continue to pass.

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -2371,7 +2371,9 @@ class PGCompiler(compiler.SQLCompiler):
         return f"power{self.function_argspec(fn)}"
 
     def visit_sequence(self, seq, **kw):
-        return "nextval('%s')" % self.preparer.format_sequence(seq)
+        return "nextval('%s')" % self.preparer.format_sequence(seq).replace(
+            "'", "''"
+        )
 
     def limit_clause(self, select, **kw):
         text = ""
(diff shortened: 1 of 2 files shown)


CaselIT:
hi, there is already #13430 for this, closing

PR #13440, opened 2026-07-18
Escape single quotes in PostgreSQL nextval() string literals

### LINKED ISSUE #13429: postgresql nextval() rendering does not escape single quotes in sequence/schema names

Devil1716:
﻿## Summary
- Escape `'` as `''` when embedding `format_sequence()` output in `nextval('...')` (`visit_sequence`, `fire_sequence`).
- Route `get_insert_default` through the identifier preparer the same way as `fire_sequence`.

## Test plan
- [x] `pytest test/dialect/postgresql/test_compiler.py::SequenceTest`
- [x] Compiling a Sequence name containing `'` no longer breaks out of the string literal

Fixes #13429


CaselIT:
there is already #13430 in progress. closing

Issue #13448, opened 2026-07-22
mssql Uuid literal renderer collapses quotes instead of escaping them; add suite for audit of all literal_execute renderers


dxbjavid:
the SQL Server implementation of `Uuid` overrides `literal_processor`, and in the native-uuid branch it does `str(value).replace("''", "'")` before wrapping the result in quotes. that replace is the wrong way round: it collapses a doubled quote into a single one rather than escaping a single quote into a doubled one. the generic `sqltypes.Uuid.literal_processor` does `replace("'", "''")`, so the two disagree.

with `as_uuid=False` the value handed to the processor is an ordinary string rather than a `UUID` object, so nothing else in the path removes a quote from it, and the rendered literal ends early.

```python
from sqlalchemy import Uuid, bindparam, column, select
from sqlalchemy.dialects import mssql

c = column("x", Uuid(as_uuid=False))
stmt = select(c).where(
    c == bindparam("u", type_=Uuid(as_uuid=False), literal_execute=True)
)

value = "z' OR 1=1--"

print(stmt.compile()._process_parameters_for_postcompile({"u": value}).statement)
print(
    stmt.compile(dialect=mssql.dialect())
    ._process_parameters_for_postcompile({"u": value})
    .statement
)
```

output:

```
SELECT x 
WHERE x = 'z'' OR 1=1'
SELECT x 
WHERE x = 'z' OR 1=1--'
```

the default dialect escapes and the mssql one does not, so on SQL Server the trailing text is parsed as SQL rather than as part of the value. the same difference shows up with `compile_kwargs={"literal_binds": True}`, but `literal_execute` is the case i care about since that one renders at execution time.

happy to put up a patch, it looks like just reversing the two arguments to `replace()` so the mssql renderer escapes the way the base type already does.

**SQLAlchemy version**: 2.0 / 2.1 (main)

zzzeek:
the use case here is invalid input combined with literal_execute preventing SQL injection, correct?

zzzeek:
because frankly I would want to approach this in a much more generalized way.   we dont have a test suite right now that exercises the literal_execute of all types with invalid data and that would be more of what we need here.  the trivial fixes would flow from that.

zzzeek:
here's that suite.   in general, there's around 22 types we need to make a similar fix.  however if you run this suite the way i have it at the moment there's >250 fails because i have it running every type with every valid dialect impl for it.   also some of these combinations aren't working yet, i need to work on this further

[...]

PR #13449, opened 2026-07-22
escape quotes in mssql uuid literal renderer

dxbjavid:
### Description

`MSUUid.literal_processor` renders the native `UNIQUEIDENTIFIER` form with `replace("''", "'")`, which collapses doubled quotes rather than escaping single ones, so it does the opposite of the generic `sqltypes.Uuid.literal_processor` it overrides. With `as_uuid=False` the value is a plain string rather than a `UUID` object, so a quote inside it ends the rendered literal and the rest is parsed as SQL, both under `literal_binds` and for parameters marked `literal_execute=True`, which renders at execution time. Reversing the two arguments to `replace()` brings it back in line with the base type; a full example and the before/after output are in #13448.

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/doc/build/changelog/unreleased_20/13448.rst
+++ b/doc/build/changelog/unreleased_20/13448.rst
@@ -1,0 +1,14 @@
+.. change::
+    :tags: bug, mssql
+    :tickets: 13448
+
+    Fixed quoting bug in the SQL Server implementation of the
+    :class:`_sqltypes.Uuid` datatype, where the literal renderer used for the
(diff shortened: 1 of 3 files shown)


zzzeek:
ive created a general purpose auditing suite to catch all of these at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6843 .  this fixes the UUID issue as well as NumericCommon and JSONPATH

PR #13457, opened 2026-07-26
Add support for multiple on-conflict clauses in inserts on SQLite

### LINKED ISSUE #13113: SQLite insert supports multiple on-conflict clauses in inserts
### Describe the use case

Apparently sqlite supports multiple on-conflict clauses in inserts: https://sqlite.org/lang_upsert.html

> If the insert operation would cause the conflict target uniqueness constraint to fail, then the insert is omitted and the corresponding DO NOTHING or DO UPDATE operation is performed instead. The ON CONFLICT clauses are checked in the order specified. If the last ON CONFLICT clause omits the conflict target, then it will fire if any uniqueness constraint fails which is not captured by prior ON CONFLICT clauses.

> Only a single ON CONFLICT clause, specifically the first ON CONFLICT clause with a matching conflict target, may run for each row of the INSERT. When an ON CONFLICT clause fires, all subsequent ON CONFLICT clauses are bypassed for that one row. 

sqlite

### Example Use

```py
from sqlalchemy import Column, Integer, MetaData, Table
from sqlalchemy.dialects.sqlite import dialect, insert

t = Table(
    "test",
    MetaData(),
    Column("id", Integer, primary_key=True),
    Column("a", Integer),
    Column("b", Integer),
)

stmt = (
    insert(t)
    .values({"id": 1})
    .on_conflict_do_update(index_elements="x", set_={"id": 2})
    .on_conflict_do_update(index_elements="y", set_={"a": 3})
    .on_conflict_do_update(index_elements="z", set_={"b": 3})
)
```

### Additional context

The syntax extension feature makes supporting this very easy since it's mostly a question of updating the apply_syntax_extension_point call:

```diff
diff --git a/lib/sqlalchemy/dialects/sqlite/dml.py b/lib/sqlalchemy/dialects/sqlite/dml.py
index f877f0a12a..17238bd82d 100644
--- a/lib/sqlalchemy/dialects/sqlite/dml.py
+++ b/lib/sqlalchemy/dialects/sqlite/dml.py
@@ -110,7 +110,7 @@ class Insert(StandardInsert):
         },
     )

-    @_on_conflict_exclusive
+    # @_on_conflict_exclusive
     def on_conflict_do_update(
         self,
         index_elements: _OnConflictIndexElementsT = None,
@@ -222,7 +222,7 @@ class OnConflictClause(SyntaxExtension, ClauseElement):

     def apply_to_insert(self, insert_stmt: StandardInsert) -> None:
         insert_stmt.apply_syntax_extension_point(
-            self.append_replacing_same_type, "post_values"
+            lambda ex: [*ex, self], "post_values"
         )
```

and the above snippet would render in sqlite:
```sql
INSERT INTO test (id) VALUES (1)
ON CONFLICT (x) DO UPDATE SET id = ?
ON CONFLICT (y) DO UPDATE SET a = ?
ON CONFLICT (z) DO UPDATE SET b = ?
```

MasoNord:
Multiple on-conflict clauses support for SQLite

Fixes: #13113

### Description
Follows up to [#13113](https://github.com/sqlalchemy/sqlalchemy/issues/13113)
Add support for multiple on-conflict clauses in insert statements for SQLite database
As [CaselIT](https://github.com/CaselIT) proposed, this required a small change in `apply_to_insert()` function. I've replaced `self.append_replacing_same_type` to `lambda ex: [*ex, self]` and removed `@_on_conflict_exclusive` from `on_conflict_do_update()` and `on_conflict_do_nothing()` methods

This pull request is:
    - Tests included

**Have a nice day!**

### DIFF
--- a/doc/build/changelog/unreleased_21/13113.rst
+++ b/doc/build/changelog/unreleased_21/13113.rst
@@ -1,0 +1,7 @@
+.. change::
+    :tags: dml, sqlite, usecase
+    :tickets: 13113
+
+    Add support for multiple on-conflict clauses in inserts for sqlite database.
+    Now you can use two or more `on conflict do update` or `on conflict do nothing`
+    clauses in a single statement
--- a/doc/build/changelog/unreleased_21/13113.rst
+++ b/doc/build/changelog/unreleased_21/13113.rst
@@ -2,6 +2,7 @@
     :tags: dml, sqlite, usecase
     :tickets: 13113
 
-    Add support for multiple on-conflict clauses in inserts for sqlite database.
+    Add support for multiple on-conflict clauses in inserts for sqlite database
+    :func:`sqlalchemy.dialects.sqlite.insert`.
     Now you can use two or more `on conflict do update` or `on conflict do nothing`
-    clauses in a single statement
+    clauses in a single statement. Pull request courtesy Diemid Berozkin
--- a/lib/sqlalchemy/dialects/sqlite/dml.py
+++ b/lib/sqlalchemy/dialects/sqlite/dml.py
@@ -110,7 +110,6 @@ class Insert(StandardInsert):
         },
     )
 
-    @_on_conflict_exclusive
     def on_conflict_do_update(
         self,
         index_elements: _OnConflictIndexElementsT = None,
@@ -161,7 +160,6 @@ class Insert(StandardInsert):
             OnConflictDoUpdate(index_elements, index_where, set_, where)
         )
 
-    @_on_conflict_exclusive
     def on_conflict_do_nothing(
         self,
         index_elements: _OnConflictIndexElementsT = None,
@@ -222,7 +220,7 @@ class OnConflictClause(SyntaxExtension, ClauseElement):
 
     def apply_to_insert(self, insert_stmt: StandardInsert) -> None:
         insert_stmt.apply_syntax_extension_point(
-            self.append_replacing_same_type, "post_values"
+            lambda ex: [*ex, self], "post_values"
         )
 
 
--- a/lib/sqlalchemy/dialects/sqlite/dml.py
+++ b/lib/sqlalchemy/dialects/sqlite/dml.py
@@ -21,7 +21,6 @@ from ...sql import coercions
 from ...sql import roles
 from ...sql import schema
 from ...sql._typing import _DMLTableArgument
-from ...sql.base import _exclusive_against
 from ...sql.base import ColumnCollection
 from ...sql.base import ReadOnlyColumnCollection
 from ...sql.base import SyntaxExtension
@@ -102,14 +101,6 @@ class Insert(StandardInsert):
         """
         return alias(self.table, name="excluded").columns
 
-    _on_conflict_exclusive = _exclusive_against(
-        "_post_values_clause",
-        msgs={
-            "_post_values_clause": "This Insert construct already has "
-            "an ON CONFLICT clause established"
-        },
-    )
-
     def on_conflict_do_update(
         self,
         index_elements: _OnConflictIndexElementsT = None,
@@ -119,6 +110,8 @@ class Insert(StandardInsert):
     ) -> Self:
         r"""
         Specifies a DO UPDATE SET action for ON CONFLICT clause.
+
+        Multiple clauses can be added, and they are evaluated in order
 
         :param index_elements:
          A sequence consisting of string column names, :class:`_schema.Column`
@@ -167,6 +160,10 @@ class Insert(StandardInsert):
     ) -> Self:
         """
         Specifies a DO NOTHING action for ON CONFLICT clause.
+
+        This clause may be used together with :func:`on_conflict_do_update()`
+        but only one :func:`on_conflict_do_nothing()` may be specified,
+        and it must be the last one
 
         :param index_elements:
          A sequence consisting of string column names, :class:`_schema.Column`
--- a/test/dialect/sqlite/test_on_conflict.py
+++ b/test/dialect/sqlite/test_on_conflict.py
@@ -81,26 +81,6 @@ class OnConflictTest(fixtures.TablesTest):
         with expect_raises(ValueError):
             insert(self.tables.users).on_conflict_do_update()
 
-    def test_on_conflict_do_no_call_twice(self):
-        users = self.tables.users
-
-        for stmt in (
-            insert(users).on_conflict_do_nothing(),
-            insert(users).on_conflict_do_update(
-                index_elements=[users.c.id], set_=dict(name="foo")
-            ),
-        ):
-            for meth in (
-                stmt.on_conflict_do_nothing,
-                stmt.on_conflict_do_update,
-            ):
-                with testing.expect_raises_message(
-                    exc.InvalidRequestError,
-                    "This Insert construct already has an "
-                    "ON CONFLICT clause established",
-                ):
-                    meth()
-
     def test_on_conflict_do_nothing(self, connection):
         users = self.tables.users
 
@@ -771,4 +751,87 @@ class OnConflictTest(fixtures.TablesTest):
                 sql.select(t.c.name, t.c.data).order_by(t.c.name)
             ).fetchall(),
             expected_updated,
+        )
+
+    def test_on_conflict_do_update_multiple_clauses(self, connection):
+        users = self.tables.users_xtra
+
+        self._exotic_targets_fixture(connection)
+
+        i = (
+            insert(users)
+            .on_conflict_do_update(
+                index_elements=["login_email"],
+                set_=dict(login_email="nord1@gmail.com"),
+            )
+            .on_conflict_do_update(
+                index_elements=[users.c.id],
+                set_=dict(name="name3"),
+            )
+            .values()
+        )
+
+        connection.execute(
+            i,
+            [
+                dict(
+                    id=1,
+                    name="name1",
+                    login_email="name1@gmail.com",
+                    lets_index_this="not",
+                ),
+                dict(
+                    id=2,
+                    name="name2",
+                    login_email="name3@gmail.com",
+                    lets_index_this="not",
+                ),
+            ],
+        )
+        eq_(
+            connection.execute(users.select()).fetchall(),
+            [
+                (1, "name1", "nord1@gmail.com", "not"),
+                (2, "name3", "name2@gmail.com", "not"),
+            ],
+        )
+
+    def test_on_conflict_do_nothing_multiple_clauses(self, connection):
+        users = self.tables.users_xtra
+
+        self._exotic_targets_fixture(connection)
+
+        i = (
+            insert(users)
+            .on_conflict_do_update(
+                index_elements=["login_email"],
+                set_=dict(login_email="nord1@gmail.com"),
+            )
+            .on_conflict_do_nothing(index_elements=[users.c.id])
+            .values()
+        )
+
+        connection.execute(
+            i,
+            [
+                dict(
+                    id=1,
+                    name="name1",
+                    login_email="name1@gmail.com",
+                    lets_index_this="not",
+                ),
+                dict(
+                    id=2,
+                    name="name2",
+                    login_email="name3@gmail.com",
+                    lets_index_this="not",
+                ),
+            ],
+        )
+        eq_(
+            connection.execute(users.select()).fetchall(),
+            [
+                (1, "name1", "nord1@gmail.com", "not"),
+                (2, "name2", "name2@gmail.com", "not"),
+            ],
         )


doc/build/changelog/unreleased_21/13113.rst
```diff
@@ -0,0 +1,7 @@
+.. change::
+    :tags: dml, sqlite, usecase
+    :tickets: 13113
+
+    Add support for multiple on-conflict clauses in inserts for sqlite database.
+    Now you can use two or more `on conflict do update` or `on conflict do nothing`
+    clauses in a single statement
```
CaselIT: let's also link to the sqlite insert here, using
```
:func:`sqlalchemy.dialects.sqlite.insert`
```

Also feel free to add yourself as author using something like "Pull request courtesy <your name>"

lib/sqlalchemy/dialects/sqlite/dml.py
CaselIT: we can remove this

lib/sqlalchemy/dialects/sqlite/dml.py
```diff
@@ -161,7 +160,6 @@ def on_conflict_do_update(
             OnConflictDoUpdate(index_elements, index_where, set_, where)
         )
 
-    @_on_conflict_exclusive
     def on_conflict_do_nothing(
```
CaselIT, line 156: ok, while we removed `_on_conflict_exclusive` let's add some documentation:

- in on_conflict_do_update mention that multiple clauses can be added and they are evaluated in order
- in on_conflict_do_nothing let's mention that it can be combined with on_conflict_do_update but only one on_conflict_do_nothing may be specified and it must be the last one


I've tried this on sqlite 3.51.2 and is seems that this is fine:
```sql
insert into t(id) values (1) on conflict (id) do update set id = 42 on conflict do nothing;
```
these are not:
```sql
insert into t(id) values (1) on conflict do nothing on conflict do nothing;
Parse error: near "on": syntax error
  sert into t(id) values (1) on conflict do nothing on conflict do nothing;
                                      error here ---^


insert into t(id) values (1) on conflict do nothing on conflict (id) do update set id = 42;
Parse error: near "on": syntax error
  sert into t(id) values (1) on conflict do nothing on conflict (id) do update s
                                      error here ---^
```

CaselIT (commented):
Looks mostly fine, left a couple of minor suggestion. Thanks!

@zzzeek I'm not sure if you want to add an explicit guard, I think a doc note is fine, if an user chooses to ignore it sqlite will complain but I think that's fine

MasoNord:
Thank you for your review. I'll update the changelog, remove `_on_conflict_exclusive`, and update the documentation to explain how to use `on_conflict_do_update` and `on_conflict_do_nothing`, as you suggested.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 356885a839850344cfdc0818b1d6bb64ee61c417 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 356885a839850344cfdc0818b1d6bb64ee61c417: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6861

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 356885a839850344cfdc0818b1d6bb64ee61c417 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 356885a839850344cfdc0818b1d6bb64ee61c417 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6861

doc/build/changelog/unreleased_21/13113.rst
```diff
@@ -0,0 +1,8 @@
+.. change::
+    :tags: dml, sqlite, usecase
```
sqla-tester, line 2: **AI Robot 1** (ai-robot-1) wrote:

Tag order convention in this project puts the change type first, then the category -- e.g. "bug, general", "usecase, postgresql", "feature, reflection, performance" elsewhere in unreleased_21. Should be `:tags: usecase, sqlite, dml`.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6861

doc/build/changelog/unreleased_21/13113.rst
```diff
@@ -0,0 +1,8 @@
...
+    :tickets: 13113
+
+    Add support for multiple on-conflict clauses in inserts for sqlite database
+    :func:`sqlalchemy.dialects.sqlite.insert`.
+    Now you can use two or more `on conflict do update` or `on conflict do nothing`
```
sqla-tester, line 7: **AI Robot 1** (ai-robot-1) wrote:

Two style points for changelog/non-tutorial docs. "Now you can use" -- these are written in passive voice without addressing the reader as "you". And single backticks in RST render as a *title reference*, not as literal text; ``on conflict do update`` needs double backticks.

Patch set 3 rewrites the entry, also spelling out the targetless-clause-must-be-last rule and linking the new documentation section.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6861

lib/sqlalchemy/dialects/sqlite/dml.py
```diff
@@ -121,6 +111,8 @@ def on_conflict_do_update(
         r"""
         Specifies a DO UPDATE SET action for ON CONFLICT clause.
 
+        Multiple clauses can be added, and they are evaluated in order
```
sqla-tester, line 114: **AI Robot 1** (ai-robot-1) wrote:

Nit: sentence has no terminating period, and since this is a new capability for methods that date back to 1.4 it should carry a `.. versionadded:: 2.1` marker plus a `.. seealso::` pointing at the narrative docs section.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6861

lib/sqlalchemy/dialects/sqlite/dml.py
```diff
@@ -170,6 +161,10 @@ def on_conflict_do_nothing(
         """
         Specifies a DO NOTHING action for ON CONFLICT clause.
 
+        This clause may be used together with :func:`on_conflict_do_update()`
```
sqla-tester, line 164: **AI Robot 1** (ai-robot-1) wrote:

This states a rule that SQLite does not actually impose. Both halves of "only one on_conflict_do_nothing() may be specified, and it must be the last one" are false:

    from sqlalchemy import Column, Integer, MetaData, String, Table, UniqueConstraint, create_engine
    from sqlalchemy.dialects.sqlite import insert

    m = MetaData()
    t = Table(
        "t", m,
        Column("id", Integer, primary_key=True),
        Column("name", String(50)),
        Column("email", String(50)),
        UniqueConstraint("email", name="uq_email"),
    )

    e = create_engine("sqlite://")
    with e.begin() as conn:
        m.create_all(conn)
        conn.execute(t.insert(), dict(id=1, name="a", email="e1"))

        # two DO NOTHING clauses: accepted
        conn.execute(
            insert(t)
            .on_conflict_do_nothing(index_elements=["email"])
            .on_conflict_do_nothing(index_elements=["id"]),
            dict(id=1, name="b", email="e2"),
        )

        # DO NOTHING not last: accepted
        conn.execute(
            insert(t)
            .on_conflict_do_nothing(index_elements=["email"])
            .on_conflict_do_update(index_elements=["id"], set_=dict(name="z")),
            dict(id=1, name="b", email="e2"),
        )

The actual constraint from https://sqlite.org/lang_upsert.html is about the *conflict target*, not about DO NOTHING: only the last ON CONFLICT clause may omit its conflict target. Since on_conflict_do_update() always renders a target, that reduces in practice to "a targetless on_conflict_do_nothing() must come last".

Also, `:func:` is the wrong Sphinx role for a method and the unqualified name will not resolve -- these need to be `:meth:`_sqlite.Insert.on_conflict_do_nothing`` etc.

Rewritten in patch set 3 to describe the real rule.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6861

sqla-tester (commented):
**AI Robot 1** (ai-robot-1) wrote:

File-level, since it concerns a file this patch does not touch: the `.. _sqlite_on_conflict_insert:` section in the lib/sqlalchemy/dialects/sqlite/base.py module docstring is the user-facing documentation for this feature -- it is what the on_conflict_do_update()/on_conflict_do_nothing() docstrings and the ORM DML query guide link to -- and it says nothing about multiple clauses, so the feature is effectively undiscoverable outside the changelog.

Patch set 3 adds a "Specifying Multiple ON CONFLICT Clauses" subsection there, with pycon+sql examples and the two semantics users will trip over: SQLite checks the clauses in the order given and applies at most the first one whose conflict target matches, and only the last clause may omit its target.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6861

* test/dialect/sqlite/test_on_conflict.py (line 80): test_on_conflict_do_no_call_twice was removed here with nothing put in its place, so there is no longer any construction-level coverage of what happens when the methods are called repeatedly. Patch set 3 adds a test asserting that a targetless on_conflict_do_nothing() followed by any further clause raises InvalidRequestError (see the comment on dml.py).

lib/sqlalchemy/dialects/sqlite/dml.py
```diff
@@ -222,7 +217,7 @@ def __init__(
 
     def apply_to_insert(self, insert_stmt: StandardInsert) -> None:
         insert_stmt.apply_syntax_extension_point(
-            self.append_replacing_same_type, "post_values"
+            lambda ex: [*ex, self], "post_values"
```
sqla-tester, line 220: **AI Robot 1** (ai-robot-1) wrote:

Dropping the exclusivity check means a clause that omits its conflict target can now land in a non-final position, where it silently compiles to SQL that SQLite rejects. Previously this raised a clear InvalidRequestError; now:

    >>> stmt = insert(t).on_conflict_do_nothing().on_conflict_do_update(
    ...     index_elements=["id"], set_=dict(name="y")
    ... )
    >>> print(stmt)
    INSERT INTO t (id, name, email) VALUES (?, ?, ?) ON CONFLICT DO NOTHING ON CONFLICT (id) DO UPDATE SET name = ?

    # ... and at execution time:
    (sqlite3.OperationalError) near "ON": syntax error

Trading a good error message for a raw DBAPI syntax error is a real usability regression, and the ordering is fully known at construction time, so the check is cheap to keep. Patch set 3 replaces the lambda with a named method on OnConflictClause that raises InvalidRequestError when the preceding clause has `inferred_target_elements is None`, i.e. only the targetless-clause-must-be-last rule is enforced; everything else this change enables stays permitted.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6861

test/dialect/sqlite/test_on_conflict.py
```diff
@@ -772,3 +752,86 @@ def test_on_conflict_do_update_bindparam(
             ).fetchall(),
             expected_updated,
         )
+
+    def test_on_conflict_do_update_multiple_clauses(self, connection):
```
sqla-tester, line 756: **AI Robot 1** (ai-robot-1) wrote:

Two gaps in what these tests pin down.

First, there is no compile-level assertion anywhere -- both new tests are execution round-trips, so nothing locks in the rendered string, the clause order, or the space separator between clauses. Patch set 3 adds an OnConflictCompileTest class with assert_compile tests covering three chained clauses and a trailing targetless DO NOTHING.

Second, order-dependence is the actual feature, and it is not asserted. Row id=1 here conflicts on both `id` and `login_email`, so swapping the two clauses should produce a different result -- with the login_email clause first, name stays "name1" and login_email is rewritten; with the id clause first, name becomes "name3" and login_email is left alone. Patch set 3 turns this into a testing.variation over both orderings so the "first matching clause wins" rule is what is being tested.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6861

test/dialect/sqlite/test_on_conflict.py
```diff
@@ -772,3 +752,86 @@ def test_on_conflict_do_update_bindparam(
...
+            insert(users)
+            .on_conflict_do_update(
+                index_elements=["login_email"],
+                set_=dict(login_email="nord1@gmail.com"),
+            )
+            .on_conflict_do_update(
+                index_elements=[users.c.id],
+                set_=dict(name="name3"),
+            )
+            .values()
```
sqla-tester, line 771: **AI Robot 1** (ai-robot-1) wrote:

`.values()` with no arguments is a no-op here (the rows come from the execute() parameter list) -- removed in patch set 3. Same at line 811.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6861

test/dialect/sqlite/test_on_conflict.py
```diff
@@ -772,3 +752,86 @@ def test_on_conflict_do_update_bindparam(
...
+        )
+        eq_(
+            connection.execute(users.select()).fetchall(),
+            [
+                (1, "name1", "nord1@gmail.com", "not"),
+                (2, "name3", "name2@gmail.com", "not"),
+            ],
+        )
+
+    def test_on_conflict_do_nothing_multiple_clauses(self, connection):
```
sqla-tester, line 799: **AI Robot 1** (ai-robot-1) wrote:

The name says "do_nothing_multiple_clauses" but the statement has one DO UPDATE and one DO NOTHING, so it reads as coverage that does not exist. Renamed to test_on_conflict_do_update_and_do_nothing in patch set 3, alongside a new test_on_conflict_untargeted_do_nothing_last that executes the targetless-trailing-DO-NOTHING form end to end.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6861

sqla-tester:

note I had claude 1. review 2. make those changes and 3. apply resolved comments that represent the review

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6861

zzzeek:
note we've made these changes on our gerrit server, no further action from the submitter is needed

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6861 has been **merged**. Congratulations! :)

PR #13461, opened 2026-07-28
escape backslashes in mysql enum/set value rendering

dxbjavid:
### Description

`MySQLTypeCompiler._visit_enumerated_values` renders ENUM and SET values by doubling single quotes but leaves backslashes untouched, so under the default sql_mode where a backslash is an escape character a value containing one is emitted wrongly, and a value ending in a backslash escapes the closing quote and breaks out of the string literal in the generated DDL. `render_literal_value` already doubles backslashes for ordinary string literals whenever the server is not in `NO_BACKSLASH_ESCAPES` mode, so this applies the same handling to enum and set values; ordinary values compile identically.

Before:

    >>> from sqlalchemy.dialects import mysql
    >>> print(mysql.ENUM("a", "b\\").compile(dialect=mysql.dialect()))
    ENUM('a','b\')

the trailing backslash swallows the closing quote. After the change it renders `ENUM('a','b\\')`, and the value is left as-is when the server has `NO_BACKSLASH_ESCAPES` set.

A compile-level regression test and a changelog entry are included.

### Checklist

This pull request is:

- [x] A short code fix
	- tests are included; a reproducible demonstration is shown above.

**Have a nice day!**

### DIFF
--- a/lib/sqlalchemy/dialects/mysql/base.py
+++ b/lib/sqlalchemy/dialects/mysql/base.py
@@ -2677,11 +2677,12 @@ class MySQLTypeCompiler(
     def _visit_enumerated_values(
         self, name: str, type_: _StringType, enumerated_values: Sequence[str]
     ) -> str:
+        preparer = self.dialect.identifier_preparer
         quoted_enums = []
         for e in enumerated_values:
-            if self.dialect.identifier_preparer._double_percents:
+            if preparer._double_percents:
                 e = e.replace("%", "%%")
-            quoted_enums.append("'%s'" % e.replace("'", "''"))
+            quoted_enums.append(preparer._format_string_literal(e))
         return self._extend_string(
             type_, {}, "%s(%s)" % (name, ",".join(quoted_enums))
         )
@@ -2718,6 +2719,21 @@ class MySQLIdentifierPreparer(
         """Unilaterally identifier-quote any number of strings."""
 
         return tuple([self.quote_identifier(i) for i in ids if i is not None])
+
+    def _format_string_literal(self, value: str) -> str:
+        """Render a plain string value as a SQL string literal.
+
+        Used for places such as ENUM and SET member values that are emitted
+        into DDL as string literals rather than identifiers. Single quotes are
+        doubled and, unless the server runs with ``NO_BACKSLASH_ESCAPES``,
+        backslashes are escaped as well, matching the handling in
+        :meth:`.MySQLCompiler.render_literal_value`.
+
+        """
+        value = value.replace("'", "''")
+        if self.dialect._backslash_escapes:
+            value = value.replace("\\", "\\\\")
+        return "'%s'" % value
 
 
 class MySQLDialect(_mariadb_shim.MariaDBShim, default.DefaultDialect):
--- a/lib/sqlalchemy/dialects/mysql/base.py
+++ b/lib/sqlalchemy/dialects/mysql/base.py
@@ -2678,11 +2678,10 @@ class MySQLTypeCompiler(
         self, name: str, type_: _StringType, enumerated_values: Sequence[str]
     ) -> str:
         preparer = self.dialect.identifier_preparer
-        quoted_enums = []
-        for e in enumerated_values:
-            if preparer._double_percents:
-                e = e.replace("%", "%%")
-            quoted_enums.append(preparer._format_string_literal(e))
+        quoted_enums = [
+            preparer._format_string_literal(e)  # type: ignore[attr-defined]
+            for e in enumerated_values
+        ]
         return self._extend_string(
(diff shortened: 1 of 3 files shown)


CaselIT:
Hi,

Thanks for looking into it!

I've given a quick glance but I think the solution should be the more similar what was done here bafba7cd3a4d83340463bc0662f6dfa68b801960, so use the identifier preparer for this.

Could you please open an issue also, since this is a bug even if improbable.

if you have time and/or motivation it may make sense to audit all usages for .replace when doing sql things and see if they make sense or if we actually should be using the preparer instead. (this would be another issue with the result of the audit so that we can decide what needs fixing and what is instead fine)

dxbjavid:
Makes sense. I've moved the escaping onto the preparer as _format_string_literal and routed the enum/set rendering through it, so _visit_enumerated_values no longer does the quote and backslash replaces inline. Ordinary values compile identically and it still leaves the backslash alone under NO_BACKSLASH_ESCAPES.

I've opened #13466 to track the bug, and pointed the changelog at it.

On the wider audit, I had a quick look and most of the other .replace calls in the dialects are either identifier quoting that already goes through the preparer or paramstyle percent doubling, but a few are worth a closer read. If it's useful I can do a proper pass and open a separate issue with the findings, as you suggested, rather than fold it into this one.

CaselIT:
> If it's useful I can do a proper pass and open a separate issue with the findings, as you suggested, rather than fold it into this one.

I think it would be, if you have the time. I would prefer a separate issue, so that we can track it better. thank you!

dxbjavid:
Done, opened #13469 with the audit. The main thing that stood out is the reflected DEFAULT rendering in the mysql _describe_to_create path, which has the same doubled-quotes-but-not-backslashes issue, though it only runs on the DESCRIBE fallback for view reflection. The rest I think are fine and I noted why in the issue. I left everything untouched there so the fix can be decided separately.

lib/sqlalchemy/dialects/mysql/base.py
```diff
@@ -2677,11 +2677,12 @@ def visit_LONGBLOB(self, type_: LONGBLOB, **kw: Any) -> str:
     def _visit_enumerated_values(
         self, name: str, type_: _StringType, enumerated_values: Sequence[str]
     ) -> str:
+        preparer = self.dialect.identifier_preparer
         quoted_enums = []
         for e in enumerated_values:
-            if self.dialect.identifier_preparer._double_percents:
+            if preparer._double_percents:
```
CaselIT: how about we move this too into format_string_literal? it seems like something that would apply to other usages of that method, no?

test/dialect/mysql/test_types.py
```diff
@@ -337,6 +337,25 @@ def test_bit_50(self, type_, expected):
 
         self.assert_compile(type_, expected)
 
+    @testing.combinations(
```
CaselIT, line 340: we could also try a _double_percents = True case, to verify that we actually do that too (since from a quick look we don't seem to have any test for it)

feel free to mock it in the preparer. could also be in another test, not necessarily in a combination here

CaselIT (commented):
thanks for the updated. left a couple of minor comments. 

also it seems to fail the type ci with
> lib/sqlalchemy/dialects/mysql/base.py:2685: error: "IdentifierPreparer" has no attribute "_format_string_literal"  [attr-defined]
lib/sqlalchemy/dialects/mysql/base.py:2734: error: "Dialect" has no attribute "_backslash_escapes"  [attr-defined]

feel free to just ignore theme using `# type: ignore[attr-defined]`

zzzeek:
this is stuck on "awaiting revision" for about a month.    I'm basically going through PRs that have no replies and closing them, but this one seems like @dxbjavid was making an effort, can we finish this up?

lib/sqlalchemy/dialects/mysql/base.py
dxbjavid: good point, moved it in. _format_string_literal now does the percent doubling too, keyed off the preparer's _double_percents, so the caller just formats the value and any other user of the method gets the same handling.

test/dialect/mysql/test_types.py
dxbjavid: added test_enum_set_double_percents, which mocks _double_percents on the preparer and checks both the true and false case, since as you said there wasn't any coverage for the percent path before.

dxbjavid:
Sorry for the long gap on this. I've pushed the revision now: the percent doubling moved into _format_string_literal so the caller just hands it the raw value, and there's a test covering both the double_percents on and off paths. For the type ci failures I added # type: ignore[attr-defined] on the two lines you pointed at, since both attributes only exist on the mysql subclasses. Tests pass locally.

PR #13464, opened 2026-07-28
Widen any_()/all_() typing to accept literal values

### LINKED ISSUE #13357: Type check error when passing lists to `any_`/`all_`

- [x] No sqlalchemy stub packages is installed (both `sqlalchemy-stubs` and `sqlalchemy2-stubs` are not compatible with v2)

- [x] The api is not in a module listed in [#6810](https://github.com/sqlalchemy/sqlalchemy/issues/6810) so it should pass type checking

### Describe the typing issue

the runtime SQL is good, but the parameter type of `any_` rejects lists.

### To Reproduce

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "sqlalchemy[postgresql]>=2.0.50",
# ]
# ///
from sqlalchemy import Integer, Text, any_, create_engine, select
from sqlalchemy.orm import DeclarativeBase, Mapped, Session, mapped_column


class Base(DeclarativeBase):
    pass


class Foo(Base):
    __tablename__ = "foo"

    id: Mapped[int] = mapped_column(Integer, primary_key=True)
    name: Mapped[str] = mapped_column(Text)


engine = create_engine("postgresql:///anyarrayissue", echo=True)

with engine.begin() as conn:
    Base.metadata.create_all(bind=conn)

    session = Session(bind=conn)
    session.add_all([Foo(name="Alice"), Foo(name="Bob"), Foo(name="Carol")])

    names = ["Alice", "Carol"]
    stmt = select(Foo.name).where(Foo.name == any_(names)).order_by(Foo.name)
    assert session.scalars(stmt).all() == names

    conn.rollback()
```

### Error

```
mcve_anyarrayissue.py:31: error: Argument 1 to "any_" has incompatible type "list[str]"; expected "ColumnElement[Never] | _HasClauseElement[Never] | SQLCoreOperations[Never] | ExpressionElementRole[Never] | TypedColumnsClauseRole[Never] | Callable[[], ColumnElement[Never]] | LambdaElement"  [arg-type]
``` 

### Versions

- OS: macOS 26
- Python: 3.12
- SQLAlchemy: 2.0.50
- Type checker (eg: mypy 0.991, pyright 1.1.290, etc): mypy 1.20.0

### Additional context

_No response_

Itachi-0xAI:
This pull request is:

- [x] A short code fix
	- includes issue number and tests
	- `Fixes: #13357` is in the commit message

### Description

`any_()` and `all_()` accept plain Python values at runtime (e.g. a
list), same as `between()` and a few other constructors in
`_elements_constructors.py` — they all go through
`coercions.expect(roles.ExpressionElementRole, ...)`, which handles
literal coercion. But `any_()`/`all_()` were typed with
`_ColumnExpressionArgument`, which doesn't include plain literals,
only `between()` and similar were typed with
`_ColumnExpressionOrLiteralArgument`. So mypy rejects
`any_(["a", "b"])` even though it runs fine and is a normal way to use
it against an `ARRAY` column.

Changed both signatures to `_ColumnExpressionOrLiteralArgument` to
match. Added a couple of typing-test lines covering the list case
alongside the existing column-expression one.

Fixes: #13357

### DIFF
--- a/lib/sqlalchemy/sql/_elements_constructors.py
+++ b/lib/sqlalchemy/sql/_elements_constructors.py
@@ -69,7 +69,9 @@ if typing.TYPE_CHECKING:
 _T = TypeVar("_T")
 
 
-def all_(expr: _ColumnExpressionArgument[_T]) -> CollectionAggregate[bool]:
+def all_(
+    expr: _ColumnExpressionOrLiteralArgument[_T],
+) -> CollectionAggregate[bool]:
     """Produce an ALL expression.
 
     For dialects such as that of PostgreSQL, this operator applies
@@ -252,7 +254,9 @@ if not TYPE_CHECKING:
         return BooleanClauseList.and_(*clauses)
 
 
-def any_(expr: _ColumnExpressionArgument[_T]) -> CollectionAggregate[bool]:
+def any_(
+    expr: _ColumnExpressionOrLiteralArgument[_T],
+) -> CollectionAggregate[bool]:
     """Produce an ANY expression.
 
     For dialects such as that of PostgreSQL, this operator applies
--- a/test/typing/plain_files/sql/operators.py
+++ b/test/typing/plain_files/sql/operators.py
@@ -4,6 +4,8 @@ from typing import Any
 from typing import assert_type
 from typing import List
 
+from sqlalchemy import all_ as all_fn
+from sqlalchemy import any_ as any_fn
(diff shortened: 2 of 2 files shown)


CaselIT:
Hi solution for #13357 was to document that it's python dbapi specifit, not to widen the type.
closing in favour of #13456

PR #13465, opened 2026-07-28
Add deprecation warnings for SQLite mode= pool-selection heuristics

### LINKED ISSUE #13433: Deprecate guess-the-pool in sqlite memory
IMO the bug is that we are looking in the query string at all, which leads us into guessing:

```
 @classmethod
    def _is_url_file_db(cls, url: URL) -> bool:
        if (url.database and url.database != ":memory:") and (
            url.query.get("mode", None) != "memory"
        ):
            return True
        else:
            return False
```

That code is **not even checking that URI mode is in use** when it interprets what "mode" means.

Im certain that long ago, that code only looked at url.database.   but now we're looking in at url.query, which officially means "we're guessing what they want", and I dont want to do that.  because now someone who wants to know what's happening needs to know all of our heuristics which are themselves in flux.   we could key this off of "cache=shared" being present but I dont want to be on the hook to know every possible new option SQLite is adding that impacts caching on/off/global etc. whatever.

I would propose the change is:

**When using URI mode, all pool switching logic is disabled, you get QueuePool/AsyncAdaptedQueuePool unconditionally unless you specified a pool class**

this would be new deprecation logic in 2.1.

for 2.1, there would be new logic there which:

1. continues to look at "mode" all by itself, because it already is and we need to emit deprecation warnings
2. with "mode" present, it then looks to see if "uri=True" is present.  if not, warns like crazy that you should not be using "mode" without URI mode present
3. with uri=True, looks to see if "mode=memory" is present, and "cache=shared" not present.  if so, warns that it's automatic choosing of pools is deprecated, in 2.2 this will be selecting [AsyncAdapted]QueuePool .  it should recommend using plain ":memory:" format instead

in 2.2 we remove the "url.query.get("mode")" thing up above.   we go back to only looking at "url.database == ':memory:'" for auto-choosing a singleton pool. it's a convenience for small test apps.  if you're using URI mode, you have to push the buttons yourself and know what you're doing (which you will because you're following our docs).

_Originally posted by @zzzeek in https://github.com/sqlalchemy/sqlalchemy/discussions/13428#discussioncomment-17623064_

Itachi-0xAI:
This pull request is:

- [x] A short code fix
	- includes issue number and tests
	- `Fixes: #13433` is in the commit message

### Description

`get_pool_class()` for the pysqlite and aiosqlite dialects picks the
pool class by inspecting the `mode` query string parameter, but as
noted in the issue, that code doesn't check whether URI mode
(`uri=true`) is actually in effect, and doesn't check that
`cache=shared` is present for `mode=memory` to actually mean a shared,
poolable in-memory database.

This implements the phase "2.1 (Current)" step from the issue: add
deprecation warnings for the two cases described there:

- `mode` present without `uri=true`
- `mode=memory` present without `cache=shared`

Pool selection itself is unchanged in this PR — only the warnings are
new. The follow-up behavior change (only honoring `mode` under
`uri=true`) is for a later release per the issue.

Added a changelog entry and a dedicated test
(`test_mode_query_param_deprecations`) covering all four combinations
(warn / warn / correct usage / unaffected file db), and updated the
existing `test_pool_class` test since one of its cases now triggers
one of these warnings.

Fixes: #13433

### DIFF
--- a/doc/build/changelog/unreleased_21/13433.rst
+++ b/doc/build/changelog/unreleased_21/13433.rst
@@ -1,0 +1,49 @@
+.. change::
+    :tags: deprecated, sqlite
+    :tickets: 13433
+
+    Deprecated the selection of a single-connection pool class, i.e.
+    :class:`.SingletonThreadPool` for pysqlite or :class:`.StaticPool` for
+    aiosqlite, based on the presence of the ``mode=memory`` query string
+    argument in a SQLite URL.  Pool selection for SQLite is intended to be
+    based on the database name alone, where only ``:memory:`` or an empty
+    database name indicate a memory database; interpreting the query string
+    additionally requires that assumptions be made regarding whether or not
+    the resulting database can be shared among multiple connections.  In a
+    future release, such URLs will make use of :class:`.QueuePool` or
+    :class:`.AsyncAdaptedQueuePool` as would any other URL.  This notably
+    includes the shared cache form
+    ``sqlite:///file:mydb?mode=memory&cache=shared&uri=true``, for which a
+    queue pool is in fact the appropriate class, as a shared cache database
+    supports multiple concurrent connections, whereas a single-connection
+    pool causes such connections to share one transaction state.
+    Applications that rely upon the present behavior should indicate the
+    intended pool using the :paramref:`_sa.create_engine.poolclass`
+    parameter.  Pull request courtesy Itachi-0xAI.
+
+.. change::
+    :tags: bug, sqlite
+    :tickets: 13433
+
+    Added a warning for query string arguments that are passed to a SQLite
+    URL without the ``uri=true`` argument also being present, and which are
+    not accepted by the ``sqlite3`` driver itself.  SQLite URI arguments
+    such as ``mode`` or ``cache`` take effect only when URI mode is in use;
+    without it they were previously discarded silently, so that a URL such
+    as ``sqlite:///file:mydb?mode=memory`` would connect to a file on disk
+    named ``file:mydb``.  Arguments intended for the driver itself may be
+    passed using the :paramref:`_sa.create_engine.connect_args` parameter.
+
+.. change::
+    :tags: bug, documentation, sqlite
+    :tickets: 13433
+
+    Corrected the SQLite documentation regarding shared cache memory
+    databases, which incorrectly indicated that the named form
+    ``sqlite:///file:mydb?mode=memory&cache=shared&uri=true`` makes use of
+    :class:`.QueuePool`; a single-connection pool is used for this form.
+    Documentation has also been added noting that a shared cache database
+    exists only for as long as at least one connection to it remains open,
+    so that ordinary pool operations such as
+    :meth:`_engine.Engine.dispose` or use of
+    :paramref:`_sa.create_engine.pool_recycle` will discard its contents.
--- a/doc/build/errors.rst
+++ b/doc/build/errors.rst
@@ -246,6 +246,95 @@ in the 2.x series of SQLAlchemy when any SQL statements are emitted.  When a con
 that was in progress is now in an invalid state, and must be explicitly rolled
 back in order to remove it from the :class:`_engine.Connection`.
 
+.. _error_sqmp:
+
+Selection of the <pool> pool class based on the 'mode=memory' query string argument is deprecated
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The SQLite dialects select a connection pool class automatically, based on
+whether the URL refers to a file database or to a memory database.  A memory
+database is indicated by the database name ``:memory:``, or by an empty
+database name, and receives a single-connection pool class, being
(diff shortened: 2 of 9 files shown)


sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision fc50a9837fd50334450ee6c9dd645db1b27b4dad of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change fc50a9837fd50334450ee6c9dd645db1b27b4dad: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6893

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6893

* lib/sqlalchemy/dialects/sqlite/pysqlite.py (line 662): wordy.  should we have a "code" link to the errors.rst page where we can do longer form explanation?
* lib/sqlalchemy/dialects/sqlite/pysqlite.py (line 819): same, also wordy here
* test/engine/test_parseconnect.py (line 1231): should we make a mock dialect for these tests so they arent corrupted by the sqlite behaviors?

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6893

* lib/sqlalchemy/dialects/sqlite/pysqlite.py (line 360): ```suggestion
    :paramref:`_sa.create_engine.poolclass` explicitly will no longer be
```
* lib/sqlalchemy/dialects/sqlite/pysqlite.py (line 380): ```suggestion
* :meth:`_engine.Engine.dispose`, which closes all connections currently in the pool
```
* lib/sqlalchemy/dialects/sqlite/pysqlite.py (line 662): I'm not sure if we do code links in the warnings.

but I agree it's wordy, I think we can remove from "To share one in-memory"

sqla-tester (commented):
**AI Robot 1** (ai-robot-1) wrote:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6893

* lib/sqlalchemy/dialects/sqlite/pysqlite.py (line 360): Applied, thanks.
* lib/sqlalchemy/dialects/sqlite/pysqlite.py (line 380): Applied, with the line wrapped across two lines. This text is part of the module level docstring in pysqlite.py, so it is subject to the 79 character limit enforced by flake8, and the suggested line is 83 characters. It now reads:

    * :meth:`_engine.Engine.dispose`, which closes all connections currently
      in the pool
* lib/sqlalchemy/dialects/sqlite/pysqlite.py (line 662): Error codes on warnings are supported and have existing precedent. HasDescriptionCode is the mixin that supplies the 'code' argument and renders the sqlalche.me link, and SAWarning sets `_what_are_we = "warning"` specifically so that warnings render as "Background on this warning at" rather than "...this error at". Existing examples that pass a code to util.warn():

- lib/sqlalchemy/sql/type_api.py, code="cprf", the cache_ok warning
- lib/sqlalchemy/orm/context.py, code="xaj1" and code="xaj2", the automatic aliasing warnings

The message is also trimmed further than suggested here. Rather than cutting only from "To share one in-memory", everything after the first two sentences is gone and the detail now lives in doc/build/errors.rst under the 'sqmp' code. The full message is now:

    Selection of the SingletonThreadPool pool class based on the
    'mode=memory' query string argument is deprecated; a future release
    will use QueuePool for this URL.  Indicate the intended pool class
    using the create_engine.poolclass parameter.
    (Background on this warning at: https://sqlalche.me/e/21/sqmp)
* lib/sqlalchemy/dialects/sqlite/pysqlite.py (line 819): Trimmed, with the detail moved to doc/build/errors.rst under the code 'squa'. The message is now:

    Query string argument(s) 'charset', 'nolock' are not accepted by the
    pysqlite driver and are being ignored; SQLite URI arguments require
    that 'uri=true' also be present in the URL.
    (Background on this warning at: https://sqlalche.me/e/21/squa)

The errors.rst entry carries what was removed, namely that the arguments are appended to the SQLite URI filename only when uri=true is present, a before and after example on the same URL, and a pointer to create_engine.connect_args for arguments meant for the driver itself.
* test/engine/test_parseconnect.py (line 1231): Done. test_plugin_url_registration, test_plugin_multiple_url_registration and test_plugin_arg_registration now register the MockDialect already defined at the bottom of this module and use 'mockdialect://' URLs, so no SQLite code runs at all.

These tests use a leftover 'foo=bar' query argument to show that a plugin does not strip arguments belonging to someone else, and that argument is what tripped the new pysqlite warning. With MockDialect the assertions are unaffected by dialect behavior, so the expect_warnings() blocks that patchset 2 needed are gone, along with @testing.requires.sqlite and the sqlalchemy.dialects.sqlite imports.

The assertions on the resolved dialect changed from sqlite.dialect to MockDialect accordingly, and eq_(e.dialect.name, "sqlite") was dropped as it no longer says anything about plugin handling.

zzzeek:
just a heads up @Itachi-0xAI we've taken this over in our own gerrit and will give you credit for the submission

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6893 has been **merged**. Congratulations! :)

Issue #9930, opened 2023-06-10
need docs /warnings that `__init__` is not called on load


zzzeek:
### Discussed in https://github.com/sqlalchemy/sqlalchemy/discussions/9928

since we retired https://docs.sqlalchemy.org/en/20/orm/constructors.html#mapping-constructors this information has been lost.  also update links to that page such as https://docs.sqlalchemy.org/en/20/orm/events.html#sqlalchemy.orm.InstanceEvents.load which are still pointing to a dead section.

this behavior sohuld be noted anywhere we are discussing `__init__` methods, including at https://docs.sqlalchemy.org/en/20/orm/mapping_styles.html#mapped-class-behavior and https://docs.sqlalchemy.org/en/20/orm/dataclasses.html#using-non-mapped-dataclass-fields

LeonxLJX:
I'd like to take this one.

Reading the issue, this looks like a documentation-clarity gap rather than a code defect: the current page doesn't give readers enough to self-diagnose, which is why it surfaces here as an issue instead of being answered by the docs.

Plan: locate the relevant section, confirm what the code actually does today (so the docs match behavior, not intent), tighten the wording, and scan neighboring pages for the same inconsistency — otherwise the fix just relocates the confusion.

One thing to confirm before writing: minimal correction scoped to this page, or also cross-link from related guides if the same gap shows up there?

(claiming via @LeonxLJX)

Issue #9493, opened 2023-03-15
MappedAsDataclass inherits dataclasses attributes, while the spec says they don't


CaselIT:
We also seem to make inherit dataclasses attributes, while we probably should not. Not sure if it's connected to this other issue #9350

```py
from dataclasses import dataclass
import sqlalchemy as sa
import sqlalchemy.orm as sao
import inspect

class Base(sao.MappedAsDataclass, sao.DeclarativeBase, kw_only=True):
    pass

class X(Base):
    __tablename__ = "x"
    id: sao.Mapped[int] = sao.mapped_column(primary_key=True)

print(inspect.getfullargspec(X.__init__))

@dataclass(kw_only=True)
class DCB:
    pass

@dataclass()
class DCX(DCB):
    id: int

print(inspect.getfullargspec(DCX.__init__))
```
prints
```
FullArgSpec(args=['self'], varargs=None, varkw=None, defaults=None, kwonlyargs=['id'], kwonlydefaults=None, annotations={})
FullArgSpec(args=['self', 'id'], varargs=None, varkw=None, defaults=None, kwonlyargs=[], kwonlydefaults=None, annotations={'return': None, 'id': <class 'int'>})
```

kwonly is inherited when using sqlalchemy, while it's not when using dataclasses (also annotations seem to be missing for some reason)

The type checkers seem to agree with dataclass behaviour since pyright/pylance think it's not inherited:
![image](https://user-images.githubusercontent.com/16175304/225416675-45b02b3b-1733-4e12-a79a-4bbe1f423fb6.png)

changing it like so
```py
class X(Base, kw_only=True):
    __tablename__ = "x"
    id: sao.Mapped[int] = sao.mapped_column(primary_key=True)
```
understands that's kw-only
![image](https://user-images.githubusercontent.com/16175304/225416829-76b5ab3b-3f1b-4d21-be27-a0039356e4de.png)

_Originally posted by @CaselIT in https://github.com/sqlalchemy/sqlalchemy/issues/9410#issuecomment-1470618377_

zzzeek:
we have a lot to go out for 2.0.8 already so not sure if i want to do this yet, but it's another thing that likely has to warn somehow, not really sure

barsikus007:
any updates on this?

zzzeek:
the latest update would be that someone named @barsikus007 posted a comment asking for updates.    Since this page is where you in fact would see any and all updates!

Issue #9410, opened 2023-03-02
document dataclass gotchas so far


zzzeek:
1. relationship = None issue :  #9383 , #9856
   - issue created: https://github.com/sqlalchemy/sqlalchemy/issues/12168
2. declared_attr on a mixin w/ dataclasses:   #9408
3. order matters regarding the default value and init generated. this all comes from dataclasses, but it may be worth reiterate that 
4. mention that all the dataclass limitations apply. see #9563 also
5. #9785 , back_populates / recursion issue 
6. when using mixins, it is likely a good idea to use kw_only on those fields so that they are not considered in field ordering, see #10105
1 and 2 have potential new features we might add but for now they also both have workarounds.   
7. using session.merge with default and/or default_factory may have unwanted effect, see [this comment](https://github.com/sqlalchemy/sqlalchemy/issues/9410#issuecomment-2989758246)

CaselIT:
> 1. relationship = None issue :

this would be worked around by the NoArg idea we were talking about at the dev meeting, right?

zzzeek:
in theory.  still mulling over other ways to deal w it

CaselIT:
in any case it makes sense to document current behaviour. if and when we have a work around we can update to say "to work around this limitation ..."

CaselIT:
We also seem to make inherit dataclasses attributes, while we probably should not. Not sure if it's connected to this other issue #9350

```py
from dataclasses import dataclass
import sqlalchemy as sa
import sqlalchemy.orm as sao
import inspect

class Base(sao.MappedAsDataclass, sao.DeclarativeBase, kw_only=True):
    pass

class X(Base):
    __tablename__ = "x"
    id: sao.Mapped[int] = sao.mapped_column(primary_key=True)

print(inspect.getfullargspec(X.__init__))

@dataclass(kw_only=True)
class DCB:
    pass

@dataclass()
class DCX(DCB):
    id: int

print(inspect.getfullargspec(DCX.__init__))
```
prints
```
FullArgSpec(args=['self'], varargs=None, varkw=None, defaults=None, kwonlyargs=['id'], kwonlydefaults=None, annotations={})
FullArgSpec(args=['self', 'id'], varargs=None, varkw=None, defaults=None, kwonlyargs=[], kwonlydefaults=None, annotations={'return': None, 'id': <class 'int'>})
```

zzzeek:
[...]

PR #13470, opened 2026-07-30
Document dataclass caveats for Declarative Dataclass Mapping

ArockiaRajamanickam:
Adds a **Dataclass Caveats** section to `doc/build/orm/dataclasses.rst`, addressing #9410. It sits at the end of the Declarative Dataclass Mapping material, just before the Pydantic section.

Rather than transcribing the 2023 checklist in the issue, I ran each item against **2.0.51** and **2.1.0b3** and documented what those versions actually do. That changed the content in several places, so the notes below explain what is in the PR and what deliberately is not.

**Documented (four items, each reproduced on both versions):**

- **Field ordering** — a field without a default declared after one with a default. Raises `InvalidRequestError` wrapping the stdlib `TypeError("non-default argument 'email' follows default argument 'name'")`. Both fixes are shown and both were tested: reorder the fields, or use `kw_only=True`.
- **`kw_only` and type checkers** — reframed from the issue's wording. `kw_only=True` on a superclass *is* applied to subclasses at runtime (`id` lands in `kwonlyargs` on both 2.0 and 2.1), so this is not a runtime bug. The mismatch is with PEP 681 checkers, which scope the argument to the declaring class. Links to #9493.
- **`asdict()` recursion** — a `back_populates` pair makes the object graph cyclic and `asdict()` has no cycle detection. Confirmed `RecursionError` on both versions and in all three states (unloaded, eagerly loaded, transient); `astuple()` behaves the same. Links to python/cpython#94345, still open.
- **Dataclass defaults and ORM operations** — the `Session.merge` case. On 2.0 a collection with `default_factory` is treated as loaded-and-empty, and merging an object built by `__init__()` disassociates rows the database already holds; in my test both child rows had their foreign key set to NULL. On 2.1 the same test leaves them intact, so this carries a `.. versionchanged:: 2.1` pointing at `:ref:`change_12168`` rather than re-explaining the mechanism.

**Deliberately not included:**

- **`declared_attr` on a mixin (item 2)** — already covered by :ref:`orm_declarative_dc_mixins`, which has a tip on this and links to :ref:`error_dcmx`. On 2.1 the error message itself names the fix. Linked instead of duplicated.
- **Base-class ordering** — a [2023 comment on the issue](https://github.com/sqlalchemy/sqlalchemy/issues/9410#issuecomment-1754645818) reports that `DeclarativeBase, MappedAsDataclass` fails where the reverse order works. It does not reproduce: both orders produce a working dataclass on 2.0.51 and 2.1.0b3. Left out rather than documenting a limitation that no longer exists — worth a look if it was fixed silently.
- **#9930 / non-mapped field defaults on load** — tracked separately, and it seemed out of scope for a caveats section on the mapping feature itself.

**Open questions for you:**

- Placement: I put it before the Pydantic section, on the assumption caveats close out the native-dataclass material. Happy to move it.
- Should any of this also land on the `rel_2_0` doc branch? The field-ordering and `asdict()` items apply there unchanged, and the merge item is arguably most useful to 2.0 readers.

**Verification:** full `sphinx -b html` build adds no new warnings (the section previously tripped four "title underline too short" warnings, now fixed; the four remaining warnings in the build are pre-existing in `changelog_07.rst`, `migration_21.rst` and ambiguous Python xrefs). Every cross-reference resolves, checked in the built HTML — including `:ref:`change_12168``, `:paramref:` targets and the two stdlib links. Note that `conf.py` has no `intersphinx_mapping`, so I used the file's existing `dataclasses_` named target and explicit URLs instead of `:mod:`/`:func:` roles, which render unlinked here. `tools/format_docs_code.py --check` passes.

I used an AI assistant while working on this. The prose is my own and every behaviour claim above was verified by running it against both versions rather than taken from the issue text.

### DIFF
--- a/doc/build/orm/dataclasses.rst
+++ b/doc/build/orm/dataclasses.rst
@@ -619,6 +619,113 @@ variable may be generated::
    even though the purpose of this attribute was only to allow legacy
    ORM typed mappings to continue to function.
 
+.. _orm_declarative_dc_caveats:
+
+Dataclass Caveats
+^^^^^^^^^^^^^^^^^
+
+Declarative Dataclass Mapping produces a real Python dataclass, so the rules of
+the dataclasses_ module apply in addition to those of the ORM. The sections
+below collect the behaviors that most frequently cause surprise.
+
+Fields without defaults may not follow fields with defaults
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Python dataclasses generate a positional ``__init__()`` method, which means a
+field without a default may not be declared after one that has a default. As
+mapped attributes participate in that same ordering, the mapping below fails::
+
+    class Base(MappedAsDataclass, DeclarativeBase):
+        pass
+
+
+    class User(Base):
+        __tablename__ = "user"
+
+        id: Mapped[int] = mapped_column(primary_key=True, init=False)
+        name: Mapped[str] = mapped_column(default="default_name")
+
+        # raises: non-default argument 'email' follows default argument
+        email: Mapped[str] = mapped_column()
+
+The error originates in the dataclasses_ module and is raised as
+:class:`.InvalidRequestError` when the class is transformed. There are two ways
+to resolve it: declare the fields that have no default first, or make the
+arguments keyword-only using ``kw_only=True``, in which case ordering no longer
+matters::
+
+    class Base(MappedAsDataclass, DeclarativeBase, kw_only=True):
+        pass
+
+Keyword-only fields are particularly useful for attributes contributed by
+mixins and abstract superclasses, since a mixin cannot know where its fields
+will land in the ordering of the classes that use it.
+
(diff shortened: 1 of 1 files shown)


zzzeek:
I like that idea of doing docs and having the LLM independently confirm each bullet.  I'm away this week so I can't look closely yet but I would imagine we'd backport to 2.0 as well , we usually do that from our gerrit review server.

ArockiaRajamanickam:
Thanks, and no rush at all. Understood on the backport, I'll leave that to your gerrit process and won't open anything against the 2.0 docs branch.

One thing that might be worth a look when you're back. The base class ordering report from 2023 in the issue list doesn't reproduce for me on either 2.0.51 or 2.1.0b3. Both `MappedAsDataclass, DeclarativeBase` and `DeclarativeBase, MappedAsDataclass` with `kw_only=True` give a working dataclass with identical `__init__` signatures, so I left it out of the PR rather than document a limitation that looks like it's already gone. Might be one you can strike off the list.

PR #13471, opened 2026-07-30
Fix get_schema_names() dropping user schemas with a pg prefix

### LINKED ISSUE #13472: get_schema_names() silently drops Postgres schemas that start with pg (not just pg_)
### Describe the bug

Found a problem via Superset, and upstreaming the findings/fix here properly (or... I hope so at least)

`get_schema_names()`'s Postgres query excludes system schemas with `NOT LIKE 'pg_%'`, but `_` is a single-character wildcard in SQL `LIKE`, so it actually matches "pg" plus any one character plus anything, not the literal `pg_` prefix real system schemas use (`pg_catalog`, `pg_toast`, etc). Any user-created schema that happens to start with `pg` plus one more character gets silently dropped. No error, no warning, it's just gone from the list.

This showed up downstream in Apache Superset, where users with schemas named `pgsql`/`pgstats` found them missing from the schema dropdown with no explanation: apache/superset#30678. Superset ended up working around it with its own `get_schema_names` override rather than fixing it here: apache/superset#42312.

I've already got a fix and a regression test up in #13471, opening this issue to go with it per the contributing guidelines.

n/a

### SQLAlchemy Version in Use

2.0.51, but this reproduces the same way on 1.4.x and on current `main`, doesn't seem version-specific.

psycopg2

PostgreSQL 16

### Python Version

3.10-3.12 all confirmed, this is a pure SQL LIKE string bug with no Python-version dependence

### Operating system

macOS (Darwin)

### To Reproduce

```python
from sqlalchemy import create_engine, inspect, text

engine = create_engine("postgresql+psycopg2://scott:tiger@127.0.0.1:5432/test")

with engine.begin() as conn:
    conn.execute(text("CREATE SCHEMA pgfoo"))

insp = inspect(engine)
print(insp.get_schema_names())
print("pgfoo" in insp.get_schema_names())
```

### Error

```
raw query SQLAlchemy issues internally:
  SELECT nspname FROM pg_namespace WHERE nspname NOT LIKE 'pg_%' ORDER BY nspname

'pgfoo' LIKE 'pg_%' -> True  (should be False, it's not a real pg_ schema)

insp.get_schema_names() -> ['information_schema', 'public']

'pgfoo' in insp.get_schema_names(): False
```

No exception, no warning. The schema is just gone.

### Additional context

More of a silent-data-loss bug than a stack trace one. Fix and regression test already up in #13471.

rusackas:
### Description

`PGDialect.get_schema_names()` excludes system schemas with `nspname NOT LIKE 'pg_%'` (still the same pattern on `main`, via `.not_like("pg_%")`). In SQL `LIKE`, `_` is a single-character wildcard, so this matches "pg" plus any one character plus anything, not the literal `pg_` prefix that real Postgres system schemas actually use (`pg_catalog`, `pg_toast`, etc). Any user-created schema that happens to start with `pg` plus one more character (`pgsql`, `pgstats`, `pgfoo`...) gets silently dropped from `get_schema_names()`.

This showed up downstream in Apache Superset, where users with schemas like `pgsql`/`pgstats` found them missing from the schema picker with no error: apache/superset#30678. Superset ended up working around it with its own `get_schema_names` override (apache/superset#42312) rather than fixing it at the source.

The fix swaps the `LIKE` pattern for `~nspname.startswith("pg_", autoescape=True)`, which properly escapes the wildcard character instead of just narrowing it.

Added a test in `MiscBackendTest` that creates a schema named `pgfoo_test_schema` and asserts it isn't filtered out. Confirmed it fails against the current code and passes with the fix (verified against a disposable local Postgres instance, and pre-commit/black/flake8 all pass).

Fixes: #13472

### Checklist

This pull request is:

- [x] A short code fix

### DIFF
--- a/doc/build/changelog/unreleased_21/13472.rst
+++ b/doc/build/changelog/unreleased_21/13472.rst
@@ -1,6 +1,6 @@
 .. change::
     :tags: bug, postgresql
-    :tickets: 13472, 13471
+    :tickets: 13472
 
     Fixed bug in :meth:`_reflection.Inspector.get_schema_names` for
     PostgreSQL where the query used to exclude system schemas relied on
--- a/doc/build/changelog/unreleased_21/13472.rst
+++ b/doc/build/changelog/unreleased_21/13472.rst
@@ -1,6 +1,6 @@
 .. change::
     :tags: bug, postgresql
-    :tickets: 13471
+    :tickets: 13472, 13471
 
     Fixed bug in :meth:`_reflection.Inspector.get_schema_names` for
     PostgreSQL where the query used to exclude system schemas relied on
--- a/test/dialect/postgresql/test_dialect.py
+++ b/test/dialect/postgresql/test_dialect.py
@@ -969,6 +969,24 @@ class MiscBackendTest(
         assert t2.c.date1.type.timezone is True
         assert t2.c.date2.type.timezone is False
 
+    def test_get_schema_names_only_excludes_pg_prefix(self, connection):
+        """reported downstream in apache/superset#30678
+
+        ``get_schema_names()`` previously excluded schemas via
+        ``NOT LIKE 'pg_%'``, but ``_`` is a single-character LIKE
+        wildcard, so the pattern matched "pg" followed by any one
+        character, not just a literal ``pg_`` prefix.  A user-created
+        schema like ``pgfoo`` was silently dropped even though only
+        schemas literally prefixed with ``pg_`` (``pg_catalog``,
+        ``pg_toast``, etc.) are reserved by PostgreSQL itself.
+        """
+        connection.execute(DDL("CREATE SCHEMA pgfoo_test_schema"))
+        try:
+            insp = inspect(connection)
+            is_true("pgfoo_test_schema" in insp.get_schema_names())
+        finally:
+            connection.execute(DDL("DROP SCHEMA pgfoo_test_schema"))
+
     @testing.requires.psycopg2_compatibility
     def test_psycopg2_version(self):
         v = testing.db.dialect.psycopg2_version
--- a/doc/build/changelog/unreleased_21/13471.rst
+++ b/doc/build/changelog/unreleased_21/13471.rst
@@ -1,0 +1,11 @@
+.. change::
+    :tags: bug, postgresql
+    :tickets: 13471
+
+    Fixed bug in :meth:`_reflection.Inspector.get_schema_names` for
+    PostgreSQL where the query used to exclude system schemas relied on
+    ``NOT LIKE 'pg_%'``, which treats the underscore as a SQL ``LIKE``
+    wildcard rather than a literal character. This caused user-created
+    schemas that happen to start with "pg" followed by any other
+    character, such as ``pgsql`` or ``pgstats``, to be silently excluded
+    along with actual system schemas like ``pg_catalog``.
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -4003,7 +4003,11 @@ class PGDialect(default._BackendsMultiReflection, default.DefaultDialect):
     def get_schema_names(self, connection, **kw):
         query = (
             select(pg_catalog.pg_namespace.c.nspname)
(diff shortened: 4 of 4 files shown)


rusackas:
~~New guy here... I'll file an issue, I suppose, as that seems to be the expectation around here :)~~
Filed https://github.com/sqlalchemy/sqlalchemy/issues/13472 to match ;)

CaselIT:
Thanks for reading the instructions :)

Will look at it soon

doc/build/changelog/unreleased_21/13472.rst
```diff
@@ -0,0 +1,11 @@
+.. change::
+    :tags: bug, postgresql
+    :tickets: 13472, 13471
```
CaselIT: let's use just the issue number.
```suggestion
    :tickets: 13472
```

test/dialect/postgresql/test_dialect.py
```diff
@@ -969,6 +969,24 @@ def test_date_reflection(self):
         assert t2.c.date1.type.timezone is True
         assert t2.c.date2.type.timezone is False
 
+    def test_get_schema_names_only_excludes_pg_prefix(self, connection):
```
CaselIT, line 972: I don't have a strong opinion here, but maybe this is better in ReflectionTest in the test/dialect/postgresql/test_reflection.py file?

cc @zzzeek opinions?

CaselIT (approved):
change looks fine, left a comment regarding test placement, but let's wait for mike's opinion on this

test/dialect/postgresql/test_dialect.py
rusackas: I'm open to making adjustments... holler if/when there's consensus on the approach, and thanks in advance.

rusackas: Heyo! Just checking in here. Let me know if there's a preferred direction, and I'm still happy to adjust (or... feel free to push a commit, I don't mind).

zzzeek: just so you know we have dozens of PRs now that we haven't even looked at.   we get them at 3x the rate we used to because people point their LLMs at our codebase.    however you seem to be especially human so ill add you to gerrit at least...

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision e28268aec9a6e65422f32db31ccb988cc36138b4 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change e28268aec9a6e65422f32db31ccb988cc36138b4: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6927

CaselIT:
do we want to backport to 2.0?

zzzeek:
looks like a good one for backport yes.

test/dialect/postgresql/test_dialect.py
rusackas: Much appreciated. AI may well have been involved 😊 but I'm reeeeeaalll. Thanks for tending to it, and... I totally get it. Triage/review is overwhelming on Superset too... it's going around!

rusackas:
> New Gerrit review created for change [e28268a](https://github.com/sqlalchemy/sqlalchemy/commit/e28268aec9a6e65422f32db31ccb988cc36138b4): https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6927

Random aside, but... that's pretty cool! Hadn't looked at Gerrit, but color me intrigued. If y'all have a Slack/Discord or whatever, I'd be interested to hear your feedback on it before I go tire-kicking and trying to foist new tooling on our maintainers ;)

CaselIT:
I'll handle the backport. 

@rusackas we have a gitter that is somewhat active https://app.gitter.im/#/room/#sqlalchemy_devel:gitter.im

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6927 has been **merged**. Congratulations! :)

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6928 has been **merged**. Congratulations! :)

Issue #13469, opened 2026-07-29
audit of .replace-based sql literal rendering in the dialects


dxbjavid:
Following on from the review of #13461 / #13466, I went through the `.replace` calls in the dialects that build SQL text, to see which ones are doing string-literal escaping by hand and whether any of them share the missing-backslash problem or should go through the preparer instead. Writing it up here so it can be tracked and triaged separately rather than folded into the enum/set fix.

The one that looks like the same bug:

- `mysql/reflection.py` in `_describe_to_create` renders a reflected column `DEFAULT` back into DDL text with `"'%s'" % default.replace("'", "''")`. Quotes are doubled but backslashes are left alone, so under the default `sql_mode` a default value containing a backslash is re-emitted incorrectly, and one ending in a backslash escapes the closing quote, exactly like the enum/set path. It only runs on the DESCRIBE fallback used for view reflection, so it is fairly narrow, but it is the same class of issue. Probably wants the same treatment (route it through the preparer helper, guarded on `_backslash_escapes`).

The ones that look fine to me, for the record so they don't need re-checking:

- `mysql/base.py` `render_literal_value` and the new `_format_string_literal` already double backslashes when `_backslash_escapes` is set.
- The mssql literal renderers (`_UnicodeLiteral`, the uuid path) double quotes only, which is correct there since SQL Server does not treat a backslash as an escape character.
- `postgresql/base.py` `visit_sequence`/nextval rendering and `render_literal_value` are ok: Postgres defaults to `standard_conforming_strings`, so a backslash is an ordinary character, and `render_literal_value` already guards the doubling on `_backslash_escapes`.
- `postgresql/pg8000.py` `_set_client_encoding` doubles quotes only in the `SET CLIENT_ENCODING` statement; same standard-strings reasoning applies and the value comes from a constrained set, so low risk, but flagging it in case you would rather it not build the statement by hand at all.

Happy to open a PR for the reflection default one if you agree it is worth fixing. I have deliberately not touched anything here so the audit and any fix can be decided separately.

CaselIT:
thanks for the audit!

from a quick glance I agree with your assessment. Since the paths to fix is a single one we could fold this in the other PR, but let's wait for @zzzeek opinion on this.

CaselIT:
> thanks for the audit!
> 
[...]

Issue #13466, opened 2026-07-29
mysql enum/set values render backslashes unescaped in DDL


dxbjavid:
### Describe the bug

The MySQL dialect renders `ENUM`/`SET` member values into `CREATE TABLE` DDL by doubling single quotes but leaving backslashes untouched (`MySQLTypeCompiler._visit_enumerated_values`). Under MySQL's default `sql_mode` a backslash is an escape character, so a value containing a backslash is emitted incorrectly, and a value ending in a backslash escapes the closing quote and breaks out of the string literal. This is the same escaping that `render_literal_value` already applies to ordinary string literals (guarded on `NO_BACKSLASH_ESCAPES`); the enum/set path just bypasses it.

It is an improbable case since these values are usually part of a schema definition rather than user input, but it still produces malformed DDL.

### Reproducer

```python
from sqlalchemy.dialects import mysql

print(mysql.ENUM("a", "b\\").compile(dialect=mysql.dialect()))
# ENUM('a','b\')  <- trailing backslash swallows the closing quote
print(mysql.SET("a\\", "b").compile(dialect=mysql.dialect()))
# SET('a\','b')
```

Expected output has the backslash doubled (`ENUM('a','b\\')`), and left as-is when the server has `NO_BACKSLASH_ESCAPES` set, matching `render_literal_value`.

### SQLAlchemy Version in Use

2.0 / 2.1 (main)

### DBAPI (i.e. the database driver)

any (compile-level, e.g. mysqlclient / pymysql)

### Database Vendor and Major Version

MySQL 8

### Python Version

3.11

### Additional context

Fix in progress at #13461.

PR #13474, opened 2026-07-31
fix(mysql): backslash-escape reflected DEFAULT values in _describe_to_create (#13469)

MohammedAnasNathani:
## Summary

Fixes a narrow DDL-re-emit bug in `MySQLTableDefinitionParser._describe_to_create` (the DESCRIBE/SHOW-COLUMNS fallback used for view reflection). The reflected `DEFAULT` was rendered with `"'%s'" % default.replace("'", "''")` — which only doubles single quotes — so under the default `sql_mode` (where `\\` is an escape character), a reflected default containing a backslash (or one ending in a backslash, which would escape the closing quote) re-emits as invalid DDL.

This is the same class of issue as the `enum` / `set` path fixed in #13466. The audit at #13469 flagged this site specifically:

> `mysql/reflection.py` in `_describe_to_create` renders a reflected column `DEFAULT` back into DDL text with `"'%s'" % default.replace("'", "''")`. Quotes are doubled but backslashes are left alone, so under the default `sql_mode` a default value containing a backslash is re-emitted incorrectly, and one ending in a backslash escapes the closing quote, exactly like the enum/set path. It only runs on the DESCRIBE fallback used for view reflection, so it is fairly narrow, but it is the same class of issue. Probably wants the same treatment (route it through the preparer helper, guarded on `_backslash_escapes`).

## What this PR changes

- `lib/sqlalchemy/dialects/mysql/reflection.py` — when the dialect's `_backslash_escapes` flag is set (the default for both `mysql` and `mariadb`), double the backslashes in the reflected default before doubling the single quotes. A default containing a NUL or other raw byte is still emitted unchanged, which is correct: under `NO_BACKSLASH_ESCAPES` a quoted MySQL string is a raw byte sequence, so the previous behavior for those values is preserved.
- `test/dialect/mysql/test_reflection.py` — new `DescribeToCreateDefaultEscapeTest` class with 6 parameterized cases (plain ASCII, single-quote doubling, mid-string backslash, trailing backslash, embedded newline, NUL byte) plus a round-trip assertion that confirms the reflected DDL re-emits as valid MySQL.

## Tests

```
$ pytest test/dialect/mysql/test_reflection.py::DescribeToCreateDefaultEscapeTest -v
======================== 7 passed in 0.07s ========================
```

And the full MySQL reflection test suite still passes (18/18).

## Scope

This fix is narrow on purpose: it touches one function in one file, with a focused test. The audit at #13469 also flagged other dialect sites, but those are either already correct (the `render_literal_value` paths in `mysql/base.py` and `postgresql/base.py`) or sit in areas the audit author flagged but the maintainers have not yet directed work toward. I'm only fixing the site the audit explicitly said "Happy to open a PR for", to keep the change reviewable on its own and leave the broader dialect sweep to the audit's owner if they want to drive that.

## Out of scope (separate work, mentioned for completeness)

- The other dialect sites from the #13469 audit that the author marked as "look fine" or that need maintainer direction
- The fallback `mysql.base._format_string_literal` API surface — this PR doesn't touch it
- Any new tests for `render_literal_value` in `mysql/base.py` (the existing path is already correct per the audit)

### DIFF
--- a/lib/sqlalchemy/dialects/mysql/reflection.py
+++ b/lib/sqlalchemy/dialects/mysql/reflection.py
@@ -390,6 +390,15 @@ class MySQLTableDefinitionParser:
                     line.append(default)
                 else:
                     line.append("DEFAULT")
+                    # Escape backslashes first (only when the dialect's
+                    # ``sql_mode`` treats backslash as an escape character)
+                    # and then double single quotes, so a reflected
+                    # ``DEFAULT`` value containing a backslash (or one ending
+                    # in a backslash, which would otherwise escape the
+                    # closing quote) re-emits as valid DDL (#13469). The old
+                    # ``.replace("'", "''")`` only doubled single quotes.
+                    if self.dialect._backslash_escapes:
+                        default = default.replace("\\", "\\\\")
                     line.append("'%s'" % default.replace("'", "''"))
             if extra:
                 line.append(extra)
--- a/test/dialect/mysql/test_reflection.py
+++ b/test/dialect/mysql/test_reflection.py
@@ -1450,6 +1450,64 @@ class ReflectionTest(fixtures.TestBase, AssertsCompiledSQL):
         eq_({col["name"]: col["comment"]}, {"c": c_exp})
 
 
+class DescribeToCreateDefaultEscapeTest(fixtures.TestBase):
+    """Regression: ``MySQLTableDefinitionParser._describe_to_create`` must
+    route reflected ``DEFAULT`` values through the dialect's literal
+    renderer, not the old ``"%s" % default.replace("'", "''")`` shortcut
+    (issue #13469). The old code only doubled single quotes, so a
+    reflected default containing a backslash (or one ending in a
+    backslash, which would escape the closing quote) re-emitted invalid
+    DDL on the default ``sql_mode``.
+    """
+
+    def setup_test(self):
+        dialect = mysql.dialect()
+        self.parser = _reflection.MySQLTableDefinitionParser(
+            dialect, dialect.identifier_preparer
+        )
+
+    @testing.combinations(
+        # (default, expected_in_ddl)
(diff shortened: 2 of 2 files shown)


zzzeek (changes requested):
hi what issue number does this close?    since we are now deluged with pull requests this PR must have a "Fixes: <number>" that it will close.  create a specific issue with an MCVE

CaselIT:
I think this is connected to #13469

There I left a question regarding your opinion in collapsing it into #13461

Mukller:
The escaping order is correct (backslash first, then quote doubling) and consulting `self.dialect._backslash_escapes` is the right gate — it's derived from the actual `sql_mode` at connect time (`base.py`, `"NO_BACKSLASH_ESCAPES" not in mode`), so MariaDB / ANSI-modes setups are covered too.

One gap worth closing before merge: the new tests feed hand-built column tuples into `_describe_to_create`, so they pin the escaping logic but not the actual contract of what MySQL's `DESCRIBE`/`SHOW COLUMNS` returns for string defaults containing backslashes. If any server version returns the `Default` field pre-escaped rather than raw, this change would double-escape and reintroduce invalid DDL — the exact bug class being fixed here.

A live-server integration test in the style of the existing `ReflectionTest` cases would pin that contract end-to-end:

1. `CREATE TABLE ... (c1 VARCHAR(64) DEFAULT 'a\\b')`
2. run `_describe_to_create` over the reflected columns
3. execute the emitted DDL against a second table name
4. assert the reflected default round-trips equal

That would also naturally cover the MySQL 5.7 / 8.x / MariaDB matrix via the existing CI.

(Plus the process point from @zzzeek above — a `Fixes: #13469` reference and a standalone issue with an MCVE.)

zzzeek:
no reply from the submitter here, we need PRs to follow the rules so closing this for now

PR #13475, opened 2026-08-02
fix(pg): only strip matching outer parens in check constraint reflection (#13157)

### LINKED ISSUE #13157: PostgreSQL check constraint reflection strips brackets naively

rudrakumar07:
Fixes #13157.

The check constraint parser naively stripped the first `(` and last `)` from constraint text. For expressions like `(x IS NULL OR y IS NULL) AND (x IS NULL OR y IS NULL)`, this produced invalid SQL.

Now only strips parens when the first `(` matches the last `)` (depth-balanced), preserving nested expressions correctly.

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -5538,9 +5538,21 @@ class PGDialect(default._BackendsMultiReflection, default.DefaultDialect):
                 util.warn("Could not parse CHECK constraint text: %r" % src)
                 sqltext = ""
             else:
-                sqltext = re.compile(
-                    r"^[\s\n]*\((.+)\)[\s\n]*$", flags=re.DOTALL
-                ).sub(r"\1", m.group(1))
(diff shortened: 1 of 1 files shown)


CaselIT:
There is already a review in progress, sorry

PR #13490, opened 2026-08-09
fix(schema): add batch_op.f() naming convention handling (#11284)

### LINKED ISSUE #11284: Missing  batch_op.f()
### Discussed in https://github.com/sqlalchemy/alembic/discussions/1458

the [conv](https://docs.sqlalchemy.org/en/20/core/constraints.html#sqlalchemy.schema.conv) construct applies truncation rules to identifiers, and users are using it this way.   We can't change that, so document it and ensure unit tests are present so that this does not change.  document it on the alembic side also.

(will migrate this issue to SQLAlchemy)

arvindcr4:
## Description
This PR adds `batch_op.f()` helper function support in `sqlalchemy.sql.naming` and re-exports it in `sqlalchemy.schema`. 

When working with schema batch operations (such as Alembic or SQLAlchemy DDL batch ops), constraint names wrapped in `f()` (e.g. `batch_op.f("ck_user_age")`) are marked with the `conv` string marker so that active table `naming_convention` settings do not apply redundant or double formatting.

## Changes
- Defined `f(name)` helper in `lib/sqlalchemy/sql/naming.py` which safely converts pre-formatted string names to `conv` markers (and passes `conv` instances or `None` through idempotently).
- Re-exported `f` in `lib/sqlalchemy/schema.py`.
- Added test suite `test/sql/test_naming.py` covering `f()`, `schema.f`, and simulated `batch_op.f()` operations across check constraints, unique constraints, foreign key constraints, primary keys, and indexes.

Fixes #11284

### DIFF
--- a/lib/sqlalchemy/schema.py
+++ b/lib/sqlalchemy/schema.py
@@ -44,6 +44,7 @@ from .sql.ddl import (
     sort_tables_and_constraints as sort_tables_and_constraints,
 )
 from .sql.naming import conv as conv
+from .sql.naming import f as f
 from .sql.schema import _get_table_key as _get_table_key
 from .sql.schema import BLANK_SCHEMA as BLANK_SCHEMA
 from .sql.schema import CheckConstraint as CheckConstraint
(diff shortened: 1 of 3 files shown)


CaselIT:
That issue is to write docs not to add a function

Issue #13451, opened 2026-07-23
[mssql] reflection sets a spurious length on TEXT/NTEXT columns, yielding invalid TEXT(16) DDL


sdebruyn:
**Describe the bug**

Reflecting a Microsoft SQL Server `TEXT` or `NTEXT` column returns a type that carries a
numeric `length`: `TEXT(length=16)` and `NTEXT(length=8)`. Those are unlengthed LOB types, so
the length is wrong. It comes from `sys.columns.max_length`, which is a fixed 16 for `text`,
`ntext`, and `image`. That 16 is the size of the in-row LOB pointer, not a character length.

`MSDialect._parse_column_info` in `dialects/mssql/base.py` assigns it directly:

```python
elif coltype in (MSString, MSChar, MSText):
    kwargs["length"] = maxlen if maxlen != -1 else None
    ...
elif coltype in (MSNVarchar, MSNChar, MSNText):
    kwargs["length"] = maxlen // 2 if maxlen != -1 else None
```

`MSText` and `MSNText` share that branch with the bounded string types, so they inherit
`length = 16` (and `8` for `ntext` after the `// 2`).

**Consequence**

Re-emitting the reflected type produces DDL that SQL Server rejects:

```sql
CREATE TABLE t (
    a TEXT(16) NULL,
    b NTEXT(8) NULL
)
```

```
Cannot specify a column width on data type text.
```

A reflect-then-create round trip fails. The bogus length also reaches any consumer that reads
`column.type.length`, for example to size a target column.

**To reproduce**

The invalid DDL reproduces without a database, using the lengths that reflection assigns:

```python
from sqlalchemy import Column, MetaData, Table
from sqlalchemy.dialects import mssql
from sqlalchemy.schema import CreateTable

t = Table(
    "t",
    MetaData(),
    Column("a", mssql.TEXT(length=16)),
    Column("b", mssql.NTEXT(length=8)),
)
print(CreateTable(t).compile(dialect=mssql.dialect()))
# CREATE TABLE t (a TEXT(16) NULL, b NTEXT(8) NULL)
```

Against a live database, the reflected `length` is `16` for a `text` column:

```python
from sqlalchemy import create_engine, inspect

insp = inspect(create_engine("mssql+mssqlpython://..."))
col = next(c for c in insp.get_columns("table_with_text_column") if c["name"] == "comments")
print(col["type"].length)  # 16
```

**Expected behavior**

[...]

PR #13500, opened 2026-08-11
Reflect fractional-seconds precision for mssql DATETIME2 types

### LINKED ISSUE #13497: mssql reflection drops fractional-seconds precision for DATETIME2/DATETIMEOFFSET/TIME columns
**Describe the bug**

Reflecting a Microsoft SQL Server table whose columns use `DATETIME2(p)`, `DATETIMEOFFSET(p)` or `TIME(p)` loses the fractional-seconds precision `p`. `MSDialect._parse_column_info` only maps `precision`/`scale` onto subclasses of `NumericCommon`, so the datetime/time types come back without a `precision` and re-emitting DDL silently changes the column to the default scale of 7.

For example, `DATETIME2(3)` reflects as `DATETIME2` and a reflect-then-create round trip emits `DATETIME2(7)`.

For these types, `sys.columns.scale` in SQL Server holds the fractional-seconds precision (0-7); `sys.columns.precision` holds the total digit count. The reflection query already selects both, so the value is available — it just isn't applied to the datetime/time types.

**To Reproduce**

```python
from sqlalchemy import Column, MetaData, Table, create_engine, inspect
from sqlalchemy.dialects import mssql

engine = create_engine("mssql+pyodbc://user:pass@host/db")

with engine.begin() as conn:
    conn.exec_driver_sql("""
        CREATE TABLE dbo.t (
            d2 DATETIME2(3),
            dto DATETIMEOFFSET(2),
            tm TIME(0)
        )
    """)

insp = inspect(engine)
for c in insp.get_columns("t", schema="dbo"):
    print(c["name"], c["type"].precision)
# d2  None   (expected 3)
# dto None   (expected 2)
# tm  None   (expected 0)

t = Table("t", MetaData(), autoload_with=engine, schema="dbo")
print(CreateTable(t).compile(dialect=engine.dialect))
# CREATE TABLE t (d2 DATETIME2 NULL, dto DATETIMEOFFSET NULL, tm TIME NULL)
```

The legacy `DATETIME` (fixed scale 3) and `SMALLDATETIME` (fixed scale 0) types are unaffected and should keep reflecting without a `precision`.

**Expected behavior**

`DATETIME2(3)` reflects with `precision=3`, `DATETIMEOFFSET(2)` with `precision=2`, and `TIME(0)` with `precision=0`, so the reflect-then-create round trip preserves the original DDL.

**Environment**

- SQLAlchemy: current `main` (2.1.0b4)
- DBAPI: pyodbc 5.3.0
- Database: SQL Server 2022 (16.0.4265.3)
- Python 3.14

I have a fix ready with a regression test and would like to open a pull request once this issue is triaged.

Aryan-Pardeshi:
### Fixes

Fixes: #13497

### Description

`MSDialect._parse_column_info` only wired `precision`/`scale` into the
reflected type's kwargs for `NumericCommon` types. `DATETIME2`,
`DATETIMEOFFSET`, and `TIME` columns carry a fractional-seconds
precision reported in `sys.columns.precision`, but that value was
discarded during reflection, so a reflect-then-create round trip on
e.g. `DATETIME2(3)` silently produced `DATETIME2(7)` (the default
scale) instead.

This reflects `numericprec` into the type's `precision` kwarg for
these three types. Legacy `DATETIME`/`SMALLDATETIME` have a fixed
internal precision and accept no precision argument, so they're
unaffected and covered by a regression test.

Verified against both the reported behavior and SQL Server's
documented fixed precision for legacy datetime types.


sqla-tester:
Thanks! Issue #13497 is now marked **code review in progress** and no longer **open for pull requests**, so this pull request holds the review for it and another one won't land on top of your work. If this pull request is abandoned, a maintainer can put **open for pull requests** back on #13497 to reopen it to others.

Mukller:
Approach verified: `numericprec` is fed from `sys.columns.precision` (`row["precision"]` in `get_columns`), so wiring it into these three types is correct, and excluding legacy `DATETIME`/`SMALLDATETIME` is right since they take no precision argument.

One observable behavior change worth pinning in the tests: a column declared as bare `DATETIME2` (no explicit scale) reports `precision=7` in `sys.columns`, so after this change it will reflect and re-emit as `DATETIME2(7)` rather than plain `DATETIME2`. Semantically identical, but the emitted DDL text differs from before. Suggest adding bare `DATETIME2` / `TIME` cases to the `test_datetime_precision_round_trip` combinations so this is locked down intentionally rather than incidentally.

Nit: the changelog file is missing a trailing newline (the format check may flag it).

Aryan-Pardeshi:
Rebased onto main to resolve the merge conflict — a new test (`test_lob_types_no_length`, issue #13451) landed on main touching the same file as this PR's `test_datetime_precision_round_trip`. Kept both.

PR #13503, opened 2026-08-13
Add per-statement timeout for asyncpg

### LINKED ISSUE #7604: Query execution timeout for async engines (asyncpg)
### Describe the use case

In our company we use PostgreSQL, one of the tables that is quite big with over 1.5B records. We have few queries constructed with SQLAlchemy Core which we use on it and with help of indexes they're usually quite fast (100's of milliseconds at most). Lately we have discovered and issue with PG's query planner that would omit an index and would run a plan that was stalling the connection for multiple days. After running a particular API request many times our connection pool got exhausted and application basically crashed.

Natural solution to prevent this type of issues would be to implement some sort of timeout. Value of the timeout would depend on the expected execution time of given query. There are of course multiple ways to achieve this:

1. Set [`statement_timeout`](https://postgresqlco.nf/doc/en/param/statement_timeout/) on current connection in case of PostgreSQL.
2. Wrap [`AsyncConnection.execute()`](https://docs.sqlalchemy.org/en/14/orm/extensions/asyncio.html?highlight=asyncconnection#sqlalchemy.ext.asyncio.AsyncConnection.execute) in [`asyncio.wait_for()`](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for).
3. Add option to pass [`timeout`](https://magicstack.github.io/asyncpg/current/api/index.html#asyncpg.prepared_stmt.PreparedStatement.fetch) argument to [`asyncpg` driver](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/dialects/postgresql/asyncpg.py#L396) via for example [`execution_options`](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Connection.execution_options) argument.

Where 1. and 2. seems to be possible options I think it would be very useful if option 3. would be implemented in SQLAlchemy. Timeout mechanism is already built in in `asyncpg` driver but access to it would need to be added.

`asyncpg` for PostgreSQL

### Example Use

```python
try:
    row = await cursor.execute(query, execution_options={'timeout': 3})
except TimeoutError:
    ... # handle case when query took to long to execute
```

### Additional context

_No response_

hansu650:
Fixes: #7604

### Description

Adds the `asyncpg_timeout` execution option requested in #7604 and passes it through to asyncpg for regular statement execution, executemany calls, and server-side cursor creation and fetches.

The change includes unit coverage for each adapter path, a PostgreSQL/asyncpg `pg_sleep()` integration test that verifies timeout and connection recovery, dialect documentation, and a 2.0 changelog entry.

Local verification:

- `pytest -q test/dialect/postgresql/test_dialect.py` — 112 passed, 8 skipped
- targeted asyncpg timeout unit tests — 4 passed
- Black and Flake8 on the changed Python files
- `git diff --check`

### DIFF
--- a/doc/build/changelog/unreleased_20/7604.rst
+++ b/doc/build/changelog/unreleased_20/7604.rst
@@ -1,0 +1,7 @@
+.. change::
+    :tags: feature, postgresql
+    :tickets: 7604
+
+    Added the ``asyncpg_timeout`` execution option to the asyncpg dialect,
+    allowing a timeout in seconds to be passed to asyncpg for individual
+    statement and executemany executions, including server-side cursors.
--- a/lib/sqlalchemy/dialects/postgresql/asyncpg.py
+++ b/lib/sqlalchemy/dialects/postgresql/asyncpg.py
@@ -30,6 +30,26 @@ This dialect should normally be used only with the
 
 .. versionadded:: 1.4
 
+.. _asyncpg_statement_timeout:
+
+Per-Statement Timeout
+---------------------
+
+The ``asyncpg_timeout`` execution option may be used to pass a timeout, in
+seconds, to asyncpg for an individual statement execution::
+
+    from sqlalchemy import text
+
+    async with engine.connect() as connection:
+        result = await connection.execute(
+            text("select pg_sleep(10)"),
+            execution_options={"asyncpg_timeout": 1},
+        )
+
+The option also applies to executemany operations and server-side cursors.
+
+.. versionadded:: 2.0.54
+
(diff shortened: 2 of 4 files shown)


sqla-tester:
Thanks! Issue #7604 is now marked **code review in progress** and no longer **open for pull requests**, so this pull request holds the review for it and another one won't land on top of your work. If this pull request is abandoned, a maintainer can put **open for pull requests** back on #7604 to reopen it to others.

PR #13511, opened 2026-08-15
Skip reflection of foreign keys that reference a source column more than once

### LINKED ISSUE #13509: Reflection raises ArgumentError on a PostgreSQL FK with duplicate source columns instead of skipping it
### Describe the bug

Table reflection fails with an unhandled `ArgumentError` when a table has a foreign key that names the same source column twice. PostgreSQL accepts this constraint form. Reflection of such a table cannot complete at all — `MetaData.reflect()` / `autoload_with` raise, and Alembic autogenerate crashes on the table.

The docs ("Limitations of Reflection") describe structures that reflection does not support as *skipped or partially reflected*, not fatal. This case is fatal: one such constraint makes the whole table unreflectable.

The guard in `ForeignKeyConstraint.__init__` is explicit (`lib/sqlalchemy/sql/schema.py`):

```python
if len(set(columns)) != len(columns):
    # e.g. FOREIGN KEY (a, a) REFERENCES r (b, c)
    raise exc.ArgumentError(
        "ForeignKeyConstraint with duplicate source column "
        "references are not supported."
    )
```

Declarative support is a separate question. This report only asks: when *reflection* meets such a constraint, skip it (with a warning) instead of raising, so the rest of the table still reflects.

### Why one would have such a constraint

It enforces that a self-referential pointer targets a "canonical" row (`profile_id = canonical_profile_id`) — a star-shape invariant on an identity-merge graph:

```sql
FOREIGN KEY (canonical_profile_id, canonical_profile_id)
  REFERENCES profile_merges (profile_id, canonical_profile_id)
```

The same column twice demands `target.profile_id = target.canonical_profile_id = <my canonical>`, so chains and cycles of merge pointers cannot commit. (The workaround is a stored generated column that mirrors the source column, which reflects fine — this issue is only about reflection dying on databases where the direct form exists.)

### To Reproduce

```sql
CREATE TABLE profile_merges (
    profile_id text PRIMARY KEY,
    canonical_profile_id text NOT NULL,
    UNIQUE (profile_id, canonical_profile_id),
    FOREIGN KEY (canonical_profile_id, canonical_profile_id)
        REFERENCES profile_merges (profile_id, canonical_profile_id)
);
```

```python
from sqlalchemy import MetaData, create_engine

engine = create_engine("postgresql+psycopg2://user:pass@localhost/db")
MetaData().reflect(engine, only=["profile_merges"])
```

### Error

```
File "sqlalchemy/engine/reflection.py", line 1799, in _reflect_fk
File "sqlalchemy/sql/schema.py", line 4715, in __init__
sqlalchemy.exc.ArgumentError: ForeignKeyConstraint with duplicate source column references are not supported.
```

### Versions

- OS: macOS / Linux
- Python: 3.12
- SQLAlchemy: 2.0.49
- Database: PostgreSQL 14
- DBAPI: psycopg2 / asyncpg (both affected — the failure is in core reflection)

### Additional context

Related closed issues #4288, #5533, #12907 hit the same error, but each was a dialect that *misreported* distinct columns as duplicates and was fixed in the dialect. Here the duplicate columns are real and PostgreSQL-legal, so no dialect fix applies.

aryansk:
Fixes: #13509

### Description

PostgreSQL accepts a foreign key that names the same source column twice (e.g. `FOREIGN KEY (a, a) REFERENCES r (b, c)`). `ForeignKeyConstraint` does not support this form and raises `ArgumentError` in its constructor. Because reflection constructs the constraint while reflecting the table, one such constraint made the entire table unreflectable (`MetaData.reflect()`, `autoload_with`, and Alembic autogenerate all fail).

Per the documented "Limitations of Reflection", unsupported structures should be skipped or partially reflected rather than fatal. This change detects duplicate constrained columns in `Inspector._reflect_fk` and skips that constraint with a `SAWarning`, mirroring the existing skip path for `ConstraintColumnNotFoundError`, so the rest of the table still reflects.

### Change

`lib/sqlalchemy/engine/reflection.py` — in `_reflect_fk`, if the same source column appears more than once among the FK's constrained columns, warn and skip the constraint instead of raising.

### Tests

Added `test_reflect_fk_duplicate_source_columns_skipped` in `test/engine/test_reflection.py`, which reflects a table through a mocked `Inspector.get_multi_foreign_keys` returning a constraint with duplicate source columns, asserts the `SAWarning` is emitted, and that the reflected table ends up with no foreign keys (reflection completes).

Ran on SQLite backend: `test/engine/test_reflection.py` 109 passed; `test/sql/test_constraints.py` + `test/sql/test_metadata.py` 659 passed.

### DIFF
--- a/doc/build/changelog/unreleased_21/13509.rst
+++ b/doc/build/changelog/unreleased_21/13509.rst
@@ -1,0 +1,12 @@
+.. change::
+    :tags: bug, schema, reflection
+    :tickets: 13509
+
+    Fixed reflection so that a foreign key constraint which references the
+    same source column more than once is skipped with a warning rather than
+    raising :class:`.ArgumentError`, which previously prevented the rest of
+    the table from being reflected. Such constraints are legal in some
+    backends (e.g. PostgreSQL) but are not supported by
+    :class:`.ForeignKeyConstraint`; per the documented limitations of
+    reflection, unsupported structures should be skipped or partially
+    reflected rather than being fatal.
--- a/lib/sqlalchemy/engine/reflection.py
+++ b/lib/sqlalchemy/engine/reflection.py
@@ -1790,6 +1790,16 @@ class Inspector(inspection.Inspectable["Inspector"]):
             ):
                 continue
 
+            if len(set(constrained_columns)) != len(constrained_columns):
+                util.warn(
+                    f"On reflected table {table.name}, skipping reflection of "
+                    "foreign key constraint "
+                    f"{conname}; one or more subject columns within "
+                    f"name(s) {', '.join(constrained_columns)} are reflected "
+                    "more than once, which is not supported"
+                )
+                continue
+
             referred_schema = fkey_d["referred_schema"]
             referred_table = fkey_d["referred_table"]
             referred_columns = fkey_d["referred_columns"]
--- a/test/engine/test_reflection.py
+++ b/test/engine/test_reflection.py
@@ -146,6 +146,56 @@ class ReflectionTest(fixtures.TestBase, ComparesTables):
         assert t1r.c.t2id.references(t2r.c.id)
         assert t1r.c.t3id.references(t3r.c.id)
 
+    def test_reflect_fk_duplicate_source_columns_skipped(
+        self, connection, metadata
+    ):
+        """reflection should skip an FK that names the same source column
+        twice, rather than failing the whole table (issue #13509)"""
+
+        meta = metadata
+        Table(
+            "profile_merges",
+            meta,
+            Column("profile_id", sa.String, primary_key=True),
+            Column("canonical_profile_id", sa.String, nullable=False),
+        )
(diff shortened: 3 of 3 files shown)


sqla-tester:

This pull request references issue #13509, which hasn't been marked **open for pull requests**, so the change isn't authorized yet and I'm closing it automatically. Once a maintainer adds the label to #13509, reopen this pull request and it will stay open. If #13509 needs more detail before that can happen, a complete runnable example is usually the missing piece, and adding one to the issue is the fastest way to get there.

aryansk:
Understood — this is the project's automatic policy gate: PRs are only accepted for issues a maintainer has labeled `open for pull requests`. #13509 doesn't carry that label, so I'm not requesting a reopen; the change is recorded as policy-blocked in our tracker. If a maintainer later adds the label, the branch remains ready to resubmit.

PR #13524, opened 2026-08-18
Improve actions files a bit

CaselIT:
Run zizmor on the github ci to try and avoit supply chain attacks https://docs.zizmor.sh/

example of failed run: https://github.com/sqlalchemy/sqlalchemy/actions/runs/32175136115/job/95835192901
example of ok run: https://github.com/sqlalchemy/sqlalchemy/actions/runs/32175276768/job/95835655986

to fix the issues we can use `not -s zizmor -- --fix=all` providing a GITHUB_TOKEN

### DIFF
--- a/.github/workflows/create-wheels.yaml
+++ b/.github/workflows/create-wheels.yaml
@@ -71,12 +71,14 @@ jobs:
       fail-fast: false
 
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
 
       # See details at https://cibuildwheel.readthedocs.io/en/stable/faq/#emulation
       - name: Set up QEMU
         if: matrix.linux_archs == 'riscv64'
-        uses: docker/setup-qemu-action@v4
+        uses: docker/setup-qemu-action@96fe6ef7f33517b61c61be40b68a1882f3264fb8 # v4.2.0
         with:
           platforms: riscv64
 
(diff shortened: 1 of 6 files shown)


.github/workflows/create-wheels.yaml
```diff
@@ -71,12 +71,14 @@ jobs:
       fail-fast: false
 
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```
zzzeek: oh.  does this mean we cant use dependabot?   that would be a big deal?

pyproject.toml
```diff
@@ -199,6 +199,10 @@ mypy = [
     "types-greenlet",
 ]
 
+zizmor = [
+    "zizmor>=1.22.0",
```
zzzeek: if the point of zizmor is to pin build tools to exact hashes then why aren't we pinning to an exact verison here?

noxfile.py
```diff
@@ -17,6 +17,8 @@
     nox.needs_version = ">=2025.10.16"
 
 nox.options.default_venv_backend = "venv"
+nox.options.reuse_venv = "yes"
```
zzzeek: err how come?   IIUC this means we're going to be sitting on older versions of things all the time

zzzeek (changes requested):
the requirement, if there is one, to pin GH actions to hashes rather than versions seems very strange.    overall there's a lot I dont understand here.

.github/workflows/create-wheels.yaml
CaselIT: dunno. would have to look

noxfile.py
CaselIT: i think it will re-install everything that's specified, but not update the dependencies if not needed. but that was something I did locally, it's not needed here

CaselIT:
Yeah I agree that it doesn't seems to add much value. I propose we remove everything but
- the change in `.github/dependabot.yml`
- the `with: persist-credentials: false`

zzzeek:
ok

vytas7:
> the requirement, if there is one, to pin GH actions to hashes rather than versions seems very strange. overall there's a lot I dont understand here.

@zzzeek As I understand the idea is that GH Actions are not published to an immutable store like PyPI; if an attacker manages to take over the repo, the version (Git tag) in question can be replaced, and you would get a compromised script for the same version. Crafting a hash collision for the pinned treeish is much harder.

PR #13532, opened 2026-08-24
Fix __allow_unmapped__ not honored for Column/MappedColumn on mixins

### LINKED ISSUE #9369: allow_unmapped seems to not be honored in all cases
### Discussed in https://github.com/sqlalchemy/sqlalchemy/discussions/9368

```py
from typing import List
from sqlalchemy import Column, ForeignKey
from sqlalchemy.types import Integer


# qualify the base with __allow_unmapped__.  Can also be
# applied to classes directly if preferred
class Base:
    __allow_unmapped__ = True

    # raises, but shouldn't (right?)
    id: int = Column(Integer, primary_key=True)


Base = declarative_base(cls=Base)


# existing mapping proceeds, Declarative will ignore any annotations
# which don't include ``Mapped[]``
class Foo(Base):
    __tablename__ = "foo"

    bars: List["Bar"] = relationship("Bar", back_populates="foo")


class Bar(Base):
    __tablename__ = "bar"
    foo_id = Column(ForeignKey("foo.id"))

    foo: Foo = relationship(Foo, back_populates="bars", cascade="all")
```

Faseeh06:
Fixes #9369

`_produce_column_copies()`, used when a `Column` or `MappedColumn` is inherited from a mixin/non-declarative base and copied onto the class being mapped, calls `_collect_annotation()` with `expect_mapped` hardcoded to `True` at both of its call sites. This bypasses `self.allow_unmapped_annotations` entirely, so a bare (non-`Mapped[]`) type-annotated attribute paired with such a column — inherited from a mixin — raises `MappedAnnotationError` even when `__allow_unmapped__ = True` is set on the mapped class or a superclass, despite the error's own text claiming that flag should suppress it:

```python
class Base(DeclarativeBase):
    __allow_unmapped__ = True

class Mixin:
    id: int = Column(Integer, primary_key=True)

class MyModel(Mixin, Base):
    __tablename__ = "my_model"
    # currently raises MappedAnnotationError despite __allow_unmapped__
```

By contrast, the equivalent path for attributes declared directly on the mapped class passes `expect_mapped=None`, which correctly defers to the flag via the dynamic computation already present in `_collect_annotation`.

## Changes
- Pass `not self.allow_unmapped_annotations` instead of a hardcoded `True` at both `_produce_column_copies()` call sites, preserving the strict default while honoring the flag, matching the direct-attribute path's behavior.
- Added a regression test (`test_allow_unmapped_column_on_mixin`) alongside the existing `test_allow_unmapped_on_mixin`/`test_allow_unmapped_cols` tests.
- Added a changelog entry.

## Test plan
- [x] Reproduced the exact repro from the issue against current `main`, confirmed fixed
- [x] Confirmed the strict default (no `__allow_unmapped__`) still correctly raises `MappedAnnotationError` for the same mixin shape — no over-broadening
- [x] New test fails on unmodified `main` and passes with the fix
- [x] `pytest test/orm/declarative/` — 2852 passed
- [x] `pytest test/orm/` (full ORM suite) — 10937 passed
- [x] `flake8` / `black --check` clean on both changed files

### DIFF
--- a/lib/sqlalchemy/orm/decl_base.py
+++ b/lib/sqlalchemy/orm/decl_base.py
@@ -1430,7 +1430,11 @@ class _DeclarativeMapperConfig(_MapperConfig, _ClassScanAbstractConfig):
                     continue
 
                 collected_annotation = self._collect_annotation(
-                    name, annotation, originating_class, True, obj
+                    name,
+                    annotation,
+                    originating_class,
+                    not self.allow_unmapped_annotations,
+                    obj,
                 )
                 obj = (
                     collected_annotation.attr_value
@@ -1452,7 +1456,11 @@ class _DeclarativeMapperConfig(_MapperConfig, _ClassScanAbstractConfig):
                     continue
 
                 collected_annotation = self._collect_annotation(
-                    name, annotation, originating_class, True, obj
+                    name,
+                    annotation,
+                    originating_class,
+                    not self.allow_unmapped_annotations,
+                    obj,
                 )
                 obj = (
                     collected_annotation.attr_value
--- a/test/orm/declarative/test_typed_mapping.py
+++ b/test/orm/declarative/test_typed_mapping.py
@@ -490,6 +490,23 @@ class MappedColumnTest(fixtures.TestBase, testing.AssertsCompiledSQL):
 
             status: int
 
+    def test_allow_unmapped_column_on_mixin(self, decl_base):
+        """test #9369"""
+
+        class Mixin:
+            # a bare (non-``Mapped[]``) annotation paired with a plain
+            # ``Column``, inherited from a mixin.  this goes through
+            # _produce_column_copies(), which previously hardcoded
+            # ``expect_mapped=True`` and ignored __allow_unmapped__.
+            id: int = Column(Integer, primary_key=True)
+            status: int = Column(Integer)
+
+        class MyClass(Mixin, decl_base):
+            __tablename__ = "mytable"
+            __allow_unmapped__ = True
+
+        eq_(MyClass.__table__.c.keys(), ["id", "status"])
+
     def test_allow_unmapped_on_base(self):
         class Base(DeclarativeBase):
             __allow_unmapped__ = True
--- a/doc/build/changelog/unreleased_21/9369.rst
+++ b/doc/build/changelog/unreleased_21/9369.rst
@@ -1,0 +1,11 @@
+.. change::
+    :tags: bug, orm
+    :tickets: 9369
+
+    Fixed bug in Declarative where ``__allow_unmapped__`` was not honored for
+    a :class:`_schema.Column` or :class:`_orm.MappedColumn` object inherited
+    from a mixin class and paired with a non-``Mapped[]`` (bare) type
+    annotation. Such an annotation would unconditionally raise
+    ``MappedAnnotationError`` even when ``__allow_unmapped__ = True`` was set
+    on the mapped class or a superclass, since the annotation-collection path
+    used for mixin attributes did not consult the flag.


sqla-tester:

This pull request references issue #9369, which hasn't been marked **open for pull requests**, so the change isn't authorized yet and I'm closing it automatically. Once a maintainer adds the label to #9369, reopen this pull request and it will stay open. If #9369 needs more detail before that can happen, a complete runnable example is usually the missing piece, and adding one to the issue is the fastest way to get there.

Mukller:
Traced this through both functions to be sure, and the two-line fix is semantically exact:

- With `expect_mapped=False` (i.e. `__allow_unmapped__ = True`), a **bare** annotation takes the `return annotated, None` branch in `_extract_mapped_subtype` (`orm/util.py`) instead of raising — the column then copies over with its plain type, exactly what #9369 asks for.
- A **`Mapped[...]`** annotation on a mixin-inherited `Column`/`MappedColumn` is unaffected: the raise-block only triggers when the annotation's `__origin__` isn't `_MappedAnnotationBase`, so extraction proceeds identically under either `expect_mapped` value.
- With the flag off, `expect_mapped=True` reproduces the old behavior bit-for-bit, so non-mixin / strict-mode paths see no change.

The regression test targets the right path (mixin + explicit `Column` + bare annotation goes through `_produce_column_copies`, unlike the adjacent non-Column test), and asserting on `__table__.c.keys()` verifies both attributes survived rather than just absence-of-exception.

Nothing else to flag — clean fix for a long-standing gap.

lib/sqlalchemy/orm/decl_base.py
```diff
@@ -1430,7 +1430,11 @@ def _produce_column_copies(
                     continue
 
                 collected_annotation = self._collect_annotation(
-                    name, annotation, originating_class, True, obj
+                    name,
+                    annotation,
+                    originating_class,
+                    not self.allow_unmapped_annotations,
```
CaselIT, line 1436: I think this and below could instead pass None.

@zzzeek thoughts?

zzzeek: im not following.  "None" for which field?   the True that's changed here to use allow_unmapped_annotations is the "expect_mapped" boolean.   by setting it to False, an error is not raised.    i havent looked at this code in a long time but what are you considering exactly?

test/orm/declarative/test_typed_mapping.py
```diff
@@ -490,6 +490,23 @@ class MyClass(AllowsUnmapped, decl_base):
 
             status: int
 
+    def test_allow_unmapped_column_on_mixin(self, decl_base):
+        """test #9369"""
+
+        class Mixin:
```
zzzeek, line 496: this is not what the test case in the issue says....the test case has the `__allow_unmapped__` on the Mixin.

in fact we should test it in all combinations:  on just the mixin, on just the concrete class, on both.  use testing.variations

test/orm/declarative/test_typed_mapping.py
```diff
@@ -490,6 +490,23 @@ class MyClass(AllowsUnmapped, decl_base):
...
+            # a bare (non-``Mapped[]``) annotation paired with a plain
+            # ``Column``, inherited from a mixin.  this goes through
+            # _produce_column_copies(), which previously hardcoded
+            # ``expect_mapped=True`` and ignored __allow_unmapped__.
+            id: int = Column(Integer, primary_key=True)
+            status: int = Column(Integer)
+
+        class MyClass(Mixin, decl_base):
+            __tablename__ = "mytable"
+            __allow_unmapped__ = True
```
zzzeek, line 506: this file will also need to be copied using tools/sync_test_files.py

zzzeek (changes requested):

lib/sqlalchemy/orm/decl_base.py
CaselIT: in other places for such a case we pass none not False

zzzeek: he's correct.  use None please, instead of True

also we need to document what "allow_unmapped" means, determine if False means anything and if not we need to change this parameter

zzzeek: what im getting at is this is a bigger change that warrants a lot of care and caution.  it's not a one liner.  especially in the tests.  needs a lot more and they need to be integrated with what we have already.

Faseeh06: @zzzeek Should I check it at broader level and then let you know ?

zzzeek: if you want, sure

Issue #5919, opened 2021-02-04
primary_key=True, in combination with naming_convention, can result in a blank name attribute on the constraint object.


tcatut:
**Describe the bug**
We have several declarative_base, and associated Table classes, which are failing on `__table__.create()`.  Example class definition and traceback will be below.

I was able to examine the traceback and add a print statement at line 3238 in `sqlalchemy/sql/compiler.py`.  From this I was able to see that `constraint.name` passes the None check, but is being created as a blank string.  This results in the traceback below.

I did try a few things and found two work arounds, but I feel like these shouldn't be needed.

The first one is to explicitly declare `PrimaryKeyConstraint(idx)` into the class definition, instead of relying on `primary_key=True` in the Column definition.  If the constraint object is explicitly declared in the class, then it gets a proper name attribute and works without exception.

The other work around is to remove our naming_convention definition.  It appears this exception ONLY happens when using both `primary_key=True` and naming_convention together.  Obviously, for us, removing the naming_convention isn't an option.

**Expected behavior**
It's my understanding that `primary_key=True` is supposed to form a complete PrimaryKeyConstraint object automatically, and we should not have to explicitly declare this in every table class.

If nothing else, you should probably have some blank string protection here.  It would be nice to get a more descriptive error message.

**To Reproduce**
```
import sqlalchemy
from sqlalchemy import Column, String, TIMESTAMP, text, MetaData, PrimaryKeyConstraint
from sqlalchemy.dialects.mysql import BIGINT
from sqlalchemy.ext.declarative import declarative_base

naming_convention={
    'ix': '%(column_0N_name)s',
    'uq': '%(column_0N_name)s',
    'ck': '%(column_0N_name)s',
    'fk': 'fk_%(table_name)s_%(referred_table_name)s',
    'pk': '%(column_0N_name)s',
}

ADS_EQUIP_DATA_base = declarative_base(metadata=MetaData(naming_convention=naming_convention))
metadata = ADS_EQUIP_DATA_base.metadata


class ADS_EQUIP_DATA(ADS_EQUIP_DATA_base):
    __tablename__ = 'ADS_EQUIP_DATA'
    __table_args__ = {'schema': 'ADS_EQUIP_CNTL', 'mysql_engine': 'InnoDB'}

    deviceId = Column(String(32), index=True)
    input_column = Column('input', String(32))
    generalType = Column(String(32), index=True)
    category = Column(String(32))
    station = Column(String(32))
[...]

PR #13536, opened 2026-08-24
Handle an empty identifier in IdentifierPreparer.quote()

### LINKED ISSUE #13535: IdentifierPreparer.quote() raises IndexError on an empty identifier, reachable through reflection
### Describe the bug

`IdentifierPreparer.quote("")` raises `IndexError: string index out of range` from `_requires_quotes`, which reads `value[0]` without checking for an empty string:

```python
def _requires_quotes(self, value: str) -> bool:
    lc_value = value.lower()
    return (
        lc_value in self.reserved_words
        or value[0] in self.illegal_initial_characters   # <-- IndexError when value == ""
        ...
```

The reason this is worth reporting rather than dismissing as "don't use an empty name" is that **reflection can hand you one**. SQLite accepts `CREATE TABLE "" (...)`, `MetaData.reflect()` reflects it happily, and only when the reflected table is *used* does the compiler raise a bare `IndexError` with no indication of what was wrong.

It also looks like the same failure that was reported in #5919, where a naming convention produced a blank constraint name. That was fixed on the producing side; the preparer still raises `IndexError` for any other route to a blank identifier.

`Column("")` is already rejected with a clear message — *"Column must be constructed with a non-blank name or assign a non-blank .name"* — so blank identifiers are handled deliberately in at least one place. `Table("")` is accepted and then fails on use.

### To Reproduce

```python
from sqlalchemy import MetaData, create_engine, pool, select

e = create_engine("sqlite://", poolclass=pool.StaticPool)
with e.begin() as c:
    c.exec_driver_sql('CREATE TABLE "" (c INTEGER)')

md = MetaData()
md.reflect(bind=e)          # succeeds, table is reflected under the name ""
print(sorted(md.tables))    # ['']

select(md.tables[""]).compile(e)
```

Direct routes to the same line:

```python
e.dialect.identifier_preparer.quote("")                 # IndexError
Table("", MetaData(), Column("c", Integer)).create(e)   # IndexError
```

### Error

```
  File "lib/sqlalchemy/sql/compiler.py", line 8096, in _requires_quotes
    or value[0] in self.illegal_initial_characters
       ~~~~~^^^
IndexError: string index out of range
```

### Versions

- OS: Windows 11
- Python: 3.14.6
- SQLAlchemy: 2.1.0b4 (main)
- Database: SQLite / pysqlite 3.50.4

### Additional context

Two ways to resolve it, and the choice is yours rather than mine:

1. Treat an empty identifier as always requiring quotes, so `quote("")` returns `""`. That is what the database itself holds, and it makes the reflect-then-query round trip work: `SELECT "".c FROM ""`. I have this working locally against the reproduction above, with `test/sql` green (8306 passed).
2. Refuse a blank name at `Table` construction, matching what `Column` already does. That leaves the reflected table unusable, so it would need reflection to skip or rename it.

Happy to send a PR for whichever you prefer — I have (1) ready with a test.

**Have a nice day!**

hamodywe:
Fixes #13535.

### The bug

`_requires_quotes` tests the initial character of an identifier without first checking that there is one:

```python
or value[0] in self.illegal_initial_characters
```

so `IdentifierPreparer.quote("")` raised `IndexError: string index out of range` from inside the compiler — not a result, and not an error anyone could act on.

The reason this is worth more than "don't use a blank name" is that **reflection can hand you one**. SQLite accepts `CREATE TABLE "" (...)`, `MetaData.reflect()` reflects it under the name `""`, and only when the reflected table is used does the `IndexError` appear:

```python
with e.begin() as c:
    c.exec_driver_sql('CREATE TABLE "" (c INTEGER)')

md = MetaData()
md.reflect(bind=e)        # fine — md.tables has ''
select(md.tables[""]).compile(e)   # IndexError
```

This is the same failure reported in #5919, where a naming convention produced a blank constraint name. That was fixed on the producing side; the preparer itself still indexed past the end of the string, so any other route to a blank identifier reached it again.

### The change

An empty identifier has no unquoted form, so it is treated as requiring quotes and renders as `""` — the form the database itself holds. The reflect-then-query round trip then works: `SELECT "".c FROM ""`.

Four lines, and no behaviour changes for a non-empty identifier: the early return only fires on `""`, which previously could not reach the end of the function at all.

### Why quoting rather than refusing

`Column("")` already refuses a blank name — *"Column must be constructed with a non-blank name or assign a non-blank .name"* — so there is a stated policy in one place, and refusing at `Table` construction would match it. I did not do that here, because it would leave a reflected table permanently unusable and would need reflection to skip or rename it, which is a design decision rather than a bug fix. This PR only stops the preparer from reading past the end of the string; if you would rather blank names were rejected up front instead, say so on #13535 and I will redo it that way.

### Tests

`PreparerTest.test_quote_empty_identifier` in `test/sql/test_quote.py`.

- **Revert-checked**: with `compiler.py` restored to `main` and the test kept, exactly that one test fails, with the `IndexError` above; the other 65 pass.
- `pytest test/sql` — **8307 passed, 354 skipped**, against 8306 passed on an unmodified tree (the one added test).
- Verified end to end on the reproduction from the issue: reflection, compile, execute and `CreateTable` all succeed.

Tested on SQLite / pysqlite 3.50.4, Python 3.14.6, Windows.

---

*Written with assistance from Claude (Opus 5); the diagnosis, the fix and the results above were run and checked by me.*

### DIFF
--- a/doc/build/changelog/unreleased_21/13535.rst
+++ b/doc/build/changelog/unreleased_21/13535.rst
@@ -1,0 +1,10 @@
+.. change::
+    :tags: bug, sql
+    :tickets: 13535
+
+    Fixed issue where :meth:`.IdentifierPreparer.quote` would raise
+    ``IndexError`` when given an empty string, as the check for an illegal
+    initial character indexed the value without testing its length.  An empty
+    identifier is now treated as requiring quotes, which is the form the
+    database itself holds; this is reachable through reflection, as SQLite
+    accepts a table created with an empty name.
--- a/lib/sqlalchemy/sql/compiler.py
+++ b/lib/sqlalchemy/sql/compiler.py
@@ -8090,6 +8090,10 @@ class IdentifierPreparer:
 
     def _requires_quotes(self, value: str) -> bool:
         """Return True if the given identifier requires quoting."""
+        if not value:
+            # an empty identifier has no unquoted form; it also has no first
+            # character, which the illegal-initial test below would index
+            return True
         lc_value = value.lower()
         return (
             lc_value in self.reserved_words
--- a/test/sql/test_quote.py
+++ b/test/sql/test_quote.py
@@ -20,6 +20,7 @@ from sqlalchemy.testing import AssertsCompiledSQL
 from sqlalchemy.testing import eq_
 from sqlalchemy.testing import fixtures
 from sqlalchemy.testing import is_
+from sqlalchemy.testing import is_true
 from sqlalchemy.testing.util import picklers
 
 
@@ -874,6 +875,18 @@ class QuoteTest(fixtures.TestBase, AssertsCompiledSQL):
 
 class PreparerTest(fixtures.TestBase):
     """Test the db-agnostic quoting services of IdentifierPreparer."""
+
+    def test_quote_empty_identifier(self):
+        """an empty identifier has no unquoted form, and no first character
+        for the illegal-initial-character test to look at.
+
+        reflection can produce one: SQLite accepts CREATE TABLE "" (...).
+
+        """
+        prep = compiler.IdentifierPreparer(default.DefaultDialect())
+
+        eq_(prep.quote(""), '""')
+        is_true(prep._requires_quotes(""))
 
     def test_unformat(self):
         prep = compiler.IdentifierPreparer(default.DefaultDialect())


sqla-tester:

This pull request references issue #13535, which hasn't been marked **open for pull requests**, so the change isn't authorized yet and I'm closing it automatically. Once a maintainer adds the label to #13535, reopen this pull request and it will stay open. If #13535 needs more detail before that can happen, a complete runnable example is usually the missing piece, and adding one to the issue is the fastest way to get there.

PR #13544, opened 2026-08-28
fix: avoid catastrophic backtracking in SQLite FK reflection regex

### LINKED ISSUE #13530: catastrophic backtracking in sqlite foreign key reflection regex
while reflecting foreign keys on sqlite, `get_foreign_keys` scans the stored
`CREATE TABLE` text (taken verbatim from `sqlite_master.sql`) with
`FK_PATTERN` to recover constraint names and options. the referred-columns
part of that pattern is
`((?:(?:"[^"]+"|[a-z0-9_]+)\s*(?:,\s*)?)+)`, which repeats a bare column
token `[a-z0-9_]+` while the separator `(?:,\s*)?` between iterations is
optional. that is a nested quantifier of the `(a+)+` shape, so a run of word
characters that is not eventually followed by a closing `)` can be split
across the iterations in exponentially many ways, and the engine tries them
all before giving up.

sqlite keeps the original text of a `CREATE TABLE` statement in
`sqlite_master`, and the pattern is applied with `re.finditer` over the whole
statement, so the run does not need to be an actual foreign key. any account
that can create a table can leave such a run inside e.g. a column `DEFAULT`
string literal, and later reflection of that ordinary table hangs.

a small reproduction:

```python
import time
from sqlalchemy import create_engine, inspect, pool

e = create_engine("sqlite://", poolclass=pool.StaticPool)
with e.begin() as c:
    c.exec_driver_sql(
        "CREATE TABLE t (id INTEGER, note TEXT DEFAULT "
        "'FOREIGN KEY (a) REFERENCES b(" + ("a" * 26) + "')"
    )

start = time.perf_counter()
inspect(e).get_foreign_keys("t")
print("get_foreign_keys took %.2fs" % (time.perf_counter() - start))
```

timings scale exponentially with the length of the run (measured here):
`n=18` 0.04s, `n=22` 0.56s, `n=26` 9.0s, and a couple more characters make it
effectively hang.

the run can be made unambiguous by requiring a non-empty separator (a comma
clause or whitespace) between column tokens, so a single word run only parses
one way; valid foreign key clauses reflect exactly as before. happy to send a
PR with a regression test if that's useful.

TrueFurina:
Fixes #13530

**Problem**: the referred-columns part of FK_PATTERN is a nested quantifier of the (a+)+ shape, because the separator between iterations is optional. A run of word characters that is not eventually matched causes catastrophic backtracking (ReDoS) when reflecting SQLite foreign keys.

**Fix**: make the comma separator mandatory between referred columns, so matching is linear — each subsequent column must be preceded by a comma, and there is no nested quantifier to backtrack into.

### DIFF
--- a/lib/sqlalchemy/dialects/sqlite/base.py
+++ b/lib/sqlalchemy/dialects/sqlite/base.py
@@ -2717,7 +2717,7 @@ class SQLiteDialect(default.DefaultDialect):
             FK_PATTERN = (
                 r'(?:CONSTRAINT\s+(?:"(.+?)"|(\w+))\s+)?'
                 r"FOREIGN\s+KEY\s*\(\s*(.+?)\s*\)\s+"
-                r'REFERENCES\s+(?:(?:"(.+?)")|([a-z0-9_]+))\s*\(\s*((?:(?:"[^"]+"|[a-z0-9_]+)\s*(?:,\s*)?)+)\)\s*'  # noqa: E501
+                r'REFERENCES\s+(?:(?:"(.+?)")|([a-z0-9_]+))\s*\(\s*((?:"[^"]+"|[a-z0-9_]+)(?:\s*,\s*(?:"[^"]+"|[a-z0-9_]+))*)\)\s*'  # noqa: E501
                 r"((?:ON\s+(?:DELETE|UPDATE)\s+"
                 r"(?:SET\s+NULL|SET\s+DEFAULT|CASCADE|RESTRICT|"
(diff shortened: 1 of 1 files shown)


sqla-tester:

This pull request references issue #13530, which is already marked **code review in progress** -- someone is working on it, either in another pull request or directly in gerrit -- so I'm closing this one automatically to keep the two from landing on top of each other. If you think your approach is better than the one in review, the place to make that case is on #13530.

PR #13545, opened 2026-08-28
fix: surface STRICT and WITHOUT ROWID in SQLite reflection

### LINKED ISSUE #13543: SQLite reflection drops STRICT and WITHOUT ROWID; Table.kwargs is empty after autoload
### Describe the bug

Reflecting a SQLite table created with `STRICT` and/or `WITHOUT ROWID` produces a `Table` whose `.kwargs` does not carry `sqlite_strict`/`sqlite_with_rowid` at all -- nothing in reflection surfaces either modifier back onto the `Table` object, even though the dialect can render both (`SQLiteDDLCompiler` appends `"WITHOUT ROWID"` / `"STRICT"` to `table_options` when compiling `CREATE TABLE`) and even *parses past* them during column reflection.

`Dialect.get_table_options()` exists for exactly this ("Return a dictionary of options specified when the table of the given name was created"), but SQLite's dialect doesn't implement it, so it falls through to the base `Dialect.get_table_options` and raises `NotImplementedError`.

Separately, `SQLiteDialect.get_columns()` already parses the raw `CREATE TABLE` text from `sqlite_master` for generated-column handling, and its own regex explicitly matches (and discards) the trailing modifier:

https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/dialects/sqlite/base.py#L2417-L2426

```python
match = re.match(
    (
        r"create table .*?\((.*)\)"
        r"(?:\s*,?\s*(?:WITHOUT\s+ROWID|STRICT))*$"
    ),
    tablesql.strip(),
    re.DOTALL | re.IGNORECASE,
)
assert match, f"create table not found in {tablesql}"
tablesql = match.group(1).strip()
```

Only `match.group(1)` (the column list) is kept; whether `WITHOUT ROWID`/`STRICT` was actually present is never captured or returned to the caller. Round-tripping a table through reflection silently drops both modifiers -- and any downstream tool that recreates a table from its reflected form (Alembic's batch mode being the concrete case that surfaced this) reproduces the table *without* them.

**Why this matters beyond metadata inspection:** Alembic's batch-mode "recreate" (used to work around SQLite's limited `ALTER TABLE`) builds the new table from `Table.kwargs` on the reflected table (`alembic/operations/batch.py`, `ApplyBatchImpl._grab_table_elements`/`_transfer_elements_to_new_table`). Since reflection never populates `sqlite_strict`/`sqlite_with_rowid`, any batch operation on a `STRICT`/`WITHOUT ROWID` table silently drops both keywords -- reported against Alembic as sqlalchemy/alembic#1758, but the root cause is this reflection gap, not anything in Alembic's own batch implementation.

sqlalchemy/alembic#1758

### SQLAlchemy Version in Use

2.0.44 (also reproduces on `main`)

pysqlite (stdlib sqlite3)

SQLite 3.45+ (any version supporting `STRICT`/`WITHOUT ROWID`, i.e. 3.37+)

### Python Version

3.11

### Operating system

macOS (reproduces regardless of platform -- this is pure SQL text parsing, no OS-specific behavior)

### To Reproduce

```python
import sqlalchemy as sa

e = sa.create_engine("sqlite://")
with e.begin() as conn:
    conn.exec_driver_sql(
        'CREATE TABLE "t" (id INTEGER NOT NULL, PRIMARY KEY (id)) '
        'WITHOUT ROWID, STRICT'
    )

meta = sa.MetaData()
with e.begin() as conn:
    reflected = sa.Table("t", meta, autoload_with=conn)
    print(dict(reflected.kwargs))

    insp = sa.inspect(conn)
    try:
        print(insp.get_table_options("t"))
    except NotImplementedError:
        print("get_table_options raised NotImplementedError")
```

### Error

No exception is raised for the core bug -- this is silent data loss, not a crash. Output of the reproduction above:

```
{}
get_table_options raised NotImplementedError
```

Expected `dict(reflected.kwargs)` to be `{'sqlite_with_rowid': False, 'sqlite_strict': True}` (or equivalent), matching what one would pass when *creating* such a table.

### Additional context

Minimal, dependency-free verification that reflection alone (no Alembic involved) drops the information; Alembic's own issue has the full downstream repro showing the recreated table's `CREATE TABLE` losing both keywords after a batch operation.

TrueFurina:
Fixes #13543

**Problem**: reflecting a SQLite table created with `STRICT` and/or `WITHOUT ROWID` produces a `Table` whose `.kwargs` does not carry `sqlite_strict`/`sqlite_with_rowid` — the dialect renders both when compiling CREATE TABLE but reflection never surfaces them.

**Fix**: implement `get_table_options` on the SQLite dialect, parsing the CREATE TABLE text for `STRICT` and `WITHOUT ROWID` and returning them as `sqlite_strict` / `sqlite_with_rowid`, mirroring the MySQL/MSSQL dialect pattern.

### DIFF
--- a/lib/sqlalchemy/dialects/sqlite/base.py
+++ b/lib/sqlalchemy/dialects/sqlite/base.py
@@ -2443,6 +2443,19 @@ class SQLiteDialect(default.DefaultDialect):
             )
 
     @reflection.cache
+    def get_table_options(
+        self, connection, table_name, schema=None, **kw
+    ) -> dict[str, Any]:
+        """Return the table options for the given table."""
+        options: dict[str, Any] = {}
+        tablesql = self._get_table_sql(connection, table_name, schema, **kw)
+        if re.search(r"WITHOUT\s+ROWID", tablesql, re.IGNORECASE):
+            options["sqlite_with_rowid"] = False
+        if re.search(r"STRICT", tablesql, re.IGNORECASE):
+            options["sqlite_strict"] = True
+        return options
+
+    @reflection.cache
     def get_columns(self, connection, table_name, schema=None, **kw):
         pragma = "table_info"
         # computed columns are threaded as hidden, they require table_xinfo


sqla-tester:

This pull request references issue #13543, which hasn't been marked **open for pull requests**, so the change isn't authorized yet and I'm closing it automatically. Once a maintainer adds the label to #13543, reopen this pull request and it will stay open. If #13543 needs more detail before that can happen, a complete runnable example is usually the missing piece, and adding one to the issue is the fastest way to get there.

PR #13546, opened 2026-08-28
fix: escape underscore in Postgres get_schema_names LIKE filter

### LINKED ISSUE #13472: get_schema_names() silently drops Postgres schemas that start with pg (not just pg_)

TrueFurina:
Fixes #13472

**Problem**: get_schema_names() on Postgres filters system schemas with NOT LIKE 'pg_%', but _ is a single-character wildcard in SQL LIKE — so any user schema starting with pg plus one more character (e.g. pgx) is silently dropped, not just the literal pg_ prefix used by real system schemas.

**Fix**: escape the underscore (NOT LIKE 'pg\_%') so the filter matches the literal pg_ prefix only.

### DIFF
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -4066,7 +4066,7 @@ class PGDialect(default._BackendsMultiReflection, default.DefaultDialect):
     def get_schema_names(self, connection, **kw):
         query = (
             select(pg_catalog.pg_namespace.c.nspname)
-            .where(pg_catalog.pg_namespace.c.nspname.not_like("pg_%"))
+            .where(pg_catalog.pg_namespace.c.nspname.not_like("pg\_%"))
             .order_by(pg_catalog.pg_namespace.c.nspname)
         )
         return connection.scalars(query).all()


sqla-tester:

This pull request references issue #13472, which is already marked **code review in progress** -- someone is working on it, either in another pull request or directly in gerrit -- so I'm closing this one automatically to keep the two from landing on top of each other. If you think your approach is better than the one in review, the place to make that case is on #13472.

PR #13565, opened 2026-09-04
Release disposed mappers from declared-events listeners and sync-target tracking

### LINKED ISSUE #13564: Disposed declarative classes are kept alive by _track_overlapping_sync_targets and by the __declare_last__ "after_configured" listener
### Describe the bug

I rebuild ORM classes at runtime (tenant-defined tables, so a class gets re-created when its schema changes) and noticed the process grows by several MB every time, even though the old registry/mappers are disposed and nothing of mine references the old classes any more. Tracking it down with `gc.get_referrers` ended in two places inside SQLAlchemy that hold on to mapped classes for the whole process lifetime:

1. **`JoinCondition._track_overlapping_sync_targets`** (`_JoinCondition` on main). It is a class-level `WeakKeyDictionary` that only exists for the "relationship X will copy column Q to column P, which conflicts with ..." warning. The keys (`to_` columns) are weak, but the values are plain `Column` objects (`from_`). As soon as the `from_` column can reach the `to_` column through strong references, the weak key is kept alive by the dictionary's own value and the entry never expires. That happens for example when an `Index` built from an ORM attribute (`Index("ix", Parent.id.desc())`) is attached to the table: the expression is annotated with the mapper, so Table → mapper → registry → the other classes → the `to_` column. In my case every rebuilt class left ~1000 `Column` objects, the mapper and the class behind, roughly 7–9 MB per rebuild per process.

2. **`__declare_last__` / `__declare_first__`**. `_ClassScanMapperConfig._setup_declared_events()` registers a closure on `Mapper` `"after_configured"` for every class that defines the hook. The closure holds the config object (and through it the class), it is never removed — `registry.dispose()` doesn't know about it — and the listener list on `Mapper` grows by one per class ever mapped and is run on every `configure_mappers()`.

Reproduced on 1.4.54, 2.0.52 and current main.

### To Reproduce

```python
import gc, weakref
import sqlalchemy as sa
from sqlalchemy.orm import registry, relationship, configure_mappers, Mapper
try:
    from sqlalchemy.orm.relationships import JoinCondition
except ImportError:  # main
    from sqlalchemy.orm.relationships import _JoinCondition as JoinCondition


def build(orm_index=False, declare_last=False):
    reg = registry()
    Base = reg.generate_base()

    class Parent(Base):
        __tablename__ = "parent"
        id = sa.Column(sa.Integer, primary_key=True)
        if declare_last:
            @classmethod
            def __declare_last__(cls):
                pass

    class Child(Base):
        __tablename__ = "child"
        id = sa.Column(sa.Integer, primary_key=True)
        parent_id = sa.Column(sa.ForeignKey("parent.id"))
        parent = relationship(Parent)

    if orm_index:
        sa.Index("ix_parent_id_desc", Parent.id.desc())

    configure_mappers()
    refs = (weakref.ref(Parent), weakref.ref(Child), weakref.ref(Parent.__table__))
    reg.dispose()
    return refs


def listeners():
    return sum(len(c) for c in Mapper.dispatch.after_configured._clslevel.values())


print(sa.__version__)
for kwargs in ({}, {"orm_index": True}, {"declare_last": True}):
    alive = []
    for _ in range(5):
        refs = build(**kwargs)
        gc.collect(); gc.collect()
        alive.append(tuple(r() is not None for r in refs))
    print(kwargs, "Parent/Child/Table alive:", alive,
          "| sync_targets:", len(JoinCondition._track_overlapping_sync_targets),
          "| after_configured listeners:", listeners())
    if kwargs.get("orm_index"):
        JoinCondition._track_overlapping_sync_targets.clear(); gc.collect()
        print("   after clearing _track_overlapping_sync_targets -> Parent alive:", refs[0]() is not None)
```

### Error

Output is the same on 1.4.54, 2.0.52 and main:

```
{}                      Parent/Child/Table alive: [(False, False, False), ...]  | sync_targets: 0 | after_configured listeners: 0
{'orm_index': True}     Parent/Child/Table alive: [(True, True, True), ...]     | sync_targets: 5 | after_configured listeners: 0
   after clearing _track_overlapping_sync_targets -> Parent alive: False
{'declare_last': True}  Parent/Child/Table alive: [(True, False, True), ...]    | sync_targets: 5 | after_configured listeners: 5
```

Expected: after `registry.dispose()` (and dropping my own references) the classes, mappers and tables are collectable, and nothing on the `Mapper` class grows with the number of registries ever created.

### Versions

- OS: Linux
- Python: 3.10 / 3.11
- SQLAlchemy: 1.4.54, 2.0.52, main
- Database: PostgreSQL (not needed for the reproducer)
- DBAPI: psycopg2

### Additional context

I have a patch that makes both structures hold the class/column weakly and removes the declared-events listeners when the class is unmapped (`registry._dispose_manager_and_mapper`); the reproducer above prints all `False` with it and the ORM test suites I ran still pass. I'll open a PR referencing this issue.

For anyone hitting this before a release: clearing the mapper's entries from `_track_overlapping_sync_targets` and removing the `after_configured` listener whose closure holds the class, at the point where the class is disposed, works around it.

chervachek1:
### Description

Fixes #13564.

Two process-wide structures kept disposed mappers (and everything reachable from them) alive:

- `_JoinCondition._track_overlapping_sync_targets` stored the `from_` column strongly. Whenever that column could reach the weak `to_` key (e.g. via an `Index` built from an ORM attribute, whose expression is annotated with the mapper), the entry never expired and pinned the whole mapping. The inner dictionary now stores `weakref.ref(from_)`; dead references are skipped when the conflict warning is evaluated.
- The `after_configured` / `before_configured` listeners created for `__declare_last__` / `__declare_first__` captured the `_ClassScanMapperConfig` (and the class) and were never removed. They now hold the class through a weak reference and are recorded per class so that `registry._dispose_manager_and_mapper()` (used by `registry.dispose()` and `clear_mappers()`) removes them.

Two tests added (`predictable_gc`): a class with either hook is collected after `registry.dispose()` and the listener count on `Mapper` returns to its previous value; a disposed registry with an ORM-attribute `Index` is collected.

### Checklist

This pull request is:

- [x] A bug fix
  - The changelog entry is in `doc/build/changelog/unreleased_20/`; ticket #13564.

### DIFF
--- a/doc/build/changelog/unreleased_20/13564.rst
+++ b/doc/build/changelog/unreleased_20/13564.rst
@@ -1,0 +1,15 @@
+.. change::
+    :tags: bug, orm
+    :tickets: 13564
+
+    Fixed two process-wide references that kept the classes, mappers and
+    tables of a disposed :class:`_orm.registry` (or of classes removed via
+    :func:`_orm.clear_mappers`) in memory indefinitely.  The
+    ``after_configured`` / ``before_configured`` listeners that declarative
+    registers for ``__declare_last__`` and ``__declare_first__`` now refer to
+    the class weakly and are removed when the class is unmapped, instead of
+    accumulating one per mapped class for the lifetime of the process.  The
+    internal bookkeeping used to warn about relationships writing the same
+    column now holds its columns weakly as well, so that a column expression
+    attached to a table (such as an :class:`.Index` built from an ORM
+    attribute) no longer pins the whole mapping.
--- a/lib/sqlalchemy/orm/decl_api.py
+++ b/lib/sqlalchemy/orm/decl_api.py
@@ -53,6 +53,7 @@ from .decl_base import _DeclarativeMapperConfig
 from .decl_base import _DeferredDeclarativeConfig
 from .decl_base import _del_attribute
 from .decl_base import _ORMClassConfigurator
+from .decl_base import _remove_declared_events
 from .decl_base import MappedClassProtocol
 from .descriptor_props import Composite
 from .descriptor_props import Synonym
@@ -1549,6 +1550,7 @@ class registry(EventTarget):
 
         class_ = manager.class_
         self._dispose_cls(class_)
+        _remove_declared_events(class_)
         instrumentation._instrumentation_factory.unregister(class_)
 
     def generate_base(
--- a/lib/sqlalchemy/orm/decl_base.py
+++ b/lib/sqlalchemy/orm/decl_base.py
@@ -251,6 +251,23 @@ def _check_declared_props_nocascade(
         return False
 
 
+_declared_event_listeners: weakref.WeakKeyDictionary[
+    Type[Any], List[Tuple[str, Callable[[], None]]]
+] = weakref.WeakKeyDictionary()
+"""Mapper-level listeners created for ``__declare_last__`` /
+``__declare_first__``, keyed by the mapped class, so that they can be
+removed when the class is unmapped."""
+
+
+def _remove_declared_events(cls: Type[Any]) -> None:
+    """Remove the Mapper listeners that were registered for ``cls`` by
+    :meth:`._ClassScanMapperConfig._setup_declared_events`."""
+
+    for identifier, fn in _declared_event_listeners.pop(cls, ()):
+        if event.contains(Mapper, identifier, fn):
+            event.remove(Mapper, identifier, fn)
+
+
 class _ORMClassConfigurator:
     """Object that configures a class that's potentially going to be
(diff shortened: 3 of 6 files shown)


sqla-tester:

This pull request references issue #13564, which hasn't been marked **open for pull requests**, so the change isn't authorized yet and I'm closing it automatically. Once a maintainer adds the label to #13564, reopen this pull request and it will stay open. If #13564 needs more detail before that can happen, a complete runnable example is usually the missing piece, and adding one to the issue is the fastest way to get there.

PR #13597, opened 2026-09-19
Document synonym()'s implicit descriptor detection for classical mappings

### LINKED ISSUE #4706: synonym descriptors automatically blend existing properties - undocumented?
While migration from classical to declarative mapping, I've run across this feature in SQLAlchemy which are are using - but I can't find it in the documentation.

Basically, we have a property

```python
    @property
    def foo(self):
        return self._foo

    @foo.setter
    def foo(self, value):
        # Validation code
        self._foo = value
```

Configuring the mapper with `mapper(Something, some_table, properties={"foo": synonym("_foo")})` seems to detect the existing property and merge it into the synonym.

Full example here: https://gist.github.com/tlandschoff-scale/83f6e1691d92236786a824cc79e359d1#file-synonym_with_setter-py

So does this work only by accident or is it just missing from the documentation!?

Greetings, Torsten

hikmetba-bit:
## Summary

Fixes #4706.

`synonym()` silently reuses an existing same-named descriptor (e.g. a `@property`) as its `descriptor` when none is passed explicitly, but this was undocumented anywhere - both the `synonym()` docstring and the [Synonyms](https://docs.sqlalchemy.org/en/latest/orm/mapped_attributes.html#synonyms) doc page only ever showed the explicit `descriptor=...` form.

This is exactly the gap the reporter ran into in #4706 ("does this work only by accident or is it just missing from the documentation?"), and the maintainer's own follow-up comment on the issue confirms it's intentional, points at the exact code (`descriptor_props.py`), and explicitly asks for "extra verbiage... to the actual docstring for `synonym()`". This PR adds that verbiage, plus a matching section on the doc page.

## What was verified, and how

I didn't want to just restate the maintainer's 6-year-old comment without checking it still holds on current `main`, so:

1. Read the actual mechanism: `SynonymProperty.instrument_class` in `lib/sqlalchemy/orm/descriptor_props.py` -  when `self.descriptor is None`, it does `desc = getattr(mapper.class_, self.key, None)` and adopts it if `Mapper._is_userland_descriptor(...)` returns `True` (i.e. it's a plain, non-ORM-mapped class attribute).
2. Installed this branch's `sqlalchemy` from source into a throwaway venv (`pip install -e .`) and wrote two real, executed test scripts (not just reasoning about the code):
   - **Imperative mapping** (`map_imperatively`), matching the reporter's exact scenario: a class with a plain `@property`/`@x.setter` pair, mapped via `properties={"status": synonym("_status")}` with no `descriptor=` passed. Confirmed the *custom* getter/setter logic actually executes (a getter that prefixes the value, a setter that uppercases it) - not just a passthrough mirror.
   - **A naive Declarative-class-body port of the same pattern** - this does *not* pick up the descriptor (custom logic never runs), because `status = synonym(...)` on the last line overwrites the earlier `@property def status` in the class `__dict__` before mapping occurs. This matches the maintainer's explanation in the issue thread of exactly why the declarative case differs.

Both doc additions spell out this classical-vs-declarative distinction explicitly, so nobody copies the pattern into a Declarative class body expecting it to work.

## Verification

Documentation-only change (plus one docstring). Verified by:
- Reading the real implementation (`descriptor_props.py`, `mapper.py::_is_userland_descriptor`).
- Two executed Python scripts against `sqlalchemy` installed from this exact branch, confirming both the working (imperative) and non-working (declarative-body) cases behave as described.
- `python -m py_compile` on the edited docstring file, and a live `import` + `.__doc__` check, to confirm the docstring edit didn't break anything.
- Did not run the full `sqlalchemy` test suite (large, dialect/DB-dependent) or build the Sphinx docs (`zzzeeksphinx` theme not installed in this environment) - the `.rst` syntax was checked by hand against the surrounding, already-working examples on the same page (same directive/code-block/cross-reference patterns).

## Test plan
- [ ] CI: docs build (Sphinx) succeeds
- [ ] Maintainer confirms the classical-vs-declarative distinction is accurately described

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sqla-tester:

This pull request references issue #4706, which hasn't been marked **open for pull requests**, so the change isn't authorized yet and I'm closing it automatically. Once a maintainer adds the label to #4706, reopen this pull request and it will stay open. If #4706 needs more detail before that can happen, a complete runnable example is usually the missing piece, and adding one to the issue is the fastest way to get there.

PR #13599, opened 2026-09-21
Fix tautological assert start == start in test_from_linter.py

### LINKED ISSUE #13598: test_from_linter.py: assert start == start is always true after the loop variable is reassigned
Three tests in `test/sql/test_from_linter.py` reuse the loop variable as the tuple-unpack target, so the assertion that's supposed to check the returned value ends up comparing the same freshly-assigned object to itself:

```python
def test_disconnect_between_ab_cd(self):
    ...
    for start in self.a, self.b:
        froms, start = find_unmatching_froms(query, start)
        assert start == start
        assert froms == {self.c, self.d}
```

By the time `assert start == start` runs, `start` has already been overwritten by the tuple unpack on the line above, so both sides of `==` are the same object. The assertion passes no matter what `find_unmatching_froms` returns as its second element, so it isn't actually checking anything.

The same file shows what this was clearly meant to look like, in the non-loop calls right after each of these loops:

```python
froms, start = find_unmatching_froms(query, self.c)
assert start == self.c
```

Here the comparison is against the value that was passed in, which is a real check. The loop versions lost that because the loop variable and the unpack target share the same name.

Minimal illustration of why this always passes, independent of SQLAlchemy:

```python
def demo(start):
    froms, start = ("whatever", "anything, does not matter at all")
    assert start == start  # always true, start was just reassigned above
    return froms

demo("self.a")  # passes no matter what demo() returns as its 2nd element
```

Affected spots (current main):
- `test_disconnect_between_ab_cd`, both loops (lines ~278-284)
- `test_c_and_d_both_disconnected`, the loop (line ~294-296)
- `test_right_nested_join_with_an_issue`, the loop (line ~401-403)

I checked with a quick mutation test: if I make `find_unmatching_froms` return `None` for the second tuple element instead of the real `start`, `test_disconnect_between_ab_cd` still passes -- its two loops give zero coverage of that return value. The other two tests happen to also make a non-loop call afterwards that does catch it, so only `test_disconnect_between_ab_cd` is silently blind right now, but all three loops have the same dead assertion.

Fix is small: give the loop variable a different name than the unpack target, e.g. `for start_in in self.a, self.b: froms, start = find_unmatching_froms(query, start_in); assert start == start_in`, mirroring the pattern already used elsewhere in the same file. Happy to open a PR once this is labeled open for pull requests, per CONTRIBUTING.md.

mathewOracle:
Three loops reused the loop variable as the tuple-unpack target:

    for start in self.a, self.b:
        froms, start = find_unmatching_froms(query, start)
        assert start == start

By the time the assert runs, start has already been overwritten by the unpack on the line above, so it always compares an object to itself regardless of what find_unmatching_froms returns. The non-loop calls in the same file do this correctly by comparing against the value that was passed in (assert start == self.c), which is what these loops were clearly meant to do too.

Renamed the loop variable so the comparison is meaningful again:

    for start_in in self.a, self.b:
        froms, start = find_unmatching_froms(query, start_in)
        assert start == start_in

Verified with a mutation test: forcing find_unmatching_froms to return None as its second element is caught by all three fixed assertions. Against the original code, test_disconnect_between_ab_cd's two loops did not catch it at all -- they gave zero coverage of the return value. All 50 tests in test_from_linter.py (3 skipped, unrelated backend limitation) pass identically before and after this change.

Fixes #13598

**This project accepts pull requests only for issues that a maintainer has
marked with the `open for pull requests` label.**

A pull request that doesn't reference such an issue **is closed
automatically**. That isn't a judgment on your change: it's how we settle on
an approach before anyone spends time writing code, and how we keep two
people from doing the same work at once.

The issue you reference has to be:

* in this repository,
* open,
* labeled `open for pull requests`, and
* not already labeled `code review in progress`, which means someone is
  working on it already.

If there's no such issue yet, open one describing the problem or the feature,
including a complete, runnable example, and wait for a maintainer to add the
label. Please don't open a pull request first and ask for the label
afterwards.

### Fixes

Fixes: #


sqla-tester:

This pull request references issue #13598, which hasn't been marked **open for pull requests**, so the change isn't authorized yet and I'm closing it automatically. Once a maintainer adds the label to #13598, reopen this pull request and it will stay open. If #13598 needs more detail before that can happen, a complete runnable example is usually the missing piece, and adding one to the issue is the fastest way to get there.

CaselIT:
it's fine

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 2b0fa41ddb3d508deaf9b15a5da0134e23c2cbb0 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 2b0fa41ddb3d508deaf9b15a5da0134e23c2cbb0: https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6986

test/sql/test_from_linter.py
```diff
@@ -275,13 +275,13 @@ def test_disconnect_between_ab_cd(self):
             .where(self.c.c.col_c == self.d.c.col_d)
             .where(self.c.c.col_c == 5)
         )
-        for start in self.a, self.b:
-            froms, start = find_unmatching_froms(query, start)
-            assert start == start
```
sqla-tester, line 280: **Michael Bayer** (zzzeek) wrote:

wow, look at that vibe coding....seven years ago 

(the joke is this is not vibe coded...)

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6986

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6986

sqla-tester:

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6986

test/sql/test_from_linter.py
sqla-tester: **Michael Bayer** (zzzeek) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6986

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/sqlalchemy/+/6986 has been **merged**. Congratulations! :)

